PR #1, opened 2017-02-02
A kafkaesque room event consumer for the roomserver.

NegativeMjark:
Implement the main input loop for the roomserver.

It will receive events from a kafkaesque event source and track
where it is in the stream.

It currently does nothing with the events it consumes.

### DIFF
--- a/src/github.com/matrix-org/dendrite/roomserver/input/consumer.go
+++ b/src/github.com/matrix-org/dendrite/roomserver/input/consumer.go
@@ -1,0 +1,91 @@
+// Package input contains the code that writes
+package input
+
+import (
+	"github.com/matrix-org/dendrite/roomserver/types"
+	sarama "gopkg.in/Shopify/sarama.v1"
+)
+
+// A ConsumerDatabase has the storage APIs needed by the consumer.
+type ConsumerDatabase interface {
+	// PartitionOffsets returns the offsets the consumer has reached for each partition.
+	PartitionOffsets(topic string) ([]types.PartitionOffset, error)
+	// SetPartitionOffset records where the consumer has reached for a partition.
+	SetPartitionOffset(topic string, partition int32, offset int64) error
+}
+
+// An ErrorHandler handles the errors encountered by the consumer.
+type ErrorHandler interface {
+	OnError(err error)
+}
+
+// A Consumer consumes a kafkaesque stream of room events.
+type Consumer struct {
+	// A kafkaesque stream consumer.
+	Consumer sarama.Consumer
+	// The database used to store the room events.
+	DB ConsumerDatabase
+	// The kafkaesque topic to consume room events from.
+	RoomEventTopic string
+	// The ErrorHandler for this consumer.
+	// If left as nil then the consumer will panic with that error.
+	ErrorHandler ErrorHandler
+}
+
+// Start starts the consumer consuming.
+func (c *Consumer) Start() error {
+	offsets := map[int32]int64{}
+
+	partitions, err := c.Consumer.Partitions(c.RoomEventTopic)
+	if err != nil {
+		return err
+	}
+	for _, partition := range partitions {
+		// Default all the offsets to the beginning of the stream.
+		offsets[partition] = sarama.OffsetOldest
+	}
+
+	storedOffsets, err := c.DB.PartitionOffsets(c.RoomEventTopic)
+	if err != nil {
+		return err
+	}
+	for _, offset := range storedOffsets {
+		// We've already processed events from this partition so advance the offset to where we got to.
+		offsets[offset.Partition] = offset.Offset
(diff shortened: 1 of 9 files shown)


Kegsay:
Kafkaesque:
> having a nightmarishly complex, bizarre, or illogical quality 

/me tries not to make jokes about dendrite being kafkaesque :p

I assume you mean this in the context of Apache Kafka.

src/github.com/matrix-org/dendrite/roomserver/input/consumer.go
```diff
@@ -0,0 +1,91 @@
...
+type ConsumerDatabase interface {
+	// PartitionOffsets returns the offsets the consumer has reached for each partition.
+	PartitionOffsets(topic string) ([]types.PartitionOffset, error)
+	// SetPartitionOffset records where the consumer has reached for a partition.
+	SetPartitionOffset(topic string, partition int32, offset int64) error
+}
+
+// An ErrorHandler handles the errors encountered by the consumer.
+type ErrorHandler interface {
+	OnError(err error)
```
Kegsay: What's the intent behind this? Why aren't we just logging errors or whatever? There's no return type, so it's not like the entity which listens for this can *do* anything about it.

src/github.com/matrix-org/dendrite/roomserver/input/consumer.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// A Consumer consumes a kafkaesque stream of room events.
+type Consumer struct {
+	// A kafkaesque stream consumer.
+	Consumer sarama.Consumer
```
Kegsay: What does this do? A brief summary of its purpose would be super.

src/github.com/matrix-org/dendrite/roomserver/input/consumer.go
```diff
@@ -0,0 +1,91 @@
...
+		// We've already processed events from this partition so advance the offset to where we got to.
+		offsets[offset.Partition] = offset.Offset
+	}
+
+	var partitionConsumers []sarama.PartitionConsumer
+	for partition, offset := range offsets {
+		pc, err := c.Consumer.ConsumePartition(c.RoomEventTopic, partition, offset)
+		if err != nil {
+			for _, pc := range partitionConsumers {
+				pc.Close()
```
Kegsay: Presumably the intent here is to either start up ALL consumers or start up none of them, but you don't state this anywhere. I don't like the variable shadowing.

src/github.com/matrix-org/dendrite/roomserver/input/consumer.go
```diff
@@ -0,0 +1,91 @@
...
+	// The database used to store the room events.
+	DB ConsumerDatabase
+	// The kafkaesque topic to consume room events from.
+	RoomEventTopic string
+	// The ErrorHandler for this consumer.
+	// If left as nil then the consumer will panic with that error.
+	ErrorHandler ErrorHandler
+}
+
+// Start starts the consumer consuming.
```
Kegsay: Unhelpful. What is it actually trying to do? It loads up the stored partition offsets and f.e partition it starts consuming messages in separate goroutines.

src/github.com/matrix-org/dendrite/roomserver/input/consumer.go
```diff
@@ -0,0 +1,91 @@
...
+		go c.consumePartition(pc)
+	}
+
+	return nil
+}
+
+// consumePartition consumes the room events for a single partition of the kafkaesque stream.
+func (c *Consumer) consumePartition(pc sarama.PartitionConsumer) {
+	for message := range pc.Messages() {
+		// Do stuff with message.
```
Kegsay: Presumably this should be `// TODO: Do stuff with message.`

src/github.com/matrix-org/dendrite/roomserver/input/consumer.go
Kegsay: Also, https://godoc.org/gopkg.in/Shopify/sarama.v1#PartitionConsumer

> You MUST call Close() or AsyncClose() on a PartitionConsumer to avoid leaks, it will not be garbage-collected automatically when it passes out of scope.

We are not doing this in the normal case.

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -0,0 +1,70 @@
+package storage
+
+import (
+	"database/sql"
+	"github.com/matrix-org/dendrite/roomserver/types"
+)
+
+type stmts struct {
```
Kegsay: Just call it `statements`?

Kegsay:
LGTM :star:

PR #3, opened 2017-02-06
Store outlier room events.

NegativeMjark:


src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -68,3 +103,267 @@ func (s *statements) upsertPartitionOffset(topic string, partition int32, offset
...
+    (5, 'm.room.member'),
+    (6, 'm.room.redaction'),
+    (7, 'm.room.history_visibility') ON CONFLICT DO NOTHING;
+`
+
+const insertEventTypeNIDSQL = "" +
+	"INSERT INTO event_types (event_type) VALUES ($1)" +
+	" ON CONFLICT ON CONSTRAINT event_type_unique" +
+	" DO UPDATE SET event_type = $1" +
+	" RETURNING (event_type_nid)"
```
NegativeMjark: Can this SQL be followed by sane humans? Or should I be adding some comments/references to the postgresql docs to explain it. (The `DO UPDATE SET event_type = $1` is particularly gnarly and probably requires some explanation)

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -8,25 +8,60 @@ import (
...
+	if err = s.preparePartitionOffsets(db); err != nil {
 		return err
 	}
 
-	if s.selectPartitionOffsetsStmt, err = db.Prepare(selectPartitionOffsetsSQL); err != nil {
+	if err = s.prepareEventTypes(db); err != nil {
 		return err
 	}
-	if s.upsertPartitionOffsetStmt, err = db.Prepare(upsertPartitionOffsetsSQL); err != nil {
+
```
NegativeMjark: I had to split the `prepare` func up to evade the cyclomatic complexity restrictions, however this approach isn't sustainable. What do you think I should do about it? Should make a list of `prepare` functions somewhere and run them in a loop?

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -68,3 +103,267 @@ func (s *statements) upsertPartitionOffset(topic string, partition int32, offset
...
+-- Well known state keys are pre-assigned numeric IDs:
+--   1 -> "" (the empty string)
+CREATE SEQUENCE IF NOT EXISTS event_state_key_nid_seq START 65536;
+CREATE TABLE IF NOT EXISTS event_state_keys (
+    -- Local numeric ID for the state key.
+    event_state_key_nid BIGINT PRIMARY KEY DEFAULT nextval('event_state_key_nid_seq'),
+    event_state_key TEXT NOT NULL CONSTRAINT event_state_key_unique UNIQUE
+);
+INSERT INTO event_state_keys (event_state_key_nid, event_state_key) VALUES
+    (1, '') ON CONFLICT DO NOTHING;
```
NegativeMjark: What do you think about the comments being inside the schema? I left them in there when I was C+Ping from https://github.com/matrix-org/dendrite/blob/markjh/dendrite_wip/roomserver.sql. Are they okay or would you prefer to see them rewritten as go comments?

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
```diff
@@ -35,3 +36,68 @@ func (d *Database) PartitionOffsets(topic string) ([]types.PartitionOffset, erro
...
+	if eventTypeNID, err = d.assignEventTypeNID(event.Type()); err != nil {
+		return err
+	}
+
+	eventStateKey := event.StateKey()
+	if eventStateKey != nil {
+		if eventStateKeyNID, err = d.assignStateKeyNID(*eventStateKey); err != nil {
+			return err
+		}
+	}
```
NegativeMjark: Is using the default value of `0` for the eventStateKeyNID for non-state events to subtle? Would you like to see a comment or an explicit assignment in an `else` block here?

src/github.com/matrix-org/dendrite/roomserver/input/consumer.go
```diff
@@ -87,7 +90,14 @@ func (c *Consumer) Start() error {
...
 	for message := range pc.Messages() {
-		// TODO: Do stuff with message.
+		var input api.InputRoomEvent
+		if err := json.Unmarshal(message.Value, &input); err != nil {
+			c.logError(message, err)
+		} else {
+			if err := processRoomEvent(c.DB, input); err != nil {
+				c.logError(message, err)
+			}
+		}
```
NegativeMjark: The control flow is a bit subtle here. I want to stop processing the message if it's invalid JSON, but I want to advance the partition offset so that we can move onto the next message.

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
```diff
@@ -35,3 +36,68 @@ func (d *Database) PartitionOffsets(topic string) ([]types.PartitionOffset, erro
...
 	return d.statements.upsertPartitionOffset(topic, partition, offset)
 }
+
+// StoreEvent implements input.EventDatabase
+func (d *Database) StoreEvent(event gomatrixserverlib.Event) error {
+	var (
+		roomNID          int64
+		eventTypeNID     int64
+		eventStateKeyNID int64
+		eventNID         int64
```
Kegsay: What's a NID? Sounds exciting.

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
```diff
@@ -35,3 +36,68 @@ func (d *Database) PartitionOffsets(topic string) ([]types.PartitionOffset, erro
...
+	}
+	return
+}
+
+func (d *Database) assignEventTypeNID(eventType string) (eventTypeNID int64, err error) {
+	if eventTypeNID, err = d.statements.selectEventTypeNID(eventType); err != nil {
+		return
+	}
+	if eventTypeNID == 0 {
+		return d.statements.insertEventTypeNID(eventType)
```
Kegsay: Boilerplate is boilerplatey, and unclear boilerplate at that.

I understand that we're exchanging "stuff" for NIDs, but I don't understand:
 - Which things can be exchanged for NIDs.
 - Why we want to do this.

As an aside, I dislike how we "mostly" use named-returns, except when we don't (this line). I'd rather we just used normal returns:
```go
whateverNID, err := d.statements.selectWhateverNID(whatever)
if err != nil {
  return 0, err
}
if whateverNID == 0 {
  return d.statements.insertWhateverNID(whatever)
}
return whateverNID, err
```

Also, the idea that "0" is a special sentinel value for NID feels like a weird choice, given it's more typical to return error consts like `sql.ErrNoRows`.

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
```diff
@@ -35,3 +36,68 @@ func (d *Database) PartitionOffsets(topic string) ([]types.PartitionOffset, erro
...
+	}
+	return
+}
+
+func (d *Database) assignStateKeyNID(eventStateKey string) (eventStateKeyNID int64, err error) {
+	if eventStateKeyNID, err = d.statements.selectEventStateKeyNID(eventStateKey); err != nil {
+		return
+	}
+	if eventStateKeyNID == 0 {
+		return d.statements.insertEventTypeNID(eventStateKey)
```
Kegsay: You don't really meant to `insertEventTypeNID` I hope? :cry:

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
Kegsay: Never mind, you explain this in the schema.

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -68,3 +103,267 @@ func (s *statements) upsertPartitionOffset(topic string, partition int32, offset
...
+    event_type TEXT NOT NULL CONSTRAINT event_type_unique UNIQUE
+);
+INSERT INTO event_types (event_type_nid, event_type) VALUES
+    (1, 'm.room.create'),
+    (2, 'm.room.power_levels'),
+    (3, 'm.room.join_rules'),
+    (4, 'm.room.third_party_invite'),
+    (5, 'm.room.member'),
+    (6, 'm.room.redaction'),
+    (7, 'm.room.history_visibility') ON CONFLICT DO NOTHING;
```
Kegsay: How do you not have any `iota` style enum constants for this *anywhere* in this PR?

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -68,3 +103,267 @@ func (s *statements) upsertPartitionOffset(topic string, partition int32, offset
...
+	}
+	return
+}
+
+const eventTypesSchema = `
+-- Numeric versions of the event "type"s. Event types tend to be taken from a
+-- small common pool. Assigning each a numeric ID should reduce the amount of
+-- data that needs to be stored and fetched from the database.
+-- It also means that many operations can work with int64 arrays rather than
+-- string arrays which may help reduce GC pressure.
```
Kegsay: > Assigning each a numeric ID **should** reduce the amount of data that needs to be stored and fetched from the database. It also means that many operations can work with int64 arrays rather than string arrays which **may** help reduce GC pressure.

This feels like a massive premature optimisation. You're going to need to give me some actual evidence to support your theories: benchmarks done by yourself or others, or something better than just a hand-wavey comment.

I dislike this because look at how much of this PR is dedicated to doing this mapping, and no real "useful" work. All the `assign*` functions and these schema files just add boilerplate which I'd rather not have unless there was an extremely compelling reason for it. I don't find these comments compelling enough.

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -68,3 +103,267 @@ func (s *statements) upsertPartitionOffset(topic string, partition int32, offset
...
+--   3 -> m.room.join_rules
+--   4 -> m.room.third_party_invite
+--   5 -> m.room.member
+--   6 -> m.room.redaction
+--   7 -> m.room.history_visibility
+-- Picking well-known numeric IDs for the events types that require special
+-- attention during state conflict resolution means that we write that code
+-- using numeric constants.
+-- It also means that the numeric IDs for common event types should be
+-- consistent between different instances which might make ad-hoc debugging
```
Kegsay: You know what else would be consistent and make ad-hoc debugging easier? Strings. :scream:

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
Kegsay: > Can this SQL be followed by sane humans? 

If you have to ask this, the answer is no. A one-liner to explain your intent would be great, rather than relying on Postgres' suboptimal mechanical syntax.

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -68,3 +103,267 @@ func (s *statements) upsertPartitionOffset(topic string, partition int32, offset
...
+}
+
+const eventStateKeysSchema = `
+-- Numeric versions of the event "state_key"s. State keys tend to be reused so
+-- assigning each string a numeric ID should reduce the amount of data that
+-- needs to be stored and fetched from the database.
+-- It also means that many operations can work with int64 arrays rather than
+-- string arrays which may help reduce GC pressure.
+-- Well known state keys are pre-assigned numeric IDs:
+--   1 -> "" (the empty string)
```
Kegsay: This is basically the **only** reused state key. The rest in the spec are either server names or user IDs. It feels like a weak argument to "numerify" state keys when you now need to SELECT and possibly INSERT just to "convert" state keys into a useful form for Dendrite.

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -68,3 +103,267 @@ func (s *statements) upsertPartitionOffset(topic string, partition int32, offset
...
+--   5 -> m.room.member
+--   6 -> m.room.redaction
+--   7 -> m.room.history_visibility
+-- Picking well-known numeric IDs for the events types that require special
+-- attention during state conflict resolution means that we write that code
+-- using numeric constants.
+-- It also means that the numeric IDs for common event types should be
+-- consistent between different instances which might make ad-hoc debugging
+-- easier.
+CREATE SEQUENCE IF NOT EXISTS event_type_nid_seq START 65536;
```
Kegsay: Why `START 65536`?

src/github.com/matrix-org/dendrite/roomserver/input/events.go
```diff
@@ -0,0 +1,30 @@
...
+	if err != nil {
+		return err
+	}
+
+	if err := db.StoreEvent(event); err != nil {
+		return err
+	}
+
+	if input.Kind == api.KindOutlier {
+		// For outlier events we only need to store the event JSON.
```
Kegsay: Why?

src/github.com/matrix-org/dendrite/roomserver/input/consumer.go
```diff
@@ -87,7 +90,14 @@ func (c *Consumer) Start() error {
 func (c *Consumer) consumePartition(pc sarama.PartitionConsumer) {
 	defer pc.Close()
 	for message := range pc.Messages() {
-		// TODO: Do stuff with message.
+		var input api.InputRoomEvent
+		if err := json.Unmarshal(message.Value, &input); err != nil {
+			c.logError(message, err)
+		} else {
+			if err := processRoomEvent(c.DB, input); err != nil {
+				c.logError(message, err)
```
Kegsay: You don't explain why you only log the error instead of something more drastic. Please explain.

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
```diff
@@ -35,3 +36,68 @@ func (d *Database) PartitionOffsets(topic string) ([]types.PartitionOffset, erro
...
+
+	eventStateKey := event.StateKey()
+	if eventStateKey != nil {
+		if eventStateKeyNID, err = d.assignStateKeyNID(*eventStateKey); err != nil {
+			return err
+		}
+	}
+
+	eventID := event.EventID()
+	referenceSHA256 := event.EventReference().EventSHA256
```
Kegsay: Seems weird to extract this for use in precisely 1 place. I would understand if the temp variable had a clearer name, but it isn't the case. Why don't we just make `insertEvent` multi-line?

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
NegativeMjark: However the user IDs and server names will likely appear multiple times in the database so reducing them to numeric IDs will provide a valuable level of compression.

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
Kegsay: A one-liner along the lines of:
```go
// work out eventStateKeyNID if there's a state key, else default to 0
```

would be nice.

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
```diff
@@ -35,3 +36,71 @@ func (d *Database) PartitionOffsets(topic string) ([]types.PartitionOffset, erro
...
+
+	if eventNID, err = d.statements.insertEvent(roomNID, eventTypeNID, eventStateKeyNID, eventID, referenceSHA256); err != nil {
+		return err
+	}
+
+	return d.statements.insertEventJSON(eventNID, event.JSON())
```
Kegsay: Why do we have 2 functions: one for `insertEvent` and one for `insertEventJSON`? Why not just use 1?

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
Kegsay: I like it. It allows you to scope comments to particular bits of SQL (like on :219 where you say `-- Local numeric ID for the state key.`)

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
Kegsay: Much better now, thanks.

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -68,3 +103,283 @@ func (s *statements) upsertPartitionOffset(topic string, partition int32, offset
...
+// We peform a update that does nothing rather that doing nothing at all because
+// postgres won't return anything unless we touch a row in the table.
+const insertEventTypeNIDSQL = "" +
+	"INSERT INTO event_types (event_type) VALUES ($1)" +
+	" ON CONFLICT ON CONSTRAINT event_type_unique" +
+	" DO UPDATE SET event_type = $1" +
+	" RETURNING (event_type_nid)"
+
+const selectEventTypeNIDSQL = "" +
+	"SELECT event_type_nid FROM event_types WHERE event_type = $1"
```
Kegsay: Why don't you just do:
```go
const selectEventTypeNIDSQL = (
    "SELECT event_type_nid FROM event_types WHERE event_type = $1"
)
```

rather than zero-length concats?

Kegsay: https://play.golang.org/p/DVrTRwanvj

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -68,3 +103,283 @@ func (s *statements) upsertPartitionOffset(topic string, partition int32, offset
...
+
+func (s *statements) insertEventTypeNID(eventType string) (eventTypeNID int64, err error) {
+	err = s.insertEventTypeNIDStmt.QueryRow(eventType).Scan(&eventTypeNID)
+	return
+}
+
+func (s *statements) selectEventTypeNID(eventType string) (eventTypeNID int64, err error) {
+	err = s.selectEventTypeNIDStmt.QueryRow(eventType).Scan(&eventTypeNID)
+	if err == sql.ErrNoRows {
+		eventTypeNID = 0
```
Kegsay: Do you ever explain what a return value of `0` actually means? Please can you?

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -68,3 +103,283 @@ func (s *statements) upsertPartitionOffset(topic string, partition int32, offset
...
+
+func (s *statements) insertEventStateKeyNID(eventStateKey string) (eventStateKeyNID int64, err error) {
+	err = s.insertEventStateKeyNIDStmt.QueryRow(eventStateKey).Scan(&eventStateKeyNID)
+	return
+}
+
+func (s *statements) selectEventStateKeyNID(eventStateKey string) (eventStateKeyNID int64, err error) {
+	err = s.selectEventStateKeyNIDStmt.QueryRow(eventStateKey).Scan(&eventStateKeyNID)
+	if err == sql.ErrNoRows {
+		eventStateKeyNID = 0
```
Kegsay: Explain the significance of 0.

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -68,3 +103,283 @@ func (s *statements) upsertPartitionOffset(topic string, partition int32, offset
...
+		return
+	}
+	return
+}
+
+const roomsSchema = `
+CREATE SEQUENCE IF NOT EXISTS room_nid_seq;
+CREATE TABLE IF NOT EXISTS rooms (
+    -- Local numeric ID for the room.
+    room_nid BIGINT PRIMARY KEY DEFAULT nextval('room_nid_seq'),
```
Kegsay: Isn't there an easier form using `AUTOINCREMENT` or something?

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -68,3 +103,283 @@ func (s *statements) upsertPartitionOffset(topic string, partition int32, offset
...
+    -- Textual ID for the room.
+    room_id TEXT NOT NULL CONSTRAINT room_id_unique UNIQUE
+);
+`
+
+// Same as insertEventTypeNIDSQL
```
Kegsay: :+1: I like this: it means you can use the same kinda of SQL voodoo and document it once.

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -68,3 +103,283 @@ func (s *statements) upsertPartitionOffset(topic string, partition int32, offset
...
+	}
+	if s.insertEventJSONStmt, err = db.Prepare(insertEventJSONSQL); err != nil {
+		return
+	}
+	return
+}
+
+const eventJSONSchema = `
+-- Stores the JSON for each event. This kept separate from the main events
+-- table to keep the rows in the main events table small.
```
Kegsay: Aha, this answers https://github.com/matrix-org/dendrite/pull/3/files#r99818915 I guess.

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -68,3 +103,283 @@ func (s *statements) upsertPartitionOffset(topic string, partition int32, offset
...
+CREATE TABLE IF NOT EXISTS event_json (
+    -- Local numeric ID for the event.
+    event_nid bigint NOT NULL PRIMARY KEY,
+    -- The JSON for the event.
+    -- Stored as TEXT because this should be valid UTF-8.
+    -- Not stored as a JSONB because we always just pull the entire event.
```
Kegsay: But this does not validate that `event_json` contains **valid** JSON then? This feels wrong.

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
NegativeMjark: There is `bigserial`. But since I need to use `CREATE SEQUENCE` elsewhere in the schema I've preserved this form for consistency.

NegativeMjark:
@Kegsay PTAL

src/github.com/matrix-org/dendrite/roomserver/input/events.go
```diff
@@ -20,11 +21,26 @@ func processRoomEvent(db RoomEventDatabase, input api.InputRoomEvent) error {
...
 	}
 
+	// TODO:
+	//  * Check that the event passes authentication checks.
+
 	if input.Kind == api.KindOutlier {
-		// For outlier events we only need to store the event JSON.
+		// For outliers we can stop after we've stored the event itself as it
+		// doesn't have any associated state to store and we don't need to
+		// notify anyone it.
```
Kegsay: > notify anyone **about** it.

surely?

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -68,3 +103,274 @@ func (s *statements) upsertPartitionOffset(topic string, partition int32, offset
...
+-- The events table holds metadata for each event, the actual JSON is stored
+-- separately to keep the size of the rows small.
+CREATE SEQUENCE IF NOT EXISTS event_nid_seq;
+CREATE TABLE IF NOT EXISTS events (
+    -- Local numeric ID for the event.
+    event_nid BIGINT PRIMARY KEY DEFAULT nextval('event_nid_seq'),
+    -- Local numeric ID for the room the event is in.
+    room_nid BIGINT NOT NULL,
+	-- Local numeric ID for the type of the event.
+    event_type_nid BIGINT NOT NULL,
```
Kegsay: Can you state that there should never be an `event_type_nid` of `0` please? (to contrast `event_state_key_nid` which *can* be `0`).

Kegsay:
lgtm2

PR #4, opened 2017-02-08
Check that events pass authentication checks.

NegativeMjark:
Record the list of events that the event passes authentication checks
against.

### DIFF
--- a/src/github.com/matrix-org/dendrite/roomserver/input/events.go
+++ b/src/github.com/matrix-org/dendrite/roomserver/input/events.go
@@ -15,10 +15,10 @@ type RoomEventDatabase interface {
 	// Returns a sorted list of state entries.
 	StateEntriesForEventIDs(eventIDs []string) ([]types.StateEntry, error)
 	// Lookup the numeric IDs for a list of string event state keys.
-	// Returns a sorted list of state entries.
+	// Returns a map from string state key to numeric ID for the state key.
 	EventStateKeyNIDs(eventStateKeys []string) (map[string]int64, error)
 	// Lookup the Events for a list of numeric event IDs.
-	// Returns a sorted list of state entries.
+	// Returns a sorted list of events.
 	Events(eventNIDs []int64) ([]types.Event, error)
 }
 
--- a/src/github.com/matrix-org/dendrite/roomserver/input/events.go
+++ b/src/github.com/matrix-org/dendrite/roomserver/input/events.go
@@ -104,22 +104,27 @@ type authEvents struct {
 	events         eventMap
 }
 
+// Create implements gomatrixserverlib.AuthEvents
 func (ae *authEvents) Create() (*gomatrixserverlib.Event, error) {
 	return ae.lookupEventWithEmptyStateKey(types.MRoomCreateNID), nil
 }
 
+// PowerLevels implements gomatrixserverlib.AuthEvents
 func (ae *authEvents) PowerLevels() (*gomatrixserverlib.Event, error) {
 	return ae.lookupEventWithEmptyStateKey(types.MRoomPowerLevelsNID), nil
 }
 
+// JoinRules implements gomatrixserverlib.AuthEvents
 func (ae *authEvents) JoinRules() (*gomatrixserverlib.Event, error) {
 	return ae.lookupEventWithEmptyStateKey(types.MRoomJoinRulesNID), nil
 }
 
+// Memmber implements gomatrixserverlib.AuthEvents
 func (ae *authEvents) Member(stateKey string) (*gomatrixserverlib.Event, error) {
 	return ae.lookupEvent(types.MRoomMemberNID, stateKey), nil
 }
 
+// ThirdPartyInvite implements gomatrixserverlib.AuthEvents
 func (ae *authEvents) ThirdPartyInvite(stateKey string) (*gomatrixserverlib.Event, error) {
 	return ae.lookupEvent(types.MRoomThirdPartyInviteNID, stateKey), nil
 }
--- a/src/github.com/matrix-org/dendrite/roomserver/input/events.go
+++ b/src/github.com/matrix-org/dendrite/roomserver/input/events.go
@@ -17,7 +17,7 @@ type RoomEventDatabase interface {
 	StateEntriesForEventIDs(eventIDs []string) ([]types.StateEntry, error)
 	// Lookup the numeric IDs for a list of string event state keys.
 	// Returns a sorted list of state entries.
-	EventStateKeyNIDs(eventStateKeys []string) ([]types.IDPair, error)
+	EventStateKeyNIDs(eventStateKeys []string) (map[string]int64, error)
 	// Lookup the Events for a list of numeric event IDs.
 	// Returns a sorted list of state entries.
 	Events(eventNIDs []int64) ([]types.Event, error)
@@ -99,9 +99,9 @@ func checkAuthEvents(db RoomEventDatabase, event gomatrixserverlib.Event, authEv
 }
 
 type authEvents struct {
-	stateNIDMap idMap
-	state       stateEntryMap
-	events      eventMap
+	stateKeyNIDMap map[string]int64
+	state          stateEntryMap
+	events         eventMap
 }
 
 func (ae *authEvents) Create() (*gomatrixserverlib.Event, error) {
@@ -137,7 +137,7 @@ func (ae *authEvents) lookupEventWithEmptyStateKey(typeNID int64) *gomatrixserve
 }
 
 func (ae *authEvents) lookupEvent(typeNID int64, stateKey string) *gomatrixserverlib.Event {
-	stateKeyNID, ok := ae.stateNIDMap.lookup(stateKey)
+	stateKeyNID, ok := ae.stateKeyNIDMap[stateKey]
 	if !ok {
 		return nil
 	}
@@ -162,32 +162,28 @@ func loadAuthEvents(
 	var eventStateKeys []string
 	eventStateKeys = append(eventStateKeys, needed.Member...)
 	eventStateKeys = append(eventStateKeys, needed.ThirdPartyInvite...)
-	stateKeyNIDs, err := db.EventStateKeyNIDs(eventStateKeys)
-	if err != nil {
+	if result.stateKeyNIDMap, err = db.EventStateKeyNIDs(eventStateKeys); err != nil {
 		return
 	}
-	result.stateNIDMap = newIDMap(stateKeyNIDs)
 
 	// Load the events we need.
-	keysNeeded := stateKeysNeeded(result.stateNIDMap, needed)
+	result.state = state
 	var eventNIDs []int64
-	result.state = newStateEntryMap(state)
+	keysNeeded := stateKeysNeeded(result.stateKeyNIDMap, needed)
 	for _, keyNeeded := range keysNeeded {
 		eventNID, ok := result.state.lookup(keyNeeded)
 		if ok {
 			eventNIDs = append(eventNIDs, eventNID)
 		}
 	}
-	events, err := db.Events(eventNIDs)
-	if err != nil {
+	if result.events, err = db.Events(eventNIDs); err != nil {
 		return
 	}
-	result.events = newEventMap(events)
 	return
 }
 
 // stateKeysNeeded works out which numeric state keys we need to authenticate some events.
-func stateKeysNeeded(stateNIDMap idMap, stateNeeded gomatrixserverlib.StateNeeded) []types.StateKeyTuple {
+func stateKeysNeeded(stateKeyNIDMap map[string]int64, stateNeeded gomatrixserverlib.StateNeeded) []types.StateKeyTuple {
 	var keys []types.StateKeyTuple
 	if stateNeeded.Create {
 		keys = append(keys, types.StateKeyTuple{types.MRoomCreateNID, types.EmptyStateKeyNID})
@@ -199,13 +195,13 @@ func stateKeysNeeded(stateNIDMap idMap, stateNeeded gomatrixserverlib.StateNeede
 		keys = append(keys, types.StateKeyTuple{types.MRoomJoinRulesNID, types.EmptyStateKeyNID})
 	}
 	for _, member := range stateNeeded.Member {
-		stateKeyNID, ok := stateNIDMap.lookup(member)
+		stateKeyNID, ok := stateKeyNIDMap[member]
 		if ok {
 			keys = append(keys, types.StateKeyTuple{types.MRoomMemberNID, stateKeyNID})
 		}
 	}
 	for _, token := range stateNeeded.ThirdPartyInvite {
-		stateKeyNID, ok := stateNIDMap.lookup(token)
+		stateKeyNID, ok := stateKeyNIDMap[token]
 		if ok {
 			keys = append(keys, types.StateKeyTuple{types.MRoomThirdPartyInviteNID, stateKeyNID})
 		}
@@ -213,34 +209,9 @@ func stateKeysNeeded(stateNIDMap idMap, stateNeeded gomatrixserverlib.StateNeede
 	return keys
 }
 
-type idMap map[string]int64
-
-func newIDMap(ids []types.IDPair) idMap {
-	result := make(map[string]int64)
-	for _, pair := range ids {
-		result[pair.ID] = pair.NID
-	}
-	return idMap(result)
-}
-
-// lookup an entry in the id map.
-func (m idMap) lookup(id string) (nid int64, ok bool) {
-	// Use a hash map here.
-	// We could use binary search here like we do for the maps below as it
-	// would be faster for small lists.
-	// However the benefits of binary search aren't as strong here and it's
-	// possible that we could encounter sets of pathological strings since
-	// the state keys are ultimately controlled by user input.
-	nid, ok = map[string]int64(m)[id]
-	return
-}
-
+// Map from event type, state key tuple to numeric event ID.
+// Implemented using binary search on a sorted array.
 type stateEntryMap []types.StateEntry
-
-// newStateEntryMap creates a map from a sorted list of state entries.
-func newStateEntryMap(stateEntries []types.StateEntry) stateEntryMap {
-	return stateEntryMap(stateEntries)
-}
 
 // lookup an entry in the event map.
 func (m stateEntryMap) lookup(stateKey types.StateKeyTuple) (eventNID int64, ok bool) {
@@ -259,12 +230,9 @@ func (m stateEntryMap) lookup(stateKey types.StateKeyTuple) (eventNID int64, ok 
 	return
 }
 
+// Map from numeric event ID to event.
+// Implemented using binary search on a sorted array.
 type eventMap []types.Event
-
-// newEventMap creates a map from a sorted list of events.
-func newEventMap(events []types.Event) eventMap {
-	return eventMap(events)
-}
 
 // lookup an entry in the event map.
 func (m eventMap) lookup(eventNID int64) (event *types.Event, ok bool) {
--- a/src/github.com/matrix-org/dendrite/roomserver/input/events.go
+++ b/src/github.com/matrix-org/dendrite/roomserver/input/events.go
@@ -1,7 +1,6 @@
 package input
 
 import (
-	"fmt"
 	"github.com/matrix-org/dendrite/roomserver/api"
 	"github.com/matrix-org/dendrite/roomserver/types"
 	"github.com/matrix-org/gomatrixserverlib"
@@ -70,10 +69,6 @@ func checkAuthEvents(db RoomEventDatabase, event gomatrixserverlib.Event, authEv
 	if err != nil {
 		return nil, err
 	}
-	if len(authStateEntries) < len(authEventIDs) {
-		return nil, fmt.Errorf("input: Some of the auth event IDs were missing from the database")
-	}
-
 	// TODO: check for duplicate state keys here.
 
 	// Work out which of the state events we actually need.
--- a/src/github.com/matrix-org/dendrite/roomserver/input/events.go
+++ b/src/github.com/matrix-org/dendrite/roomserver/input/events.go
@@ -158,20 +158,20 @@ func loadAuthEvents(
 	needed gomatrixserverlib.StateNeeded,
 	state []types.StateEntry,
 ) (result authEvents, err error) {
-	// Lookup the numeric IDs for the state keys
-	var eventStateKeys []string
-	eventStateKeys = append(eventStateKeys, needed.Member...)
-	eventStateKeys = append(eventStateKeys, needed.ThirdPartyInvite...)
-	if result.stateKeyNIDMap, err = db.EventStateKeyNIDs(eventStateKeys); err != nil {
+	// Lookup the numeric IDs for the state keys needed for auth.
+	var neededStateKeys []string
+	neededStateKeys = append(neededStateKeys, needed.Member...)
+	neededStateKeys = append(neededStateKeys, needed.ThirdPartyInvite...)
+	if result.stateKeyNIDMap, err = db.EventStateKeyNIDs(neededStateKeys); err != nil {
 		return
 	}
 
 	// Load the events we need.
 	result.state = state
 	var eventNIDs []int64
-	keysNeeded := stateKeysNeeded(result.stateKeyNIDMap, needed)
-	for _, keyNeeded := range keysNeeded {
-		eventNID, ok := result.state.lookup(keyNeeded)
+	keyTuplesNeeded := stateKeyTuplesNeeded(result.stateKeyNIDMap, needed)
+	for _, keyTuple := range keyTuplesNeeded {
+		eventNID, ok := result.state.lookup(keyTuple)
 		if ok {
 			eventNIDs = append(eventNIDs, eventNID)
 		}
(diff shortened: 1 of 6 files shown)


src/github.com/matrix-org/dendrite/roomserver/types/types.go
```diff
@@ -1,10 +1,82 @@
...
 type PartitionOffset struct {
 	// The ID of the partition.
 	Partition int32
 	// The offset into the partition.
 	Offset int64
 }
+
+// A StateKey is a pair of a numeric event type and a numeric state key.
+// It is used to lookup state entries.
+type StateKey struct {
```
Kegsay: Naming: If I see a function:
```go
func foo(a StateKey) {

}
```

I am *always* going to think `a` is `event.state_key`, no matter how much you doc otherwise. Can we call this something else, like `StateTuple`, `StateKeyTuple`, `StateTupleNID`, etc?

src/github.com/matrix-org/dendrite/roomserver/types/types.go
```diff
@@ -1,10 +1,82 @@
...
+	// The numeric ID for the event type.
+	EventTypeNID int64
+	// The numeric ID for the state key or 0 if the event is not a state event.
```
Kegsay: Somewhat weird that you can have a `StateKey` and it isn't a state event? What's the deal with that?

src/github.com/matrix-org/dendrite/roomserver/types/types.go
```diff
@@ -1,10 +1,82 @@
...
+
+// A StateEntry is an entry in the room state of a matrix room.
+type StateEntry struct {
+	StateKey
+	// The numeric ID for the event.
+	EventNID int64
+}
+
+// LessThan returns true if this state entry is less than the other state entry.
+func (a StateEntry) LessThan(b StateEntry) bool {
```
Kegsay: I think you should add a comment here to explain how you plan on using sorting to detect and remove duplicates. The `LessThan` implementation exists only for this reason, and that there is no semantic meaning to the actual sort output.

src/github.com/matrix-org/dendrite/roomserver/types/types.go
```diff
@@ -1,10 +1,82 @@
...
+	if a.StateKey != b.StateKey {
+		return a.StateKey.LessThan(b.StateKey)
+	}
+	return a.EventNID < b.EventNID
+}
+
+// An IDPair is a pair of a string ID and the corresponding numeric ID.
```
Kegsay: Unhelpful. I can tell it's a pair of string ID, int ID from the struct definition. What do the ints/strings represent? I suspect they represent multiple things, which is why you're being vague, but this doesn't help inform the reader what its purpose in life is. The bulk lookup comment is somewhat more useful, maybe flesh that out a bit more?

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -2,20 +2,24 @@ package storage
...
-	selectRoomNIDStmt          *sql.Stmt
-	insertEventStmt            *sql.Stmt
-	insertEventJSONStmt        *sql.Stmt
+	selectPartitionOffsetsStmt  *sql.Stmt
+	upsertPartitionOffsetStmt   *sql.Stmt
+	insertEventTypeNIDStmt      *sql.Stmt
+	selectEventTypeNIDStmt      *sql.Stmt
+	insertEventStateKeyNIDStmt  *sql.Stmt
+	selectEventStateKeyNIDStmt  *sql.Stmt
+	selectEventStateKeyNIDsStmt *sql.Stmt
```
Kegsay: Can we rename this to something which won't be confused with the almost identical `selectEventStateKeyNIDStmt` please? Like `selectMultiEventStateKeyNIDStmt`?

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -230,6 +237,11 @@ const insertEventStateKeyNIDSQL = "" +
 const selectEventStateKeyNIDSQL = "" +
 	"SELECT event_state_key_nid FROM event_state_keys WHERE event_state_key = $1"
 
+const selectEventStateKeyNIDsSQL = "" +
+	"SELECT event_state_key, event_state_key_nid FROM event_state_keys" +
+	" WHERE event_state_key = ANY($1)" +
```
Kegsay: `ANY($1)` is like SQLite3's `IN` right? Is there a limit to the number of args you can give? If there is, is it possible for us to reach the limit?

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -230,6 +237,11 @@ const insertEventStateKeyNIDSQL = "" +
...
+	" ORDER BY event_state_key ASC"
```
Kegsay: Why?

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -240,6 +252,24 @@ func (s *statements) selectEventStateKeyNID(eventStateKey string) (eventStateKey
...
 
+func (s *statements) selectEventStateKeyNIDs(eventStateKeys []string) ([]types.IDPair, error) {
+	rows, err := s.selectEventStateKeyNIDsStmt.Query(pq.StringArray(eventStateKeys))
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+
+	results := make([]types.IDPair, len(eventStateKeys))
+	i := 0
```
Kegsay: Why do you do this? In other projects, like [go-neb](https://github.com/matrix-org/go-neb/blob/master/src/github.com/matrix-org/go-neb/database/schema.go#L258), we just have a slice and call `append()`. I imagine this is being done for allocation reasons (once vs many with `append()` if it has to resize) but you don't comment that this is the case.

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -307,17 +337,24 @@ CREATE TABLE IF NOT EXISTS events (
     event_id TEXT NOT NULL CONSTRAINT event_id_unique UNIQUE,
     -- The sha256 reference hash for the event.
     -- Needed for setting reference hashes when sending new events.
-    reference_sha256 BYTEA NOT NULL
+    reference_sha256 BYTEA NOT NULL,
+    -- A list of numeric IDs for events that can authenticate this event.
```
Kegsay: > numeric IDs for events

You mean:

> numeric event IDs

As in `Event.EventNID` on https://github.com/matrix-org/dendrite/pull/4/files#diff-991257c04f932031b519c68ccb3ebdcaR58 ?

Kegsay: Also, can we be consistent and suffix `nid` onto this? We do that everywhere else where we have `BIGINT` stuff.

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -307,17 +337,24 @@ CREATE TABLE IF NOT EXISTS events (
...
+	"INSERT INTO events (room_nid, event_type_nid, event_state_key_nid, event_id, reference_sha256, auth_events)" +
+	" VALUES ($1, $2, $3, $4, $5, $6)" +
 	" ON CONFLICT ON CONSTRAINT event_id_unique" +
 	" DO UPDATE SET event_id = $1" +
 	" RETURNING event_nid"
 
+const selectStateEventsByIDSQL = "" +
+	"SELECT event_type_nid, event_state_key_nid, event_nid FROM events" +
+	" WHERE event_id = ANY($1)" +
+	" ORDER BY event_type_nid, event_state_key_nid ASC"
```
Kegsay: Why ASC? Why not DESC? Does it matter?

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -326,20 +363,46 @@ func (s *statements) prepareEvents(db *sql.DB) (err error) {
...
+		return
+	}
 	return
 }
 
 func (s *statements) insertEvent(
 	roomNID, eventTypeNID, eventStateKeyNID int64,
 	eventID string,
 	referenceSHA256 []byte,
+	authEventNIDs []int64,
```
Kegsay: What do you think about having a type alias of:
```go
type NID int64
```

Do you think this would be useful and clearer to the reader? That means you could drop all the `NID` suffixes on your variable names.

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -326,20 +363,46 @@ func (s *statements) prepareEvents(db *sql.DB) (err error) {
...
 
+func (s *statements) selectStateEventsByID(eventIDs []string) ([]types.StateEntry, error) {
+	results := make([]types.StateEntry, len(eventIDs))
+	rows, err := s.selectStateEventsByIDStmt.Query(pq.StringArray(eventIDs))
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+	i := 0
+	for ; rows.Next(); i++ {
```
Kegsay: Can you do this [here](https://github.com/matrix-org/dendrite/pull/4/files#diff-8e2932ebbde743c3da4d0d9a88d5b205R268) please? Be consistent, and I prefer the in-line form.

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -348,6 +411,9 @@ func (s *statements) prepareEventJSON(db *sql.DB) (err error) {
 	if s.insertEventJSONStmt, err = db.Prepare(insertEventJSONSQL); err != nil {
 		return
 	}
+	if s.selectEventJSONsStmt, err = db.Prepare(selectEventJSONsSQL); err != nil {
```
Kegsay: https://github.com/matrix-org/dendrite/pull/4/files#r100104642

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -372,7 +438,35 @@ const insertEventJSONSQL = "" +
 	"INSERT INTO event_json (event_nid, event_json) VALUES ($1, $2)" +
 	" ON CONFLICT DO NOTHING"
 
+const selectEventJSONsSQL = "" +
+	"SELECT event_nid, event_json FROM event_json" +
+	" WHERE event_nid = ANY($1)" +
+	" ORDER BY event_nid ASC"
```
Kegsay: Why ASC? Why not DESC or unsorted?

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
```diff
@@ -38,7 +38,7 @@ func (d *Database) SetPartitionOffset(topic string, partition int32, offset int6
 }
 
 // StoreEvent implements input.EventDatabase
-func (d *Database) StoreEvent(event gomatrixserverlib.Event) error {
+func (d *Database) StoreEvent(event gomatrixserverlib.Event, authEventNIDS []int64) error {
```
Kegsay: Nit: `authEventNIDS` => `authEventNIDs`

Or even better: `authEvents []NID` mayhaps? :angel:

Kegsay: From an API perspective, if feels weird that the auth event NIDs aren't bundled with the `event`. Why is this?

src/github.com/matrix-org/dendrite/roomserver/input/events.go
```diff
@@ -44,3 +58,222 @@ func processRoomEvent(db RoomEventDatabase, input api.InputRoomEvent) error {
...
+// checkAuthEvents checks that the event passes authentication checks
+// Returns the numeric IDs for the auth events.
+func checkAuthEvents(db RoomEventDatabase, event gomatrixserverlib.Event, authEventIDs []string) ([]int64, error) {
+	// Grab the numeric IDs for the supplied auth state events from the database.
+	authStateEntries, err := db.StateEntriesForEventIDs(authEventIDs)
+	if err != nil {
+		return nil, err
+	}
+	if len(authStateEntries) < len(authEventIDs) {
+		return nil, fmt.Errorf("input: Some of the auth event IDs were missing from the database")
```
Kegsay: It feels like this check should be the responsibility of `StateEntriesForEventIDs`. Why don't we just return an `err` if the output slice != input slice?

Kegsay: Also, as always, more specific error messages would be great: along the lines of test failures:

> expect X, got Y

src/github.com/matrix-org/dendrite/roomserver/input/events.go
```diff
@@ -44,3 +58,222 @@ func processRoomEvent(db RoomEventDatabase, input api.InputRoomEvent) error {
...
+	}
+
+	// Work out which of the state events we actaully need.
```
Kegsay: > actaully

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
NegativeMjark: It is similar but the parameter is an array. https://www.postgresql.org/docs/9.1/static/arrays.html
This means that the entire list is passed as a parameter to the query rather than being part of the query string. I've yet to find a limit on the size of query parameter in postgresql.

src/github.com/matrix-org/dendrite/roomserver/input/events.go
```diff
@@ -44,3 +61,224 @@ func processRoomEvent(db RoomEventDatabase, input api.InputRoomEvent) error {
...
+	return &event.Event
+}
+
+// loadAuthEvents loads the events needed for authentication from the supplied room state.
+func loadAuthEvents(
+	db RoomEventDatabase,
+	needed gomatrixserverlib.StateNeeded,
+	state []types.StateEntry,
+) (result authEvents, err error) {
+	// Lookup the numeric IDs for the state keys
```
Kegsay: More like:

> Lookup the numeric IDs for the state keys which are needed for auth

src/github.com/matrix-org/dendrite/roomserver/input/events.go
```diff
@@ -44,3 +61,224 @@ func processRoomEvent(db RoomEventDatabase, input api.InputRoomEvent) error {
...
+	var eventStateKeys []string
+	eventStateKeys = append(eventStateKeys, needed.Member...)
+	eventStateKeys = append(eventStateKeys, needed.ThirdPartyInvite...)
+	stateKeyNIDs, err := db.EventStateKeyNIDs(eventStateKeys)
```
Kegsay: We have 2 lots of state key NIDs flying around now:
 - The ones needed for auth (this line)
 - The ones pulled from the auth event IDs

Please be specific in your naming as to which you are talking about. In this case `neededStateKeyNIDs` may be a better name.

src/github.com/matrix-org/dendrite/roomserver/input/events.go
```diff
@@ -44,3 +61,224 @@ func processRoomEvent(db RoomEventDatabase, input api.InputRoomEvent) error {
...
+	if err != nil {
+		return
+	}
+	result.stateNIDMap = newIDMap(stateKeyNIDs)
```
Kegsay: It's not obvious if this is mapping NID -> String or String -> NID. A one-liner comment would be good (or as we discussed IRL, drop `idMap` and just have the actual type in-lined: `map[string]int64` which makes it clear.

src/github.com/matrix-org/dendrite/roomserver/input/events.go
```diff
@@ -44,3 +61,224 @@ func processRoomEvent(db RoomEventDatabase, input api.InputRoomEvent) error {
...
+	// size and are controlled by us.
+	list := []types.StateEntry(m)
+	i := sort.Search(len(list), func(i int) bool {
+		return !list[i].StateKeyTuple.LessThan(stateKey)
+	})
+	if i < len(list) && list[i].StateKeyTuple == stateKey {
+		ok = true
+		eventNID = list[i].EventNID
+	}
+	return
```
Kegsay: I have no confidence that this function looks up things correctly. Please add unit tests (should be trivial).

Kegsay:
Mostly LGTM, few things I'm unclear on though.

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
NegativeMjark: It might, but I worry than I'd run into sadness when I wanted to pass slices of NIDs into database functions. (Currently I can just cast the slices to Int64Array and postgres will do the rest, but I can't cast []NID to []int64.)

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
NegativeMjark: The event type comes from gomatrixserverlib, The concept of NIDs exists only within the roomserver, so the list of NIDs doesn't really belong on the Event type.

The auth events roughly correspond to the auth_events key in the event JSON, but there exist events in the wild that are missing necessary events from the auth_events key, so we keep the concepts separate in the API so that we can fix the list for those events.

src/github.com/matrix-org/dendrite/roomserver/types/types.go
NegativeMjark: Renamed to StateKeyTuple.

src/github.com/matrix-org/dendrite/roomserver/types/types.go
NegativeMjark: Done

src/github.com/matrix-org/dendrite/roomserver/types/types.go
NegativeMjark: Good question, I've removed that comment for now.

src/github.com/matrix-org/dendrite/roomserver/types/types.go
NegativeMjark: I've removed IDPair in favour of passing a map[string]int64 around.

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
NegativeMjark: Done

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
NegativeMjark: I've renamed them to bulk.* d29e50b

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
NegativeMjark: I removed this when I removed the IDPair.

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
NegativeMjark: I know the maximum size of the list upfront, so might as well do a single allocation. I've added a comment

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
NegativeMjark: Done

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
NegativeMjark: It's necessary for the binary search, I've commented.

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
NegativeMjark: Done

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
NegativeMjark: Done

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
NegativeMjark: For binary search. Comment added.

src/github.com/matrix-org/dendrite/roomserver/input/events.go
NegativeMjark: I've moved the check inside the database.

src/github.com/matrix-org/dendrite/roomserver/input/events.go
NegativeMjark: Done

src/github.com/matrix-org/dendrite/roomserver/input/events.go
NegativeMjark: Done

src/github.com/matrix-org/dendrite/roomserver/input/events.go
NegativeMjark: I've changed some of the names now, it might be better?

src/github.com/matrix-org/dendrite/roomserver/input/events.go
NegativeMjark: Done

src/github.com/matrix-org/dendrite/roomserver/input/events.go
NegativeMjark: Done

Kegsay:
Is the information in https://github.com/matrix-org/dendrite/pull/4#discussion_r100307723 contained in the code somewhere?

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
```diff
@@ -115,3 +116,32 @@ func (d *Database) assignStateKeyNID(eventStateKey string) (int64, error) {
 	}
 	return eventStateKeyNID, nil
 }
+
+// StateEntriesForEventIDs implements input.EventDatabase
+func (d *Database) StateEntriesForEventIDs(eventIDs []string) ([]types.StateEntry, error) {
+	return d.statements.bulkSelectStateEventByIDID(eventIDs)
```
Kegsay: `bulkSelectStateEventByIDID` ?

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -240,6 +254,25 @@ func (s *statements) selectEventStateKeyNID(eventStateKey string) (eventStateKey
...
+	defer rows.Close()
+
+	result := make(map[string]int64, len(eventStateKeys))
+	for rows.Next() {
+		var stateKey string
+		var stateKeyNID int64
+		if err := rows.Scan(&stateKey, &stateKeyNID); err != nil {
+			return nil, err
+		}
+		result[stateKey] = stateKeyNID
```
Kegsay: What should we do if `result[stateKey]` already exists in `results`? Panic? Log? Ignore? It feels like we've got ourselves into a right state if the array of tuples contains dupes?

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -326,20 +369,58 @@ func (s *statements) prepareEvents(db *sql.DB) (err error) {
...
 ) (eventNID int64, err error) {
 	err = s.insertEventStmt.QueryRow(
 		roomNID, eventTypeNID, eventStateKeyNID, eventID, referenceSHA256,
+		pq.Int64Array(authEventNIDs),
 	).Scan(&eventNID)
 	return
 }
 
+func (s *statements) bulkSelectStateEventByIDID(eventIDs []string) ([]types.StateEntry, error) {
+	rows, err := s.bulkSelectStateEventByIDIDStmt.Query(pq.StringArray(eventIDs))
```
Kegsay: `IDID` c/p fail

src/github.com/matrix-org/dendrite/roomserver/input/events_test.go
```diff
@@ -0,0 +1,108 @@
...
+		inputStateKey int64
+		wantOK        bool
+		wantEventNID  int64
+	}{
+		{0, 0, false, 0},
+		{1, 1, true, 1},
+		{1, 3, true, 2},
+		{2, 1, true, 3},
+		{1, 2, false, 0},
+		{3, 1, false, 0},
```
Kegsay: Please comment why the values of `wantOK` and `wantEventNID` are the values mentioned here like you did on previous tests you've done in Go. I have no idea why `3,1` is not valid for example...

src/github.com/matrix-org/dendrite/roomserver/input/events_test.go
```diff
@@ -0,0 +1,108 @@
...
+		{0, false, nil},
+		{1, true, &events[0]},
+		{2, true, &events[1]},
+		{3, true, &events[2]},
+		{4, false, nil},
+		{5, true, &events[3]},
+		{6, false, nil},
+		{7, false, nil},
+		{8, true, &events[4]},
+		{9, false, nil},
```
Kegsay: Please comment what each test case is testing so we know why the output is expected to be XYZ.

src/github.com/matrix-org/dendrite/roomserver/input/events.go
```diff
@@ -2,12 +2,24 @@ package input
...
 type RoomEventDatabase interface {
-	StoreEvent(event gomatrixserverlib.Event) error
+	// Stores a matrix room event in the database
+	StoreEvent(event gomatrixserverlib.Event, authEventNIDs []int64) error
+	// Lookup the state entries for a list of string event IDs
+	// Returns a sorted list of state entries.
+	StateEntriesForEventIDs(eventIDs []string) ([]types.StateEntry, error)
+	// Lookup the numeric IDs for a list of string event state keys.
+	// Returns a sorted list of state entries.
+	EventStateKeyNIDs(eventStateKeys []string) (map[string]int64, error)
```
Kegsay: Return comment lies.

src/github.com/matrix-org/dendrite/roomserver/input/events.go
```diff
@@ -2,12 +2,24 @@ package input
...
+	// Lookup the Events for a list of numeric event IDs.
+	// Returns a sorted list of state entries.
+	Events(eventNIDs []int64) ([]types.Event, error)
```
Kegsay: Return comment lies.

Kegsay: You also don't explain when an error is returned, notably when there was NID which couldn't be converted into an Event, does that count as an error?

src/github.com/matrix-org/dendrite/roomserver/input/events.go
Kegsay: Also, when is an error returned? Obviously for sql errors, but what if the string cannot be mapped to a nid?

src/github.com/matrix-org/dendrite/roomserver/input/events.go
```diff
@@ -2,12 +2,24 @@ package input
...
 )
 
 // A RoomEventDatabase has the storage APIs needed to store a room event.
 type RoomEventDatabase interface {
-	StoreEvent(event gomatrixserverlib.Event) error
+	// Stores a matrix room event in the database
+	StoreEvent(event gomatrixserverlib.Event, authEventNIDs []int64) error
+	// Lookup the state entries for a list of string event IDs
+	// Returns a sorted list of state entries.
+	StateEntriesForEventIDs(eventIDs []string) ([]types.StateEntry, error)
```
Kegsay: When is an error returned?

src/github.com/matrix-org/dendrite/roomserver/input/events.go
```diff
@@ -17,12 +29,16 @@ func processRoomEvent(db RoomEventDatabase, input api.InputRoomEvent) error {
 		return err
 	}
 
-	if err := db.StoreEvent(event); err != nil {
+	// Check that the event passes authentication checks.
```
Kegsay: Technically it isn't just that, it also then converts those auth event IDs into auth event NIDs, right?

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
NegativeMjark: FIxed d59db7e

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
NegativeMjark: Fixed d59db7e

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
NegativeMjark: That can't happen cause of the UNIQUE constraint.

Kegsay:
Basically LGTM, needs a few more comments on key areas to explain to the reader why you're doing the things you're doing, but no functional changes are required.

src/github.com/matrix-org/dendrite/roomserver/input/events.go
NegativeMjark: Fixed 43846a0

src/github.com/matrix-org/dendrite/roomserver/input/events.go
NegativeMjark: Fixed 43846a0

src/github.com/matrix-org/dendrite/roomserver/input/events.go
NegativeMjark: Fixed 78606ed

src/github.com/matrix-org/dendrite/roomserver/input/events.go
NegativeMjark: Fixed ead2a80

src/github.com/matrix-org/dendrite/roomserver/input/events_test.go
NegativeMjark: Fixed 1ef475c

src/github.com/matrix-org/dendrite/roomserver/input/events_test.go
NegativeMjark: Fixed 1ef475c

NegativeMjark:
@Kegsay PTAL

Kegsay:
LGTM 🎉

PR #6, opened 2017-02-13
Calculate and store the state at each event

NegativeMjark:


src/github.com/matrix-org/dendrite/roomserver/input/events.go
```diff
@@ -21,6 +21,21 @@ type RoomEventDatabase interface {
 	// Lookup the Events for a list of numeric event IDs.
 	// Returns a sorted list of events.
 	Events(eventNIDs []int64) ([]types.Event, error)
+	// Lookup the state of a room at each event for a list of string event IDs.
+	// Returns a sorted list of state at each event.
+	// Returns an error if there is an error talking to the database
+	// or if the room state for the event IDs aren't in the database
+	StateAtEventIDs(eventIDs []string) ([]types.StateAtEvent, error)
```
Kegsay: Is `len(eventIDs) == len([]types.StateAtEvent)`? If so, I assume we're correlating based on position in the array. This makes me sad.

NegativeMjark: We don't give a way to map the string IDs to the corresponding bits of returned state. Since the call-site doesn't actually care about that mapping.

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -333,6 +340,13 @@ CREATE TABLE IF NOT EXISTS events (
     -- Local numeric ID for the state_key of the event
     -- This is 0 if the event is not a state event.
     event_state_key_nid BIGINT NOT NULL,
+	-- Local numeric ID for the state at the event.
```
Kegsay: Tab.

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -333,6 +340,13 @@ CREATE TABLE IF NOT EXISTS events (
...
+    -- This is 0 if we don't know the state at the event.
+    -- If the state is not 0 this this event is part of the contiguous
```
Kegsay: > this this

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -333,6 +340,13 @@ CREATE TABLE IF NOT EXISTS events (
...
+    -- part of the event graph
+    -- Since many different events can have the same state we store the
+    -- state into a separate state table and refer to it by numeric ID.
+    state_nid bigint NOT NULL DEFAULT 0
```
Kegsay: Naming: I don't like the name `state_nid` because it's not obvious what it conceptually represents in Matrix-land: a state event, state tuple, state key? Based on your comments, it looks like this represents the room state at a particular point in time, so perhaps `state_snapshot_id`?

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -372,6 +394,12 @@ func (s *statements) prepareEvents(db *sql.DB) (err error) {
 	if s.bulkSelectStateEventByIDStmt, err = db.Prepare(bulkSelectStateEventByIDSQL); err != nil {
 		return
 	}
+	if s.bulkSelectStateAtEventByIDStmt, err = db.Prepare(bulkSelectStateAtEventByIDSQL); err != nil {
```
Kegsay: Random thought: Why aren't we inlining the SQL here? It's getting rather tedious to see `selectFooByBarSQL` and `selectFooByBarSQLStmt` when really we just need the `Stmt`. I don't think pulling them out to `consts` buys us anything, I don't find it more or less readable (just annoying as I have to CTRL+F the variable name to see what the actual SQL is).

WDYT?

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -361,6 +375,14 @@ const bulkSelectStateEventByIDSQL = "" +
 	" WHERE event_id = ANY($1)" +
 	" ORDER BY event_type_nid, event_state_key_nid ASC"
 
+const bulkSelectStateAtEventByIDSQL = "" +
+	"SELECT event_type_nid, event_state_key_nid, event_nid, state_nid FROM events" +
+	" WHERE event_id = ANY($1)" +
+	" ORDER BY event_type_nid, event_state_key_nid ASC"
```
Kegsay: Why ASC? Why not DESC or unsorted?

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
```diff
@@ -494,3 +555,182 @@ func (s *statements) bulkSelectEventJSON(eventNIDs []int64) ([]eventJSONPair, er
...
+`
+
+const insertStateSQL = "" +
+	"INSERT INTO state (room_nid, state_data_nids)" +
+	" VALUES ($1, $2)" +
+	" RETURNING state_nid"
+
+const bulkSelectStateDataNIDsSQL = "" +
+	"SELECT state_nid, state_data_nids FROM state" +
+	" WHERE state_nid = ANY($1) ORDER BY state_nid"
```
Kegsay: This ORDER BY is going to be used for clobbering right? Maybe? Shame I don't know the significance of it because there is no comment.

src/github.com/matrix-org/dendrite/roomserver/types/types.go
```diff
@@ -75,3 +83,15 @@ const (
 	// EmptyStateKeyNID is the numeric ID for the empty state key.
 	EmptyStateKeyNID = 1
 )
+
+// StateDataNIDList is used to return the result of bulk StateDataNID lookups from the database.
+type StateDataNIDList struct {
```
Kegsay: It feels like we do this a lot: define a struct simply to represent a table schema. I think we should be more upfront with this in the type names, so people don't think there is a semantic Matrixy meaning to these structs. Some convention like appending `Result` (as in, Result Set) may be an idea? E.g `StateDataNIDListResult`

WDYT?

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
Kegsay: From RL: It actually looks surprisingly bad when it's inlined. You end up having to parse SQL and Go together and it really didn't actually make things look easier to read.

We should split up the tables into different files though (but not in this PR).

src/github.com/matrix-org/dendrite/roomserver/types/types.go
NegativeMjark: I thought that the comment was sufficient here? (I'd like to keep "Result" available for when we start adding query APIs to the /api package that are used by other dendrite components to query room information from the room server)

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
NegativeMjark: Fixed 854c98d

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
NegativeMjark: Fixed 0b68ed7

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
NegativeMjark: Fixed 91b10ef

src/github.com/matrix-org/dendrite/roomserver/storage/sql.go
NegativeMjark: I've added comments in 2333384

NegativeMjark:
@Kegsay PTAL

src/github.com/matrix-org/dendrite/roomserver/types/types.go
Kegsay: The comment does indeed explain it, but that's not the point. The point is to try to make it obvious when it is used (and you only have the type to go by) that this is just a convenience struct rather than something which we may want to pass through to other APIs.

I'm open to other naming suggestions.

src/github.com/matrix-org/dendrite/roomserver/input/events.go
```diff
@@ -9,18 +9,33 @@ import (
...
 	// Lookup the numeric IDs for a list of string event state keys.
 	// Returns a map from string state key to numeric ID for the state key.
-	EventStateKeyNIDs(eventStateKeys []string) (map[string]int64, error)
+	EventStateKeyNIDs(eventStateKeys []string) (map[string]types.EventStateKeyNID, error)
 	// Lookup the Events for a list of numeric event IDs.
 	// Returns a sorted list of events.
-	Events(eventNIDs []int64) ([]types.Event, error)
+	Events(eventNIDs []types.EventNID) ([]types.Event, error)
+	// Lookup the state of a room at each event for a list of string event IDs.
+	// Returns a sorted list of state at each event.
```
Kegsay: What are they sorted by? What is the significance of this?

src/github.com/matrix-org/dendrite/roomserver/input/events.go
```diff
@@ -9,18 +9,33 @@ import (
...
+	// Returns an error if there is an error talking to the database
+	// or if the room state for the event IDs aren't in the database
+	StateAtEventIDs(eventIDs []string) ([]types.StateAtEvent, error)
+	// Lookup the numeric state data IDs for the each numeric state ID
+	// The returned slice is sorted by numeric state ID.
+	StateDataNIDs(stateNIDs []types.StateSnapshotNID) ([]types.StateDataNIDList, error)
```
Kegsay: Comments are wrong now right? You mean "state snapshot ID" not "state ID".

Kegsay: Also saying "sorted by numeric state ID" is vague: Sorted by the order they are in the array, or sorted by the actual value of each NID?

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
```diff
@@ -38,47 +38,61 @@ func (d *Database) SetPartitionOffset(topic string, partition int32, offset int6
...
+	}
+
+	return roomNID, types.StateAtEvent{
+		BeforeStateSnapshotNID: stateNID,
+		StateEntry: types.StateEntry{
+			StateKeyTuple: types.StateKeyTuple{
+				EventTypeNID:     eventTypeNID,
+				EventStateKeyNID: eventStateKeyNID,
+			},
+			EventNID: eventNID,
```
Kegsay: I dislike how even non-state events will have a state key tuple (just with a `EventStateKeyNID` of 0). I'd like for this to be clearer, but that can wait for another PR.

src/github.com/matrix-org/dendrite/roomserver/input/state.go
```diff
@@ -0,0 +1,258 @@
...
+	"github.com/matrix-org/dendrite/roomserver/types"
+	"github.com/matrix-org/gomatrixserverlib"
+	"sort"
+)
+
+func calculateAndStoreState(
+	db RoomEventDatabase, event gomatrixserverlib.Event, roomNID types.RoomNID, stateEventIDs []string,
+) (types.StateSnapshotNID, error) {
+	if stateEventIDs != nil {
+		// 1) We've been told what the state at the event is.
```
Kegsay: What do these numbers correspond to?

src/github.com/matrix-org/dendrite/roomserver/input/state.go
```diff
@@ -0,0 +1,258 @@
...
+				roomNID, stateDataNIDs, []types.StateEntry{prevState.StateEntry},
+			)
+		}
+		// If there are too many deltas then we need to calculate the full state
+		// So fall through to calculateAndStoreStateMany
+	}
+	return calculateAndStoreStateMany(db, roomNID, prevStates)
+}
+
+const maxStateDataNIDs = 64
```
Kegsay: Rationale for 64? Why not 32? Why not 128?

src/github.com/matrix-org/dendrite/roomserver/input/state.go
```diff
@@ -0,0 +1,258 @@
...
+		// 6) There weren't any conflicts
+		state = combined
+	}
+
+	// TODO: Check if we can encode the new state as a delta against the
+	// previous state.
+	return db.AddState(roomNID, nil, state)
+}
+
+func resolveConflicts(db RoomEventDatabase, combinded, conflicted []types.StateEntry) ([]types.StateEntry, error) {
```
Kegsay: > combinded

src/github.com/matrix-org/dendrite/roomserver/input/state.go
```diff
@@ -0,0 +1,258 @@
...
+	panic(fmt.Errorf("Not implemented"))
+}
+
+func duplicateStateKeys(a []types.StateEntry) []types.StateEntry {
```
Kegsay: Naming: Is this finding duplicates or duplicating state keys? Perhaps `findDuplicateStateKeys` would be better.

src/github.com/matrix-org/dendrite/roomserver/input/state.go
```diff
@@ -0,0 +1,258 @@
...
+	var result []types.StateEntry
+	j := 0
+	for i := 1; i < len(a); i++ {
+		if a[j].StateKeyTuple != a[i].StateKeyTuple {
+			result = append(result, a[j:i]...)
```
Kegsay: Explain why you're doing `a[j:i]...`

src/github.com/matrix-org/dendrite/roomserver/input/state.go
```diff
@@ -0,0 +1,258 @@
...
+			j = i
+		}
+	}
+	if j != len(a)-1 {
+		result = append(result, a[j:]...)
```
Kegsay: Why do we append the rest of the elements if the later elements have `==` `StateKeyTuples`?

src/github.com/matrix-org/dendrite/roomserver/input/state.go
```diff
@@ -0,0 +1,258 @@
...
+func (s stateDataNIDSorter) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
+
+func uniqueStateDataNIDs(nids []types.StateDataNID) []types.StateDataNID {
+	sort.Sort(stateDataNIDSorter(nids))
+	return nids[:unique(stateDataNIDSorter(nids))]
+}
+
+// Remove duplicate items from a sorted list.
+// Takes the same interface as sort.Sort
+// Returns the length of the date without duplicates
```
Kegsay: > date

src/github.com/matrix-org/dendrite/roomserver/input/state.go
```diff
@@ -0,0 +1,258 @@
...
+func (s stateDataNIDSorter) Len() int           { return len(s) }
+func (s stateDataNIDSorter) Less(i, j int) bool { return s[i] < s[j] }
+func (s stateDataNIDSorter) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
+
+func uniqueStateDataNIDs(nids []types.StateDataNID) []types.StateDataNID {
+	sort.Sort(stateDataNIDSorter(nids))
+	return nids[:unique(stateDataNIDSorter(nids))]
+}
+
+// Remove duplicate items from a sorted list.
```
Kegsay: I'm failing to see where the removal is done in this code.

src/github.com/matrix-org/dendrite/roomserver/input/state.go
```diff
@@ -0,0 +1,258 @@
...
+
+import (
+	"fmt"
+	"github.com/matrix-org/dendrite/roomserver/types"
+	"github.com/matrix-org/gomatrixserverlib"
+	"sort"
+)
+
+func calculateAndStoreState(
+	db RoomEventDatabase, event gomatrixserverlib.Event, roomNID types.RoomNID, stateEventIDs []string,
```
Kegsay: Comment what `stateEventIDs` are. This is `prev_events`? What state events must be supplied in order for this function to work correctly?

src/github.com/matrix-org/dendrite/roomserver/input/state.go
```diff
@@ -0,0 +1,258 @@
...
+const maxStateDataNIDs = 64
+
+func calculateAndStoreStateMany(db RoomEventDatabase, roomNID types.RoomNID, prevStates []types.StateAtEvent) (types.StateSnapshotNID, error) {
+	// Conflict resolution.
+	// First stage: load the state datablocks for the prev events.
+	stateNIDs := make([]types.StateSnapshotNID, len(prevStates))
+	for i, state := range prevStates {
+		stateNIDs[i] = state.BeforeStateSnapshotNID
+	}
+	stateDataNIDLists, err := db.StateDataNIDs(uniqueStateSnapshotNIDs(stateNIDs))
```
Kegsay: You need to explain this more. You're calling `uniqueStateSnapshotNIDs` presumably because the previous `StateAtEvents` may have been non-state events and hence have the same snapshot NID.

src/github.com/matrix-org/dendrite/roomserver/input/state.go
```diff
@@ -0,0 +1,258 @@
...
+	if err != nil {
+		return 0, err
+	}
+
+	var stateDataNIDs []types.StateDataNID
+	for _, list := range stateDataNIDLists {
+		stateDataNIDs = append(stateDataNIDs, list.StateDataNIDs...)
```
Kegsay: You need to explain this more. You're presumably concatting the entire state data blocks together to form a complete picture of the room state: is this consistent? (You're using `stateDataNIDLists` which was returned by querying potentially multiple distinct state snapshots, so isn't there a time component here which we're completely ignoring?).

Kegsay: Oh but we don't care about the entire array because we never loop over it (we loop over `prevStates`), we're just loading this all up in case we need to look up the `StateEntry` for it?

src/github.com/matrix-org/dendrite/roomserver/types/types.go
```diff
@@ -13,13 +13,32 @@ type PartitionOffset struct {
...
+
+// RoomNID is a numeric ID for a room.
+type RoomNID int64
+
+// StateSnapshotNID is a numeric ID for the state at an event.
+type StateSnapshotNID int64
+
+// StateDataNID is a numeric ID for a block of state data.
+// These blocks of state data are combined to form the actual state.
+type StateDataNID int64
```
Kegsay: Naming: I really hate the use of the term "data" - it's such an overloaded term. Your comment mentions what I think it really should be called: `StateBlockNID` - WDYT about renaming this for clarity? My intention is to try to be able to disambiguate clearly whether you're dealing with State:
 - Blocks
 - Snapshots
 - Entries (which is also a term I dislike, because everything is an entry in a database!)

src/github.com/matrix-org/dendrite/roomserver/input/state.go
```diff
@@ -0,0 +1,258 @@
...
+	stateEntryLists, err := db.StateEntries(uniqueStateDataNIDs(stateDataNIDs))
+	if err != nil {
+		return 0, err
+	}
+	stateDataNIDsMap := stateDataNIDListMap(stateDataNIDLists)
+	stateEntriesMap := stateEntryListMap(stateEntryLists)
+
+	var combined []types.StateEntry
+	for _, prevState := range prevStates {
+		list, ok := stateDataNIDsMap.lookup(prevState.BeforeStateSnapshotNID)
```
Kegsay: RAWR! `list` is possibly the most unhelpful variable name ever. It's a list of WHAT (how many lists does this function deal with, be clearer!).

src/github.com/matrix-org/dendrite/roomserver/input/state.go
```diff
@@ -0,0 +1,258 @@
...
+			entries, ok := stateEntriesMap.lookup(stateDataNID)
+			if !ok {
+				// This should only get hit if the database is corrupt.
+				// It should be impossible for an event to reference a NID that doesn't exist
+				panic(fmt.Errorf("Corrupt DB: Missing state numeric ID %d", prevState.BeforeStateSnapshotNID))
+			}
+			fullState = append(fullState, entries...)
+		}
+		if prevState.EventStateKeyNID != 0 {
+			fullState = append(fullState, prevState.StateEntry)
```
Kegsay: Come on, explain why you're doing this: you want the full state *after* the event in `prev_events` but *before* the event being added in this function, amirite? It's really unclear that `prevState.EventStateKeyNID != 0` really means `prevState.IsStateEvent()` - A tiny method on `types.StateEntry` would go far in improving clarity to prevent me from nagging you to comment every time you do this otherwise.

src/github.com/matrix-org/dendrite/roomserver/input/state.go
```diff
@@ -0,0 +1,258 @@
...
+		}
+
+		// Stable sort so that the most recent entry for each state key stays
+		// remains later in the list than the older entries for the same state key.
+		sort.Stable(stateEntryByStateKeySorter(fullState))
+		// Unique returns the last entry for each state key.
```
Kegsay: "and hence the most recent entry"

src/github.com/matrix-org/dendrite/roomserver/input/state.go
```diff
@@ -0,0 +1,258 @@
...
+		fullState = fullState[:unique(stateEntryByStateKeySorter(fullState))]
+		// Add the full state for this StateSnapshotNID.
+		combined = append(combined, fullState...)
+	}
+
+	// Collect all the entries with the same type and key together.
+	// We don't care about the order here.
```
Kegsay: Why not? Because this is to do with the list of `prev_events` and we don't really care if we deal with one parent branch before another?

src/github.com/matrix-org/dendrite/roomserver/input/events.go
NegativeMjark: Fixed 96096bd

src/github.com/matrix-org/dendrite/roomserver/input/events.go
NegativeMjark: Fixed 8b71dd7

src/github.com/matrix-org/dendrite/roomserver/input/state.go
NegativeMjark: They are left over from when I was enumerating the cases I wanted to handle. In a later PR I'd like to go back and hook up metrics for each of the cases.

src/github.com/matrix-org/dendrite/roomserver/input/state.go
NegativeMjark: Done 67b261d

src/github.com/matrix-org/dendrite/roomserver/input/state.go
NegativeMjark: Done 7fc31ee

src/github.com/matrix-org/dendrite/roomserver/input/state.go
NegativeMjark: Done bed16c0

src/github.com/matrix-org/dendrite/roomserver/input/state.go
NegativeMjark: ee1b5ee might help

src/github.com/matrix-org/dendrite/roomserver/input/state.go
NegativeMjark: ee1b5ee might help

src/github.com/matrix-org/dendrite/roomserver/input/state.go
NegativeMjark: ee1b5ee might help

src/github.com/matrix-org/dendrite/roomserver/input/state.go
NegativeMjark: Done ee1b5ee

src/github.com/matrix-org/dendrite/roomserver/input/state.go
NegativeMjark: 548ba1fd might make this clearer.

src/github.com/matrix-org/dendrite/roomserver/input/state.go
NegativeMjark: 632162f9 might help

src/github.com/matrix-org/dendrite/roomserver/input/state.go
NegativeMjark: 632162f might help

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
NegativeMjark: I'd be interested in what suggestions you have for making this clearer. But hopefully it can wait for another PR.

src/github.com/matrix-org/dendrite/roomserver/types/types.go
NegativeMjark: Done 437f298

I think Entry is fairly common jargon for a key value pair in a map. So I'm inclined to keep that one.

src/github.com/matrix-org/dendrite/roomserver/input/state.go
NegativeMjark: Done 8526277

src/github.com/matrix-org/dendrite/roomserver/input/state.go
NegativeMjark: 2a33b17

src/github.com/matrix-org/dendrite/roomserver/input/state.go
NegativeMjark: Done d2952f48

src/github.com/matrix-org/dendrite/roomserver/input/state.go
NegativeMjark: Done 1f3d2c6

src/github.com/matrix-org/dendrite/roomserver/input/state.go
Kegsay: Yup it does.

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
Kegsay: It can.

src/github.com/matrix-org/dendrite/roomserver/input/state.go
```diff
@@ -0,0 +1,292 @@
...
+
+// findDuplicateStateKeys finds the state entries where the state key tuple appears more than once in a sorted list.
+// Returns a sorted list of those state entries.
+func findDuplicateStateKeys(a []types.StateEntry) []types.StateEntry {
+	var result []types.StateEntry
+	// j is the starting index of a block of entries with the same state key tuple.
+	j := 0
+	for i := 1; i < len(a); i++ {
+		// Check if the state key tuple matches the start of the block
+		if a[j].StateKeyTuple != a[i].StateKeyTuple {
```
Kegsay: I think I know why I don't find this clear. You're doing the inverse to what is obvious. We have a sorted list of things:
```
[ 1, 1, 2, 4, 5, 5, 5, 6 ]
```
and we want to return the sorted duplicates:
```
[1, 5, 5]
```

But instead of doing "if the previous element is the same as the current element, append to the result" you're "skipping over" the duplicates and then batch appending the duplicates. I can only imagine you're doing this for speed, to which I say: clarity is more important. If we profile this code later and find it to be a bottleneck, THEN we can optimise it.

Kegsay:
I've mentioned a lot of things I'd like for us to do, but the only ones which I feel are outstanding for this PR are:
 - UTs for `unique()`
 - Addressing https://github.com/matrix-org/dendrite/pull/6/files#r101097187

NegativeMjark:
>  I dislike State Entry NID: It's just a 3-uple of state key / event type / event ID, and I would love to see it renamed to something that is less overloaded than "entry".

I think Entry is acceptable jargon for a key-value pair in a map.

> Docs (in the README):

> We have a nice model of how room state is stored with clear relationships. An ASCII diagram in the README would be lovely.

I've made a start with 93b7a26

> Function grouping:
> 
> Look at calculateAndStoreStateMany. How much of this function is dedicated to retrieving and combining state together for each prev_event? Most of it. The meat of the conflict resolution only starts at :128 when you "collect all the entries with the same type and key together", and then do magic sorts to find the duplicates (and hence conflicts). I would love to see these two concepts split out more:

Hmm. The way the retrieval code loads the data is somewhat tied up with how the conflict code processes it. I don't think the shape of the returned data is useful outside of the conflict detection/resolution code.

NegativeMjark:
>> Function grouping:
>> 
>> Look at calculateAndStoreStateMany. How much of this function is dedicated to retrieving and combining state together for each prev_event? Most of it. The meat of the conflict resolution only starts at :128 when you "collect all the entries with the same type and key together", and then do magic sorts to find the duplicates (and hence conflicts). I would love to see these two concepts split out more:

> Hmm. The way the retrieval code loads the data is somewhat tied up with how the conflict code processes it. I don't think the shape of the returned data is useful outside of the conflict detection/resolution code.

Eh I split it out anyway and it does seem a little clearer to follow 2aec67d

NegativeMjark:
@Kegsay PTAL

src/github.com/matrix-org/dendrite/roomserver/input/state.go
NegativeMjark: Actually I want it to return `[1,1,5,5,5]`.

Kegsay: Case in point: I don't understand this function.

Kegsay: Okay, I now understand this and I think the following comments would've helped:

> ~~Check if the state key tuple matches the start of the block.~~ `i` is incremented each time a duplicate tuple is found. If the tuple doesn't match the start of the block (j) then we've reached the end of the block and need to add the block to the `result`.

> Check if the size of the block is bigger than one. **If it is one, then this is just a single unique entry so don't add it to the result.**

> Check if the last block had more than one event in it. **We won't have added it to `result` yet because we only do that when we find a non-match.**

Kegsay:
Change the comments outlined in https://github.com/matrix-org/dendrite/pull/6/files#r101236125 then LGTM! \o/ you got there!

src/github.com/matrix-org/dendrite/roomserver/input/state.go
NegativeMjark: Done with slightly different wording 4759f01

Kegsay:
LGTM

PR #11, opened 2017-02-17
Use gorilla/mux to route HTTP requests

Kegsay:
- Make `/sync` and `/send` use the right API paths.
- Pluck out the `roomID` as an example of how to do it.

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/routing/routing.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/routing/routing.go
@@ -12,18 +12,39 @@ import (
 
 const pathPrefixR0 = "/_matrix/client/r0"
 
-func make(metricsName string, h util.JSONRequestHandler) http.Handler {
-	return prometheus.InstrumentHandler(metricsName, util.MakeJSONAPI(h))
-}
-
 // Setup registers HTTP handlers with the given ServeMux. It also supplies the given http.Client
 // to clients which need to make outbound HTTP requests.
 func Setup(servMux *http.ServeMux, httpClient *http.Client) {
 	apiMux := mux.NewRouter()
 	r0mux := apiMux.PathPrefix(pathPrefixR0).Subrouter()
 	r0mux.Handle("/sync", make("sync", &readers.Sync{}))
-	r0mux.Handle("/rooms/{roomID}/send/{eventType}", make("send_message", &writers.SendMessage{}))
+	r0mux.Handle("/rooms/{roomID}/send/{eventType}",
+		make("send_message", wrap(func(req *http.Request) (interface{}, *util.HTTPError) {
+			vars := mux.Vars(req)
+			roomID := vars["roomID"]
+			eventType := vars["eventType"]
+			h := &writers.SendMessage{}
+			return h.OnIncomingRequest(req, roomID, eventType)
+		})),
+	)
 
 	servMux.Handle("/metrics", prometheus.Handler())
 	servMux.Handle("/api/", http.StripPrefix("/api", apiMux))
+}
+
+// make a util.JSONRequestHandler into an http.Handler
+func make(metricsName string, h util.JSONRequestHandler) http.Handler {
+	return prometheus.InstrumentHandler(metricsName, util.MakeJSONAPI(h))
+}
+
+// jsonRequestHandlerWrapper is a wrapper to allow in-line functions to conform to util.JSONRequestHandler
+type jsonRequestHandlerWrapper struct {
+	function func(req *http.Request) (interface{}, *util.HTTPError)
+}
+
+func (r *jsonRequestHandlerWrapper) OnIncomingRequest(req *http.Request) (interface{}, *util.HTTPError) {
+	return r.function(req)
+}
+func wrap(f func(req *http.Request) (interface{}, *util.HTTPError)) *jsonRequestHandlerWrapper {
+	return &jsonRequestHandlerWrapper{f}
 }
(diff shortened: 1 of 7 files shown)


src/github.com/matrix-org/dendrite/clientapi/writers/sendmessage.go
```diff
@@ -14,7 +15,9 @@ type SendMessage struct {
 // OnIncomingRequest implements util.JSONRequestHandler
 func (s *SendMessage) OnIncomingRequest(req *http.Request) (interface{}, *util.HTTPError) {
 	logger := req.Context().Value(util.CtxValueLogger).(*log.Entry)
```
NegativeMjark: I realise this isn't part of the PR but would it make sense create a type safe accessors for the logger as per the suggestions in https://golang.org/pkg/context/

src/github.com/matrix-org/dendrite/clientapi/writers/sendmessage.go
```diff
@@ -14,7 +15,9 @@ type SendMessage struct {
...
-	logger.Info("Doing stuff...")
+	vars := mux.Vars(req)
+	roomID := vars["roomID"]
+	logger.WithField("roomID", roomID).Info("Doing stuff...")
```
NegativeMjark: You should also grab the eventType from vars here.

NegativeMjark: The implementation of `mux.Vars` is "interesting". There should probably be a comment saying that this handler has to be called through a gorilla Mux in order for it to have stashed the route vars on the request context.

src/github.com/matrix-org/dendrite/clientapi/writers/sendmessage.go
Kegsay: Failing to see where that is mentioned.

src/github.com/matrix-org/dendrite/clientapi/routing/routing.go
```diff
@@ -0,0 +1,29 @@
...
+// Setup registers HTTP handlers with the given ServeMux. It also supplies the given http.Client
+// to clients which need to make outbound HTTP requests.
+func Setup(servMux *http.ServeMux, httpClient *http.Client) {
+	apiMux := mux.NewRouter()
+	r0mux := apiMux.PathPrefix(pathPrefixR0).Subrouter()
+	r0mux.Handle("/sync", make("sync", &readers.Sync{}))
+	r0mux.Handle("/rooms/{roomID}/send/{eventType}", make("send_message", &writers.SendMessage{}))
+
+	servMux.Handle("/metrics", prometheus.Handler())
+	servMux.Handle("/api/", http.StripPrefix("/api", apiMux))
```
NegativeMjark: I like having the paths all in one place. It makes it much easier to see where a request is going to be handled. :+1:

src/github.com/matrix-org/dendrite/clientapi/writers/sendmessage.go
Kegsay: It is "interesting" and I'm wondering how best to handle that going forwards. In all practicalities we WILL be using a gorilla mux so I don't feel inclined to doc this every time we pluck something out of vars.

I dislike how separated the path pattern is from where we access things, but I like how we can easily see the endpoints available in routing.go

src/github.com/matrix-org/dendrite/clientapi/writers/sendmessage.go
NegativeMjark:
```go
     // Packages that define a Context key should provide type-safe accessors
     // for the values stored using that key:
```

Kegsay: Oh okay I'll add it to the util package then.

Kegsay: https://github.com/matrix-org/util/pull/2

src/github.com/matrix-org/dendrite/clientapi/writers/sendmessage.go
Kegsay: > You should also grab the eventType from vars here.

Done.

Kegsay:
@NegativeMjark PTAL

Kegsay:
@NegativeMjark PTAL

src/github.com/matrix-org/dendrite/clientapi/routing/routing.go
```diff
@@ -0,0 +1,51 @@
...
+func Setup(servMux *http.ServeMux, httpClient *http.Client) {
+	apiMux := mux.NewRouter()
+	r0mux := apiMux.PathPrefix(pathPrefixR0).Subrouter()
+	r0mux.Handle("/sync", make("sync", wrap(func(req *http.Request) (interface{}, *util.HTTPError) {
+		return readers.Sync(req)
+	})))
+	r0mux.Handle("/rooms/{roomID}/send/{eventType}",
+		make("send_message", wrap(func(req *http.Request) (interface{}, *util.HTTPError) {
+			vars := mux.Vars(req)
+			roomID := vars["roomID"]
```
NegativeMjark: Can we inline the vars["<key>"] here? It doesn't look like we need to store these as local variables here.

NegativeMjark:
Other than maybe inlining the vars[""]. LGTM

src/github.com/matrix-org/dendrite/clientapi/routing/routing.go
Kegsay: Done.

PR #13, opened 2017-02-20
Add error package and some Matrix errors

Kegsay:
Package name is `jsonerror` and not `errors` so as not to clash with the stdlib. Terse function names are used to make usage not stutter. For example:

```go
err := jsonerror.Forbidden("you shall not pass")
```

At the moment they are all the same fundamental `MatrixError` type. This will
be bad if we ever want to `switch` based on the kind of error. I'm hoping we
won't ever need to introspect into errors like this: ideally these errors would
be created purely for immediately being returned in an HTTP response.

The error string constants have not been yanked out into `consts` because they
are referenced in exactly one place. If we ever refer to these strings elsewhere
then we should expose them and make them `const`ified.

`MatrixError` implements the `error` interface.

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/error/error.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/error/error.go
@@ -1,0 +1,62 @@
+package error
+
+import "fmt"
+
+// MatrixError represents the "standard error response" in Matrix.
+// http://matrix.org/docs/spec/client_server/r0.2.0.html#api-standards
+type MatrixError struct {
+	ErrCode string `json:"errcode"`
+	Err     string `json:"error"`
+}
+
+func (e *MatrixError) Error() string {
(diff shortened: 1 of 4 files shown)


src/github.com/matrix-org/dendrite/clientapi/error/error.go
```diff
@@ -0,0 +1,62 @@
+package error
```
NegativeMjark: Doesn't having the package name get syntax highlighted make you sad?

NegativeMjark: More to the point how will people use this when the package name clashes with the builtin error type ?

Kegsay: But but `errors` is already in use :<

/me settles with `jsonerror`.

PR #14, opened 2017-02-21
Track which events are the latest events in the room.

NegativeMjark:


src/github.com/matrix-org/dendrite/roomserver/input/events.go
```diff
@@ -34,6 +34,10 @@ type RoomEventDatabase interface {
 	AddState(roomNID types.RoomNID, stateBlockNIDs []types.StateBlockNID, state []types.StateEntry) (types.StateSnapshotNID, error)
 	// Set the state at an event.
 	SetState(eventNID types.EventNID, stateNID types.StateSnapshotNID) error
+	// Lookup the latests events in a room in preperation for an update.
```
Kegsay: > latests

> preperation

NegativeMjark: Done c422a1f

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
```diff
@@ -195,3 +195,62 @@ func (d *Database) StateBlockNIDs(stateNIDs []types.StateSnapshotNID) ([]types.S
...
+		return nil, nil, err
+	}
+	eventNIDs, err := d.statements.selectLatestEventsNIDsForUpdate(txn, roomNID)
+	if err != nil {
+		txn.Rollback()
+		return nil, nil, err
+	}
+	stateAndRefs, err := d.statements.bulkSelectStateAtEventAndReference(txn, eventNIDs)
+	if err != nil {
+		txn.Rollback()
```
Kegsay: Do we really need to `Rollback()` if they are just SELECTs?

src/github.com/matrix-org/dendrite/roomserver/input/latest_events.go
```diff
@@ -0,0 +1,72 @@
...
+) error {
+	oldLatest, updater, err := db.GetLatestEventsForUpdate(roomNID)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if err == nil {
+			// Commit if there wasn't an error.
+			// Set the returned err value if we encounter an error committing.
+			err = updater.Close(true)
```
Kegsay: I'd prefer if we used consts/words/enums/something to explain that `true` means "please commit" and `false` means "roll back".

src/github.com/matrix-org/dendrite/roomserver/input/latest_events.go
```diff
@@ -0,0 +1,72 @@
...
+		} else {
+			// Ignore any error we get rolling back since we don't want to
+			// clobber the current error
+			updater.Close(false)
```
Kegsay: What's the worst thing that can happen if we fail to rollback? Memory leak? Connection leak? I'm wary about dropping this error to the floor if it could indicate a real problem.

src/github.com/matrix-org/dendrite/roomserver/input/latest_events.go
```diff
@@ -0,0 +1,72 @@
...
+		})
+	}
+
+	if err = updater.SetLatestEvents(roomNID, newLatest); err != nil {
+		return err
+	}
+
+	// The err should be nil at this point.
+	// But when we call Close in the defer above it might set an error here.
+	return err
```
Kegsay: Uhhh, that's not how `defer` works afaik. If this was a named return function then this would be true.

https://play.golang.org/p/rY_K2WnCTZ

src/github.com/matrix-org/dendrite/roomserver/input/latest_events.go
```diff
@@ -0,0 +1,72 @@
...
+		return err
+	}
+
+	// Check if this event references any of the latest events in the room.
+	var newLatest []types.StateAtEventAndReference
+	for _, l := range oldLatest {
+		for _, prevEvent := range prevEvents {
+			if l.EventID == prevEvent.EventID && bytes.Compare(l.EventSHA256, prevEvent.EventSHA256) == 0 {
+				// This event can be removed from the latest events cause we've found an event that references it.
+				continue
```
Kegsay: I don't follow. We got an incoming event X and it states a bunch of `prev_events`. One of those prev events we have in our database. It's still the "latest" event right? Why do you filter it out?

Kegsay: It feels more like these are "events which I can't fit into the DAG"?

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
NegativeMjark: We need to end the txn somehow, and Rollback is probably clearer than Commit here?

src/github.com/matrix-org/dendrite/roomserver/storage/events_table.go
```diff
@@ -176,3 +184,43 @@ func (s *eventStatements) updateEventState(eventNID types.EventNID, stateNID typ
...
+		var (
+			eventTypeNID     int64
+			eventStateKeyNID int64
+			eventNID         int64
+			stateSnapshotNID int64
+			eventID          string
+			eventSHA256      []byte
+		)
+		if err = rows.Scan(
+			&eventTypeNID, &eventStateKeyNID, &eventNID, &stateSnapshotNID, &eventNID, &eventSHA256,
```
Kegsay: You have `&eventNID` twice. You probably want the second one to be `$eventID`. :)

src/github.com/matrix-org/dendrite/roomserver/storage/previous_events_table.go
```diff
@@ -0,0 +1,72 @@
...
+-- The previous events table stores the event_ids referenced by the events
+-- stored in the events table.
+-- This is used to tell if a new event is already referenced by an event we 
+CREATE TABLE IF NOT EXISTS previous_events (
+	// The string event ID taken from the prev_events key of an event.
+	previous_event_id TEXT NOT NULL,
+	// The SHA256 reference hash taken from the prev_events key of an event.
+	previous_reference_sha256 BYTEA NOT NULL,
+	// A list of numeric event IDs of events that reference this prev_event.
+	// This isn'
```
Kegsay: This isn't what?! TELL MEEEEEE!!!

src/github.com/matrix-org/dendrite/roomserver/input/latest_events.go
NegativeMjark: Yeah. Not sure what I want to do with this. Clobber the original error? Create a new error holding both errors? Log something? panic?

I'm tempted to just add a log line WDYT?

src/github.com/matrix-org/dendrite/roomserver/input/latest_events.go
NegativeMjark: Done a935ef6

src/github.com/matrix-org/dendrite/roomserver/storage/previous_events_table.go
```diff
@@ -0,0 +1,72 @@
...
+
+import (
+	"database/sql"
+	"github.com/matrix-org/dendrite/roomserver/types"
+)
+
+const previousEventSchema = `
+-- The previous events table stores the event_ids referenced by the events
+-- stored in the events table.
+-- This is used to tell if a new event is already referenced by an event we
```
Kegsay: More fun incomplete sentences.

Basically, this table stores all the `prev_events` event IDs for every event ever seen ever, right?

src/github.com/matrix-org/dendrite/roomserver/storage/previous_events_table.go
```diff
@@ -0,0 +1,72 @@
...
+-- This is used to tell if a new event is already referenced by an event we 
+CREATE TABLE IF NOT EXISTS previous_events (
+	// The string event ID taken from the prev_events key of an event.
+	previous_event_id TEXT NOT NULL,
+	// The SHA256 reference hash taken from the prev_events key of an event.
+	previous_reference_sha256 BYTEA NOT NULL,
+	// A list of numeric event IDs of events that reference this prev_event.
+	// This isn'
+	event_nids BIGINT[] NOT NULL,
+	CONSTRAINT previous_event_id_unique UNIQUE (region_nid, prev_event_id, prev_reference_sha256)
```
Kegsay: wut? `prev_event_id` isn't a thing, it's `previous_event_id`. Ditto with the SHA. What's a `region_nid`?

src/github.com/matrix-org/dendrite/roomserver/storage/events_table.go
NegativeMjark: Fixed b8f2999

src/github.com/matrix-org/dendrite/roomserver/storage/previous_events_table.go
```diff
@@ -0,0 +1,72 @@
...
+);
+`
+
+// Insert an entry into the previous_events table.
+// The there is already an entry indicating that an event references that previous event then
```
Kegsay: > The there

src/github.com/matrix-org/dendrite/roomserver/storage/previous_events_table.go
Kegsay: Presumably this isn't guaranteed to contain unique entries given the `array_append` you do on `:32`.

src/github.com/matrix-org/dendrite/roomserver/storage/previous_events_table.go
NegativeMjark: Yep, Maybe fixed e7b7ed3

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
Kegsay: True true. Fair point.

src/github.com/matrix-org/dendrite/roomserver/storage/previous_events_table.go
NegativeMjark: Fixed 31faf96

src/github.com/matrix-org/dendrite/roomserver/types/types.go
```diff
@@ -79,6 +79,12 @@ func (s StateAtEvent) IsStateEvent() bool {
 	return s.EventStateKeyNID != 0
 }
 
+// StateAtEventAndReference ...
```
Kegsay: What do?!

src/github.com/matrix-org/dendrite/roomserver/types/types.go
```diff
@@ -119,3 +125,19 @@ type StateEntryList struct {
 	StateBlockNID StateBlockNID
 	StateEntries  []StateEntry
 }
+
+// A RoomRecentEventsUpdater is used to update the recent events in a room.
+// (On postgresql this wraps a database transaction that holds a "FOR UPDATE"
+//  lock on the a row holding the latest events for the room.)
```
Kegsay: > on the a row

src/github.com/matrix-org/dendrite/roomserver/storage/previous_events_table.go
NegativeMjark: Fixed  e69bc73

src/github.com/matrix-org/dendrite/roomserver/types/types.go
```diff
@@ -119,3 +125,19 @@ type StateEntryList struct {
...
+type RoomRecentEventsUpdater interface {
+	// Store the previous events referenced by an event.
+	StorePreviousEvents(eventNID EventNID, previousEventReferences []gomatrixserverlib.EventReference) error
+	// Check whether the eventReference is already referenced by another matrix event.
+	IsReferenced(eventReference gomatrixserverlib.EventReference) (bool, error)
+	// Set the list of latest events for the room.
+	SetLatestEvents(roomNID RoomNID, latest []StateAtEventAndReference) error
```
Kegsay: Naming: Is there some distinction between `StoreXXX` and `SetXXX` that you're making here? "Store" looks like it is INSERTing, and `Set` looks like it is updating, but it's completely unclear. If I try to set the same thing twice will they append? Clobber? Error out?

I don't mind exposing a commit/rollback interface like this, provided it is clear what actions can be taken, and what a compliant implementation of `RoomRecentEventsUpdater` should do.

src/github.com/matrix-org/dendrite/roomserver/types/types.go
NegativeMjark: Fixed b7f1de0

src/github.com/matrix-org/dendrite/roomserver/types/types.go
NegativeMjark: Fixed 2cce7c9

src/github.com/matrix-org/dendrite/roomserver/storage/previous_events_table.go
NegativeMjark: I can't remember what I was going to type there :(. I've removed the "This isn't" for now.

src/github.com/matrix-org/dendrite/roomserver/input/latest_events.go
NegativeMjark: Fixed 75454c5 by renaming Close to Commit and Rollback.

src/github.com/matrix-org/dendrite/roomserver/storage/rooms_table.go
```diff
@@ -11,7 +12,11 @@ CREATE TABLE IF NOT EXISTS rooms (
     -- Local numeric ID for the room.
     room_nid BIGINT PRIMARY KEY DEFAULT nextval('room_nid_seq'),
     -- Textual ID for the room.
-    room_id TEXT NOT NULL CONSTRAINT room_id_unique UNIQUE
+    room_id TEXT NOT NULL CONSTRAINT room_id_unique UNIQUE,
+	-- The most recent events in the room that aren't referenced by another event.
+	-- This list may empty if the server hasn't joined the room yet.
+	-- (The server will be in that state while it stores the events for the initial state of the room)
+	latest_event_nids BIGINT[] NOT NULL
```
Kegsay: Attack of the tabs!

NegativeMjark: Fixed bab6ee4

NegativeMjark: Actually fixed fde423e

src/github.com/matrix-org/dendrite/roomserver/input/latest_events.go
NegativeMjark: Does fe2c0be help?

src/github.com/matrix-org/dendrite/roomserver/storage/rooms_table.go
```diff
@@ -11,7 +12,11 @@ CREATE TABLE IF NOT EXISTS rooms (
     -- Local numeric ID for the room.
     room_nid BIGINT PRIMARY KEY DEFAULT nextval('room_nid_seq'),
     -- Textual ID for the room.
-    room_id TEXT NOT NULL CONSTRAINT room_id_unique UNIQUE
+    room_id TEXT NOT NULL CONSTRAINT room_id_unique UNIQUE,
+	-- The most recent events in the room that aren't referenced by another event.
```
Kegsay: Naming: Whilst I appreciate you've been consistently calling this "latest" events, I don't think it's a good name. I think of latest as:
```
Time goes down the page. 1 is the m.room.create event (root).

    1
    |
    2
   /  \
  3   4
  |   |
  5   6 <--- latest
  |
  7 <----- latest
```

whereas in actuality, it seems "latest" is anything which doesn't match up to the known state of the DAG, i.e. no known parents (orphans?):

```
    1
    |
    2
   /  \
  3   4
  |   |
  5   6
  |
  7   8 <--- latest
```

Is my thinking correct? If so, why not call these `orphan_events`?

src/github.com/matrix-org/dendrite/roomserver/types/types.go
NegativeMjark: Does f70d0d4 help?

src/github.com/matrix-org/dendrite/roomserver/storage/rooms_table.go
NegativeMjark: Your first diagram is correct.

NegativeMjark: The latest_events are the events that aren't referenced by another event.
So 5 is reference by 7 so is not a latest event.
But 7 isn't referenced by anything yet so is a latest event.

NegativeMjark: I've borrowed your ascii art c1e897c

src/github.com/matrix-org/dendrite/roomserver/types/types.go
NegativeMjark: Or would you like to see comments on how often SetLatest... and StorePrevious... are called?

src/github.com/matrix-org/dendrite/roomserver/input/latest_events.go
NegativeMjark: I've added a TODO to add the logging in 8a9dbad
I'm not doing it in this PR because I'll need to thread through a context object first.

Kegsay: SGTM.

src/github.com/matrix-org/dendrite/roomserver/input/latest_events.go
Kegsay: Yup! :star:

src/github.com/matrix-org/dendrite/roomserver/types/types.go
Kegsay: > Does f70d0d4 help?

It does. Can you do that for `StorePrevious...` please?

src/github.com/matrix-org/dendrite/roomserver/storage/rooms_table.go
Kegsay: Oh I see, so the intuition with "latest" is correct, I'm just getting really confused between "references" and "is referenced by".

src/github.com/matrix-org/dendrite/roomserver/input/latest_events.go
```diff
@@ -0,0 +1,90 @@
...
+		})
+	}
+
+	if err = updater.SetLatestEvents(roomNID, newLatest); err != nil {
+		return err
+	}
+
+	// The err should be nil at this point.
+	// But when we call Close in the defer above it might set an error here.
+	return
```
Kegsay: Just because I said you need to use a named return doesn't mean that I like that! This function is way too long to comfortably use named return args IMO.

Can we please just yank out the logic to a traditional function and then `defer` rollback a level higher? E.g:

```go
func updateLatestEvents(....) (err error) {
    defer func() {
        if err == nil {
            err = Commit()
        } else {
            Rollback()
        }
    }()

    err = doTheActualLogic(...args)
}
```

Kegsay:
- https://github.com/matrix-org/dendrite/pull/14#discussion_r102217805 needs to be addressed.
- https://github.com/matrix-org/dendrite/pull/14/files#r102219692 needs to be addressed.
- Is https://github.com/matrix-org/dendrite/pull/14#discussion_r102206983 correct? If so, comment the code please.

After that, LGTM!

src/github.com/matrix-org/dendrite/roomserver/input/latest_events.go
NegativeMjark: Done 9079f96

src/github.com/matrix-org/dendrite/roomserver/types/types.go
NegativeMjark: Done e8793ed

src/github.com/matrix-org/dendrite/roomserver/storage/previous_events_table.go
NegativeMjark: It is guaranteed to contain unique entries I think cause of the WHERE clause.

Kegsay:
LGTM!

PR #15, opened 2017-02-22
Extract access tokens from HTTP requests

Kegsay:
Doesn't verify tokens yet.


src/github.com/matrix-org/dendrite/clientapi/readers/sync.go
```diff
@@ -3,13 +3,22 @@ package readers
...
 )
 
 // Sync implements /sync
 func Sync(req *http.Request) (interface{}, *util.HTTPError) {
 	logger := util.GetLogger(req.Context())
-	logger.Info("Doing stuff...")
+	userID, err := auth.VerifyAccessToken(req)
+	if err != nil {
+		return nil, &util.HTTPError{
+			Code: 403,
```
Kegsay: I would very much like your thoughts on this. This has decoupled the error JSON and the HTTP status code. Should we be bundling them into one thing or?

src/github.com/matrix-org/dendrite/clientapi/writers/sendmessage.go
```diff
@@ -3,13 +3,26 @@ package writers
...
+	log "github.com/Sirupsen/logrus"
+	"github.com/matrix-org/dendrite/clientapi/auth"
 	"github.com/matrix-org/util"
 )
 
 // SendMessage implements /rooms/{roomID}/send/{eventType}
 func SendMessage(req *http.Request, roomID, eventType string) (interface{}, *util.HTTPError) {
 	logger := util.GetLogger(req.Context())
-	logger.WithField("roomID", roomID).WithField("eventType", eventType).Info("Doing stuff...")
+	userID, err := auth.VerifyAccessToken(req)
```
Kegsay: Should we be doing this here or in `routing.go`? It feels like maybe `routing.go` but I'm unsure.

Kegsay: From IRL: Keep it here.

src/github.com/matrix-org/dendrite/clientapi/readers/sync.go
Kegsay: From IRL: Use `util.JSONResponse` instead and combine the concepts.

src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
```diff
@@ -0,0 +1,47 @@
...
+	"net/http"
+	"strings"
+
+	"github.com/matrix-org/dendrite/clientapi/jsonerror"
+)
+
+// VerifyAccessToken verifies that an access token was supplied in the given HTTP request
+// and returns the user ID it corresponds to. Returns an error if there is no access token
+// or the token is invalid.
+func VerifyAccessToken(req *http.Request) (userID string, err error) {
```
Kegsay: From IRL: Three-value return (user id, failure JSON response, everything-went-wrong error)

Kegsay:
@NegativeMjark PTAL

src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
```diff
@@ -0,0 +1,54 @@
...
+	"strings"
+
+	"github.com/matrix-org/dendrite/clientapi/jsonerror"
+	"github.com/matrix-org/util"
+)
+
+// VerifyAccessToken verifies that an access token was supplied in the given HTTP request
+// and returns the user ID it corresponds to. Returns err if there was a fatal problem checking
+// the token. Returns resErr (an error response which can be sent to the client) if the token is invalid.
+func VerifyAccessToken(req *http.Request) (userID string, resErr *util.JSONResponse, err error) {
```
NegativeMjark: Hmm Now that I see it in the flesh the three value return is a bit cumbersome.

Maybe just return `(userID string, errResp *util.JsonResponse)` and wrap the `error`s as 500 inside `VerifyAccessToken`?

PR #19, opened 2017-02-24
Add user_ips, users, access_tokens tables

Kegsay:
Based off https://github.com/matrix-org/dendrite/pull/15

Done in the same style as `roomserver`. We should probably think about
how we will do upgrades to the schema.

The schema and the indexes are cargo-culted from Synapse. I have no idea if the Synapse structure for this is sane or if we should be structuring this differently.


NegativeMjark:
As per matrix/IRL conversation I can't easily review the table schema unless I can also see the code that's going to read/write to the tables.

I'd hold off implementing schema updating until we are closer to having deployments that actually need it.

It's possible that adding the code for reading/writing code from the tables will make this PR a bit big to review. But I'm not sure I can complain too much given the size of the PRs I've been chucking your way.

Kegsay:
Closing this as I'm going to stab message sending rather than working on surrounding bits first.

PR #23, opened 2017-03-03
Add a query API to the roomserver for getting the latest events in a room.

NegativeMjark:
Implemented as a JSON POST API.

I briefly considered using net/rpc as the RPC protocol. But gave up since it doesn't handle automatically reconnecting and we have existing infrastructure for monitoring HTTP APIs.
Note that we only return non-200 responses if the input was invalid or if there as an internal error in the server. I'd prefer this as a pattern for internal APIs as it is easier to monitor.

I have kept the functions signature expected by net/rpc. That is two arguments, a request struct and a response struct, and a single error return. That should make it easier to switch to a net/rpc solution if the HTTP JSON overhead ever becomes an issue.

### DIFF
--- a/src/github.com/matrix-org/dendrite/roomserver/api/query.go
+++ b/src/github.com/matrix-org/dendrite/roomserver/api/query.go
@@ -9,8 +9,14 @@ import (
 )
 
 // StateKeyTuple is a pair of an event type and state_key.
+// This is used when requesting parts of the state of a room.
 type StateKeyTuple struct {
-	EventType     string
+	// The "type" key
+	EventType string
+	// The "state_key" of a matrix event.
+	// The empty string is a legitimate value for the "state_key" in matrix
+	// so take care to initialise this field lest you accidentally request a
+	// "state_key" with the go default of the empty string.
 	EventStateKey string
 }
 
--- a/src/github.com/matrix-org/dendrite/roomserver/api/query.go
+++ b/src/github.com/matrix-org/dendrite/roomserver/api/query.go
@@ -27,8 +27,11 @@ type QueryLatestEventsAndStateRequest struct {
 type QueryLatestEventsAndStateResponse struct {
 	// Copy of the request for debugging.
 	QueryLatestEventsAndStateRequest
+	// Does the room exist?
+	// If the room doesn't exist this will be false and LatestEvents will be empty.
+	RoomExists bool
 	// The latest events in the room.
-	LatestEvents gomatrixserverlib.EventReference
+	LatestEvents []gomatrixserverlib.EventReference
 	// The state events requested.
 	StateEvents []gomatrixserverlib.Event
 }
@@ -86,9 +89,8 @@ func postJSON(httpClient http.Client, apiURL string, request, response interface
 		}
 		if err = json.NewDecoder(res.Body).Decode(&errorBody); err != nil {
 			return err
-		} else {
-			return fmt.Errorf("api: %d: %s", res.StatusCode, errorBody.Message)
 		}
+		return fmt.Errorf("api: %d: %s", res.StatusCode, errorBody.Message)
 	}
 	return json.NewDecoder(res.Body).Decode(response)
 }
--- a/src/github.com/matrix-org/dendrite/roomserver/api/query.go
+++ b/src/github.com/matrix-org/dendrite/roomserver/api/query.go
@@ -1,0 +1,71 @@
+package api
+
+import (
+	"github.com/matrix-org/gomatrixserverlib"
+)
+
+// StateKeyTuple is a pair of an event type and state_key.
+type StateKeyTuple struct {
+	EventType     string
+	EventStateKey string
+}
+
+// QueryLatestEventsAndStateRequest is a request to QueryLatestEventsAndState
+type QueryLatestEventsAndStateRequest struct {
+	// The roomID to query the latest events for.
+	RoomID string
+	// The state key tuples to fetch from the room current state.
+	// If this list is empty or nil then no events are returned.
+	StateToFetch []StateKeyTuple
+}
+
+// QueryLatestEventsAndStateResponse is a response to QueryLatestEventsAndState
+type QueryLatestEventsAndStateResponse struct {
+	// Copy of the request for debugging.
+	QueryLatestEventsAndStateRequest
+	// The latest events in the room.
+	LatestEvents gomatrixserverlib.EventReference
+	// The state events requested.
+	StateEvents []gomatrixserverlib.Event
+}
+
+// RoomserverQueryAPI is used to query information from the room server.
+type RoomserverQueryAPI interface {
+	// Query the latest events and state for a room from the room server.
+	QueryLatestEventsAndState(
+		request *QueryLatestEventsAndStateRequest,
+		response *QueryLatestEventsAndStateResponse,
+	) error
+}
+
+// RPCServer is used to register a roomserver implementation with an RPC server.
+type RPCServer interface {
+	RegisterName(name string, rcvr interface{}) error
+}
+
+// RegisterRoomserverQueryAPI registers a RoomserverQueryAPI implementation with an RPC server.
+func RegisterRoomserverQueryAPI(rpcServer RPCServer, roomserver RoomserverQueryAPI) error {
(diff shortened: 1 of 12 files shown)


src/github.com/matrix-org/dendrite/roomserver/query/query.go
```diff
@@ -0,0 +1,68 @@
...
+	// Lookup the numeric ID for the room.
+	// Returns 0 if the room doesn't exists.
+	// Returns an error if there was a problem talking to the database.
+	RoomNID(roomID string) (types.RoomNID, error)
+	// Lookup event references for the latest events in the room.
+	// Returns an error if there was a problem talking to the database.
+	LatestEventIDs(roomNID types.RoomNID) ([]gomatrixserverlib.EventReference, error)
+}
+
+// RoomserverQueryAPI is an implemenation of RoomserverQueryAPI
```
Kegsay: > implemenation

src/github.com/matrix-org/dendrite/roomserver/api/query.go
```diff
@@ -0,0 +1,96 @@
...
+	EventType     string
+	EventStateKey string
+}
+
+// QueryLatestEventsAndStateRequest is a request to QueryLatestEventsAndState
+type QueryLatestEventsAndStateRequest struct {
+	// The roomID to query the latest events for.
+	RoomID string
+	// The state key tuples to fetch from the room current state.
+	// If this list is empty or nil then no events are returned.
```
Kegsay: Why? Shouldn't we at least have `LatestEvents` in the response?

src/github.com/matrix-org/dendrite/roomserver/api/query.go
```diff
@@ -0,0 +1,96 @@
...
+type RoomserverQueryAPI interface {
+	// Query the latest events and state for a room from the room server.
+	QueryLatestEventsAndState(
+		request *QueryLatestEventsAndStateRequest,
+		response *QueryLatestEventsAndStateResponse,
+	) error
+}
+
+// RoomserverQueryLatestEventsAndStatePath is the HTTP path for the QueryLatestEventsAndState API.
+const RoomserverQueryLatestEventsAndStatePath = "/api/Roomserver/QueryLatestEventsAndState"
```
Kegsay: Nit: Capital R? Really? I understand `QueryLatestEventsAndState` matching the interface function name, but I don't see why `Roomserver` as opposed to `roomserver`.

src/github.com/matrix-org/dendrite/roomserver/api/query.go
```diff
@@ -0,0 +1,96 @@
...
+	"encoding/json"
+	"fmt"
+	"github.com/matrix-org/gomatrixserverlib"
+	"net/http"
+)
+
+// StateKeyTuple is a pair of an event type and state_key.
+type StateKeyTuple struct {
+	EventType     string
+	EventStateKey string
```
Kegsay: Given the significance of `""` I don't like the type of this being `string`, because if people forget to specify the `EventStateKey` it will default to `""` rather than doing what I think it should be doing, which is failing.

src/github.com/matrix-org/dendrite/roomserver/roomserver-integration-tests/main.go
```diff
@@ -112,6 +115,16 @@ func runAndReadFromTopic(runCmd *exec.Cmd, topic string, count int) ([]string, e
 	go func() {
 		// Read all of stdout.
 		data, err := readCmd.Output()
+		defer func() {
```
Kegsay: Shouldn't this be before `readCmd.Output()` in case that panics too?

src/github.com/matrix-org/dendrite/roomserver/roomserver-integration-tests/main.go
```diff
@@ -157,7 +170,16 @@ func deleteTopic(topic string) error {
 	return cmd.Run()
 }
 
-func testRoomServer(input []string, wantOutput []string) {
+// testRoomserver is used run integration tests against a single roomserver.
```
Kegsay: > is used run integration tests

src/github.com/matrix-org/dendrite/roomserver/roomserver-integration-tests/main.go
```diff
@@ -334,7 +360,21 @@ func main() {
...
+	testRoomserver(input, want, func(q api.RoomserverQueryAPI) {
+		var response api.QueryLatestEventsAndStateResponse
+		if err := q.QueryLatestEventsAndState(
+			&api.QueryLatestEventsAndStateRequest{RoomID: "!HCXfdvrfksxuYnIFiJ:matrix.org"},
+			&response,
+		); err != nil {
+			panic(err)
+		}
+		if !response.RoomExists {
+			panic(fmt.Errorf("Wanted room \"!HCXfdvrfksxuYnIFiJ:matrix.org\" to exist"))
```
Kegsay: Use raw strings rather than needless escaping.

src/github.com/matrix-org/dendrite/roomserver/roomserver-integration-tests/main.go
```diff
@@ -17,8 +18,10 @@ var (
 	zookeeperURI = defaulting(os.Getenv("ZOOKEEPER_URI"), "localhost:2181")
 	// The URI the kafka server is listening on.
 	kafkaURI = defaulting(os.Getenv("KAFKA_URIS"), "localhost:9092")
+	// The address the roomserver should listen on.
+	roomserverAddr = defaulting(os.Getenv("ROOMSERVER_URI"), "localhost:9876")
 	// How long to wait for the roomserver to write the expected output messages.
-	timeoutString = defaulting(os.Getenv("TIMEOUT"), "10s")
+	timeoutString = defaulting(os.Getenv("TIMEOUT"), "60s")
```
Kegsay: Why the increase? This is all localhost right, so I hope it doesn't take >10s...

src/github.com/matrix-org/dendrite/roomserver/query/query.go
NegativeMjark: Done 2e62111

src/github.com/matrix-org/dendrite/roomserver/api/query.go
NegativeMjark: Fixed d20b466

src/github.com/matrix-org/dendrite/roomserver/api/query.go
NegativeMjark: Done d39ce86

src/github.com/matrix-org/dendrite/roomserver/api/query.go
NegativeMjark: Eh. I think using ptr to string is even more confusing since it would seem to imply that sending nil is a useful thing to do.

src/github.com/matrix-org/dendrite/roomserver/roomserver-integration-tests/main.go
NegativeMjark: Commented 400dff3

src/github.com/matrix-org/dendrite/roomserver/roomserver-integration-tests/main.go
NegativeMjark: Done 400dff3

src/github.com/matrix-org/dendrite/roomserver/roomserver-integration-tests/main.go
NegativeMjark: Done 400dff3

src/github.com/matrix-org/dendrite/roomserver/roomserver-integration-tests/main.go
NegativeMjark: Dene 400dff3

src/github.com/matrix-org/dendrite/roomserver/api/query.go
NegativeMjark: Commented cce5af4

PR #27, opened 2017-03-07
Add /createRoom endpoint and unmarshal HTTP body

Kegsay:
Doesn't actually do anything with the request struct yet.

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/writers/createroom.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/writers/createroom.go
@@ -1,0 +1,42 @@
+package writers
+
+import (
+	"encoding/json"
+	"net/http"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/matrix-org/dendrite/clientapi/auth"
+	"github.com/matrix-org/dendrite/clientapi/common"
+	"github.com/matrix-org/util"
+)
+
+// http://matrix.org/docs/spec/client_server/r0.2.0.html#post-matrix-client-r0-createroom
+type createRoomRequest struct {
+	Invite          []string               `json:"invite"`
+	Name            string                 `json:"name"`
+	Visibility      string                 `json:"visibility"`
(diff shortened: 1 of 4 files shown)


src/github.com/matrix-org/dendrite/clientapi/writers/createroom.go
```diff
@@ -0,0 +1,42 @@
...
+	"encoding/json"
+	"net/http"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/matrix-org/dendrite/clientapi/auth"
+	"github.com/matrix-org/dendrite/clientapi/common"
+	"github.com/matrix-org/util"
+)
+
+// http://matrix.org/docs/spec/client_server/r0.2.0.html#post-matrix-client-r0-createroom
```
NegativeMjark: Minor nit. Can we link to the https:// docs? The green padlock gives me warm fuzzy feelings.

src/github.com/matrix-org/dendrite/clientapi/writers/createroom.go
```diff
@@ -0,0 +1,42 @@
...
+type createRoomRequest struct {
+	Invite          []string               `json:"invite"`
+	Name            string                 `json:"name"`
+	Visibility      string                 `json:"visibility"`
+	Topic           string                 `json:"topic"`
+	Preset          string                 `json:"preset"`
+	CreationContent map[string]interface{} `json:"creation_content"`
+	InitialState    json.RawMessage        `json:"initial_state"` // TODO
+	RoomAliasName   string                 `json:"room_alias_name"`
```
Kegsay: I'm approaching this like I approached `gomatrix` - no needless documentation on JSON objects, but instead links to the spec. This may not be sustainable long-term, but we'll see.

src/github.com/matrix-org/dendrite/clientapi/writers/createroom.go
Kegsay: Done.

PR #29, opened 2017-03-08
Move the functions for reading room state to a separate package.

NegativeMjark:
This should:
    1) Make the input package a bit cleaner.
    2) Allow us to reuse the state reading code from the query package.


Kegsay:
LGTM

PR #33, opened 2017-03-10
Send /createRoom events to kafka

Kegsay:
It doesn't return an error, so yay?

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/writers/createroom.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/writers/createroom.go
@@ -14,6 +14,7 @@ import (
 	"github.com/matrix-org/dendrite/clientapi/httputil"
 	"github.com/matrix-org/dendrite/clientapi/jsonerror"
 	"github.com/matrix-org/dendrite/common"
+	"github.com/matrix-org/dendrite/roomserver/api"
 	"github.com/matrix-org/gomatrixserverlib"
 	"github.com/matrix-org/util"
 	sarama "gopkg.in/Shopify/sarama.v1"
@@ -180,15 +181,11 @@ func createRoom(req *http.Request, cfg config.ClientAPI, roomID string, producer
 	}
 
 	// send events to the room server
-	var m sarama.ProducerMessage
-	value, err := json.Marshal(builtEvents)
+	msgs, err := eventsToMessages(builtEvents)
 	if err != nil {
 		return util.ErrorResponse(err)
 	}
-	m.Topic = "clientapiOutput" // TODO: Make this customisable like roomserver is?
-	m.Key = sarama.StringEncoder("")
-	m.Value = sarama.ByteEncoder(value)
-	if _, _, err = producer.SendMessage(&m); err != nil {
+	if err = producer.SendMessages(msgs); err != nil {
 		return util.ErrorResponse(err)
 	}
 
@@ -253,6 +250,35 @@ func authEventsFromStateNeeded(eventsNeeded gomatrixserverlib.StateNeeded,
 		}
 	}
 	return
+}
+
+func eventsToMessages(events []*gomatrixserverlib.Event) ([]*sarama.ProducerMessage, error) {
+	msgs := make([]*sarama.ProducerMessage, len(events))
+	for i, e := range events {
+		var m sarama.ProducerMessage
+
+		// map auth event references to IDs
+		var authEventIDs []string
+		for _, ref := range e.AuthEvents() {
+			authEventIDs = append(authEventIDs, ref.EventID)
+		}
+
+		ire := api.InputRoomEvent{
+			Kind:         api.KindNew,
+			Event:        e.JSON(),
+			AuthEventIDs: authEventIDs,
+		}
+
+		value, err := json.Marshal(ire)
+		if err != nil {
+			return nil, err
+		}
+		m.Topic = "clientapiOutput" // TODO: Make this customisable like roomserver is?
+		m.Key = sarama.StringEncoder("")
+		m.Value = sarama.ByteEncoder(value)
+		msgs[i] = &m
+	}
+	return msgs, nil
 }
 
 type authEventProvider struct {
--- a/src/github.com/matrix-org/dendrite/clientapi/writers/createroom.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/writers/createroom.go
@@ -16,6 +16,7 @@ import (
 	"github.com/matrix-org/dendrite/common"
 	"github.com/matrix-org/gomatrixserverlib"
 	"github.com/matrix-org/util"
+	sarama "gopkg.in/Shopify/sarama.v1"
 )
 
 // https://matrix.org/docs/spec/client_server/r0.2.0.html#post-matrix-client-r0-createroom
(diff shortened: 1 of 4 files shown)


src/github.com/matrix-org/dendrite/clientapi/writers/createroom.go
```diff
@@ -176,7 +177,19 @@ func createRoom(req *http.Request, cfg config.ClientAPI, roomID string) util.JSO
 		// Add the event to the list of auth events
 		builtEventMap[common.StateKeyTuple{e.Type, e.StateKey}] = ev
 		builtEvents = append(builtEvents, ev)
+	}
 
+	// send events to the room server
+	var m sarama.ProducerMessage
+	value, err := json.Marshal(builtEvents)
```
Kegsay: I assume you expect them to be in this form? An array of event objects?

src/github.com/matrix-org/dendrite/clientapi/writers/createroom.go
```diff
@@ -176,7 +177,19 @@ func createRoom(req *http.Request, cfg config.ClientAPI, roomID string) util.JSO
...
+	if err != nil {
+		return util.ErrorResponse(err)
+	}
+	m.Topic = "clientapiOutput" // TODO: Make this customisable like roomserver is?
+	m.Key = sarama.StringEncoder("")
```
Kegsay: I don't know why I need to do this.

src/github.com/matrix-org/dendrite/clientapi/writers/createroom.go
```diff
@@ -176,7 +177,19 @@ func createRoom(req *http.Request, cfg config.ClientAPI, roomID string) util.JSO
...
 		builtEvents = append(builtEvents, ev)
+	}
 
+	// send events to the room server
+	var m sarama.ProducerMessage
+	value, err := json.Marshal(builtEvents)
+	if err != nil {
+		return util.ErrorResponse(err)
+	}
+	m.Topic = "clientapiOutput" // TODO: Make this customisable like roomserver is?
```
Kegsay: This is in the same vein as `roomserverOutput` topics, but should this be customisable? Why would we ever change it?

src/github.com/matrix-org/dendrite/clientapi/writers/createroom.go
NegativeMjark: Nope I expect there to be a single "github.com/matrix-org/dendrite/roomserver/api/InputRoomEvent" object per event.

src/github.com/matrix-org/dendrite/clientapi/writers/createroom.go
```diff
@@ -176,7 +177,19 @@ func createRoom(req *http.Request, cfg config.ClientAPI, roomID string) util.JSO
...
+	m.Key = sarama.StringEncoder("")
+	m.Value = sarama.ByteEncoder(value)
+	if _, _, err = producer.SendMessage(&m); err != nil {
```
NegativeMjark: You will need to send a message per event. `producer.SendMessages` provides a bulk send API.

src/github.com/matrix-org/dendrite/clientapi/writers/createroom.go
NegativeMjark: Actually the key should probably be null or maybe the EventID? Kafka uses the key as a primary key for deduplicating messages when compacting logs.

src/github.com/matrix-org/dendrite/clientapi/writers/createroom.go
NegativeMjark: I *think* my plan was that the "m.room.create" event should be KindOutlier because the server isn't in the room until it gets a m.room.member join event. The m.room.member join event should be KindJoin with the m.room.create as the state. All the other events should be KindNew.

Kegsay: As per #dendrite-dev: I've used `KindNew` everywhere for now since it should work?

src/github.com/matrix-org/dendrite/clientapi/writers/createroom.go
Kegsay: Done. https://github.com/matrix-org/dendrite/pull/33/commits/53cd6546d3785b98721d92961de492b0b3e8ded2

src/github.com/matrix-org/dendrite/clientapi/writers/createroom.go
Kegsay: Done (event ID). https://github.com/matrix-org/dendrite/pull/33/commits/b1e975298670055b0565f2c53868522bf5173a8c

src/github.com/matrix-org/dendrite/clientapi/writers/createroom.go
```diff
@@ -255,6 +252,35 @@ func authEventsFromStateNeeded(eventsNeeded gomatrixserverlib.StateNeeded,
...
+			Kind:         api.KindNew,
+			Event:        e.JSON(),
+			AuthEventIDs: authEventIDs,
+		}
+
+		value, err := json.Marshal(ire)
+		if err != nil {
+			return nil, err
+		}
+		m.Topic = "clientapiOutput" // TODO: Make this customisable like roomserver is?
```
NegativeMjark: This topic is the one that the room server reads from. So will need to configurable like the roomserver is.

src/github.com/matrix-org/dendrite/clientapi/writers/createroom.go
```diff
@@ -242,6 +252,35 @@ func authEventsFromStateNeeded(eventsNeeded gomatrixserverlib.StateNeeded,
 	return
 }
 
+func eventsToMessages(events []*gomatrixserverlib.Event) ([]*sarama.ProducerMessage, error) {
+	msgs := make([]*sarama.ProducerMessage, len(events))
+	for i, e := range events {
+		var m sarama.ProducerMessage
```
NegativeMjark: Could write this as `m := &msgs[i]`. Dunno if you think it's a good idea.

src/github.com/matrix-org/dendrite/clientapi/writers/createroom.go
Kegsay: Why though? When/why would we ever change this?

src/github.com/matrix-org/dendrite/clientapi/writers/createroom.go
Kegsay: I don't think it's a good idea, because I don't think that it is clearer than:
```go
var m sarama.ProducerMessage
msgs[i] = m
```

NegativeMjark: Oh I see, it is an array of pointers so what you have done is fine...

src/github.com/matrix-org/dendrite/clientapi/writers/createroom.go
Kegsay: Done anyway. https://github.com/matrix-org/dendrite/pull/33/commits/dab39a30933e77e75052031098f285c99f12910e

PR #36, opened 2017-03-14
Tool for creating room events

NegativeMjark:
I'm starting a collection of tools under `matrix-org/dendrite/tools`. The first is a tool for generating lots of room events for testing purposes.

### DIFF
--- a/src/github.com/matrix-org/dendrite/tools/create-room-events/main.go
+++ b/src/github.com/matrix-org/dendrite/tools/create-room-events/main.go
@@ -57,8 +57,8 @@ func main() {
 	}
 
 	flag.Parse()
-	*userID = strings.Replace("$SERVER_NAME", *serverName, *userID, 1)
-	*roomID = strings.Replace("$SERVER_NAME", *serverName, *roomID, 1)
+	*userID = strings.Replace(*userID, "$SERVER_NAME", *serverName, 1)
+	*roomID = strings.Replace(*roomID, "$SERVER_NAME", *serverName, 1)
 
 	// Decode the ed25519 private key.
 	privateKeyBytes, err := base64.RawStdEncoding.DecodeString(*privateKeyString)
--- a/src/github.com/matrix-org/dendrite/tools/create-room-events/main.go
+++ b/src/github.com/matrix-org/dendrite/tools/create-room-events/main.go
@@ -150,6 +150,6 @@ func writeEvent(event gomatrixserverlib.Event) {
 			panic(err)
 		}
 	} else {
-		panic(fmt.Errorf("Format %q is not valid, must be %q or %q", format, "RAW", "INPUT"))
+		panic(fmt.Errorf("Format %q is not valid, must be %q or %q", format, "InputRoomEvent", "Event"))
 	}
 }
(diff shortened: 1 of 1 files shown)


Kegsay:
Tools need usage and help messages on the command prompt. This does not appear to have that. Please add (I appreciate you do have comments on things, but that's not a great help for the user).

src/github.com/matrix-org/dendrite/tools/create-room-events/main.go
```diff
@@ -0,0 +1,128 @@
...
+		panic(err)
+	}
+	writeEvent(event)
+	reference := event.EventReference()
+	b.PrevEvents = []gomatrixserverlib.EventReference{reference}
+	b.Depth++
+	return reference
+}
+
+// Write an event to the output.
```
Kegsay: one event per-line? `\n` delimited?

src/github.com/matrix-org/dendrite/tools/create-room-events/main.go
```diff
@@ -0,0 +1,128 @@
...
+	b.AuthEvents = []gomatrixserverlib.EventReference{create}
+	member := build()
+
+	// Build a number of m.room.message events.
+	b.Type = "m.room.message"
+	b.StateKey = nil
+	b.SetContent(map[string]string{"body": "Test Message"})
+	b.AuthEvents = []gomatrixserverlib.EventReference{create, member}
+	for i := 0; i < count; i++ {
+		build()
```
Kegsay: Unhelpful name given it also writes to stdout. `buildAndOutput` is a bit more mouthy but is a lot easier to follow.

src/github.com/matrix-org/dendrite/tools/create-room-events/main.go
```diff
@@ -0,0 +1,128 @@
...
+	privateKeyString = defaulting(os.Getenv("PRIVATE_KEY"), defaultKey)
+	// The room ID to generate events in.
+	roomID = defaulting(os.Getenv("ROOM_ID"), "!roomid:"+serverName)
+	// The user ID to use for as the event sender.
+	userID = defaulting(os.Getenv("USER_ID"), "@userid:"+serverName)
+	// The number of m.room.messsage events to generate.
+	messageCount = defaulting(os.Getenv("MESSAGE_COUNT"), "10")
+	// The output format to use for the messages.
+	//	INPUT -> api.InputRoomEvent
+	//  RAW   -> gomatrixserverlib.Event
```
Kegsay: Adding more terminology for a dendrite-only tool feels sadness making to me. Can we not just use `Event` and `InputRoomEvent` (the actual type names)?

Also, tabs!

NegativeMjark:
> I would like to know why we are using env vars rather than https://golang.org/pkg/flag/ though.

No particular reason, it makes implementing the defaults a bit easier. (changing SERVER_NAME changes the defaults for the other variables). But I could probably have done it with the flags module. if necessary.

Kegsay:
LGTM - We may have people whine about it being env vars though, since it is an unusual choice over `--flags`.

Kegsay:
LGTM

PR #37, opened 2017-03-14
Implement state conflict resolution

NegativeMjark:


src/github.com/matrix-org/dendrite/roomserver/input/state.go
```diff
@@ -114,8 +116,94 @@ func calculateAndStoreStateAfterManyEvents(db RoomEventDatabase, roomNID types.R
...
+		result[i] = event.Event
+		eventIDMap[event.Event.EventID()] = entries[i]
+	}
+	return result, eventIDMap, nil
+}
+
+// resolveConflicts resolves a list of conflicted state entries. It takes two lists.
+// The first is a list of all state entries both conflicted and not conflicted.
+// The second is a list of all state entries where there is more than one numeric event ID for the same state key tuple.
+// Returns a list of combining the entries without conflictes with the result of state resolution for the entries with conflicts.
```
Kegsay: > Returns a list of combining the entries without conflictes...

Better as:

> Returns a list which combines the entries without conflicts...

Kegsay: Also, *why* do we need to provide `combined` in the first place? It looks like we just get the non-conflicting ones and pass them through unaltered, so uhhh.. why can't the caller do that? The API shape feels weird currently.

src/github.com/matrix-org/dendrite/roomserver/input/state.go
```diff
@@ -114,8 +116,94 @@ func calculateAndStoreStateAfterManyEvents(db RoomEventDatabase, roomNID types.R
 	return db.AddState(roomNID, nil, state)
 }
 
+// loadStateEvents loads the matrix events for a list of state entries.
+// Returns a list of events and a map from string event ID back to state entry.
```
Kegsay: > Returns a list of events

I can tell that from the function definition. Tell me something the function definition can't say:
 - What is this list sorted by?
 - Is it a guaranteed 1:1 mapping from `[]types.StateEntry`?
 - Are the events all state events?

PR #38, opened 2017-03-14
Add option to stop processing input after a given number of messages

NegativeMjark:


src/github.com/matrix-org/dendrite/roomserver/input/consumer.go
```diff
@@ -45,6 +46,11 @@ type Consumer struct {
 	// The ErrorLogger for this consumer.
 	// If left as nil then the consumer will panic when it encounters an error
 	ErrorLogger ErrorLogger
+	// If non-nil then the consumer will stop processing messages after this
+	// many messages and will shutdown
+	StopProcessingAfter *int
```
Kegsay: Be more precise here please. It's unclear if a malformed message counts as having been "processed" or not. Something like:

> If non-nil then the consumer will stop processing messages after the given number of messages and will shut down. Malformed messages are included in this count.

Kegsay:
Why do you want this? There seems to be no justifications given for adding this feature.

Kegsay:
LGTM

PR #39, opened 2017-03-14
Implement event sending part of /rooms/$roomid/send/$type/$txnid

Kegsay:
This involves:
 - Parsing the HTTP request
 - Requesting auth events from the roomserver via the Query API
 - Building the event
 - Doing auth checks on the event
 - Sending it to the roomserver input log

Tested this with a roomserver instance and it all seems to work as expected. There's an increasing amount of duplication and pain points I'm seeing, but will fix those in subsequent PRs.

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/writers/sendmessage.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/writers/sendmessage.go
@@ -81,8 +81,8 @@ func SendMessage(req *http.Request, roomID, eventType, txnID string, cfg config.
 
 	// check to see if this user can perform this operation
 	stateEvents := make([]*gomatrixserverlib.Event, len(queryRes.StateEvents))
-	for i, e := range queryRes.StateEvents {
-		stateEvents[i] = &e
+	for i := range queryRes.StateEvents {
+		stateEvents[i] = &queryRes.StateEvents[i]
 	}
 	provider := gomatrixserverlib.NewAuthEvents(stateEvents)
 	if err = gomatrixserverlib.Allowed(e, &provider); err != nil {
--- a/src/github.com/matrix-org/dendrite/clientapi/writers/sendmessage.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/writers/sendmessage.go
@@ -3,22 +3,54 @@ package writers
 import (
 	"net/http"
 
+	"fmt"
 	log "github.com/Sirupsen/logrus"
 	"github.com/matrix-org/dendrite/clientapi/auth"
+	"github.com/matrix-org/dendrite/clientapi/config"
+	"github.com/matrix-org/dendrite/clientapi/httputil"
+	"github.com/matrix-org/dendrite/common"
+	"github.com/matrix-org/dendrite/roomserver/api"
 	"github.com/matrix-org/util"
 )
 
-// SendMessage implements /rooms/{roomID}/send/{eventType}
-func SendMessage(req *http.Request, roomID, eventType string) util.JSONResponse {
+// http://matrix.org/docs/spec/client_server/r0.2.0.html#put-matrix-client-r0-rooms-roomid-send-eventtype-txnid
+type sendMessageResponse struct {
+	EventID string `json:"event_id"`
+}
+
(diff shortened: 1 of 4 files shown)


src/github.com/matrix-org/dendrite/clientapi/writers/sendmessage.go
```diff
@@ -3,22 +3,150 @@ package writers
...
+		return httputil.LogThenError(req, err)
+	}
+
+	return util.JSONResponse{
+		Code: 200,
+		JSON: sendMessageResponse{e.EventID()},
+	}
+}
+
+func sendToRoomserver(e gomatrixserverlib.Event, producer sarama.SyncProducer, topic string) error {
```
NegativeMjark: This is presumably slightly duplicated in createRoom?

src/github.com/matrix-org/dendrite/clientapi/writers/sendmessage.go
```diff
@@ -3,22 +3,150 @@ package writers
...
+		Kind:         api.KindNew,
+		Event:        e.JSON(),
+		AuthEventIDs: authEventIDs,
+	}
+
+	value, err := json.Marshal(ire)
+	if err != nil {
+		return err
+	}
+	var m sarama.ProducerMessage
```
NegativeMjark: It might be nice if the kafka specific bits were hidden behind an interface that took an InputRoomEvent.

src/github.com/matrix-org/dendrite/clientapi/writers/sendmessage.go
```diff
@@ -3,22 +3,150 @@ package writers
...
+	m.Topic = topic
+	m.Key = sarama.StringEncoder(e.EventID())
+	m.Value = sarama.ByteEncoder(value)
+	if _, _, err := producer.SendMessage(&m); err != nil {
+		return err
+	}
+	return nil
+}
+
+func stateNeeded(builder *gomatrixserverlib.EventBuilder) (requiredStateEvents []common.StateKeyTuple, err error) {
```
NegativeMjark: Much of this logic seems duplicated in `authEventsFromStateNeeded`. Would it make sense to make authEventsFromStateNeeded use this function when working out which common.StateKeyTuple are needed?

src/github.com/matrix-org/dendrite/clientapi/writers/sendmessage.go
Kegsay: Yes, albeit this is the single event form and the other is the bulk form. I will factor it out in a separate PR.

src/github.com/matrix-org/dendrite/clientapi/writers/sendmessage.go
Kegsay: Instead of passing in a `sarama.Producer`, I'm happy to pass in an interface which will send `InputRoomEvents` to the roomserver. I'd like to do that in another PR though.

src/github.com/matrix-org/dendrite/clientapi/writers/sendmessage.go
Kegsay: Yes it would make sense. I would like to do that in a separate PR though.

src/github.com/matrix-org/dendrite/clientapi/writers/sendmessage.go
Kegsay: https://github.com/matrix-org/dendrite/pull/40

src/github.com/matrix-org/dendrite/clientapi/writers/sendmessage.go
Kegsay: https://github.com/matrix-org/dendrite/pull/40

src/github.com/matrix-org/dendrite/clientapi/writers/sendmessage.go
Kegsay: https://github.com/matrix-org/dendrite/pull/44

Kegsay:
Changes have now happened.

PR #49, opened 2017-03-22
Add metrics for the various ways of calculating room state

NegativeMjark:
Each code path through the function has a different "algorithm" label since some of the code paths will be a lot faster than others.

Whether the function succeeds or fails is recorded in an "outcome" label since the performance of a failing attempt might be different to a successful attempt.

Might be an idea to check with @leonerd whether "success" and "failure" are sensible values for the "outcome" label.


Kegsay:
LGTM! We need to document the metrics somewhere other than code though.

PR #50, opened 2017-03-24
Factor out Consumer from the roomserver

Kegsay:
I need to be able to use this same code in the sync server, so factoring it out.

### DIFF
--- a/src/github.com/matrix-org/dendrite/cmd/dendrite-roomserver/main.go
+++ b/src/github.com/matrix-org/dendrite/cmd/dendrite-roomserver/main.go
@@ -2,16 +2,18 @@ package main
 
 import (
 	"fmt"
-	"github.com/matrix-org/dendrite/roomserver/input"
-	"github.com/matrix-org/dendrite/roomserver/query"
-	"github.com/matrix-org/dendrite/roomserver/storage"
-	"github.com/prometheus/client_golang/prometheus"
-	sarama "gopkg.in/Shopify/sarama.v1"
 	"net/http"
 	_ "net/http/pprof"
 	"os"
 	"strconv"
 	"strings"
+
+	"github.com/matrix-org/dendrite/common"
+	"github.com/matrix-org/dendrite/roomserver/input"
+	"github.com/matrix-org/dendrite/roomserver/query"
+	"github.com/matrix-org/dendrite/roomserver/storage"
+	"github.com/prometheus/client_golang/prometheus"
+	sarama "gopkg.in/Shopify/sarama.v1"
 )
 
 var (
@@ -43,10 +45,13 @@ func main() {
 	}
 
 	consumer := input.Consumer{
-		Consumer:             kafkaConsumer,
+		ContinualConsumer: common.ContinualConsumer{
+			Topic:          inputRoomEventTopic,
+			Consumer:       kafkaConsumer,
+			PartitionStore: db,
+		},
 		DB:                   db,
 		Producer:             kafkaProducer,
-		InputRoomEventTopic:  inputRoomEventTopic,
 		OutputRoomEventTopic: outputRoomEventTopic,
 	}
 
--- a/src/github.com/matrix-org/dendrite/common/consumers.go
+++ b/src/github.com/matrix-org/dendrite/common/consumers.go
@@ -25,8 +25,14 @@ type PartitionStorer interface {
 // A ContinualConsumer continually consumes logs even across restarts. It requires a PartitionStorer to
 // remember the offset it reached.
 type ContinualConsumer struct {
-	Topic          string
-	Consumer       sarama.Consumer
+	// The kafkaesque topic to consume events from.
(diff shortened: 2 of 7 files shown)


src/github.com/matrix-org/dendrite/common/consumers.go
```diff
@@ -0,0 +1,102 @@
...
+type PartitionStorer interface {
+	// PartitionOffsets returns the offsets the consumer has reached for each partition.
+	PartitionOffsets(topic string) ([]PartitionOffset, error)
+	// SetPartitionOffset records where the consumer has reached for a partition.
+	SetPartitionOffset(topic string, partition int32, offset int64) error
+}
+
+// A ContinualConsumer continually consumes logs even across restarts. It requires a PartitionStorer to
+// remember the offset it reached.
+type ContinualConsumer struct {
```
richvdh: do we not put this at the top of the file (before `PartitionOffset`) since it is the main export from this file?

src/github.com/matrix-org/dendrite/common/consumers.go
```diff
@@ -0,0 +1,102 @@
...
+	Topic          string
```
richvdh: `Topic` and `Consumer` lost their docs in the refactor. It's kinda obvious I guess, but seems a shame to lose them?

src/github.com/matrix-org/dendrite/cmd/dendrite-roomserver/main.go
```diff
@@ -43,10 +45,13 @@ func main() {
 	}
 
 	consumer := input.Consumer{
-		Consumer:             kafkaConsumer,
+		ContinualConsumer: common.ContinualConsumer{
```
richvdh: It feels a bit weird for half the fields of ContinualConsumer to be populated here and the others be populated by the Consumer when it is started. It feels to me like the Consumer should be responsible for making a new `ContinualConsumer` and copying the fields in.

YMMV though, certainly happy to go with this.

src/github.com/matrix-org/dendrite/common/consumers.go
Kegsay: There is no such standard in Go for exports.

src/github.com/matrix-org/dendrite/cmd/dendrite-roomserver/main.go
Kegsay: At present, I want to keep it like this. Once we've used `ContinualConsumer` a bit more and get a feel for how best to structure it, then this may change.

src/github.com/matrix-org/dendrite/common/consumers.go
Kegsay: Done. https://github.com/matrix-org/dendrite/pull/50/commits/5c1be7c043a54d55fc11483efc15fb7e6b2b9d15

src/github.com/matrix-org/dendrite/common/consumers.go
```diff
@@ -0,0 +1,103 @@
...
+
+// ErrShutdown can be returned from ContinualConsumer.ProcessMessage to stop the ContinualConsumer.
+var ErrShutdown = fmt.Errorf("shutdown")
+
+// Start starts the consumer consuming.
+// Starts up a goroutine for each partition in the kafka stream.
+// Returns nil once all the goroutines are started.
+// Returns an error if it can't start consuming for any of the partitions.
+func (c *ContinualConsumer) Start() error {
+	// TODO: This is nearl
```
Kegsay: Done. https://github.com/matrix-org/dendrite/pull/50/commits/349b991178cc96536fe87c5083999c5933c30ff4

src/github.com/matrix-org/dendrite/common/consumers.go
richvdh: I know it's not a standard. I just find it easier to read a file so that the top-level things are at the start, and the things they use further down.

Kegsay: One can make the arguement the other way too (see the fundamental stuff first at the top, then how it's used as you progress) but equally I just don't care enough to go either way on this.

src/github.com/matrix-org/dendrite/common/consumers.go
```diff
@@ -25,8 +25,14 @@ type PartitionStorer interface {
 // A ContinualConsumer continually consumes logs even across restarts. It requires a PartitionStorer to
 // remember the offset it reached.
 type ContinualConsumer struct {
-	Topic          string
-	Consumer       sarama.Consumer
+	// The kafkaesque topic to consume events from.
```
richvdh: Should these be "Topic is..."?

Kegsay: Fields don't have the same linting rules as types or functions, which require them to start with the name. This was directly lifted from Mjark's prose, which I feel adequately describes the purpose of the field. If you prefer, I can change this to start "Topic is...".

PR #52, opened 2017-03-27
Read roomserver output log and remember position across restarts

Kegsay:
- This uses `ContinualConsumer` to remember its position.
- Fixed an embarrassing bug which has existed since time immemorial which prevented Dendrite from remembering its position in the logs (forgotten `append()`).

At present, config information is hard-coded. My next PR will factor this out into configuration files.


src/github.com/matrix-org/dendrite/clientapi/storage/syncserver.go
```diff
@@ -0,0 +1,39 @@
...
+
+// NewSyncServerDatabase creates a new sync server database
+func NewSyncServerDatabase(dataSourceName string) (*SyncServerDatabase, error) {
+	var db *sql.DB
+	var err error
+	if db, err = sql.Open("postgres", dataSourceName); err != nil {
+		return nil, err
+	}
+	partitions := common.PartitionOffsetStatements{}
+	if err := partitions.Prepare(db); err != nil {
```
richvdh: redeclaring `err`?

Kegsay: Done. https://github.com/matrix-org/dendrite/pull/52/commits/46b3fc501c92429042e92cc8f79ac4e0a8795332

Kegsay:
> one pita in synapse is that there are multiple files with the same name. That may well be less of a concern in go, and it may well be the best thing to do anyway, but have you thought about this at all?

Yes, and have yet to formulate a coherent solution. I'm aware of the problems though. Go doesn't help you on that front.

PR #54, opened 2017-03-27
Write OutputRoomEvents into the database and remember the arrival order

Kegsay:
Depends on #52 and #53.

Actual diff is https://github.com/matrix-org/dendrite/commit/edd7feecf0c6dded9e9b80abd1ae6f392c2f69b2

The arrival order will eventually be used to generate streaming tokens to return to the client as a sync token.


src/github.com/matrix-org/dendrite/clientapi/sync/syncserver.go
```diff
@@ -38,6 +45,28 @@ func (s *Server) Start() error {
...
+		log.WithError(err).Errorf("roomserver output log: event parse failure")
+		return nil
+	}
+	log.WithFields(log.Fields{
+		"event_id": ev.EventID(),
+		"room_id":  ev.RoomID(),
+	}).Info("received event from roomserver")
+
+	if err := s.db.WriteEvent(&ev, output.AddsStateEventIDs, output.RemovesStateEventIDs); err != nil {
+		log.WithError(err).Errorf("roomserver output log: write event failure")
```
richvdh: I feel like we need to worry more about this. Not sure what we can do about it though.

Kegsay: We have this problem throughout Dendrite: what do we do if the database goes away? I agree that this currently is failing very badly as it is getting into an inconsistent state: I can change this to a `panic()` so at least the process will fall over and we won't advance our token. WDYT?

richvdh: I think a panic is *probably* better than carrying on regardless.

Kegsay: Done. https://github.com/matrix-org/dendrite/pull/54/commits/f480c32038036833802fac88c594f8b45fd1fcbb

src/github.com/matrix-org/dendrite/clientapi/storage/output_room_events_table.go
```diff
@@ -19,8 +19,8 @@ CREATE TABLE IF NOT EXISTS output_room_events (
     event_json TEXT NOT NULL,
     -- A list of event IDs which represent a delta of added/removed room state. May be NULL
```
richvdh: comment is now a lie :/

Kegsay: Done. https://github.com/matrix-org/dendrite/pull/54/commits/1ee29d512ba112fe1469bf4fb27a755f3cc84f91

PR #55, opened 2017-03-30
Store current room state

Kegsay:
Track the latest state of a room via a `current_room_state` table. This table will be used in the future to populate the `state` section of a `/sync` response. 

Example:
```
syncserver=> select room_id, event_id, type, state_key, membership from current_room_state ;
           room_id           |          event_id           |           type            |    state_key     | membership 
-----------------------------+-----------------------------+---------------------------+------------------+------------
 !K3G3c19vws6A5c4D:localhost | $RQteHWtOu0bvxE7r:localhost | m.room.create             |                  | 
 !K3G3c19vws6A5c4D:localhost | $7EqmmTkurN6f5DuS:localhost | m.room.member             | @alice:localhost | join
 !K3G3c19vws6A5c4D:localhost | $F5Q95GUPXudLZjgW:localhost | m.room.power_levels       |                  | 
 !K3G3c19vws6A5c4D:localhost | $bqRzNJyqzn9SjUda:localhost | m.room.join_rules         |                  | 
 !K3G3c19vws6A5c4D:localhost | $UpcGLIIu4fHE5uaV:localhost | m.room.history_visibility |                  | 
 !K3G3c19vws6A5c4D:localhost | $ESmO0e9wurIFD59c:localhost | m.room.member             | @bob:localhost   | leave
(6 rows)
```

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/storage/current_room_state_table.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/storage/current_room_state_table.go
@@ -27,6 +27,8 @@ CREATE TABLE IF NOT EXISTS current_room_state (
     -- Clobber based on 3-uple of room_id, type and state_key
     CONSTRAINT room_state_unique UNIQUE (room_id, type, state_key)
 );
+-- for event deletion
+CREATE UNIQUE INDEX IF NOT EXISTS event_id_idx ON current_room_state(event_id);
 `
 
 const upsertRoomStateSQL = "" +
--- a/src/github.com/matrix-org/dendrite/clientapi/storage/current_room_state_table.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/storage/current_room_state_table.go
@@ -1,0 +1,89 @@
+package storage
+
+import (
+	"database/sql"
+	"encoding/json"
+
+	"github.com/matrix-org/dendrite/clientapi/events"
+	"github.com/matrix-org/gomatrixserverlib"
+)
+
+const currentRoomStateSchema = `
+-- Stores the current room state for every room.
+CREATE TABLE IF NOT EXISTS current_room_state (
+    -- The 'room_id' key for the state event.
+    room_id TEXT NOT NULL,
+    -- The state event ID
+    event_id TEXT NOT NULL,
+    -- The state event type e.g 'm.room.member'
+    type TEXT NOT NULL,
+    -- The state_key value for this state event e.g ''
+    state_key TEXT NOT NULL,
+    -- The JSON for the event. Stored as TEXT because this should be valid UTF-8.
+    event_json TEXT NOT NULL,
+    -- The 'content.membership' value if this event is an m.room.member event. For other
+    -- events, this will be NULL.
+    membership TEXT,
+    -- Clobber based on 3-uple of room_id, type and state_key
+    CONSTRAINT room_state_unique UNIQUE (room_id, type, state_key)
+);
+`
+
+const upsertRoomStateSQL = "" +
+	"INSERT INTO current_room_state (room_id, event_id, type, state_key, event_json, membership) VALUES ($1, $2, $3, $4, $5, $6)" +
+	" ON CONFLICT ON CONSTRAINT room_state_unique" +
+	" DO UPDATE SET event_id = $2, event_json = $5, membership = $6"
+
+const deleteRoomStateByEventIDSQL = "" +
+	"DELETE FROM current_room_state WHERE event_id = $1"
+
+type currentRoomStateStatements struct {
+	upsertRoomStateStmt          *sql.Stmt
+	deleteRoomStateByEventIDStmt *sql.Stmt
+}
+
+func (s *currentRoomStateStatements) prepare(db *sql.DB) (err error) {
+	_, err = db.Exec(currentRoomStateSchema)
+	if err != nil {
+		return
+	}
+	if s.upsertRoomStateStmt, err = db.Prepare(upsertRoomStateSQL); err != nil {
+		return
+	}
+	if s.deleteRoomStateByEventIDStmt, err = db.Prepare(deleteRoomStateByEventIDSQL); err != nil {
+		return
+	}
+	return
+}
+
+func (s *currentRoomStateStatements) UpdateRoomState(txn *sql.Tx, added []gomatrixserverlib.Event, removedEventIDs []string) error {
+	// remove first, then add, as we do not ever delete state, but do replace state which is a remove followed by an add.
+	for _, eventID := range removedEventIDs {
+		_, err := txn.Stmt(s.deleteRoomStateByEventIDStmt).Exec(eventID)
+		if err != nil {
+			return err
+		}
+	}
+
+	for _, event := range added {
+		if event.StateKey() == nil {
+			// ignore non state events
+			continue
+		}
+		var membership *string
+		if event.Type() == "m.room.member" {
+			var memberContent events.MemberContent
+			if err := json.Unmarshal(event.Content(), &memberContent); err != nil {
+				return err
+			}
+			membership = &memberContent.Membership
+		}
+		_, err := txn.Stmt(s.upsertRoomStateStmt).Exec(
+			event.RoomID(), event.EventID(), event.Type(), *event.StateKey(), event.JSON(), membership,
+		)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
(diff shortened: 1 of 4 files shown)


src/github.com/matrix-org/dendrite/clientapi/storage/output_room_events_table.go
```diff
@@ -13,21 +15,28 @@ CREATE TABLE IF NOT EXISTS output_room_events (
...
     room_id TEXT NOT NULL,
     -- The JSON for the event. Stored as TEXT because this should be valid UTF-8.
     event_json TEXT NOT NULL,
-    -- A list of event IDs which represent a delta of added/removed room state.
-    add_state_ids TEXT[] NOT NULL,
-    remove_state_ids TEXT[] NOT NULL
+    -- A list of event IDs which represent a delta of added/removed room state. This can be NULL
+    -- if there is no delta.
+    add_state_ids TEXT[],
+    remove_state_ids TEXT[]
```
Kegsay: It turns out giving it nil slices really makes it insert `NULL`. Since the Go idiom is to pass around `nil` slices, I've reverted the previous change which had these set to `NOT NULL` so we can actually insert rows into this table.

Kegsay:
As per #dendrite-dev:

> you've added a whole load of stuff to an existing PR?

> No? Just look at the diff. That branch has existed since before you looked at the other PRs which is why the commit history is messy and because you commented on a commit, it's dragged the comments into that new PR

Given this is clearly confusing you, I shall attempt to wrangle git into force rebasing off master so these commits disappear. This may take some time for me to figure out so please bear with me...

Kegsay:
Right, done via:
```
git rebase -i 09fba1a487b27040583196e39d485b294288c804 --onto master
```

src/github.com/matrix-org/dendrite/clientapi/storage/current_room_state_table.go
```diff
@@ -0,0 +1,89 @@
...
+    -- The 'room_id' key for the state event.
+    room_id TEXT NOT NULL,
+    -- The state event ID
+    event_id TEXT NOT NULL,
+    -- The state event type e.g 'm.room.member'
+    type TEXT NOT NULL,
+    -- The state_key value for this state event e.g ''
+    state_key TEXT NOT NULL,
+    -- The JSON for the event. Stored as TEXT because this should be valid UTF-8.
+    event_json TEXT NOT NULL,
```
richvdh: should we not get this via the `output_room_events` table?

src/github.com/matrix-org/dendrite/clientapi/storage/current_room_state_table.go
```diff
@@ -0,0 +1,89 @@
...
+);
+`
+
+const upsertRoomStateSQL = "" +
+	"INSERT INTO current_room_state (room_id, event_id, type, state_key, event_json, membership) VALUES ($1, $2, $3, $4, $5, $6)" +
+	" ON CONFLICT ON CONSTRAINT room_state_unique" +
+	" DO UPDATE SET event_id = $2, event_json = $5, membership = $6"
+
+const deleteRoomStateByEventIDSQL = "" +
+	"DELETE FROM current_room_state WHERE event_id = $1"
```
richvdh: we probably need an index to make this work?

src/github.com/matrix-org/dendrite/clientapi/storage/current_room_state_table.go
```diff
@@ -0,0 +1,89 @@
...
+	for _, event := range added {
+		if event.StateKey() == nil {
+			// ignore non state events
+			continue
+		}
+		var membership *string
+		if event.Type() == "m.room.member" {
+			var memberContent events.MemberContent
+			if err := json.Unmarshal(event.Content(), &memberContent); err != nil {
+				return err
```
richvdh: Is there a danger of getting a malformed member event which fails to unmarshal, and therefore fucks up the update mid-transaction somehow? Or can that not happen?

src/github.com/matrix-org/dendrite/clientapi/storage/output_room_events_table.go
```diff
@@ -13,21 +15,28 @@ CREATE TABLE IF NOT EXISTS output_room_events (
...
 );
 `
 
 const insertEventSQL = "" +
-	"INSERT INTO output_room_events (room_id, event_json, add_state_ids, remove_state_ids) VALUES ($1, $2, $3, $4)"
+	"INSERT INTO output_room_events (room_id, event_id, event_json, add_state_ids, remove_state_ids) VALUES ($1, $2, $3, $4, $5)"
+
+const selectEventsSQL = "" +
+	"SELECT event_json FROM output_room_events WHERE event_id = ANY($1)"
```
richvdh: needs an index?

src/github.com/matrix-org/dendrite/clientapi/storage/syncserver.go
```diff
@@ -30,13 +31,38 @@ func NewSyncServerDatabase(dataSourceName string) (*SyncServerDatabase, error) {
...
 func (d *SyncServerDatabase) WriteEvent(ev *gomatrixserverlib.Event, addStateEventIDs, removeStateEventIDs []string) error {
-	return d.events.InsertEvent(ev.RoomID(), ev.JSON(), addStateEventIDs, removeStateEventIDs)
+	return runTransaction(d.db, func(txn *sql.Tx) error {
+		if err := d.events.InsertEvent(txn, ev, addStateEventIDs, removeStateEventIDs); err != nil {
+			return err
+		}
+		// Update the current room state based on the added/removed state event IDs.
+		// In the common case there is a single added event ID which is the state event itself, assuming `ev` is a state event.
+		// However, conflict resolution may result in there being different events being added, or even some removed.
+		if len(removeStateEventIDs) == 0 && len(addStateEventIDs) == 1 && addStateEventIDs[0] == ev.EventID() {
```
richvdh: isn't `len(addStateEventIDs) == 0` also a common case (if the event is not a state event)?

Or are non-state events included in `addStateEventIDs` (which would be confusing)?

src/github.com/matrix-org/dendrite/clientapi/storage/current_room_state_table.go
Kegsay: To make it work fast yeah. Will add.

src/github.com/matrix-org/dendrite/clientapi/storage/current_room_state_table.go
Kegsay: > Is there a danger of getting a malformed member event which fails to unmarshal

Always a possibility. The event originates from the room server though, which would likely have already failed in this way.

> fucks up the update mid-transaction somehow

Not possible. By returning `err`, the `runTransaction` code will cause `txn.Rollback()` to be called, even in the event of a panic.

src/github.com/matrix-org/dendrite/clientapi/storage/current_room_state_table.go
Kegsay: It'd save some space, but at this point I don't care. I want a dead simple working version first.

src/github.com/matrix-org/dendrite/clientapi/storage/output_room_events_table.go
Kegsay: Yep.

src/github.com/matrix-org/dendrite/clientapi/storage/syncserver.go
Kegsay: You're absolutely right. I should check for 0 and bail earlier.

Kegsay: Done. https://github.com/matrix-org/dendrite/pull/55/commits/9520c718cc925967d9048dcd4e70c404492b12f1

src/github.com/matrix-org/dendrite/clientapi/storage/output_room_events_table.go
Kegsay: Done. https://github.com/matrix-org/dendrite/pull/55/commits/9520c718cc925967d9048dcd4e70c404492b12f1

src/github.com/matrix-org/dendrite/clientapi/storage/current_room_state_table.go
Kegsay: Done. https://github.com/matrix-org/dendrite/pull/55/commits/9520c718cc925967d9048dcd4e70c404492b12f1

PR #56, opened 2017-04-06
Add structs for HTTP long-polling

Kegsay:
This does nothing useful yet. It just hooks into the right places and extract some data from the HTTP request.


src/github.com/matrix-org/dendrite/clientapi/routing/routing.go
```diff
@@ -49,11 +49,11 @@ func Setup(servMux *http.ServeMux, httpClient *http.Client, cfg config.ClientAPI
 }
 
 // SetupSyncServerListeners configures the given mux with sync-server listeners
-func SetupSyncServerListeners(servMux *http.ServeMux, httpClient *http.Client, cfg config.Sync) {
+func SetupSyncServerListeners(servMux *http.ServeMux, httpClient *http.Client, cfg config.Sync, srp sync.RequestPool) {
```
richvdh: I'm a bit surprised by the types here. Shouldn't `srp` be a pointer argument? I realise the current impl just has a single pointer, so can be copied and still end up representing the same "pool", but it seems wrong to rely on that?

(likewise for other `RequestPool` params)

src/github.com/matrix-org/dendrite/clientapi/sync/requestpool.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	i, err := strconv.Atoi(timeoutMS)
+	if err != nil {
+		return defaultSyncTimeout
+	}
+	return time.Duration(i) * time.Millisecond
+}
+
+// NewRequestPool makes a new RequestPool
+func NewRequestPool(db *storage.SyncServerDatabase) RequestPool {
```
richvdh: feel like this should be at the top, from the point of view of making sense when reading through the file.

src/github.com/matrix-org/dendrite/clientapi/sync/requestpool.go
```diff
@@ -0,0 +1,100 @@
...
+func (rp *RequestPool) OnIncomingSyncRequest(req *http.Request) util.JSONResponse {
+	// Extract values from request
+	logger := util.GetLogger(req.Context())
+	userID, resErr := auth.VerifyAccessToken(req)
+	if resErr != nil {
+		return *resErr
+	}
+	since := req.URL.Query().Get("since")
+	timeout := getTimeout(req.URL.Query().Get("timeout"))
+	wantFullState := req.URL.Query().Get("full_state") != ""
```
richvdh: spec says `false` should be treated as, uh, false here.

src/github.com/matrix-org/dendrite/clientapi/storage/output_room_events_table.go
```diff
@@ -72,7 +72,7 @@ func (s *outputRoomEventsStatements) Events(txn *sql.Tx, eventIDs []string) ([]g
 	}
 	defer rows.Close()
 
-	result := make([]gomatrixserverlib.Event, len(eventIDs))
+	var result []gomatrixserverlib.Event
```
Kegsay: It does work, but when there is a mismatch then it panics rather than returning a sensible error message, which is also why I edited the error message.

src/github.com/matrix-org/dendrite/clientapi/routing/routing.go
Kegsay: Yes, I found this out when building the next PR. The current PR doesn't actually care about this.

src/github.com/matrix-org/dendrite/clientapi/sync/requestpool.go
Kegsay: Okay, I'll change this in the next PR or else I'm going to conflict badly. Where do you want it? Just below the `struct` definition?

src/github.com/matrix-org/dendrite/clientapi/sync/requestpool.go
Kegsay: Done. https://github.com/matrix-org/dendrite/pull/56/commits/f531abed8992fe390ce162309c7e08be7015aefa

src/github.com/matrix-org/dendrite/clientapi/storage/output_room_events_table.go
```diff
@@ -86,7 +86,7 @@ func (s *outputRoomEventsStatements) Events(txn *sql.Tx, eventIDs []string) ([]g
 		result = append(result, ev)
 	}
 	if i != len(eventIDs) {
-		return nil, fmt.Errorf("failed to map all event IDs to events: (%d != %d)", i, len(eventIDs))
+		return nil, fmt.Errorf("failed to map all event IDs to events: (got %d, wanted, %d)", i, len(eventIDs))
```
Kegsay: Done. https://github.com/matrix-org/dendrite/pull/56/commits/f531abed8992fe390ce162309c7e08be7015aefa

src/github.com/matrix-org/dendrite/clientapi/sync/requestpool.go
richvdh: yup that would be fine

Kegsay: Done. https://github.com/matrix-org/dendrite/commit/622b7af3111a67e0169ae1a38cd8f9f60840fa1f

PR #57, opened 2017-04-10
Return events based on 'since' value

Kegsay:
This PR implements sync tokens as an integer offset from the stream of events received from the room server. If the offset is in the future, the HTTP request is blocked until that event arrives or the timeout expires.

This PR doesn't concern itself with:
 - Visibility of events.
 - Format of HTTP response.
 - Optimisations when servicing very old offsets.

Those will come in a different PR.

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/jsonerror/jsonerror.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/jsonerror/jsonerror.go
@@ -64,6 +64,12 @@ func UnknownToken(msg string) *MatrixError {
 	return &MatrixError{"M_UNKNOWN_TOKEN", msg}
 }
 
+// InvalidSync is an error when the client tries to hit /sync with an invalid
+// ?since= parameter.
+func InvalidSync(msg string) *MatrixError {
+	return &MatrixError{"M_BAD_SYNC", msg}
+}
+
 // LimitExceededError is a rate-limiting error.
 type LimitExceededError struct {
 	MatrixError
--- a/src/github.com/matrix-org/dendrite/clientapi/jsonerror/jsonerror.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/jsonerror/jsonerror.go
@@ -2,6 +2,7 @@ package jsonerror
 
 import (
 	"fmt"
+
 	"github.com/matrix-org/util"
 )
 
@@ -62,12 +63,6 @@ func MissingToken(msg string) *MatrixError {
 // requires authentication and supplies a valid, but out-of-date token.
 func UnknownToken(msg string) *MatrixError {
 	return &MatrixError{"M_UNKNOWN_TOKEN", msg}
-}
-
-// InvalidSync is an error when the client tries to hit /sync with an invalid
-// ?since= parameter.
-func InvalidSync(msg string) *MatrixError {
-	return &MatrixError{"M_BAD_SYNC", msg}
 }
 
 // LimitExceededError is a rate-limiting error.
--- a/src/github.com/matrix-org/dendrite/clientapi/storage/output_room_events_table.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/storage/output_room_events_table.go
@@ -36,13 +36,17 @@ const insertEventSQL = "" +
 const selectEventsSQL = "" +
 	"SELECT event_json FROM output_room_events WHERE event_id = ANY($1)"
 
+const selectEventsInRangeSQL = "" +
+	"SELECT event_json FROM output_room_events WHERE id >= $1 AND id <= $2"
+
 const selectMaxIDSQL = "" +
 	"SELECT MAX(id) FROM output_room_events"
 
 type outputRoomEventsStatements struct {
-	insertEventStmt  *sql.Stmt
-	selectEventsStmt *sql.Stmt
-	selectMaxIDStmt  *sql.Stmt
+	insertEventStmt         *sql.Stmt
+	selectEventsStmt        *sql.Stmt
+	selectMaxIDStmt         *sql.Stmt
+	selectEventsInRangeStmt *sql.Stmt
 }
 
 func (s *outputRoomEventsStatements) prepare(db *sql.DB) (err error) {
@@ -59,6 +63,9 @@ func (s *outputRoomEventsStatements) prepare(db *sql.DB) (err error) {
 	if s.selectMaxIDStmt, err = db.Prepare(selectMaxIDSQL); err != nil {
 		return
 	}
+	if s.selectEventsInRangeStmt, err = db.Prepare(selectEventsInRangeSQL); err != nil {
+		return
+	}
 	return
 }
 
@@ -67,6 +74,34 @@ func (s *outputRoomEventsStatements) prepare(db *sql.DB) (err error) {
 func (s *outputRoomEventsStatements) MaxID() (id int64, err error) {
 	err = s.selectMaxIDStmt.QueryRow().Scan(&id)
 	return
+}
+
+func (s *outputRoomEventsStatements) InRange(oldPos, newPos int64) ([]gomatrixserverlib.Event, error) {
+	rows, err := s.selectEventsInRangeStmt.Query(oldPos, newPos)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+
+	var result []gomatrixserverlib.Event
+	var i int64
+	for ; rows.Next(); i++ {
+		var eventBytes []byte
+		if err := rows.Scan(&eventBytes); err != nil {
+			return nil, err
+		}
+		ev, err := gomatrixserverlib.NewEventFromTrustedJSON(eventBytes, false)
+		if err != nil {
+			return nil, err
+		}
+		result = append(result, ev)
+	}
+	// Expect one event per position, inclusive eg old=3, new=5, expect 3,4,5 so 3 events.
+	wantNum := (1 + newPos - oldPos)
+	if i != wantNum {
+		return nil, fmt.Errorf("failed to map all positions to events: (got %d, wanted, %d)", i, wantNum)
+	}
+	return result, nil
 }
 
 // InsertEvent into the output_room_events table. addState and removeState are an optional list of state event IDs.
--- a/src/github.com/matrix-org/dendrite/clientapi/storage/output_room_events_table.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/storage/output_room_events_table.go
@@ -36,9 +36,13 @@ const insertEventSQL = "" +
 const selectEventsSQL = "" +
 	"SELECT event_json FROM output_room_events WHERE event_id = ANY($1)"
 
+const selectMaxIDSQL = "" +
+	"SELECT MAX(id) FROM output_room_events"
+
 type outputRoomEventsStatements struct {
 	insertEventStmt  *sql.Stmt
 	selectEventsStmt *sql.Stmt
+	selectMaxIDStmt  *sql.Stmt
 }
 
 func (s *outputRoomEventsStatements) prepare(db *sql.DB) (err error) {
@@ -52,6 +56,16 @@ func (s *outputRoomEventsStatements) prepare(db *sql.DB) (err error) {
 	if s.selectEventsStmt, err = db.Prepare(selectEventsSQL); err != nil {
 		return
 	}
+	if s.selectMaxIDStmt, err = db.Prepare(selectMaxIDSQL); err != nil {
+		return
+	}
+	return
+}
+
+// MaxID returns the ID of the last inserted event in this table. This should only ever be used at startup, as it will
+// race with inserting events if it is done afterwards.
+func (s *outputRoomEventsStatements) MaxID() (id int64, err error) {
+	err = s.selectMaxIDStmt.QueryRow().Scan(&id)
 	return
 }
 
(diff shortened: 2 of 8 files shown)


src/github.com/matrix-org/dendrite/clientapi/storage/output_room_events_table.go
```diff
@@ -52,15 +60,60 @@ func (s *outputRoomEventsStatements) prepare(db *sql.DB) (err error) {
...
+	}
+	// Expect one event per position, exclusive of old. eg old=3, new=5, expect 4,5 so 2 events.
+	wantNum := (newPos - oldPos)
+	if i != wantNum {
+		return nil, fmt.Errorf("failed to map all positions to events: (got %d, wanted, %d)", i, wantNum)
+	}
+	return result, nil
+}
+
 // InsertEvent into the output_room_events table. addState and removeState are an optional list of state event IDs.
```
richvdh: it'd be nice to document what it returns

src/github.com/matrix-org/dendrite/clientapi/storage/syncserver.go
```diff
@@ -81,6 +85,16 @@ func (d *SyncServerDatabase) SetPartitionOffset(topic string, partition int32, o
 	return d.partitions.UpsertPartitionOffset(topic, partition, offset)
 }
 
+// SyncStreamPosition returns the latest position in the sync stream
```
richvdh: again, nice to document what it returns if there are no events yet

src/github.com/matrix-org/dendrite/clientapi/storage/syncserver.go
```diff
@@ -81,6 +85,16 @@ func (d *SyncServerDatabase) SetPartitionOffset(topic string, partition int32, o
...
+func (d *SyncServerDatabase) SyncStreamPosition() (int64, error) {
```
richvdh: should this not return a, uh, `syncStreamPosition` ?

src/github.com/matrix-org/dendrite/clientapi/sync/requestpool.go
```diff
@@ -95,7 +152,13 @@ func getTimeout(timeoutMS string) time.Duration {
 	return time.Duration(i) * time.Millisecond
 }
 
-// NewRequestPool makes a new RequestPool
-func NewRequestPool(db *storage.SyncServerDatabase) RequestPool {
-	return RequestPool{db}
+func getSyncStreamPosition(since string) (syncStreamPosition, error) {
+	if since == "" {
+		return syncStreamPosition(0), nil
```
richvdh: assuming the first event in the db gets an id of zero (does it?), and given `oldPos` is exclusive, I think this will mean the first event won't get returned?

src/github.com/matrix-org/dendrite/clientapi/sync/requestpool.go
```diff
@@ -54,34 +75,70 @@ func (rp *RequestPool) OnIncomingSyncRequest(req *http.Request) util.JSONRespons
...
+		}
+	case res := <-done: // received a response
+		return res
 	}
 }
 
-// OnNewEvent is called when a new event is received from the room server
-func (rp *RequestPool) OnNewEvent(ev *gomatrixserverlib.Event) {
+// OnNewEvent is called when a new event is received from the room server. Must only be
+// called from a single goroutine.
```
richvdh: might be worth noting why (presumably to avoid races between updates which end up in the wrong order?)

src/github.com/matrix-org/dendrite/clientapi/sync/requestpool.go
```diff
@@ -54,34 +75,70 @@ func (rp *RequestPool) OnIncomingSyncRequest(req *http.Request) util.JSONRespons
...
+	// Whichever returns first is the one we will serve back to the client.
+	// TODO: Currently this means that cpu work is timed, which may not be what we want long term.
+	timeoutChan := make(chan struct{})
+	timer := time.AfterFunc(timeout, func() {
+		close(timeoutChan) // signal that the timeout has expired
+	})
+
+	done := make(chan util.JSONResponse)
+	go func() {
+		syncData, err := rp.currentSyncForUser(syncReq)
```
richvdh: do we not need to cancel this after a timeout? Otherwise we're going to end up stacking them up forever if a user keeps calling sync but never gets any events?

richvdh: I suspect fixing this ties into your TODO above...

src/github.com/matrix-org/dendrite/common/consumers.go
```diff
@@ -67,7 +67,9 @@ func (c *ContinualConsumer) Start() error {
 	}
 	for _, offset := range storedOffsets {
 		// We've already processed events from this partition so advance the offset to where we got to.
-		offsets[offset.Partition] = offset.Offset
+		// Offsets are provided with each message, so if we use the same offset on startup then we'll
```
richvdh: the logic looks right, but I find the comment confusing... isn't the point that the offset passed to `ConsumePartition` is inclusive?

src/github.com/matrix-org/dendrite/clientapi/jsonerror/jsonerror.go
```diff
@@ -64,6 +64,12 @@ func UnknownToken(msg string) *MatrixError {
 	return &MatrixError{"M_UNKNOWN_TOKEN", msg}
 }
 
+// InvalidSync is an error when the client tries to hit /sync with an invalid
+// ?since= parameter.
+func InvalidSync(msg string) *MatrixError {
+	return &MatrixError{"M_BAD_SYNC", msg}
```
Kegsay: What would you prefer? `M_UNKNOWN`?

src/github.com/matrix-org/dendrite/clientapi/storage/output_room_events_table.go
```diff
@@ -31,14 +31,22 @@ CREATE UNIQUE INDEX IF NOT EXISTS event_id_idx ON output_room_events(event_id);
...
 
 const insertEventSQL = "" +
-	"INSERT INTO output_room_events (room_id, event_id, event_json, add_state_ids, remove_state_ids) VALUES ($1, $2, $3, $4, $5)"
+	"INSERT INTO output_room_events (room_id, event_id, event_json, add_state_ids, remove_state_ids) VALUES ($1, $2, $3, $4, $5) RETURNING id"
 
 const selectEventsSQL = "" +
 	"SELECT event_json FROM output_room_events WHERE event_id = ANY($1)"
 
+const selectEventsInRangeSQL = "" +
+	"SELECT event_json FROM output_room_events WHERE id > $1 AND id <= $2"
```
Kegsay: > is that covered under "optimisations when servicing very old offsets"?

Yes, I'm not caring about this currently.

src/github.com/matrix-org/dendrite/clientapi/storage/output_room_events_table.go
```diff
@@ -52,15 +60,60 @@ func (s *outputRoomEventsStatements) prepare(db *sql.DB) (err error) {
...
+		return
+	}
+	if s.selectEventsInRangeStmt, err = db.Prepare(selectEventsInRangeSQL); err != nil {
+		return
+	}
+	return
+}
+
+// MaxID returns the ID of the last inserted event in this table. This should only ever be used at startup, as it will
+// race with inserting events if it is done afterwards.
```
Kegsay: Yes, yes we should.

src/github.com/matrix-org/dendrite/clientapi/storage/syncserver.go
Kegsay: Yes, it should, but that's going to be painful to do because it'll introduce a circular loop in the `import` graph which Go will die on. I'll need to move it out to a 3rd package. I was going to punt that for another PR.

src/github.com/matrix-org/dendrite/clientapi/sync/requestpool.go
Kegsay: There's currently no cancellation mechanism at all built into the actual `/sync` work. This was done on purpose, because whilst we could reasonably recheck the exit condition on `rp.cond.Wait()` what we actually want is to be able to shoot the entire request in the head if we've already timed out. That's the entire point of using `Context` objects, where we can periodically check `ctx.Err()` but to add this into this PR was getting excessive, so I punted it.

src/github.com/matrix-org/dendrite/clientapi/sync/requestpool.go
Kegsay: Yep, your assumption is bang on. We don't assert that `currPos` is monotonically increasing, it's fastest finger first, so there's a chance we could set `currPos` to an older value than intended if this was called on different goroutines.

src/github.com/matrix-org/dendrite/clientapi/sync/requestpool.go
Kegsay: > assuming the first event in the db gets an id of zero (does it?)

It doesn't. It's set to 1.

src/github.com/matrix-org/dendrite/common/consumers.go
Kegsay: Yes, it is inclusive. How should I re-phrase this?

src/github.com/matrix-org/dendrite/clientapi/sync/requestpool.go
Kegsay: Done. https://github.com/matrix-org/dendrite/pull/57/commits/ee7099f67fe02321a5759173aeda1350ed271e8e

src/github.com/matrix-org/dendrite/clientapi/storage/syncserver.go
Kegsay: Done. https://github.com/matrix-org/dendrite/pull/57/commits/ee7099f67fe02321a5759173aeda1350ed271e8e

src/github.com/matrix-org/dendrite/clientapi/storage/output_room_events_table.go
Kegsay: Done. https://github.com/matrix-org/dendrite/pull/57/commits/ee7099f67fe02321a5759173aeda1350ed271e8e

src/github.com/matrix-org/dendrite/clientapi/storage/output_room_events_table.go
```diff
@@ -52,15 +60,60 @@ func (s *outputRoomEventsStatements) prepare(db *sql.DB) (err error) {
...
+func (s *outputRoomEventsStatements) MaxID() (id int64, err error) {
+	var nullableID sql.NullInt64
+	err = s.selectMaxIDStmt.QueryRow().Scan(&nullableID)
+	if nullableID.Valid {
+		id = nullableID.Int64
+	}
 	return
 }
 
+func (s *outputRoomEventsStatements) InRange(oldPos, newPos int64) ([]gomatrixserverlib.Event, error) {
```
Kegsay: Done. https://github.com/matrix-org/dendrite/pull/57/commits/ee7099f67fe02321a5759173aeda1350ed271e8e

src/github.com/matrix-org/dendrite/clientapi/storage/output_room_events_table.go
Kegsay: Done. https://github.com/matrix-org/dendrite/pull/57/commits/ee7099f67fe02321a5759173aeda1350ed271e8e

src/github.com/matrix-org/dendrite/clientapi/jsonerror/jsonerror.go
richvdh: I don't know. Maybe. WDYT?

src/github.com/matrix-org/dendrite/clientapi/sync/requestpool.go
```diff
@@ -112,7 +112,8 @@ func (rp *RequestPool) OnIncomingSyncRequest(req *http.Request) util.JSONRespons
 }
 
 // OnNewEvent is called when a new event is received from the room server. Must only be
-// called from a single goroutine.
+// called from a single goroutine, or else the current position in the stream may be
+// set incorrectly as it is blindly clobbered.
```
richvdh: I think it would be clearer if you explicitly mentioned a race between goroutines here; "... single goroutine, to avoid races between updates which could set the current position in the stream incorrectly"

src/github.com/matrix-org/dendrite/common/consumers.go
richvdh: How about "ConsumePartition will start streaming from the message with the given offset (inclusive), so increment 1 to avoid getting the same message a second time" ?

src/github.com/matrix-org/dendrite/clientapi/sync/requestpool.go
Kegsay: Done. https://github.com/matrix-org/dendrite/pull/57/commits/13f8237c5429e73626a04991e6c7a6b0b6b9e26f

src/github.com/matrix-org/dendrite/common/consumers.go
Kegsay: SGTM. Done https://github.com/matrix-org/dendrite/pull/57/commits/13f8237c5429e73626a04991e6c7a6b0b6b9e26f

src/github.com/matrix-org/dendrite/clientapi/jsonerror/jsonerror.go
Kegsay: We can always change it to something else some other time I guess. Done. https://github.com/matrix-org/dendrite/pull/57/commits/13f8237c5429e73626a04991e6c7a6b0b6b9e26f

PR #72, opened 2017-04-19
/sync: Handle missing state events and return events in the correct order

Kegsay:


src/github.com/matrix-org/dendrite/syncserver/storage/output_room_events_table.go
```diff
@@ -232,7 +248,14 @@ func (s *outputRoomEventsStatements) Events(txn *sql.Tx, eventIDs []string) ([]g
 	return result, nil
 }
 
-func rowsToEvents(rows *sql.Rows) ([]gomatrixserverlib.Event, error) {
+type rowOrdering uint8
+
+const (
+	rowOrderForward = rowOrdering(iota)
+	rowOrderReverse
+)
```
NegativeMjark: I'm not sure this is a good idea.
It might be clearer to just add the correct order by to the selects in the first place?

src/github.com/matrix-org/dendrite/syncserver/storage/output_room_events_table.go
```diff
@@ -244,7 +267,14 @@ func rowsToEvents(rows *sql.Rows) ([]gomatrixserverlib.Event, error) {
 		if err != nil {
 			return nil, err
 		}
-		result = append(result, ev)
+		if order == rowOrderForward {
+			result = append(result, ev)
+		} else if order == rowOrderReverse {
+			result = append([]gomatrixserverlib.Event{ev}, result...)
```
NegativeMjark: My main complaint is with the clarity of the rowOrdering options, but it's worth noting that "appending" to the front of a slice is less efficient than appending to the end of the slice.

src/github.com/matrix-org/dendrite/syncserver/storage/output_room_events_table.go
Kegsay: I don't understand what you mean. I want the latest X messages, ordered from old -> new. I can't do that with a single `ORDER BY`, as getting the latest X messages is done by ordering the `id` from high -> low, but then within those returned results I want to reverse them so it's low -> high.

src/github.com/matrix-org/dendrite/syncserver/storage/output_room_events_table.go
Kegsay: Source?

src/github.com/matrix-org/dendrite/syncserver/storage/output_room_events_table.go
NegativeMjark: Hmm, I see. you want to apply an order by with a limit to get the last N, and then reverse the order. that kind of makes sense. I'd suggest writing a reverse function that takes a list of events and reverses the order of the list. Rather than adding the enum to this function.

src/github.com/matrix-org/dendrite/syncserver/storage/output_room_events_table.go
NegativeMjark: It's a consequence of how go represents slices internally. They are represented as a pointer to the start of the list, a length and a capacity. When you append to the end of a slice it only needs to reallocate if there isn't enough capacity. If it does need to reallocate then it'll need to copy the contents of the backing array somewhere. When it reallocates it'll probably double the size of the backing array. So it'll take O(n log(n)) operations to fill out a list by appending to the end. (If you pre-allocated the backing array with the correct length then it will be O(n) operations).

However if you "append" to the beginning then you have to copy the entire array for every insertion which will be O(n^2) operations.

NegativeMjark: Sorry I guess it's technically O(n) for the appending to the end each time. I think sum of 2^k for k from 0 to log(n) is n?

Kegsay: I was under the impression that the `append()` built-in did various fun optimisations for prepend operations, but I may be hallucinating this.

src/github.com/matrix-org/dendrite/syncserver/storage/output_room_events_table.go
Kegsay: Done. https://github.com/matrix-org/dendrite/pull/72/commits/f62751ef10964d7b69afa6150917a2816f27f51f

src/github.com/matrix-org/dendrite/syncserver/storage/output_room_events_table.go
Kegsay: Either way, it's gone now. https://github.com/matrix-org/dendrite/pull/72/commits/f62751ef10964d7b69afa6150917a2816f27f51f

PR #73, opened 2017-04-20
tool: Add client-api-proxy

Kegsay:
Currently it's not possible to point a Matrix client at Dendrite because there isn't any 1 host:port address for the CS API.

Add a tool which will act as a reverse proxy for CS API endpoints. This means:
 - Matrix clients can be pointed at the proxy and all CS API endpoints will work.
 - Matrix clients don't need to know about the `/api` prefix since we add that on for them.

Manually tested and it all works.


Kegsay:
Done.

PR #74, opened 2017-04-20
Make it possible to point Riot at Dendrite

Kegsay:
This PR makes it possible to point an instance of Riot-Web at  `client-api-proxy` and have some degree of usability. It adds:
 - A dumb implementation of `/login` which lets anyone login as anyone else, the password is ignored and the `access_token` is just the user ID so it works with other Dendrite endpoints.
 - A bunch of random stub endpoints have been added. This was done by trial-and-error in order to get Riot to load the right things.
 - Default CORS handling (from matrix-org/util).


Kegsay:
Done.

NegativeMjark:
I'd prefer it even more if the TODO markers said what was needed to be done, and were in the implementation of the API stubs.

PR #75, opened 2017-04-20
Improve logging configuration

superdump:
The commit log should be explanatory of what's been done here. Please use scrutiny and treat me like the golang newb that I am.


src/github.com/matrix-org/dendrite/common/log.go
```diff
@@ -0,0 +1,42 @@
...
+		_ = os.Mkdir(logDir, os.ModePerm)
+		logrus.AddHook(dugong.NewFSHook(
+			filepath.Join(logDir, "info.log"),
+			filepath.Join(logDir, "warn.log"),
+			filepath.Join(logDir, "error.log"),
+			formatter,
+			&dugong.DailyRotationSchedule{GZip: true},
+		))
+	} else {
+		logrus.SetFormatter(formatter)
```
NegativeMjark: Does this mean we don't get colours on stdout anymore?

superdump: Yes, but I can change that.

superdump:
Colours are back. Now always logs to stderr even when LOG_DIR is set. The format is slightly different between stderr and LOG_DIR but that seems to be intentional.

PR #76, opened 2017-04-20
Rename dendrite-sync-server to dendrite-syncserver for consistency

NegativeMjark:
Makes the name of the binary consistent with the syncserver package name.


Kegsay:
LGTM

PR #77, opened 2017-04-20
Yank out clientapi config options to env vars for now

Kegsay:
Specifically the ones needed to run a Dendrite, since some of them are annoying to pull out (private key).

### DIFF
--- a/src/github.com/matrix-org/dendrite/cmd/dendrite-clientapi/main.go
+++ b/src/github.com/matrix-org/dendrite/cmd/dendrite-clientapi/main.go
@@ -4,6 +4,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"golang.org/x/crypto/ed25519"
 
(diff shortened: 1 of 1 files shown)


src/github.com/matrix-org/dendrite/cmd/dendrite-clientapi/main.go
```diff
@@ -52,9 +69,9 @@ func main() {
 		ServerName:           "localhost",
 		KeyID:                "ed25519:something",
 		PrivateKey:           privKey,
-		KafkaProducerURIs:    []string{"localhost:9092"},
-		ClientAPIOutputTopic: "roomserverInput",
-		RoomserverURL:        "http://localhost:7777",
+		KafkaProducerURIs:    kafkaURIs,
+		ClientAPIOutputTopic: clientAPIOutputTopic,
```
NegativeMjark: Hmm, I really need to come up with better names for these.
Ideally they should be the same environment variable for all processes.

Kegsay: Yeah, will need to have a think about this as it's only going to get worse. I feel it needs to be agnostic to the thing which is producing/consuming them (i.e remove the input/output distinction) but I don't know what a clear name would be for some topics..

PR #80, opened 2017-04-20
Add Apache Version 2.0 license and headers to all golang files

superdump:


ara4n:
lgtm. we'll need the same for gomatrixserverlib too.

PR #82, opened 2017-04-21
Reworked CLI

superdump:
This is being done mostly in parallel to the existing `main`s. The commit log hopefully makes things clear but I would welcome input on:

* CLI usage
* the way the client API has been integrated
* how we might be able to spawn and supervise the components all from one binary/command

Also, I would strongly note that I think the following is a good approach:

* use environment variables for secrets
* use command line options and/or environment variables for configuring things that change between each process (basic service mode of operation and service discovery type stuff)
* use configuration files for things that rarely need changing including things that remain the same for an entire clustered deployment

I don't mind being able to configure everything from everywhere, but things can get cluttered or awkward outside the bounds of the above IMO.

I don't really intend this to be merged yet, but I made a PR out of it just to get it some attention so we can figure out how we want to evolve these things. It improves consistency, but there are some things that may very well be contentious. 😄


Kegsay:
General Qs:
 - Why docopt rather than [flag](https://golang.org/pkg/flag/)? It feels like a lot of extra stuff *just* to do CLI flags...
 - Why make into a single binary? I know we want to share code, but we can do that in many ways, not just by having a single `cmd`. What motivated you to do it like this rather than having a helper package?

I agree with your good approach notes.

ara4n:
@superdump @kegsay whatever happened to this?

erikjohnston:
Closing for now due to lack of consensus and progress. We can/should revisit this down the line.

PR #94, opened 2017-05-08
Factor out creating/deleting/writing to kafka topics

Kegsay:
Spoke with @NegativeMjark before he turned into a pumpkin and we agreed that putting this in `common/test` was the right thing to do.

### DIFF
--- a/src/github.com/matrix-org/dendrite/cmd/roomserver-integration-tests/main.go
+++ b/src/github.com/matrix-org/dendrite/cmd/roomserver-integration-tests/main.go
@@ -48,7 +48,7 @@ var (
 	testDatabase = defaulting(os.Getenv("DATABASE"), fmt.Sprintf("dbname=%s binary_parameters=yes", testDatabaseName))
 )
 
-var env = test.KafkaEnv{
+var exe = test.KafkaExecutor{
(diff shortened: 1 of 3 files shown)


superdump:
I don't really like the use of `env`. Maybe it's familiarity rather than simplicity but I associate it very much with environment for processes such as environment variables.

My first thought was `config(uration)` but the output writer makes me think perhaps `context` would be more appropriate. A `KafkaContext` which contains configuration and instances of things for use of Kafka. What do you think?

richvdh:
I see Rob's point about `env` but tbh I'm not that fussed

Kegsay:
@superdump `KafkaContext` makes me think it is a `Context` but it isn't. `Config` is typically used to represent structs which feed in input params so I guess `KafkaConfig` works.

superdump:
Other than the bikeshedding about the name of the structure (and variable), LGTM.

Kegsay:
It isn't just a plain old config file since there are functions which execute kafka scripts on it so `KafkaConfig` is misleading. I went with `KafkaExecutor` in the end since you know, it executes kafka scripts.

Also, as fun as it is, it's probably not the best use of time to have multiple people review the same PR 0:)

PR #96, opened 2017-05-09
Split out notifying /sync requests and calculating sync responses

Kegsay:
The logic for notifying /sync requests is about to get really
complicated as we optimise when to wake up requests, so split
out that code into a separate struct to isolate it and make
it easier to unit test.

There are no functional changes.

### DIFF
--- a/src/github.com/matrix-org/dendrite/syncapi/consumers/roomserver.go
+++ b/src/github.com/matrix-org/dendrite/syncapi/consumers/roomserver.go
@@ -32,11 +32,11 @@ import (
 type Server struct {
 	roomServerConsumer *common.ContinualConsumer
 	db                 *storage.SyncServerDatabase
-	rp                 *sync.RequestPool
+	notifier           *sync.Notifier
 }
 
 // NewServer creates a new sync server. Call Start() to begin consuming from room servers.
-func NewServer(cfg *config.Sync, rp *sync.RequestPool, store *storage.SyncServerDatabase) (*Server, error) {
+func NewServer(cfg *config.Sync, n *sync.Notifier, store *storage.SyncServerDatabase) (*Server, error) {
 	kafkaConsumer, err := sarama.NewConsumer(cfg.KafkaConsumerURIs, nil)
 	if err != nil {
 		return nil, err
@@ -50,7 +50,7 @@ func NewServer(cfg *config.Sync, rp *sync.RequestPool, store *storage.SyncServer
 	s := &Server{
 		roomServerConsumer: &consumer,
 		db:                 store,
-		rp:                 rp,
+		notifier:           n,
 	}
 	consumer.ProcessMessage = s.onMessage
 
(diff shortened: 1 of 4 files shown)


src/github.com/matrix-org/dendrite/syncapi/sync/notifier.go
```diff
@@ -0,0 +1,66 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sync
+
+import (
+	"github.com/matrix-org/dendrite/syncapi/types"
+	"github.com/matrix-org/gomatrixserverlib"
+	"sync"
```
superdump: Don't we normally list stdlib first, then external deps, then internal deps? Although because of internal deps looking like external deps with gb, they get listed together.

src/github.com/matrix-org/dendrite/syncapi/consumers/roomserver.go
```diff
@@ -96,7 +96,7 @@ func (s *Server) onMessage(msg *sarama.ConsumerMessage) error {
 		}).Panicf("roomserver output log: write event failure")
 		return nil
 	}
-	s.rp.OnNewEvent(&ev, types.StreamPosition(syncStreamPos))
+	s.notifier.OnNewEvent(&ev, types.StreamPosition(syncStreamPos))
```
superdump: This used to use a request pool but now it uses a notifier directly? This feels like a change that is potentially deeper than I would expect, if a request pool is actually some kind of pool of requests as the name would suggest. What happened here?

src/github.com/matrix-org/dendrite/syncapi/sync/notifier.go
Kegsay: I don't think about it. `goimports` *should* re-order to be somewhat sane, it's possible that my editor isn't configured right / has been broken. Will have a look.

src/github.com/matrix-org/dendrite/syncapi/consumers/roomserver.go
Kegsay: This function is called whenever there is a new event from the room server.

Previously, `RequestPool` would handle both the notification of this event and the calculation of responses. This PR changes it so `RequestPool` only needs to care about calculation of responses, NOT notifying any waiting requests.

As a result, `Server` doesn't need to know that `RequestPool` even exists, it just needs to know about the `Notifier`, which is why this has changed from being a `RequestPool` to a `Notifier`.

superdump: Okey poke. Thanks for the clarification.

src/github.com/matrix-org/dendrite/syncapi/sync/notifier.go
Kegsay: I've moved `sync` up to the top. I didn't see anything wrong in my editor configuration.

Generally speaking, we don't really focus on formatting when reviewing Go code, because `go fmt`, `goimports`, etc generally do a pretty good job at it. I accept that in this case it seems to have erred since we use `gb`, but in general we just don't think about it and that hasn't been a problem for us. On the contrary, it means we can focus on the code instead.

Kegsay:
@superdump PTAL

superdump:
I was just doing so. LGTM.

PR #98, opened 2017-05-10
Allow multiple sync server test requests

Kegsay:
- Test data for the sync server is now in its own file.
- Rejig the sync server tests to support multiple `/sync` requests and corresponding
  assertions.
- Fixed a minor bug which resulted in state events to appear twice in /sync
  responses when syncing without a `since` parameter.

There's a bunch of other events in the test data which we don't go through yet, because I'm conscious of how large this PR is.


src/github.com/matrix-org/dendrite/syncapi/storage/syncserver.go
```diff
@@ -169,6 +169,30 @@ func (d *SyncServerDatabase) CompleteSync(userID string, numRecentEventsPerRoom
 			if err != nil {
 				return err
 			}
+
+			// There may be some overlap where events in stateEvents are already in recentEvents, so filter
+			// them out so we don't include them twice in the /sync response. They should be in recentEvents
+			// only, so clients get to the correct state once they have rolled forward.
+			for i := 0; i < len(recentEvents); i++ {
```
Kegsay: Done. https://github.com/matrix-org/dendrite/pull/98/commits/1bad29d95de5741fd1d6d8ad0e54c0ea6c28bca9

PR #101, opened 2017-05-15
Only wake up /sync requests which the event is for

Kegsay:
Really long PR, sorry :disappointed: 

The gist is:
 - Add logic to `Notifier` to determine which /sync streams to wake up (based on who is joined to which room)
 - Add `database.AllJoinedUsersInRooms()` so the `Notifier` can know who is joined to which rooms at startup. Kept up-to-date by the Kafka consumer.

In my defence, 300 lines are purely for unit tests :(

### DIFF
--- a/src/github.com/matrix-org/dendrite/syncapi/sync/notifier.go
+++ b/src/github.com/matrix-org/dendrite/syncapi/sync/notifier.go
@@ -122,7 +122,7 @@ func (n *Notifier) WaitForEvents(req syncRequest) types.StreamPosition {
 	// give up the stream lock prior to waiting on the user lock
 	stream := n.fetchUserStream(req.userID, true)
 	n.streamLock.Unlock()
-	return stream.Wait()
+	return stream.Wait(currentPos)
 }
 
 // Load the membership states required to notify users correctly.
--- a/src/github.com/matrix-org/dendrite/syncapi/sync/notifier.go
+++ b/src/github.com/matrix-org/dendrite/syncapi/sync/notifier.go
@@ -20,6 +20,7 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/matrix-org/dendrite/clientapi/events"
+	"github.com/matrix-org/dendrite/syncapi/storage"
 	"github.com/matrix-org/dendrite/syncapi/types"
 	"github.com/matrix-org/gomatrixserverlib"
 )
@@ -46,7 +47,7 @@ type Notifier struct {
 
 // NewNotifier creates a new notifier set to the given stream position.
 // In order for this to be of any use, the Notifier needs to be told all rooms and
-// the joined users within each of them by calling Notifier.UsersJoinedToRooms().
+// the joined users within each of them by calling Notifier.LoadFromDatabase().
 func NewNotifier(pos types.StreamPosition) *Notifier {
 	return &Notifier{
 		currPos:                  pos,
@@ -127,10 +128,20 @@ func (n *Notifier) WaitForEvents(req syncRequest) types.StreamPosition {
 	}
 }
 
-// UsersJoinedToRooms marks the given users as 'joined' to the given rooms, such that new events from
+// Load the membership states required to notify users correctly.
+func (n *Notifier) Load(db *storage.SyncServerDatabase) error {
+	roomToUsers, err := db.AllJoinedUsersInRooms()
+	if err != nil {
+		return err
+	}
+	n.usersJoinedToRooms(roomToUsers)
+	return nil
+}
+
+// usersJoinedToRooms marks the given users as 'joined' to the given rooms, such that new events from
 // these rooms will wake the given users /sync requests. This should be called prior to ANY calls to
 // OnNewEvent (eg on startup) to prevent racing.
-func (n *Notifier) UsersJoinedToRooms(roomIDToUserIDs map[string][]string) {
+func (n *Notifier) usersJoinedToRooms(roomIDToUserIDs map[string][]string) {
 	// This is just the bulk form of userJoined where we only lock once.
 	n.roomIDToJoinedUsersMutex.Lock()
 	defer n.roomIDToJoinedUsersMutex.Unlock()
--- a/src/github.com/matrix-org/dendrite/syncapi/sync/notifier.go
+++ b/src/github.com/matrix-org/dendrite/syncapi/sync/notifier.go
@@ -99,8 +99,10 @@ func (n *Notifier) OnNewEvent(ev *gomatrixserverlib.Event, pos types.StreamPosit
 	}
 }
 
-// WaitForEvents blocks until there are new events for this request.
-func (n *Notifier) WaitForEvents(req syncRequest) types.StreamPosition {
+// WaitForEvents blocks until there are new events for this request. If forceBlock is true, the request
+// will be forcibly waited until a new event wakes it up. This is typically only useful for testing
+// blocking code.
+func (n *Notifier) WaitForEvents(req syncRequest, forceBlock bool) types.StreamPosition {
 	// Do what synapse does: https://github.com/matrix-org/synapse/blob/v0.20.0/synapse/notifier.py#L298
 	// - Bucket request into a lookup map keyed off a list of joined room IDs and separately a user ID
 	// - Incoming events wake requests for a matching room ID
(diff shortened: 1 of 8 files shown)


src/github.com/matrix-org/dendrite/syncapi/sync/notifier.go
```diff
@@ -15,53 +15,217 @@
...
+		return err
+	}
+	n.usersJoinedToRooms(roomToUsers)
+	return nil
+}
+
+// usersJoinedToRooms marks the given users as 'joined' to the given rooms, such that new events from
+// these rooms will wake the given users /sync requests. This should be called prior to ANY calls to
+// OnNewEvent (eg on startup) to prevent racing.
+func (n *Notifier) usersJoinedToRooms(roomIDToUserIDs map[string][]string) {
```
NegativeMjark: memmbe `setUsersJoinedToRooms()` instead?

src/github.com/matrix-org/dendrite/syncapi/sync/notifier.go
```diff
@@ -15,53 +15,217 @@
...
+	stream, ok := n.userStreams[userID]
+	if !ok {
+		// TODO: Unbounded growth of streams (1 per user)
+		stream = NewUserStream(userID)
+		n.userStreams[userID] = stream
+	}
+	return stream
+}
+
+func (n *Notifier) userJoined(roomID, userID string) {
```
NegativeMjark: I wonder if this should be `addJoinedUser()` or something?

Kegsay: Done. https://github.com/matrix-org/dendrite/pull/101/commits/258f1a604f6a9ac81eec246ce5e273bfde6d16fd

src/github.com/matrix-org/dendrite/syncapi/sync/notifier.go
Kegsay: Done. https://github.com/matrix-org/dendrite/pull/101/commits/258f1a604f6a9ac81eec246ce5e273bfde6d16fd

src/github.com/matrix-org/dendrite/syncapi/sync/notifier.go
```diff
@@ -15,53 +15,217 @@
...
+	// update the current position in a guard and then notify relevant /sync streams.
+	// This needs to be done PRIOR to waking up users as they will read this value.
+	n.currPosMutex.Lock()
 	n.currPos = pos
-	n.cond.L.Unlock()
+	n.currPosMutex.Unlock()
 
-	n.cond.Broadcast() // notify ALL waiting goroutines
+	// Map this event's room_id to a list of joined users, and wake them up.
+	userIDs := n.joinedUsers(ev.RoomID())
```
NegativeMjark: Does this race with the updates in `userJoined` and `userLeft`

src/github.com/matrix-org/dendrite/syncapi/sync/notifier.go
```diff
@@ -15,53 +15,217 @@
...
+func (n *Notifier) addJoinedUser(roomID, userID string) {
+	n.roomIDToJoinedUsersMutex.Lock()
+	defer n.roomIDToJoinedUsersMutex.Unlock()
+	if _, ok := n.roomIDToJoinedUsers[roomID]; !ok {
+		n.roomIDToJoinedUsers[roomID] = make(set)
+	}
+	n.roomIDToJoinedUsers[roomID].add(userID)
+}
+
+func (n *Notifier) userLeft(roomID, userID string) {
```
NegativeMjark: prolly should call this `removeJoinedUser` to match `addJoinedUser`

src/github.com/matrix-org/dendrite/syncapi/sync/notifier.go
Kegsay: No, because those functions are only called by the same goroutine that call this function (it all happens on the consumer goroutine).

src/github.com/matrix-org/dendrite/syncapi/sync/notifier.go
```diff
@@ -15,53 +15,217 @@
...
-	currPos types.StreamPosition
-	// A condition variable to notify all waiting goroutines of a new sync stream position
-	cond *sync.Cond
+	// The latest sync stream position: guarded by 'currPosMutex' which is RW to allow
+	// for concurrent reads on /sync requests
+	currPos      types.StreamPosition
+	currPosMutex *sync.RWMutex
+	// A map of RoomID => Set<UserID> : Map access is guarded by roomIDToJoinedUsersMutex.
+	roomIDToJoinedUsers      map[string]set
+	roomIDToJoinedUsersMutex *sync.Mutex
```
NegativeMjark: Why do you need a mutex for roomIDToJoinedUsers? OnNewEvent seems to only be called in a single thread anyway?

src/github.com/matrix-org/dendrite/syncapi/sync/notifier.go
```diff
@@ -15,53 +15,217 @@
...
+	n.roomIDToJoinedUsersMutex.Lock()
+	defer n.roomIDToJoinedUsersMutex.Unlock()
+	if _, ok := n.roomIDToJoinedUsers[roomID]; !ok {
+		return
+	}
+	return n.roomIDToJoinedUsers[roomID].values()
+}
+
+// A string set, mainly existing for improving clarity of structs in this file.
+type set map[string]bool
```
NegativeMjark: Maybe call this `userIDSet`?

NegativeMjark: (it'd be awesome if there were `UserID` and `RoomID` named types.)

Kegsay:
As per RL, there exists a race whereby the `currPos` is updated with an event which *would* notify but this happens *prior* to calling `stream.Wait()`. Wrapping the `currPos` logic and the code up to the `Wait()` into a single mutex fixes this.

However, this introduces a different race whereby just after we release the global lock and before we lock the stream the `pos` for the stream is updated, meaning we not immediately return and instead wait for another event to notify us. We should be re-checking the `pos` to see if it has been updated by something else.

src/github.com/matrix-org/dendrite/syncapi/sync/notifier.go
```diff
@@ -15,53 +15,217 @@
...
+
+func (s set) add(str string) {
+	s[str] = true
+}
+
+func (s set) remove(str string) {
+	delete(s, str)
+}
+
+func (s set) has(str string) bool {
```
NegativeMjark: Is this used anywhere?

Kegsay: No. Fixed https://github.com/matrix-org/dendrite/pull/101/commits/d7de6c3f337647143898e57e94fc0f94b3b67bc1

src/github.com/matrix-org/dendrite/syncapi/sync/notifier.go
Kegsay: Done. https://github.com/matrix-org/dendrite/pull/101/commits/d7de6c3f337647143898e57e94fc0f94b3b67bc1

src/github.com/matrix-org/dendrite/syncapi/sync/notifier.go
Kegsay: Done. https://github.com/matrix-org/dendrite/pull/101/commits/d7de6c3f337647143898e57e94fc0f94b3b67bc1

src/github.com/matrix-org/dendrite/syncapi/sync/notifier.go
```diff
@@ -15,53 +15,205 @@
...
+		n.roomIDToJoinedUsers[roomID] = make(set)
+	}
+	n.roomIDToJoinedUsers[roomID].add(userID)
+}
+
+// Not thread-safe: must be called on the OnNewEvent goroutine only
+func (n *Notifier) removeJoinedUser(roomID, userID string) {
+	if _, ok := n.roomIDToJoinedUsers[roomID]; !ok {
+		n.roomIDToJoinedUsers[roomID] = make(set)
+	}
```
NegativeMjark: The check here is probably unnecessary as it's safe to call `delete` on a `nil` map, but it probably doesn't matter too much.

Kegsay:
@NegativeMjark PTAL https://github.com/matrix-org/dendrite/pull/101/commits/3f334fac75bf5668775400115ec9ea18f81eb5cd

PR #103, opened 2017-05-15
Add invites to /sync responses

Kegsay:
Does not handle `invite_state` because it's horrendous :sob:


src/github.com/matrix-org/dendrite/cmd/syncserver-integration-tests/main.go
```diff
@@ -508,6 +508,35 @@ func main() {
 	// $ curl -XPUT -d '{"name":"A Different Custom Room Name"}' "http://localhost:8009/_matrix/client/r0/rooms/%21PjrbIMW2cIiaYF4t:localhost/state/m.room.name?access_token=@alice:localhost"
 	// $ curl -XPUT -d '{"msgtype":"m.text","body":"hello bob"}' "http://localhost:8009/_matrix/client/r0/rooms/%21PjrbIMW2cIiaYF4t:localhost/send/m.room.message/2?access_token=@alice:localhost"
 	// $ curl -XPUT -d '{"membership":"invite"}' "http://localhost:8009/_matrix/client/r0/rooms/%21PjrbIMW2cIiaYF4t:localhost/state/m.room.member/@charlie:localhost?access_token=@bob:localhost"
+	if err := exe.WriteToTopic(inputTopic, canonicalJSONInput(outputRoomEventTestData[11:14])); err != nil {
```
NegativeMjark: Is this a sensible way to supply test data? It seems a bit fragile to refer to things by numeric index.

Kegsay: It's the best way I have currently. I'm considering adding in aliases for the indexes but because they need to be injected in the right order my hands are somewhat tied and making the dev know the indices feels the right way to go here.

PR #104, opened 2017-05-16
Handle room transitions to 'leave' in incremental /sync requests

Kegsay:
With a test.

### DIFF
--- a/src/github.com/matrix-org/dendrite/cmd/syncserver-integration-tests/main.go
+++ b/src/github.com/matrix-org/dendrite/cmd/syncserver-integration-tests/main.go
@@ -540,6 +540,38 @@ func main() {
 	// $ curl -XPUT -d '{"membership":"join"}' "http://localhost:8009/_matrix/client/r0/rooms/%21PjrbIMW2cIiaYF4t:localhost/state/m.room.member/@charlie:localhost?access_token=@charlie:localhost"
 	// $ curl -XPUT -d '{"msgtype":"m.text","body":"not charlie..."}' "http://localhost:8009/_matrix/client/r0/rooms/%21PjrbIMW2cIiaYF4t:localhost/send/m.room.message/3?access_token=@alice:localhost"
 	// $ curl -XPUT -d '{"membership":"leave"}' "http://localhost:8009/_matrix/client/r0/rooms/%21PjrbIMW2cIiaYF4t:localhost/state/m.room.member/@charlie:localhost?access_token=@alice:localhost"
+	if err := exe.WriteToTopic(inputTopic, canonicalJSONInput(outputRoomEventTestData[14:17])); err != nil {
+		panic(err)
+	}
+	// Check transitions to leave work
+	testSyncServer(syncServerCmdChan, "@charlie:localhost", "15", `{
+		"account_data": {
+			"events": []
+		},
+		"next_batch": "17",
+		"presence": {
+			"events": []
+		},
+		"rooms": {
+			"invite": {},
+			"join": {},
+			"leave": {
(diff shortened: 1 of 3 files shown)


NegativeMjark:
Would it be possible to add a test to check that users don't see state changes in rooms after they have left?

NegativeMjark:
Also I find the way the test data is generated using numeric indexes quite hard to follow. Would it be possible to add named variables for the entries of `outputRoomEventTestData` and `clientEventTestData` ?

Kegsay:
>Would it be possible to add a test to check that users don't see state changes in rooms after they have left?

Done and fixed. https://github.com/matrix-org/dendrite/pull/104/commits/2f2a0d594022e04a2bc675c28ca155ac86e40238 Well spotted, I got confused as to what `StateBetween` was returning 😞

Kegsay:
> Would it be possible to add named variables for the entries

Not really. It needs to be injected in order, so I can't convert it into a struct. I can alias them on use though e.g:

```go
// $ curl -XPUT -d '{"msgtype":"m.text","body":"why did you kick charlie"}' "http://localhost:8009/_matrix/client/r0/rooms/%21PjrbIMW2cIiaYF4t:localhost/send/m.room.message/3?access_token=@bob:localhost"
// $ curl -XPUT -d '{"name":"No Charlies"}' "http://localhost:8009/_matrix/client/r0/rooms/%21PjrbIMW2cIiaYF4t:localhost/state/m.room.name?access_token=@alice:localhost"
    bobMsgThenRoomNameChange := canonicalJSONInput(outputRoomEventTestData[17:19])
	if err := exe.WriteToTopic(inputTopic, bobMsgThenRoomNameChange); err != nil {
		panic(err)
	}
```

Alternatively, I can alias the indices and add a variadic function to extract them e.g:
```go
events := getEvents(idxBobMsg17, idxRoomNameChange18)
```
so you don't need to think about the weird off-by-one on the slice args (`[17:19]` in this case).

NegativeMjark:
Hmm, the `getEvents` with named indexes could be nice.
Alternatively you could use named values when constructing the outputRoomEventTestData slice. So:

```go
outputRoomEventTestData = []string{"{value1...}", "{value2...}"}
```

Could become:

```go
event1 = "{value1...}"
event2 = "{value2...}"
outputRoomEventTestData = []string{event1, event2}
```

Kegsay:
It puts the room into the `leave` section, but it leaks history. I've added a test and filed a TODO for the leak since that's going to need a lot more work than I want to do in this PR.

Kegsay:
Using named indices so it now looks like:
```go
writeToRoomServerLog(i11StateAliceRoomName, i12AliceMsg, i13StateBobInviteCharlie)
```

@NegativeMjark PTAL

src/github.com/matrix-org/dendrite/cmd/syncserver-integration-tests/main.go
```diff
@@ -540,8 +547,90 @@ func main() {
...
+					}
+				}
+			}
+		}
+	}`)
+
+	// Test joining and leaving the same room in a single /sync request puts the room in the 'leave' section.
+	// TODO: Use an earlier since value to assert that the /sync response doesn't leak messages
+	//       from before charlie was joined to the room. Currently it does leak because RecentEvents doesn't
+	//       take membership into account.
```
NegativeMjark: Does it also get added to the "join" section? or am I missing something.

Kegsay: You're missing something I think. You are not in the room by this point, so it can't be in `join`. You need to be told about the room's existence so it has to be in a section. The correct place for it is `leave`.

PR #106, opened 2017-05-17
Minor refactoring

Kegsay:
- `s/Server/Roomserver/` in `consumers` to accurately reflect what is being consumed.
- `s/set/userIDSet/` in `notifier.go` for clarity.
- Removed lying comments.

### DIFF
--- a/src/github.com/matrix-org/dendrite/syncapi/consumers/roomserver.go
+++ b/src/github.com/matrix-org/dendrite/syncapi/consumers/roomserver.go
@@ -28,15 +28,15 @@ import (
 	sarama "gopkg.in/Shopify/sarama.v1"
 )
 
-// Roomserver consumes events that originated in the room server.
-type Roomserver struct {
+// OutputRoomEvent consumes events that originated in the room server.
+type OutputRoomEvent struct {
 	roomServerConsumer *common.ContinualConsumer
 	db                 *storage.SyncServerDatabase
 	notifier           *sync.Notifier
 }
 
-// NewRoomserver creates a new room server consumer. Call Start() to begin consuming from room servers.
-func NewRoomserver(cfg *config.Sync, n *sync.Notifier, store *storage.SyncServerDatabase) (*Roomserver, error) {
+// NewOutputRoomEvent creates a new OutputRoomEvent consumer. Call Start() to begin consuming from room servers.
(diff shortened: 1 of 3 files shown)


src/github.com/matrix-org/dendrite/syncapi/consumers/roomserver.go
```diff
@@ -28,15 +28,15 @@ import (
 	sarama "gopkg.in/Shopify/sarama.v1"
 )
 
-// Server contains all the logic for running a sync server
-type Server struct {
+// Roomserver consumes events that originated in the room server.
+type Roomserver struct {
```
NegativeMjark: Could this be called `Consumer` or something along those lines to make it clearer that it consumes things?

NegativeMjark: Or `RoomEventConsumer` if you have to differentiate it from other kinds of consumer?

Kegsay: The package is already `consumers` so if the sync server uses the type it'll be `consumers.Roomserver` which is nice? Linter has the weird stuttering warning thing, so I figured this would be good.

NegativeMjark: Hmm, I guess you refer to it by `consumers.Foo` everywhere so it might be clear enough without suffixing with `Consumer`.

However I still dislike that it's named after the thing writing the event rather than the actual event it consumes.
It shouldn't care which sever is writing to that stream, it should only care about the contents.

(In this case there will only ever be one server writing to that stream, but I feel it's a dangerous precedent to set in the codebase)

NegativeMjark: So `consumers.OutputRoomEvent` or `consumers.RoomEvent` might be alright here?

Kegsay: `consumers.OutputRoomEvent` wfm

Kegsay: Done. https://github.com/matrix-org/dendrite/pull/106/commits/5b468ec16598344f4563ba794cf8bddb5ddebd23

PR #109, opened 2017-05-18
Rename "make" to "makeAPI" and factor out some more common code into it

NegativeMjark:
Naming a function the same as a go builtin function seems like a bad
idea. Also move the call to `NewJSONRequestHander` inside the function
rather than calling it everywhere.


Kegsay:
LGTM

PR #110, opened 2017-05-18
Add AccountDatabase for storing user accounts

Kegsay:
Including the ability to add new accounts with a user/password and
select accounts using a user/password. Uses bcrypt to hash passwords.

This does not:
 - Make access tokens.
 - Make devices.

Actual diff is +185. The rest is vendor.


src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts_table.go
```diff
@@ -0,0 +1,94 @@
...
+    -- TODO:
+    -- is_guest, is_admin, appservice_id, upgraded_ts, devices, any email reset stuff?
+);
+`
+
+const insertAccountSQL = "" +
+	"INSERT INTO accounts(localpart, created_ts, password_hash) VALUES ($1, $2, $3)"
+
+const selectAccountByPasswordHashSQL = "" +
+	"SELECT localpart WHERE localpart = $1 AND password_hash = $2"
```
NegativeMjark: This shouldn't work since the password hash includes a random salt.

You'll need to read the password hash from the database and then call bcrypt.CompareHashAndPassword passing the hash and the password.

Kegsay: Oh of course :( totally forgot about the salt.

Kegsay: Done. https://github.com/matrix-org/dendrite/pull/110/commits/fa076d3a07ce0439c23d0e1450c7adabc11e644f

Kegsay:
PTAL @NegativeMjark

PR #115, opened 2017-05-22
Start implementing /join for room aliases for rooms the server is not in.

NegativeMjark:
This PR just creates the join event and sends it to the remote server. This is enough for the remote server to think that dendrite is a member of the room and to involve it in federation.

Notably it doesn't send anything to the roomserver, so the client won't actually think that it has joined the room.

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/producers/roomserver.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/producers/roomserver.go
@@ -92,6 +92,7 @@ func (c *RoomserverProducer) SendEventWithState(state gomatrixserverlib.RespStat
 	return c.SendInputRoomEvents(ires, eventIDs)
 }
 
+// TODO Make this a method on gomatrixserverlib.Event
 func authEventIDs(event gomatrixserverlib.Event) (ids []string) {
 	for _, ref := range event.AuthEvents() {
 		ids = append(ids, ref.EventID)
--- a/src/github.com/matrix-org/dendrite/clientapi/producers/roomserver.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/producers/roomserver.go
@@ -57,7 +57,7 @@ func (c *RoomserverProducer) SendEvents(events []gomatrixserverlib.Event) error 
 }
 
 // SendEventWithState writes an event with KindNew to the roomserver input log
-// preceeded by the state at the event as KindOutlier.
+// with the state at the event as KindOutlier before it.
 func (c *RoomserverProducer) SendEventWithState(state gomatrixserverlib.RespState, event gomatrixserverlib.Event) error {
 	outliers, err := state.Events()
 	if err != nil {
--- a/src/github.com/matrix-org/dendrite/clientapi/producers/roomserver.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/producers/roomserver.go
@@ -45,20 +45,58 @@ func NewRoomserverProducer(kafkaURIs []string, topic string) (*RoomserverProduce
 func (c *RoomserverProducer) SendEvents(events []gomatrixserverlib.Event) error {
 	eventIDs := make([]string, len(events))
 	ires := make([]api.InputRoomEvent, len(events))
-	for i := range events {
-		var authEventIDs []string
-		for _, ref := range events[i].AuthEvents() {
-			authEventIDs = append(authEventIDs, ref.EventID)
-		}
-		ire := api.InputRoomEvent{
+	for i, event := range events {
+		ires[i] = api.InputRoomEvent{
 			Kind:         api.KindNew,
-			Event:        events[i].JSON(),
-			AuthEventIDs: authEventIDs,
+			Event:        event.JSON(),
+			AuthEventIDs: authEventIDs(event),
 		}
-		ires[i] = ire
-		eventIDs[i] = events[i].EventID()
+		eventIDs[i] = event.EventID()
 	}
 	return c.SendInputRoomEvents(ires, eventIDs)
+}
+
+// SendEventWithState writes an event with KindNew to the roomserver input log
+// preceeded by the state at the event as KindOutlier.
+func (c *RoomserverProducer) SendEventWithState(state gomatrixserverlib.RespState, event gomatrixserverlib.Event) error {
+	outliers, err := state.Events()
+	if err != nil {
+		return err
+	}
+
+	eventIDs := make([]string, len(outliers)+1)
+	ires := make([]api.InputRoomEvent, len(outliers)+1)
+	for i, outlier := range outliers {
+		ires[i] = api.InputRoomEvent{
+			Kind:         api.KindOutlier,
+			Event:        outlier.JSON(),
+			AuthEventIDs: authEventIDs(outlier),
+		}
+		eventIDs[i] = outlier.EventID()
+	}
+
+	stateEventIDs := make([]string, len(state.StateEvents))
+	for i := range state.StateEvents {
+		stateEventIDs[i] = state.StateEvents[i].EventID()
+	}
+
+	ires[len(outliers)] = api.InputRoomEvent{
+		Kind:          api.KindNew,
+		Event:         event.JSON(),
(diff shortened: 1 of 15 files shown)


src/github.com/matrix-org/dendrite/federationapi/routing/routing.go
```diff
@@ -24,19 +24,20 @@ import (
...
-	v2keysmux.Handle("/server/",
-		makeAPI("localkeys", func(req *http.Request) util.JSONResponse {
-			return readers.LocalKeys(req, cfg)
-		}),
-	)
+	localKeys := makeAPI("localkeys", func(req *http.Request) util.JSONResponse {
+		return readers.LocalKeys(req, cfg)
+	})
+
+	v2keysmux.Handle("/server/{keyID}", localKeys)
```
Kegsay: We don't ever get this from `mux.Vars`. Intentional?

src/github.com/matrix-org/dendrite/cmd/dendrite-federation-api-server/main.go
```diff
@@ -32,6 +33,14 @@ var (
 	logDir     = os.Getenv("LOG_DIR")
 	serverName = gomatrixserverlib.ServerName(os.Getenv("SERVER_NAME"))
 	serverKey  = os.Getenv("SERVER_KEY")
+	// Base64 encoded SHA256 TLS fingerprint of the X509 certificate used by
+	// the public federation listener for this server.
+	// Can be generated from a PEM certificate called "server.crt" using:
+	//
+	//  openssl x509 -noout -fingerprint -sha256 -inform pem -in server.crt |\
+	//	python -c 'print raw_input()[19:].replace(":","").decode("hex").encode("base64").rstrip("=\n")'
```
Kegsay: Tab.

src/github.com/matrix-org/dendrite/clientapi/producers/roomserver.go
```diff
@@ -45,22 +45,60 @@ func NewRoomserverProducer(kafkaURIs []string, topic string) (*RoomserverProduce
...
+		AuthEventIDs:  authEventIDs(event),
+		HasState:      true,
+		StateEventIDs: stateEventIDs,
+	}
+	eventIDs[len(outliers)] = event.EventID()
+
+	return c.SendInputRoomEvents(ires, eventIDs)
+}
+
+func authEventIDs(event gomatrixserverlib.Event) (ids []string) {
```
Kegsay: We should really make this a function on `gomatrixserverlib.Event`. Here is fine for now though.

Kegsay: Provided there is a TODO or something.

src/github.com/matrix-org/dendrite/clientapi/producers/roomserver.go
```diff
@@ -45,22 +45,60 @@ func NewRoomserverProducer(kafkaURIs []string, topic string) (*RoomserverProduce
...
 		}
-		ires[i] = ire
-		eventIDs[i] = events[i].EventID()
+		eventIDs[i] = event.EventID()
 	}
 	return c.SendInputRoomEvents(ires, eventIDs)
 }
 
+// SendEventWithState writes an event with KindNew to the roomserver input log
+// preceeded by the state at the event as KindOutlier.
```
Kegsay: > preceeded

Kegsay: I would probably drop that word since whilst it generally means "before" it can also be used as "after" in some contexts.

src/github.com/matrix-org/dendrite/clientapi/writers/joinroom.go
```diff
@@ -0,0 +1,249 @@
...
+			Code: 200,
+			// TODO: Put the response struct somewhere common.
+			JSON: struct {
+				RoomID string `json:"room_id"`
+			}{roomID},
+		}
+	}
+
+	if err != nil {
+		// TODO: Gerenerate the correct HTTP status code for all different
```
Kegsay: > Gerenerate

src/github.com/matrix-org/dendrite/clientapi/writers/joinroom.go
```diff
@@ -0,0 +1,249 @@
...
+		//   3) The remote server returned something invalid.
+		//   4) We couldn't fetch the public keys needed to verify the
+		//      signatures on the state events.
+		//   5) ...
+		return httputil.LogThenError(r.req, err)
+	}
+
+	// This should only be reached if the list of candidate servers was empty.
+	// If the list was non-empty then either the loop body would have returned
+	// or err would be non-nil.
```
Kegsay: If you're having to explain the control flow, it probably means it's too complicated. I was going to comment on the fact that we continually clobber `err` and then check it outside the `for` loop. I don't really like it. Can we think of a nicer way (even if it involves splitting some of this function up into 2 functions?

I also don't like how we silently discard errors the previous servers return. They may be non-fatal, but it's probably worth still logging it.

Kegsay:
Other than typo and my concern about the convoluted control flow in `joinRoomUsingServers`, LGTM.

src/github.com/matrix-org/dendrite/clientapi/writers/joinroom.go
Kegsay: Maybe put the current for loop in a new function and return the last error on failure?

NegativeMjark: Sort of done 0c8d3c5

src/github.com/matrix-org/dendrite/clientapi/writers/joinroom.go
```diff
@@ -0,0 +1,260 @@
...
+
+// joinRoomUsingServer tries to join a remote room using a given matrix server.
+// If there was a failure communicating with the server or the response from the
+// server was invalid this returns an error.
+// Otherwise this returns a JSONResponse.
+func (r joinRoomReq) joinRoomUsingServer(roomID string, server gomatrixserverlib.ServerName) (util.JSONResponse, error) {
+	respMakeJoin, err := r.federation.MakeJoin(server, roomID, r.userID)
+	if err != nil {
+		// TODO: Check if the user was not allowed to join the room.
+		return util.JSONResponse{}, err
```
Kegsay: I would personally be passing around pointers so it can be the idiomatic `return nil, err`

src/github.com/matrix-org/dendrite/clientapi/writers/joinroom.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+
+	if len(servers) == 0 {
+		return util.JSONResponse{
+			Code: 404,
+			JSON: jsonerror.NotFound("No candidate servers found for room"),
+		}
+	}
+
+	var response util.JSONResponse
```
Kegsay: Can we move this into the `for` body since it's only used there?

Kegsay: Bonus points if you put `var lastErr error` outside the `for` loop and then return that instead of reusing `err`. That would then be clear enough to remove the comments on `:183`.

Kegsay:
Looking a lot better, thank you :) - LGTM other than remaining tweaks.

src/github.com/matrix-org/dendrite/clientapi/writers/joinroom.go
NegativeMjark: Done d4ff1cd

src/github.com/matrix-org/dendrite/clientapi/writers/joinroom.go
NegativeMjark: Done d4ff1cd

src/github.com/matrix-org/dendrite/clientapi/writers/joinroom.go
NegativeMjark: Done be9adca

Kegsay:
LGTM

PR #116, opened 2017-05-22
Call 'go build' from the pre-commit hook to check that everything builds

NegativeMjark:
This would have caught the problem in #114 that necessitated  a7acfa5

A possible concern is that it takes a couple of seconds to build. One option would be to use `go install` instead, which would write object files in `/pkg` and `/vendor/pkg` that can be reused by subsequent checks. The name of the directory inside `/pkg` doesn't seem to conflict with the name used by gb so it should be safe to do.

### DIFF
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -9,7 +9,7 @@ gocyclo -over 12 src/
 gb test
 
 # Check that all the packages can build.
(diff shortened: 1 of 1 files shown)


Kegsay:
LGTM, sorry!

PR #118, opened 2017-05-24
Add in devices_table to store device information

Kegsay:
Not fully hooked up yet because I need to think how this works with integration tests.

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/auth/storage/devices/devices_table.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/auth/storage/devices/devices_table.go
@@ -26,15 +26,15 @@ import (
 const devicesSchema = `
 -- Stores data about devices.
 CREATE TABLE IF NOT EXISTS devices (
-	-- The access token granted to this device. This has to be the primary key
-	-- so we can distinguish which device is making a given request.
-	access_token TEXT NOT NULL PRIMARY KEY,
-	-- The device identifier. This only needs to uniquely identify a device for a given user, not globally.
-	-- access_tokens will be clobbered based on the device ID for a user.
-	id TEXT NOT NULL,
-	-- The Matrix user ID localpart for this device. This is preferable to storing the full user_id
-	-- as it is smaller, makes it clearer that we only manage devices for our own users, and may make
-	-- migration to different domain names easier.
+    -- The access token granted to this device. This has to be the primary key
+    -- so we can distinguish which device is making a given request.
+    access_token TEXT NOT NULL PRIMARY KEY,
+    -- The device identifier. This only needs to uniquely identify a device for a given user, not globally.
+    -- access_tokens will be clobbered based on the device ID for a user.
+    id TEXT NOT NULL,
+    -- The Matrix user ID localpart for this device. This is preferable to storing the full user_id
+    -- as it is smaller, makes it clearer that we only manage devices for our own users, and may make
+    -- migration to different domain names easier.
     localpart TEXT NOT NULL,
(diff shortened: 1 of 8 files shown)


src/github.com/matrix-org/dendrite/clientapi/auth/storage/devices/storage.go
```diff
@@ -15,23 +15,82 @@
...
+		return nil, err
+	}
+	return &Database{db, d}, nil
 }
 
 // GetDeviceByAccessToken returns the device matching the given access token.
+// Returns sql.ErrNoRows if no matching device was found.
 func (d *Database) GetDeviceByAccessToken(token string) (*authtypes.Device, error) {
-	// TODO: Actual implementation
+	// return d.devices.selectDeviceByToken(token) TODO: Figure out how to make integ tests pass
```
NegativeMjark: ^ Is this TODO still valid? I assume you got the integ tests passing since it passes CI?

src/github.com/matrix-org/dendrite/clientapi/auth/storage/devices/devices_table.go
```diff
@@ -0,0 +1,117 @@
...
+
+const devicesSchema = `
+-- Stores data about devices.
+CREATE TABLE IF NOT EXISTS devices (
+    -- The access token granted to this device. This has to be the primary key
+    -- so we can distinguish which device is making a given request.
+    access_token TEXT NOT NULL PRIMARY KEY,
+    -- The device identifier. This only needs to uniquely identify a device for a given user, not globally.
+    -- access_tokens will be clobbered based on the device ID for a user.
+    id TEXT NOT NULL,
```
NegativeMjark: It might be better to call this device_id? It makes it easier to do joins with other tables.

src/github.com/matrix-org/dendrite/clientapi/auth/storage/devices/storage.go
Kegsay: It is still valid. They pass because we don't actually hit the database in this function, we always return a device.

src/github.com/matrix-org/dendrite/clientapi/auth/storage/devices/devices_table.go
Kegsay: Done. https://github.com/matrix-org/dendrite/pull/118/commits/9f85ad0d0aeecaac8b2495a79d5d0d13b6be760a

src/github.com/matrix-org/dendrite/clientapi/auth/storage/devices/devices_table.go
```diff
@@ -0,0 +1,117 @@
...
+	return err
+}
+
+func (s *devicesStatements) selectDeviceByToken(accessToken string) (*authtypes.Device, error) {
+	var dev authtypes.Device
+	var localpart string
+	err := s.selectDeviceByTokenStmt.QueryRow(accessToken).Scan(&dev.ID, &localpart)
+	if err != nil {
+		dev.UserID = makeUserID(localpart, s.serverName)
+		dev.AccessToken = accessToken
```
NegativeMjark: Is the check here correct? it seems weird to set the UserID if there was an error.

Kegsay: .. Right you are. Done. https://github.com/matrix-org/dendrite/pull/118/commits/f71287edd9446889a1c0c9e637d83e8fb445e60d

Kegsay:
@NegativeMjark PTAL

PR #120, opened 2017-05-26
Skeleton for the Media API

superdump:


src/github.com/matrix-org/dendrite/mediaapi/routing/routing.go
```diff
@@ -0,0 +1,66 @@
...
+			writers.Download(w, req, gomatrixserverlib.ServerName(vars["serverName"]), types.MediaID(vars["mediaId"]), cfg, activeRemoteRequests)
+		})),
+	)
+
+	servMux.Handle("/metrics", prometheus.Handler())
+	servMux.Handle("/api/", http.StripPrefix("/api", apiMux))
+}
+
+// make a util.JSONRequestHandler function into an http.Handler.
+func makeAPI(metricsName string, f func(*http.Request) util.JSONResponse) http.Handler {
```
Kegsay: This has made its way into `common` now, please use that :)

src/github.com/matrix-org/dendrite/mediaapi/types/types.go
```diff
@@ -0,0 +1,72 @@
...
+	Base64Hash         Base64Hash
+	UserID             MatrixUserID
+}
+
+// ActiveRemoteRequests is a lockable map of media URIs requested from remote homeservers
+// It is used for ensuring multiple requests for the same file do not clobber each other.
+type ActiveRemoteRequests struct {
+	sync.Mutex
+	// The string key is an mxc:// URL
+	Set map[string]*sync.Cond
```
Kegsay: `Set` is probably a misleading field name since it isn't `map[string]bool`. Mentioning "mxc" in the field name would help given its the key for this map. Maybe something like `InProgressMXCs` or something?

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
```diff
@@ -0,0 +1,95 @@
...
+
+	// Set status code and write the body
+	w.WriteHeader(res.Code)
+	r.Logger.WithField("code", res.Code).Infof("Responding (%d bytes)", len(resBytes))
+	w.Write(resBytes)
+}
+
+// Validate validates the downloadRequest fields
+func (r *downloadRequest) Validate() *util.JSONResponse {
+	// maybe give the URL pattern in the routing, these are not even possible as the handler would not be hit...?
```
Kegsay: Is this a TODO or something else? If you're unsure about something (when the handler will be hit), then please sort it out now whilst this is fresh in your mind (test it locally), as it is unlikely that we are going to be looking at random question comments like this. Alternatively, TODO tag it please so we can grep for it.

src/github.com/matrix-org/dendrite/mediaapi/writers/upload.go
```diff
@@ -0,0 +1,149 @@
...
+		},
+		Logger: util.GetLogger(req.Context()),
+	}
+
+	if resErr := r.Validate(cfg.MaxFileSizeBytes); resErr != nil {
+		return nil, resErr
+	}
+
+	// FIXME: do we want to always override ContentDisposition here or only if
+	// there is no Content-Disposition header set?
```
Kegsay: What are the pros/cons of overriding vs not? What does Synapse do?

src/github.com/matrix-org/dendrite/mediaapi/routing/routing.go
superdump: \o/ And I'll make upload an auth API (as it is) and figure out the deviceDB part, unless you know that I shouldn't yet and should just stub it or note it.

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
superdump: It was something I was unsure about, just typoed. Meant to be "Maybe given..." I'll test and remove the comment.

It is not clear what validation needs to be applied to the origin and media ID beyond checking that they are strings of length >= 1. I don't expect to be able to test all possibilities so I'd rather just check the possibilities are within what is acceptable and safe. Whitelisting, so to speak.

src/github.com/matrix-org/dendrite/mediaapi/routing/routing.go
Kegsay: If you want to hook it up, you can do what the /sync server does currently: https://github.com/matrix-org/dendrite/blob/master/src/github.com/matrix-org/dendrite/cmd/dendrite-sync-api-server/main.go#L80:L83 - I don't mind though as I'm going to need to touch these anyway to standardise where the account database is.

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
Kegsay: iirc @NegativeMjark stated that they can't contain `/`s and such. I would consult the synapse code to see what validation it does on the upload part, given Synapse is the only working HS we can be much more strict (eg if synapse is only return [a-z0-9] then let's just use that rather than allowing arbitrary UTF-8).

Kegsay: We can then update [the spec](http://matrix.org/docs/spec/client_server/r0.2.0.html#id43) to refine "opaque ID" to something a bit more strict.

src/github.com/matrix-org/dendrite/mediaapi/writers/upload.go
superdump: Synapse seems to have bailed out never implemented parsing the `Content-Disposition` header from the upload request. https://github.com/matrix-org/synapse/blob/fad3a8433535d7de321350bbd17373138c6fd3ec/synapse/rest/media/v1/upload_resource.py#L91

It instead takes the upload name from a `filename` query parameter: https://github.com/matrix-org/synapse/blob/fad3a8433535d7de321350bbd17373138c6fd3ec/synapse/rest/media/v1/upload_resource.py#L71

It parses them for remote files downloads but only stores the upload name, if specified: https://github.com/matrix-org/synapse/blob/fad3a8433535d7de321350bbd17373138c6fd3ec/synapse/rest/media/v1/media_repository.py#L169

If the upload name is present in the db, it responds to download/thumbnail requests with a Content-Disposition header including the file name: https://github.com/matrix-org/synapse/blob/9bba6ebaa903a81cd94fada114aa71e20b685adb/synapse/rest/media/v1/_base.py#L66

The override is essentially ignoring the `Content-Disposition` header from the request and placing higher priority on the `filename` query parameter.

Given synapse ignores the `Content-Disposition` header on uploads, the override would likely result in the same behaviour as synapse. But given the TODO status in synapse for parsing the header, it's not clear what is supposed to be done.

Kegsay: @NegativeMjark the TODO is pinned to you, what should we do here?

I seem to vaguely recall some nonsense with mobile clients and not being able to set content-disposition but could set `filename` or something, but this was years ago.

NegativeMjark: I think it makes sense for the explicit filename parameter to take precedence over the Content-Disposition. I don't remember why we never got round to parsing the Content-Disposition header, I wonder if it was something silly like "Content-Disposition" not actually being a valid request header.

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
NegativeMjark: Synapse chooses the random ID on upload. I'd cheekily update the spec to match the restrictions that synapse uses.

src/github.com/matrix-org/dendrite/mediaapi/routing/routing.go
superdump: I think I'll add it as a later step then.

superdump: Done.

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
superdump: Synapse calls this: https://github.com/matrix-org/synapse/blob/master/synapse/rest/media/v1/_base.py#L35

Which just takes the first two path components that follow the prefix path from the request path as the server name and media ID. And then if someone happens to have appended some filename in the next path component, that is also used.

As @NegativeMjark noted, synapse generates a media ID as a random string as follows: https://github.com/matrix-org/synapse/blob/master/synapse/util/stringutils.py#L24

Filename stuff is not at all well-documented in the spec - neither the query parameter for upload nor appending to the path for download are documented. The only mention is the Content-Disposition header in the download response.

src/github.com/matrix-org/dendrite/mediaapi/types/types.go
superdump: Done.

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
Kegsay: Given https://github.com/matrix-org/synapse/blob/master/synapse/util/stringutils.py#L24 - I would therefore apply a regexp to the media ID for it. We should also add a regexp for the server name (origin) given it should be a domain name with optional port.

src/github.com/matrix-org/dendrite/mediaapi/writers/upload.go
Kegsay: Okay, let's ignore parsing the Content-Disposition on upload then to match Synapse. We can revisit this later if need be.

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go

I check if the media ID matches `[A-Za-z0-9_=-]+` which ensures it has non-zero length and allows all characters we use in the base64-encoded URLEncoding form of the hash sum which will come in a later PR.

I added a comment that origin is validated either by comparison to the configured server name of this homeserver or by a DNS SRV record lookup when creating the request for the remote file.

src/github.com/matrix-org/dendrite/mediaapi/writers/upload.go
superdump: Content-Disposition is not a valid request header. https://tools.ietf.org/html/rfc6266 defines it as a response header. As such, I have removed the code from upload that attempted to parse it and then override it. Instead, only the filename query parameter is used.

While I was at it, I did `url.PathEscape()` on the filename query parameter immediately and also checked for filenames beginning with `~` as that wasn't escaped but could potentially be dangerous.

Kegsay:
LGTM

PR #122, opened 2017-05-26
Hook up registration/login APIs and implement access token generation

Kegsay:
This finishes feature work on registration/login for now.

Registration:
```
$ curl -XPOST -d '{"username":"alice","password":"12345678"}' "http://localhost:7778/api/_matrix/client/r0/register"
{
	"flows": [{
		"stages": ["m.login.dummy"]
	}],
	"completed": [],
	"params": {},
	"session": "totallyuniquesessionid"
}

$ curl -XPOST -d '{"username":"alice","password":"12345678","auth":{"type":"m.login.dummy"}}' "http://localhost:7778/api/_matrix/client/r0/register"
{
	"user_id": "@alice:localhost",
	"access_token": "mV8NAhs90FwKISo5_dlSVqxnqn0AYkS11YnW4faVLxc",
	"home_server": "localhost",
	"device_id": "unknown-device"
}

$ curl -XPOST -d '{"username":"bob","password":"1","auth":{"type":"m.login.dummy"}}' "http://localhost:7778/api/_matrix/client/r0/register"
{"errcode":"M_WEAK_PASSWORD","error":"password too weak: min 8 chars"}
```

Login:
```
$ curl -XGET  "http://localhost:7778/api/_matrix/client/r0/login"
{
	"flows": [{
		"type": "m.login.password",
		"stages": ["m.login.password"]
	}]
}

$ curl -XPOST -d '{"user":"alice","password":"invalid","type":"m.login.password"}' "http://localhost:7778/api/_matrix/client/r0/login"
{"errcode":"M_BAD_JSON","error":"username or password was incorrect, or the account does not exist"}

$ curl -XPOST -d '{"user":"alice","password":"12345678","type":"m.login.password"}' "http://localhost:7778/api/_matrix/client/r0/login"
{
	"user_id": "@alice:localhost",
	"access_token": "d_4_PLGDSRdVXv4-r0b1xhU8VGhNvV1LcsYkx2nOLLI",
	"home_server": "localhost"
}

$ curl -XPOST -d '{}' "http://localhost:7778/api/_matrix/client/r0/createRoom?access_token=@alice:localhost"
{"errcode":"M_FORBIDDEN","error":"Invalid access token"}

$ curl -XPOST -d '{}' "http://localhost:7778/api/_matrix/client/r0/createRoom?access_token=d_4_PLGDSRdVXv4-r0b1xhU8VGhNvV1LcsYkx2nOLLI"
{ .... }
```

Repeated logins clobber access tokens for now (so there is only ever 1 valid access_token at any one time, meaning multi-device doesn't work). This is keyed off the device ID, which we don't yet parse out the request.

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
@@ -16,7 +16,9 @@
 package auth
 
 import (
+	"crypto/rand"
 	"database/sql"
+	"encoding/base64"
 	"fmt"
 	"net/http"
 	"strings"
@@ -26,6 +28,13 @@ import (
 	"github.com/matrix-org/dendrite/clientapi/jsonerror"
 	"github.com/matrix-org/util"
 )
+
+// UnknownDeviceID is the default device id if one is not specified.
+var UnknownDeviceID = "unknown-device"
+
+// OWASP recommends at least 128 bits of entropy for tokens: https://www.owasp.org/index.php/Insufficient_Session-ID_Length
+// 32 bytes => 256 bits
+var tokenByteLength = 32
 
 // VerifyAccessToken verifies that an access token was supplied in the given HTTP request
 // and returns the device it corresponds to. Returns resErr (an error response which can be
@@ -54,6 +63,17 @@ func VerifyAccessToken(req *http.Request, deviceDB *devices.Database) (device *a
 		}
 	}
 	return
+}
+
+// GenerateAccessToken creates a new access token. Returns an error if failed to generate
+// random bytes.
+func GenerateAccessToken() (string, error) {
+	b := make([]byte, tokenByteLength)
+	if _, err := rand.Read(b); err != nil {
+		return "", err
+	}
+	// url-safe no padding
+	return base64.RawURLEncoding.EncodeToString(b), nil
 }
 
 // extractAccessToken from a request, or return an error detailing what went wrong. The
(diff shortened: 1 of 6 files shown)


src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
```diff
@@ -27,6 +29,13 @@ import (
 	"github.com/matrix-org/util"
 )
 
+// UnknownDeviceID is the default device id if one is not specified.
+var UnknownDeviceID = "unknown-device"
```
NegativeMjark: Is this specified somewhere?
Is it okay for multiple clients to share a device ID?

NegativeMjark: It looks like synapse generates a random device ID if one isn't specified?
https://github.com/matrix-org/synapse/blob/master/synapse/handlers/device.py#L82

Kegsay: > Repeated logins clobber access tokens for now (so there is only ever 1 valid access_token at any one time, meaning multi-device doesn't work). This is keyed off the device ID, which we don't yet parse out the request.

I'm aware this deviates from Synapse for now. I would rather we didn't amass a huge list of *valid* access tokens for an account, so I'm limiting it to 1 unknown device for now. Clients should be giving us device IDs.

NegativeMjark: Seems reasonable. Could you add a comment explaining this?

NegativeMjark:
Other than commenting on device ID. LGTM

src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
Kegsay: Done. https://github.com/matrix-org/dendrite/pull/122/commits/1f8da07c416b4b97d1ca259ea3eb1605aca1f96c

PR #124, opened 2017-05-26
/upload handler and storage database

superdump:
It does not yet generate thumbnails nor does it do any auth.

### DIFF
--- a/src/github.com/matrix-org/dendrite/mediaapi/fileutils/fileutils.go
+++ b/src/github.com/matrix-org/dendrite/mediaapi/fileutils/fileutils.go
@@ -1,0 +1,207 @@
+// Copyright 2017 Vector Creations Ltd
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package fileutils
+
+import (
+	"bufio"
+	"crypto/sha256"
+	"encoding/base64"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/matrix-org/dendrite/mediaapi/types"
+)
+
+// RemoveDir removes a directory and logs a warning in case of errors
+func RemoveDir(dir types.Path, logger *log.Entry) {
+	dirErr := os.RemoveAll(string(dir))
+	if dirErr != nil {
+		logger.WithError(dirErr).WithField("dir", dir).Warn("Failed to remove directory")
+	}
+}
+
+// createTempDir creates a tmp/<random string> directory within baseDirectory and returns its path
+func createTempDir(baseDirectory types.Path) (types.Path, error) {
+	baseTmpDir := path.Join(string(baseDirectory), "tmp")
+	if err := os.MkdirAll(baseTmpDir, 0770); err != nil {
+		return "", fmt.Errorf("Failed to create base temp dir: %v", err)
+	}
+	tmpDir, err := ioutil.TempDir(baseTmpDir, "")
+	if err != nil {
+		return "", fmt.Errorf("Failed to create temp dir: %v", err)
+	}
+	return types.Path(tmpDir), nil
+}
+
+// createFileWriter creates a buffered file writer with a new file at directory/filename
+// Returns the file handle as it needs to be closed when writing is complete
+func createFileWriter(directory types.Path, filename types.Filename) (*bufio.Writer, *os.File, error) {
+	filePath := path.Join(string(directory), string(filename))
+	file, err := os.Create(filePath)
+	if err != nil {
+		return nil, nil, fmt.Errorf("Failed to create file: %v", err)
+	}
+
+	return bufio.NewWriter(file), file, nil
+}
+
+func createTempFileWriter(absBasePath types.Path) (*bufio.Writer, *os.File, types.Path, error) {
+	tmpDir, err := createTempDir(absBasePath)
+	if err != nil {
+		return nil, nil, "", fmt.Errorf("Failed to create temp dir: %q", err)
+	}
+	writer, tmpFile, err := createFileWriter(tmpDir, "content")
+	if err != nil {
+		return nil, nil, "", fmt.Errorf("Failed to create file writer: %q", err)
+	}
+	return writer, tmpFile, tmpDir, nil
+}
+
+var (
+	// ErrFileIsTooLarge indicates that the uploaded file is larger than the configured maximum file size
+	ErrFileIsTooLarge = fmt.Errorf("file is too large")
+	errRead           = fmt.Errorf("failed to read response from remote server")
+	errResponse       = fmt.Errorf("failed to write file data to response body")
+	errHash           = fmt.Errorf("failed to hash file data")
+	errWrite          = fmt.Errorf("failed to write file to disk")
+)
+
+// WriteTempFile writes to a new temporary file
+func WriteTempFile(reqReader io.Reader, maxFileSizeBytes types.FileSizeBytes, absBasePath types.Path) (types.Base64Hash, types.FileSizeBytes, types.Path, error) {
+	tmpFileWriter, tmpFile, tmpDir, err := createTempFileWriter(absBasePath)
+	if err != nil {
+		return "", -1, "", err
+	}
+	defer tmpFile.Close()
+
+	limitedReader := io.LimitReader(reqReader, int64(maxFileSizeBytes))
+	// Hash the file data. The hash will be returned. The hash is useful as a
+	// method of deduplicating files to save storage, as well as a way to conduct
+	// integrity checks on the file data in the repository.
+	hasher := sha256.New()
+	teeReader := io.TeeReader(limitedReader, hasher)
+	bytesWritten, err := io.Copy(tmpFileWriter, teeReader)
+	if err != nil && err != io.EOF {
+		return "", -1, "", err
+	}
+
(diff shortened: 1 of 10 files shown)


src/github.com/matrix-org/dendrite/mediaapi/fileutils/fileutils.go
```diff
@@ -0,0 +1,196 @@
...
+	case hashLen < 1:
+		return "", fmt.Errorf("Invalid filePath (Base64Hash too short): %q", base64Hash)
+	case hashLen > 255:
+		return "", fmt.Errorf("Invalid filePath (Base64Hash too long - max 255 characters): %q", base64Hash)
+	case hashLen < 2:
+		subPath = ""
+		fileName = string(base64Hash)
+	case hashLen < 3:
+		subPath = string(base64Hash[0:1])
+		fileName = string(base64Hash[1:])
```
NegativeMjark: Can we just assume that the hash will be longer than 5 characters long?

src/github.com/matrix-org/dendrite/mediaapi/fileutils/fileutils.go
```diff
@@ -0,0 +1,196 @@
...
+// RemoveDir removes a directory and logs a warning in case of errors
+func RemoveDir(dir types.Path, logger *log.Entry) {
+	dirErr := os.RemoveAll(string(dir))
+	if dirErr != nil {
+		logger.WithError(dirErr).WithField("dir", dir).Warn("Failed to remove directory")
+	}
+}
+
+// WriteTempFile writes to a new temporary file
+func WriteTempFile(reqReader io.Reader, maxFileSizeBytes types.FileSizeBytes, absBasePath types.Path) (types.Base64Hash, types.FileSizeBytes, types.Path, error) {
```
NegativeMjark: It might be nice to wrap long argument lists like this so they are easier to read on github.
```go
func WriteTempFile(
    reqReader io.Reader, maxFileSizeBytes types.FileSizeBytes, absBasePath types.Path,
) (types.Base64Hash, types.FileSizeBytes, types.Path, error)  {
```

src/github.com/matrix-org/dendrite/mediaapi/storage/media_repository_table.go
```diff
@@ -0,0 +1,108 @@
...
+    media_origin TEXT NOT NULL,
+    -- The MIME-type of the media file as specified when uploading.
+    content_type TEXT NOT NULL,
+    -- Size of the media file in bytes.
+    file_size_bytes BIGINT NOT NULL,
+    -- When the content was uploaded in UNIX epoch ms.
+    creation_ts BIGINT NOT NULL,
+    -- The file name with which the media was uploaded.
+    upload_name TEXT NOT NULL,
+    -- A golang base64 URLEncoding string representation of a SHA-256 hash sum of the file data.
```
NegativeMjark: Why is necessary to say that it is a "golang" string here?

src/github.com/matrix-org/dendrite/mediaapi/fileutils/fileutils.go
```diff
@@ -0,0 +1,196 @@
...
+	tmpFileWriter, tmpFile, tmpDir, err := createTempFileWriter(absBasePath)
+	if err != nil {
+		return "", -1, "", err
+	}
+	defer tmpFile.Close()
+
+	limitedReader := io.LimitReader(reqReader, int64(maxFileSizeBytes))
```
NegativeMjark: What happens if the limit is exceeded? Do we return an error? or do we truncate the data?

src/github.com/matrix-org/dendrite/mediaapi/fileutils/fileutils.go
```diff
@@ -0,0 +1,196 @@
...
+	// Note: in all error and success cases, we need to remove the temporary directory
+	defer RemoveDir(tmpDir, logger)
+	duplicate := false
+	finalPath, err := GetPathFromBase64Hash(mediaMetadata.Base64Hash, absBasePath)
+	if err != nil {
+		return "", duplicate, fmt.Errorf("failed to get file path from metadata: %q", err)
+	}
+
+	var stat os.FileInfo
+	if stat, err = os.Stat(finalPath); os.IsExist(err) {
```
NegativeMjark: Using `os.IsExist` is probably wrong as it checks whether the error is due to the file existing. In practise the call to os.Stat will succeed if the file exists.

src/github.com/matrix-org/dendrite/mediaapi/writers/upload.go
```diff
@@ -149,3 +209,38 @@ func (r *uploadRequest) Validate(maxFileSizeBytes types.FileSizeBytes) *util.JSO
...
+// The order of operations is important as it avoids metadata entering the database before the file
+// is ready, and if we fail to move the file, it never gets added to the database.
+// Returns a util.JSONResponse error and cleans up directories in case of error.
+func (r *uploadRequest) storeFileAndMetadata(tmpDir types.Path, absBasePath types.Path, db *storage.Database) *util.JSONResponse {
+	finalPath, duplicate, err := fileutils.MoveFileWithHashCheck(tmpDir, r.MediaMetadata, absBasePath, r.Logger)
+	if err != nil {
+		r.Logger.WithError(err).Error("Failed to move file.")
+		return &util.JSONResponse{
+			Code: 400,
+			JSON: jsonerror.Unknown(fmt.Sprintf("Failed to upload")),
```
NegativeMjark: You can probably drop the `Sprintf` for a constant string.

src/github.com/matrix-org/dendrite/mediaapi/writers/upload.go
```diff
@@ -149,3 +209,38 @@ func (r *uploadRequest) Validate(maxFileSizeBytes types.FileSizeBytes) *util.JSO
...
+		}
+	}
+	if duplicate {
+		r.Logger.WithField("dst", finalPath).Info("File was stored previously - discarding duplicate")
+	}
+
+	if err = db.StoreMediaMetadata(r.MediaMetadata); err != nil {
```
NegativeMjark: If the file was a duplicate is it safe to store the meta data again?

NegativeMjark: I'd guess not if we are using the hash as the `media_id` and have a unique constraint on `(media_origin, media_id)`.

src/github.com/matrix-org/dendrite/mediaapi/fileutils/fileutils.go
```diff
@@ -0,0 +1,196 @@
...
+	case hashLen > 255:
+		return "", fmt.Errorf("Invalid filePath (Base64Hash too long - max 255 characters): %q", base64Hash)
+	case hashLen < 2:
+		subPath = ""
+		fileName = string(base64Hash)
+	case hashLen < 3:
+		subPath = string(base64Hash[0:1])
+		fileName = string(base64Hash[1:])
+	default:
+		subPath = path.Join(
```
NegativeMjark: Should this be `filepath.Join` ?

src/github.com/matrix-org/dendrite/mediaapi/storage/media_repository_table.go
superdump: I wrote that before I saw that it was part of an RFC so I'll reference that instead.

src/github.com/matrix-org/dendrite/mediaapi/fileutils/fileutils.go
superdump: Fixed locally.

src/github.com/matrix-org/dendrite/mediaapi/fileutils/fileutils.go
superdump: I'll add a comment explaining that the LimitReader will emit EOF when the limit is reached. As such, the data would be truncated. As the client should have reported a Content-Length claiming a 0 < length <= max file size bytes, it's mostly a precaution. But a non-EOF error would be clearer. What do you think?

NegativeMjark: I wouldn't object to having our own copy of LimitReader given that it's only 20 lines of code. However, I suspect adding a comment to explain that it is okay to truncate since we check the reported Content-Length would be enough here.

src/github.com/matrix-org/dendrite/mediaapi/fileutils/fileutils.go
superdump: We could... this is kind of a relic from when I was writing the file based on a sanitised upload name. I can just check the length of the hash is >= 3 and error if not, in order to simplify the code. Done.

src/github.com/matrix-org/dendrite/mediaapi/writers/upload.go
superdump: The hash could have been due to a remote file. I think we have to try. Alternatively we could query for localhost and the hash as the media ID and if we get a hit then we bail but the unique constraint has the same effect...?

src/github.com/matrix-org/dendrite/mediaapi/fileutils/fileutils.go
superdump: Yes. Fixed everywhere.

src/github.com/matrix-org/dendrite/mediaapi/writers/upload.go
superdump: Done.

src/github.com/matrix-org/dendrite/mediaapi/fileutils/fileutils.go
superdump: OK. Done.

src/github.com/matrix-org/dendrite/mediaapi/fileutils/fileutils.go
superdump: Hmm. I'd have to go through allll the code and wrap long lines. At this point, that is probably the job of a formatter as it feels like a time sink for time that is needed elsewhere otherwise. :/ Normally I wouldn't be against doing this at all.

src/github.com/matrix-org/dendrite/mediaapi/writers/upload.go
NegativeMjark: Ah I missed that we are already ignoring duplicate local uploads.

PR #125, opened 2017-05-31
mediaapi/writers/download: Add local download support

superdump:
There are a few scattered `FIXME`s that express uncertainties in some places. Please suggest how to address the uncertainties.

### DIFF
--- a/src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
+++ b/src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
@@ -15,7 +15,6 @@
 package writers
 
 import (
-	"database/sql"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -38,13 +37,6 @@ const mediaIDCharacters = "A-Za-z0-9_=-"
 
 // Note: unfortunately regex.MustCompile() cannot be assigned to a const
 var mediaIDRegex = regexp.MustCompile("[" + mediaIDCharacters + "]+")
-
-// Error types used by downloadRequest.getMediaMetadata
-// FIXME: make into types
-var (
-	errDBQuery    = fmt.Errorf("error querying database for media")
-	errDBNotFound = fmt.Errorf("media not found")
-)
 
 // downloadRequest metadata included in or derivable from an download request
 // https://matrix.org/docs/spec/client_server/r0.2.0.html#get-matrix-media-r0-download-servername-mediaid
@@ -121,15 +113,17 @@ func (r *downloadRequest) Validate() *util.JSONResponse {
 
 func (r *downloadRequest) doDownload(w http.ResponseWriter, cfg *config.MediaAPI, db *storage.Database, activeRemoteRequests *types.ActiveRemoteRequests) *util.JSONResponse {
 	// check if we have a record of the media in our database
-	mediaMetadata, err := r.getMediaMetadata(db)
-	if err == nil {
-		// If we have a record, we can respond from the local file
-		r.MediaMetadata = mediaMetadata
-		return r.respondFromLocalFile(w, cfg.AbsBasePath)
-	} else if err == errDBNotFound {
+	mediaMetadata, err := db.GetMediaMetadata(r.MediaMetadata.MediaID, r.MediaMetadata.Origin)
+	if err != nil {
+		r.Logger.WithError(err).Error("Error querying the database.")
+		return &util.JSONResponse{
+			Code: 500,
+			JSON: jsonerror.InternalServerError(),
+		}
+	}
+	if mediaMetadata == nil {
 		if r.MediaMetadata.Origin == cfg.ServerName {
 			// If we do not have a record and the origin is local, the file is not found
-			r.Logger.WithError(err).Warn("Failed to look up file in database")
 			return &util.JSONResponse{
 				Code: 404,
 				JSON: jsonerror.NotFound(fmt.Sprintf("File with media ID %q does not exist", r.MediaMetadata.MediaID)),
@@ -141,35 +135,9 @@ func (r *downloadRequest) doDownload(w http.ResponseWriter, cfg *config.MediaAPI
 			JSON: jsonerror.NotFound(fmt.Sprintf("File with media ID %q does not exist", r.MediaMetadata.MediaID)),
 		}
 	}
-	// Another error from the database
-	r.Logger.WithError(err).WithFields(log.Fields{
-		"MediaID": r.MediaMetadata.MediaID,
-		"Origin":  r.MediaMetadata.Origin,
-	}).Error("Error querying the database.")
-	return &util.JSONResponse{
-		Code: 500,
-		JSON: jsonerror.InternalServerError(),
-	}
-}
-
-// getMediaMetadata queries the database for media metadata
-func (r *downloadRequest) getMediaMetadata(db *storage.Database) (*types.MediaMetadata, error) {
-	mediaMetadata, err := db.GetMediaMetadata(r.MediaMetadata.MediaID, r.MediaMetadata.Origin)
-	if err != nil {
-		if err == sql.ErrNoRows {
-			r.Logger.WithFields(log.Fields{
-				"Origin":  r.MediaMetadata.Origin,
-				"MediaID": r.MediaMetadata.MediaID,
-			}).Info("Media not found in database.")
-			return nil, errDBNotFound
-		}
-		r.Logger.WithError(err).WithFields(log.Fields{
-			"Origin":  r.MediaMetadata.Origin,
-			"MediaID": r.MediaMetadata.MediaID,
-		}).Error("Error querying database for media.")
-		return nil, errDBQuery
-	}
-	return mediaMetadata, nil
+	// If we have a record, we can respond from the local file
+	r.MediaMetadata = mediaMetadata
+	return r.respondFromLocalFile(w, cfg.AbsBasePath)
 }
 
 // respondFromLocalFile reads a file from local storage and writes it to the http.ResponseWriter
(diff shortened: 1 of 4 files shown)


src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
```diff
@@ -101,3 +118,152 @@ func (r *downloadRequest) Validate() *util.JSONResponse {
...
+		Code: 500,
+		JSON: jsonerror.InternalServerError(),
+	}
+}
+
+// getMediaMetadata queries the database for media metadata
+func (r *downloadRequest) getMediaMetadata(db *storage.Database) (*types.MediaMetadata, error) {
+	mediaMetadata, err := db.GetMediaMetadata(r.MediaMetadata.MediaID, r.MediaMetadata.Origin)
+	if err != nil {
+		if err == sql.ErrNoRows {
```
NegativeMjark: I'd prefer it if we weren't checking for sql errors outside the storage package.
I'd move the check for `sql.ErrNoRows` inside `GetMediaMetadata` and have it return `nil, nil` in that case.

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
```diff
@@ -101,3 +118,152 @@ func (r *downloadRequest) Validate() *util.JSONResponse {
...
+		r.Logger.WithError(err).Warn("Failed to open file")
+		return &util.JSONResponse{
+			Code: 404,
+			JSON: jsonerror.NotFound(fmt.Sprintf("File with media ID %q does not exist", r.MediaMetadata.MediaID)),
+		}
+	}
+	stat, err := file.Stat()
+	if err != nil {
+		// FIXME: Remove erroneous file from database?
+		r.Logger.WithError(err).Warn("Failed to stat file")
```
NegativeMjark: This code should only be hit if something has gone wrong with the filesystem. I'd probably just return a 500 error in this case.

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
```diff
@@ -101,3 +118,152 @@ func (r *downloadRequest) Validate() *util.JSONResponse {
...
+		r.Logger.WithError(err).Warn("Failed to get file path from metadata")
+		return &util.JSONResponse{
+			Code: 404,
+			JSON: jsonerror.NotFound(fmt.Sprintf("File with media ID %q does not exist", r.MediaMetadata.MediaID)),
+		}
+	}
+	file, err := os.Open(filePath)
+	defer file.Close()
+	if err != nil {
+		// FIXME: Remove erroneous file from database?
```
NegativeMjark: I'd probably just return a 500 response if the filesytem and database have become out of sync like this.

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
```diff
@@ -101,3 +118,152 @@ func (r *downloadRequest) Validate() *util.JSONResponse {
...
+	}
+	return mediaMetadata, nil
+}
+
+// respondFromLocalFile reads a file from local storage and writes it to the http.ResponseWriter
+// Returns a util.JSONResponse error in case of error
+func (r *downloadRequest) respondFromLocalFile(w http.ResponseWriter, absBasePath types.Path) *util.JSONResponse {
+	filePath, err := fileutils.GetPathFromBase64Hash(r.MediaMetadata.Base64Hash, absBasePath)
+	if err != nil {
+		// FIXME: Remove erroneous file from database?
```
NegativeMjark: I'd probably just 500 if the contents of the database are corrupt like this.

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
```diff
@@ -101,3 +118,152 @@ func (r *downloadRequest) Validate() *util.JSONResponse {
...
+		// FIXME: close the connection here or just return?
+		r.closeConnection(w)
+	}
+	return nil
+}
+
+func (r *downloadRequest) closeConnection(w http.ResponseWriter) {
+	r.Logger.WithFields(log.Fields{
+		"Origin":  r.MediaMetadata.Origin,
+		"MediaID": r.MediaMetadata.MediaID,
```
NegativeMjark: Would it make sense to set the "Origin" and "MediaID" log fields on  `r.Logger` when it is set in `Download`?

```go
Logger: util.GetLogger(req.Context()).WithFields(log.Fields{
   "Origin": origin,
   "MediaID": mediaID,
})
```

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
superdump: Good suggestion. Done.

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
```diff
@@ -101,3 +118,152 @@ func (r *downloadRequest) Validate() *util.JSONResponse {
 	}
 	return nil
 }
+
+func (r *downloadRequest) doDownload(w http.ResponseWriter, cfg *config.MediaAPI, db *storage.Database, activeRemoteRequests *types.ActiveRemoteRequests) *util.JSONResponse {
+	// check if we have a record of the media in our database
+	mediaMetadata, err := r.getMediaMetadata(db)
+	if err == nil {
```
NegativeMjark: Generally I expect the error checks to be:

```go
if err != nil {
   // Error handling.
}
// Normal handling.
```

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
superdump: ...you can do that? That's awesome! Yes. Done everywhere.

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
```diff
@@ -101,3 +118,152 @@ func (r *downloadRequest) Validate() *util.JSONResponse {
...
+		// If we have a record, we can respond from the local file
+		r.MediaMetadata = mediaMetadata
+		return r.respondFromLocalFile(w, cfg.AbsBasePath)
+	} else if err == errDBNotFound {
+		if r.MediaMetadata.Origin == cfg.ServerName {
+			// If we do not have a record and the origin is local, the file is not found
+			r.Logger.WithError(err).Warn("Failed to look up file in database")
```
NegativeMjark: I wouldn't bother logging the error here. I probably wouldn't log anything at all if the client gives a media ID that doesn't exist.

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
superdump: There was a reason for doing is this way, but it's been reworked as part of addressing the `sql.ErrNoRows` 'leak'. Done.

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
superdump: OK. Done.

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
```diff
@@ -101,3 +118,152 @@ func (r *downloadRequest) Validate() *util.JSONResponse {
...
+			JSON: jsonerror.NotFound(fmt.Sprintf("File with media ID %q does not exist", r.MediaMetadata.MediaID)),
+		}
+	}
+
+	if r.MediaMetadata.FileSizeBytes > 0 && int64(r.MediaMetadata.FileSizeBytes) != stat.Size() {
+		r.Logger.WithFields(log.Fields{
+			"fileSizeDatabase": r.MediaMetadata.FileSizeBytes,
+			"fileSizeDisk":     stat.Size(),
+		}).Warn("File size in database and on-disk differ.")
+		// FIXME: Remove erroneous file from database?
```
NegativeMjark: Again I'd probably 500 here if the database and filesystem have got out of sync.

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
NegativeMjark: If you moved the sql.ErrNoRows check inside the storage package then you could probably get rid of this function and the `errDBNotFound`, and the `errDBQuery` variables entirely.

NegativeMjark: Having done that you can probably get rid of this function and the errDBNotFound and the errDBQuery variables.

superdump: Missed those two comments, but yes, I did all of that. :)

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
```diff
@@ -101,3 +113,117 @@ func (r *downloadRequest) Validate() *util.JSONResponse {
...
+
+	if bytesResponded, err := io.Copy(w, file); err != nil {
+		r.Logger.WithError(err).Warn("Failed to copy from cache")
+		if bytesResponded == 0 {
+			return &util.JSONResponse{
+				Code: 500,
+				JSON: jsonerror.NotFound(fmt.Sprintf("Failed to respond with file with media ID %q", r.MediaMetadata.MediaID)),
+			}
+		}
+		// If we have written any data then we have already responded with 200 OK and all we can do is close the connection
```
NegativeMjark: It looks like net/http already handles this case correctly. I think it should be safe to just return from the handler here.

superdump: I'm going to explicitly `return nil` inside, at the end of the `if {}` block to avoid possible future bugs doing things after that block returns control to the remainder of the function. Done.

PR #126, opened 2017-05-31
Add support for downloading of remote files

superdump:
There is some synchronisation to pay attention to around ActiveRemoteRequests to ensure that we only have one download request for remote media which gets cached and then served to all the incoming download requests from clients.

### DIFF
--- a/src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
+++ b/src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
@@ -253,10 +253,8 @@ func (r *downloadRequest) getMediaMetadataForRemoteFile(db *storage.Database, ac
 	if activeRemoteRequestResult, ok := activeRemoteRequests.MXCToResult[mxcURL]; ok {
 		r.Logger.Info("Waiting for another goroutine to fetch the remote file.")
 
+		// NOTE: Wait unlocks and locks again internally. There is still a deferred Unlock() that will unlock this.
 		activeRemoteRequestResult.Cond.Wait()
-		activeRemoteRequests.Unlock()
-		// NOTE: there is still a deferred Unlock() that will unlock this
-		activeRemoteRequests.Lock()
 
 		// check if we have a record of the media in our database
 		mediaMetadata, err := db.GetMediaMetadata(r.MediaMetadata.MediaID, r.MediaMetadata.Origin)
--- a/src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
+++ b/src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
@@ -140,10 +140,14 @@ func (r *downloadRequest) doDownload(w http.ResponseWriter, cfg *config.MediaAPI
 			}
 		}
 		// If we do not have a record and the origin is remote, we need to fetch it and respond with that file
-		return r.respondFromRemoteFile(w, cfg, db, activeRemoteRequests)
+		resErr := r.getRemoteFile(cfg, db, activeRemoteRequests)
+		if resErr != nil {
+			return resErr
+		}
+	} else {
+		// If we have a record, we can respond from the local file
+		r.MediaMetadata = mediaMetadata
 	}
-	// If we have a record, we can respond from the local file
-	r.MediaMetadata = mediaMetadata
 	return r.respondFromLocalFile(w, cfg.AbsBasePath)
 }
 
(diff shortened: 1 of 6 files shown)


src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
```diff
@@ -202,13 +198,319 @@ func (r *downloadRequest) respondFromLocalFile(w http.ResponseWriter, absBasePat
...
+
+	// No record was found
+
+	// Check if there is an active remote request for the file
+	mxcURL := "mxc://" + string(r.MediaMetadata.Origin) + "/" + string(r.MediaMetadata.MediaID)
+	if activeRemoteRequestResult, ok := activeRemoteRequests.MXCToResult[mxcURL]; ok {
+		r.Logger.Info("Waiting for another goroutine to fetch the remote file.")
+
+		activeRemoteRequestResult.Cond.Wait()
+		activeRemoteRequests.Unlock()
```
NegativeMjark: I think we want to release the `activeRemoteRequests` before calling wait. Otherwise all the other attempts to download will block.

NegativeMjark: Oh, except that the Cond is wrapping the `activeRemoteRequests` mutex.
I wonder if that works...

NegativeMjark: Hmm, maybe it does work.
You can probably remove the Unlock and Lock calls though.

superdump: It works but I don't need unlock, lock after wait: https://golang.org/src/sync/cond.go?s=1407:1428#L43

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
```diff
@@ -202,13 +198,317 @@ func (r *downloadRequest) respondFromLocalFile(w http.ResponseWriter, absBasePat
...
+	if resErr != nil {
+		result = resErr.Code
+		return resErr
+	}
+
+	// NOTE: Writing the metadata to the media repository database and removing the mxcURL from activeRemoteRequests needs to be atomic.
+	// If it were not atomic, a new request for the same file could come in in routine A and check the database before the INSERT.
+	// Routine B which was fetching could then have its INSERT complete and remove the mxcURL from the activeRemoteRequests.
+	// If routine A then checked the activeRemoteRequests it would think it needed to fetch the file when it's already in the database.
+	// The locking below mitigates this situation.
```
NegativeMjark: I don't think this is necessary. I think you only need to hold the lock when you remove the entry from the active requests map as long as you only remove the after you insert into the database.

NegativeMjark: Oh, I guess you'd need to recheck the database before downloading.

superdump: Hmm. Probably better that we discuss this in #dendrite-dev:matrix.org as it's a bit too complicated to discuss back and forth here I think.

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
```diff
@@ -202,13 +202,281 @@ func (r *downloadRequest) respondFromLocalFile(w http.ResponseWriter, absBasePat
...
+	if resErr != nil {
+		return resErr
+	} else if mediaMetadata != nil {
+		// If we got metadata from an active request, we can respond from the local file
+		r.MediaMetadata = mediaMetadata
+	} else {
+		// Note: This is an active request that MUST broadcastMediaMetadata to wake up waiting goroutines!
+		// Note: errorResponse is the named return variable
+		// Note: broadcastMediaMetadata uses mutexes and conditions from activeRemoteRequests
+		defer r.broadcastMediaMetadata(activeRemoteRequests, errorResponse)
```
NegativeMjark: Note that it is the function call that is deferred, not the evaluation of the arguments.
https://golang.org/ref/spec#Defer_statements

NegativeMjark: You are safe to assume that the errorResponse parameter will always be set by the return stmt so that part is fine.
However it wouldn't be set if something panic'd. So this might need to be written as:

```go
defer func() {
     if err := recover(); err != nil {
          resErr := jsonerror.InternalServerError()
          r.broadcastMediaMetadata(activeRemoteRequests, &resErr)
          panic(err)
     }
     r.broadcastMediaMetadata(activeRemoteRequests, errorResponse)
}
```

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
```diff
@@ -202,13 +202,288 @@ func (r *downloadRequest) respondFromLocalFile(w http.ResponseWriter, absBasePat
...
+		if dnsErr, ok := err.(*net.DNSError); ok && dnsErr.Timeout() {
+			return nil, &util.JSONResponse{
+				Code: 504,
+				JSON: jsonerror.Unknown(fmt.Sprintf("DNS look up for homeserver at %v timed out", r.MediaMetadata.Origin)),
+			}
+		}
+		resErr := jsonerror.InternalServerError()
+		return nil, &resErr
+	}
+	url := "https://" + strings.Trim(dnsResult.SRVRecords[0].Target, ".") + ":" + strconv.Itoa(int(dnsResult.SRVRecords[0].Port))
```
NegativeMjark: You should probably use 'dnsResult.Addrs[0]' here.

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
```diff
@@ -202,13 +202,288 @@ func (r *downloadRequest) respondFromLocalFile(w http.ResponseWriter, absBasePat
...
+	}
+	if duplicate {
+		r.Logger.WithField("dst", finalPath).Info("File was stored previously - discarding duplicate")
+		// Continue on to store the metadata in the database
+	}
+
+	return types.Path(finalPath), duplicate, nil
+}
+
+func (r *downloadRequest) createRemoteRequest() (*http.Response, *util.JSONResponse) {
```
NegativeMjark: This function should probably be a method on `gomatrixserverlib.Client`.
But I'm happy to see it moved in a different PR to this one.

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
superdump: Now done with the panic handling.

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
```diff
@@ -202,13 +202,288 @@ func (r *downloadRequest) respondFromLocalFile(w http.ResponseWriter, absBasePat
...
+			JSON: jsonerror.Unknown(fmt.Sprintf("File with media ID %q could not be downloaded from %q", r.MediaMetadata.MediaID, r.MediaMetadata.Origin)),
+		}
+	}
+
+	return resp, nil
+}
+
+var contentDispositionRegex = regexp.MustCompile("filename([*])?=(utf-8'')?([A-Za-z0-9._-]+)")
+
+func contentDispositionToFilename(contentDisposition string) types.Filename {
```
NegativeMjark: Note that golang has a library function for parsing Content-Disposition headers http://golang.org/pkg/mime/#ParseMediaType that might be useful here.

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go

PR #127, opened 2017-06-01
Add API for querying events by ID.

NegativeMjark:


src/github.com/matrix-org/dendrite/roomserver/api/query.go
```diff
@@ -68,6 +68,22 @@ type QueryStateAfterEventsResponse struct {
...
+type QueryEventsByIDRequest struct {
+	// The event IDs to look up.
+	EventIDs []string
+}
+
+// QueryEventsByIDResponse is a response to QueryEventsByID
+type QueryEventsByIDResponse struct {
+	// Copy of the request for debugging.
+	QueryEventsByIDRequest
+	// A list of events with the request IDs.
```
Kegsay: `s/request/requested/`

src/github.com/matrix-org/dendrite/roomserver/api/query.go
```diff
@@ -68,6 +68,22 @@ type QueryStateAfterEventsResponse struct {
...
+	// If the roomserver does not have a copy of a requested event
+	// then it will omit that event from the list.
+	Events []gomatrixserverlib.Event
```
Kegsay: What's the ordering on this array? Same as the ones given in the request (modulo omissions?). Arbitrary? Do you guarantee there will be no duplicates with the same event ID?

Kegsay: Probably also worth explicitly stating that if the roomserver HAS a copy of the event but is unable to load it, it will return an error rather than being omitted.

PR #131, opened 2017-06-02
Clean up syncapi storage to match the coding style

NegativeMjark:
The expected coding style is that the table.go files only contain functions whose name begins with a single SQL verb wrapping a single SQL statement with the same name. The only other functions in those files should be utility functions to help execute individual queries.

The main storage file which implements the component storage interface is the only place where multiple statements can be executed in a single function.

This makes it a bit easier to read the code as it will only leave the main storage file to execute a single statement, and because the functions are prefixed with the SQL verb it's usually possible to guess what is going to happen.


Kegsay:
Tests pass so LGTM

PR #132, opened 2017-06-03
mediaapi: Add thumbnail support

superdump:


superdump:
From: https://github.com/h2non/bimg#prerequisites
> Note: libvips v8.3+ is required for GIF, PDF and SVG support.

Travis is configured to use `trusty` but `xenial` wouldn't deliver a new enough `libvips` for GIF support either:
screen shot 2017-06-03 at 11 43 55

The libvips wiki notes the ppa I've added to get a newer `libvips` on `trusty`: http://www.vips.ecs.soton.ac.uk/index.php?title=Supported#Unix

superdump:
@Kegsay previously expressed a preference for pure go libraries instead of depending on libraries that have to be installed externally and bridged into go. However, the performance of the available golang image processing libraries (https://github.com/nfnt/resize) is grossly worse than something based on libvips.

The pure golang solution with nfnt/resize does not use any smart use of box filters to downscale large JPEGs (as in resolutions used in all digital cameras today) to within some percentage of the desired output resolution, and then use some traditional interpolation algorithm like bicubic or lanczos3. nfnt/resize can only use interpolation directly which is much much slower. Also, libvips has a smart multi-threaded design which splits the image up and processes those parts in parallel.

In summary, I think trying to do any media processing in pure golang instead of using mature and appropriately-optimised libraries written in other languages like C and using ASM optimisations is currently not a good choice. Perhaps the situation will change in the future but I strongly vote for bimg + libvips today and accept the cost of additional installation complexity. (I think we should ship a docker container anyway and then it's basically just pulling down binary data to execute. 😄)

Kegsay:
> libvips has a smart multi-threaded design which splits the image up and processes those parts in parallel.

Looks like nfnt/resize already does this based on the number of max procs (defaults to # cores)  https://github.com/nfnt/resize/blob/master/resize.go#L612 - so it's a shame that there no obvious room for improvement for the golang version.

ara4n:
i don't see the point in jumping through hoops to avoid native deps - surely we just use the best tool for the job?

Kegsay:
I think that it's a bit of a double standard to say "use the best tool for the job" on the one hand, but strive to eradicate Kafka/Zookeeper, to the point of emphasising this on all the Dendrite literature, on the other hand.

NegativeMjark:
I'm really not happy with adding a system dependency, especially one that requires adding a random ppa.
I'd much rather use a native golang library for resizing. Even though the performance is lower.

We can always ship a libvips version of the media repository in a separate go project if we need the additional performance.

NegativeMjark:
Alternatively go supports tags that you could use tags to make libvips conditional.

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
```diff
@@ -41,31 +43,51 @@ const mediaIDCharacters = "A-Za-z0-9_=-"
...
 			Origin:  origin,
 		},
+		IsThumbnailRequest: isThumbnailRequest,
 		Logger: util.GetLogger(req.Context()).WithFields(log.Fields{
 			"Origin":  origin,
 			"MediaID": mediaID,
 		}),
 	}
 
+	if r.IsThumbnailRequest {
```
NegativeMjark: I have a slight bias towards using separate functions for downloading and thumbnailing instead of using a condition to wedge both into a single function. Sharing the common code is nice but usually the same effect can be better handled by lifting the common code into a separate function.

However the reason I dislike having the code for the two cases wedged together is that it makes it harder to follow what is going on in a single case. That doesn't seem to be too bad in this case since it's a single bool condition and is always tested in the same way.

So while this is usually something I'd object to, I can see being acceptable here for now.

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
```diff
@@ -335,13 +497,18 @@ func (r *downloadRequest) fetchRemoteFileAndStoreMetadata(absBasePath types.Path
 		return &resErr
 	}
 
-	// TODO: generate thumbnails
+	go func() {
+		err := thumbnailer.GenerateThumbnails(finalPath, thumbnailSizes, r.MediaMetadata, activeThumbnailGeneration, db, r.Logger)
+		if err != nil {
+			r.Logger.WithError(err).Warn("Error generating thumbnails")
+		}
+	}()
```
NegativeMjark: Do we always generate the thumbnails in the background?
What happens if this is a request for thumbnails? Will this request race with the background thumbnail generation?

src/github.com/matrix-org/dendrite/mediaapi/thumbnailer/thumbnailer.go
```diff
@@ -0,0 +1,416 @@
...
+}
+
+func boolToInt(b bool) int {
+	if b {
+		return 1
+	}
+	return 0
+}
+
+func compareThumbnailMetrics(a thumbnailMetrics, b thumbnailMetrics, desiredCrop bool) int {
```
NegativeMjark: Maybe write this as:
```go
// betterThan checks whether the thumbnail metrics are a better fit than the other metrics
func (a thumbnailMetrics) betterThan(b thumbnailMetrics, desiredCrop bool) bool {
...
}

I don't think we are distinguishing between `1` and `0` at the callsites so returning a bool might be neater.

File: src/github.com/matrix-org/dendrite/mediaapi/thumbnailer/thumbnailer.go [hunk 4]
```diff
@@ -0,0 +1,416 @@
...
+	}
+
+	return options.Width, options.Height, nil
+}
+
+// init with worst values
+func newThumbnailMetrics() thumbnailMetrics {
+	return thumbnailMetrics{
+		isSmaller:      1,
+		aspect:         float64(16384 * 16384),
```
NegativeMjark: Maybe use https://golang.org/pkg/math/#Inf for the worst value?

src/github.com/matrix-org/dendrite/mediaapi/thumbnailer/thumbnailer.go
```diff
@@ -0,0 +1,416 @@
...
+		}()
+	}
+
+	// Check if the thumbnail exists.
+	thumbnailMetadata, err := db.GetThumbnail(mediaMetadata.MediaID, mediaMetadata.Origin, config.Width, config.Height, config.ResizeMethod)
+	if err != nil {
+		logger.WithFields(log.Fields{
+			"Width":        config.Width,
+			"Height":       config.Height,
+			"ResizeMethod": config.ResizeMethod,
```
NegativeMjark: Maybe do the `WithFields` once and cache the result in a local variable?

src/github.com/matrix-org/dendrite/mediaapi/thumbnailer/thumbnailer.go
```diff
@@ -0,0 +1,416 @@
...
+	if err = bimg.Write(string(dst), newImage); err != nil {
+		logger.WithError(err).Error("Failed to resize image")
+		return -1, -1, err
+	}
+
+	return options.Width, options.Height, nil
+}
+
+// init with worst values
+func newThumbnailMetrics() thumbnailMetrics {
```
NegativeMjark: Could we have a different name than `metrics` here? I'm going to confuse this with prometheus metrics otherwise.

Maybe `thumbnailFitness` ?

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
superdump: I can refactor. It's fine.

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
superdump: Thumbnails being pre-generated according to the configuration of the server are generated in the background. There is a race in time, but it's done safely using the `activeThumbnailGeneration` synchronised map of thumbnail file name to a `sync.Cond` and the result of the thumbnail attempt, in a very similar way to the download of remote files.

For example (and I have tested this and specifically made sure that this works), say that we are running our server configured to pre-generate a bunch of thumbnail formats including 800x600 using the scale method. If a thumbnail request is received for an 800x600 scaled version of some image that is hosted on a remote server and we don't yet have that file, the server downloads the file from the remote server and then fires off a separate go routine (the code commented on here) to do the pre-generated thumbnailing. Each thumbnail configuration in turn will lock `activeThumbnailGeneration`, check if there's an active thumbnail generation job ongoing for that thumbnail, etc and if not, add a sync.Cond to the map and unlock. If there is an active thumbnail generation job ongoing, it waits on the sync.Cond. So it doesn't matter if the pre-generation job gets to generating that thumbnail first, or if the code that continues after this explicitly and synchronously requests to generate that specific thumbnail. First come, first served and the rest will wait or just find the file.

src/github.com/matrix-org/dendrite/mediaapi/thumbnailer/thumbnailer.go
superdump: Sure.

src/github.com/matrix-org/dendrite/mediaapi/thumbnailer/thumbnailer.go
superdump: I was actually looking at `MaxFloat64` and previously I was doing some `b.aspect - a.aspect` and was concerned that the precision with such a large number would be such that the subtraction would be inconsequential.

As such, I took something more reasonably within the realm of an acceptable max image resolution. However, now I am comparing the numbers directly I don't have the same precision problem.

I knew this would provoke comment but I forgot to comment it to explain. How do you want me to proceed - comment, `Inf()` or `MaxFloat64`?

src/github.com/matrix-org/dendrite/mediaapi/thumbnailer/thumbnailer.go
superdump: I did consider that but had no strong argument either way - I'll do it.

src/github.com/matrix-org/dendrite/mediaapi/thumbnailer/thumbnailer.go
superdump: OK.

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
superdump: Actually, I changed my mind. I think this is better. 😄

src/github.com/matrix-org/dendrite/mediaapi/thumbnailer/thumbnailer.go
superdump: Done.

src/github.com/matrix-org/dendrite/mediaapi/thumbnailer/thumbnailer.go
superdump: Done. Both in a few places in `download.go` and in `thumbnailer.go`.

src/github.com/matrix-org/dendrite/mediaapi/thumbnailer/thumbnailer.go
superdump: Done.

src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
NegativeMjark: What happens if dynamic thumbnail generation is turned off?

superdump: It's basically all covered here - pay attention to the control flow and hopefully it'll be clear: https://github.com/matrix-org/dendrite/blob/a1e62a2b0b9f76309a8aad96c5bd3fdeef8982f6/src/github.com/matrix-org/dendrite/mediaapi/writers/download.go#L292-L327

src/github.com/matrix-org/dendrite/mediaapi/thumbnailer/thumbnailer.go
NegativeMjark: I think Inf() makes the most sense.

superdump: Done.

superdump:
bimg:
thumbnail-96x96-crop

nfnt:
thumbnail-96x96-crop

nfnt is slightly sharper but uses more cpu and takes longer (~845ms for the matrix.org config)
bimg is less sharp, uses less cpu and takes less time (~295ms for the matrix.org config)

superdump:
I added a comment and a couple of small sample images to the README. Thanks for reviewing and glad you like it. 😄

PR #134, opened 2017-06-03
syncapi: Add configuration for the account database

superdump:


superdump:
This is broken. Some more work is needed to use the correct databases in the integration tests but I don't have time to look into it to understand which databases the create-account CLI should be using right now.

NegativeMjark:
This has been made somewhat obsolete by the config unification in #141

PR #135, opened 2017-06-05
Fetching missing state from the roomserver.

NegativeMjark:
Whenever the syncserver receives an event from the room server that adds
state that isn't in the syncserver's local database it should fetch
those state events from the roomserver.

### DIFF
--- a/src/github.com/matrix-org/dendrite/syncapi/consumers/roomserver.go
+++ b/src/github.com/matrix-org/dendrite/syncapi/consumers/roomserver.go
@@ -16,6 +16,7 @@ package consumers
 
 import (
 	"encoding/json"
+	"fmt"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/matrix-org/dendrite/common"
@@ -33,6 +34,7 @@ type OutputRoomEvent struct {
 	roomServerConsumer *common.ContinualConsumer
 	db                 *storage.SyncServerDatabase
 	notifier           *sync.Notifier
+	query              api.RoomserverQueryAPI
 }
 
 // NewOutputRoomEvent creates a new OutputRoomEvent consumer. Call Start() to begin consuming from room servers.
@@ -51,6 +53,7 @@ func NewOutputRoomEvent(cfg *config.Sync, n *sync.Notifier, store *storage.SyncS
 		roomServerConsumer: &consumer,
 		db:                 store,
 		notifier:           n,
+		query:              api.NewRoomserverQueryAPIHTTP(cfg.RoomserverURL, nil),
 	}
 	consumer.ProcessMessage = s.onMessage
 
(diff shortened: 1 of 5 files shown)


src/github.com/matrix-org/dendrite/syncapi/consumers/roomserver.go
```diff
@@ -100,3 +115,74 @@ func (s *OutputRoomEvent) onMessage(msg *sarama.ConsumerMessage) error {
 
 	return nil
 }
+
+// addsStateEvents looks up the state events that are added by a new event.
+func (s *OutputRoomEvent) addsStateEvents(
```
Kegsay: Naming: maybe `lookupStateEvents` is more accurate?

src/github.com/matrix-org/dendrite/syncapi/storage/syncserver.go
```diff
@@ -323,6 +324,49 @@ func (d *SyncServerDatabase) fetchStateEvents(txn *sql.Tx, roomIDToEventIDSet ma
...
+	stateEvents, err := d.roomstate.selectEventsWithEventIDs(txn, missing)
+
+	if err != nil {
+		return nil, err
+	}
+	if len(stateEvents) != len(missing) {
+		return nil, fmt.Errorf("failed to map all event IDs to events: (got %d, wanted %d)", len(stateEvents), len(missing))
+	}
+	for _, e := range stateEvents {
+		// Set the stream position to 0 since these events occured outside the
```
Kegsay: > occured

src/github.com/matrix-org/dendrite/syncapi/storage/syncserver.go
```diff
@@ -323,6 +324,49 @@ func (d *SyncServerDatabase) fetchStateEvents(txn *sql.Tx, roomIDToEventIDSet ma
...
+		// stream so probably happened before it.
+		// TOOD: What happens if we receive a state event from outside the
+		// timeline associated with an event in the middle of the timeline?
```
Kegsay: > TOOD

I've mulled over this. This scenario is hit when we get arbitrary state events from federation. It is the room servers job to keep a consistent current room state based on deltas to add/remove state event IDs. The room server has instructed us that this forms part of the current state and therefore should be in the current state table. However, the room server has not told us which stream position the event should be in. 

We're *assuming* they happened outside the stream (backfill from federation or equiv) and hence setting the stream position to 0. This may change when we eventually implement backfill and realise it should be `-435` or whatever. I think this behaviour is probably okay for now, but will need to be revisited when we implement backfill.

Kegsay: from rl: The stream position here needs to be the position in the kafka log which added these missing state events. We don't know this value at this point in time, so we'll need to remember it at insert time (add a column to current_room_state which has the stream position for the cases where we've fetched them from the room server).

src/github.com/matrix-org/dendrite/syncapi/storage/current_room_state_table.go
```diff
@@ -36,6 +36,9 @@ CREATE TABLE IF NOT EXISTS current_room_state (
     -- The 'content.membership' value if this event is an m.room.member event. For other
     -- events, this will be NULL.
     membership TEXT,
+	-- The serial ID of the output_room_events table when this event became
+	-- part of the current state of the room.
+	added_at BIGINT,
```
Kegsay: Tabs.

src/github.com/matrix-org/dendrite/syncapi/storage/current_room_state_table.go
```diff
@@ -45,9 +49,10 @@ CREATE INDEX IF NOT EXISTS membership_idx ON current_room_state(type, state_key,
 `
 
 const upsertRoomStateSQL = "" +
-	"INSERT INTO current_room_state (room_id, event_id, type, state_key, event_json, membership) VALUES ($1, $2, $3, $4, $5, $6)" +
+	"INSERT INTO current_room_state (room_id, event_id, type, state_key, event_json, membership, added_at_id)" +
```
Kegsay: The table schema is `added_at` not `added_at_id`. Choose one.

Kegsay:
Otherwise LGTM, I don't need to see this again.

PR #136, opened 2017-06-07
Foundation for media API testing

superdump:
Including moving some common stuff from the syncserver-integration-tests to common/test.


src/github.com/matrix-org/dendrite/cmd/mediaapi-integration-tests/main.go
```diff
@@ -0,0 +1,178 @@
...
+		host,
+		"api",
+		"_matrix",
+		"media",
+		"v1",
+		endpoint,
+	}
+	pathComponents = append(pathComponents, components...)
+	return scheme + path.Join(pathComponents...)
+}
```
NegativeMjark: I wonder if this would be nicer written as:

```go
func getMediaURI(scheme, host, endpoint string, components ...string) string {
 	pathComponents := []string{host,"/api/_matrix/media/v1", endpoint}
 	pathComponents = append(pathComponents, components...)
 	return scheme + path.Join(pathComponents...)
 }
```

src/github.com/matrix-org/dendrite/cmd/mediaapi-integration-tests/main.go
```diff
@@ -0,0 +1,178 @@
...
+
+func testDownload(suffix, origin, mediaID, wantedBody string, wantedStatusCode int, serverCmdChan chan error) {
+	req, err := http.NewRequest(
+		"GET",
+		getMediaURI("http://", "localhost:1777"+suffix, "download", []string{
+			origin,
+			mediaID,
+		}),
```
NegativeMjark: I don't really understand the point of the host suffix if you are still passing the origin in it's entirety.
I think it would be clearer to always pass a complete host string to testDownload, rather than hardcoding it in one of the places it is used.

src/github.com/matrix-org/dendrite/cmd/mediaapi-integration-tests/main.go
superdump: Yup. Done.

src/github.com/matrix-org/dendrite/cmd/mediaapi-integration-tests/main.go
```diff
@@ -0,0 +1,178 @@
...
+
+	// create server1 with only pre-generated thumbnails allowed
+	server1Cmd, server1CmdChan, server1Dir := startMediaAPI("1", false)
+	defer cleanUpServer(server1Cmd, server1Dir)
+	testDownload("1", "localhost:17771", "doesnotexist", "", 404, server1CmdChan)
+
+	// create server2 with dynamic thumbnail generation
+	server2Cmd, server2CmdChan, server2Dir := startMediaAPI("2", true)
+	defer cleanUpServer(server2Cmd, server2Dir)
+	testDownload("2", "localhost:17772", "doesnotexist", "", 404, server2CmdChan)
```
NegativeMjark: You could make startMediaAPI return the addresses it is listening on so that you could avoid hard coding the server name here?

src/github.com/matrix-org/dendrite/cmd/mediaapi-integration-tests/main.go
```diff
@@ -0,0 +1,178 @@
...
+func makeConfig(suffix, basePath string, dynamicThumbnails bool) string {
+	return fmt.Sprintf(
+		`
+server_name: "%s"
+base_path: %s
+max_file_size_bytes: %s
+database: "%s"
+dynamic_thumbnails: %s
+%s`,
+		"localhost:1777"+suffix,
```
NegativeMjark: The name of the server is the "public" address for federation traffic, so will probably be different to the address that the media-api-server is listening on. The media-api-server won't be receiving federation traffic directly since it will need to go through a proxy to add the `/api/` prefix to the path.

src/github.com/matrix-org/dendrite/cmd/mediaapi-integration-tests/main.go
superdump: An artefact of refactoring. Done.

src/github.com/matrix-org/dendrite/cmd/mediaapi-integration-tests/main.go
superdump: Done.

src/github.com/matrix-org/dendrite/cmd/mediaapi-integration-tests/main.go
superdump: That's true of a real deployment, yes. But this is for integration testing where I will try to run only what is needed to do testing. A reverse proxy is not necessary imo.

superdump: Oh, although if I want to fetch a file from the other server, it's going to need to sit behind a reverse proxy as well. 💩

superdump: Done.

superdump:
At this point, the test runs as follows: https://github.com/matrix-org/dendrite/blob/37d3a1b3052478b8361d950353307cf183c63ea0/src/github.com/matrix-org/dendrite/cmd/mediaapi-integration-tests/main.go#L151-L184

NegativeMjark:
Could you add an paragraph to `mediaapi-integration-tests/TESTS.md` to explain what it is? Is the intention that it will track our progress on implementing the tests? Is is up to date?

.travis.yml
```diff
@@ -18,6 +21,9 @@ install:
  - go get github.com/golang/lint/golint
  - go get github.com/fzipp/gocyclo
 
+before_script:
+ - openssl req -x509 -newkey rsa:4096 -keyout server.key -out server.crt -days 365 -nodes -subj /CN=localhost
```
NegativeMjark: Maybe add a comment along the lines of:
```
# Generate a self-signed X.509 certificate for TLS.
```

 I'm not entirely sure that openssl is self-explanatory enough for the casual reader.

superdump: Done.

src/github.com/matrix-org/dendrite/cmd/syncserver-integration-tests/main.go
```diff
@@ -33,23 +30,25 @@ import (
...
 	// How long to wait for the syncserver to write the expected output messages.
 	// This needs to be high enough to account for the time it takes to create
 	// the postgres database tables which can take a while on travis.
-	timeoutString = defaulting(os.Getenv("TIMEOUT"), "10s")
+	timeoutString = test.Defaulting(os.Getenv("TIMEOUT"), "10s")
 	// The name of maintenance database to connect to in order to create the test database.
-	postgresDatabase = defaulting(os.Getenv("POSTGRES_DATABASE"), "postgres")
+	postgresDatabase = test.Defaulting(os.Getenv("POSTGRES_DATABASE"), "postgres")
+	// Postgres docker container name (for running psql)
+	postgresContainerName = os.Getenv("POSTGRES_CONTAINER")
```
NegativeMjark: Maybe add a comment explaining that if this is unset then psql is run normally outside the container?

superdump: Done.

PR #140, opened 2017-06-14
Add full docker-compose setup

superdump:


superdump:
I remembered that I forgot this. 😄

docker/docker-compose.yaml
```diff
@@ -0,0 +1,170 @@
+version: '3'
```
NegativeMjark: My docker only seems to understand up to version '2' of the config format.
Replacing '3' with '2' doesn't seem to result in any obvious problems for me.
Can we use version '2' here? Or do you know that you are using version '3' features of docker?

docker/docker-compose.yaml
```diff
@@ -0,0 +1,170 @@
...
+        image: postgres:9.6
+        environment:
+            POSTGRES_PASSWORD: $POSTGRES_PASSWORD
+        volumes:
+            - ./init-dendrite-dbs.sh:/docker-entrypoint-initdb.d/init-dendrite-dbs.sh
+            - $HOME/dendrite/postgres:/var/lib/postgresql/data
+        networks:
+            - backend
+        ports:
+            - "5432:5432"
```
NegativeMjark: Umm, I already have a postgres installed so I'm seeing bind errors trying to start the dockerised postgres here. Would it be okay if we used a different port here to avoid the clash?

docker-build.sh
```diff
@@ -0,0 +1,21 @@
...
+mkdir -p docker/bin
+cp bin/*linux-amd64 docker/bin/
+
+cd docker
+
+for cli in {client,federation}-api-proxy dendrite-{{client,federation,media,sync}-api,room}-server; do
+    dockerfile=Dockerfile.$cli
+    cat <<EOF > $dockerfile
+FROM scratch
+COPY bin/$cli-linux-amd64 $cli
```
NegativeMjark: I had to copy a couple of shared libraries across to get this to work.

```bash
$ ldd docker/bin/client-api-proxy-linux-amd64 
        linux-vdso.so.1 =>  (0x00007ffdd22d8000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f47ace76000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f47acaac000)
        /lib64/ld-linux-x86-64.so.2 (0x000055fd2b791000)
```

docker/generate-keys.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+set -e
```
NegativeMjark: Maybe add a "mkdir -p certs" to create the directory if it doesn't exist.

docker/docker-compose.yaml
```diff
@@ -0,0 +1,170 @@
...
+            - /var/run/docker.sock:/var/run/docker.sock
+        networks:
+            - backend
+        expose:
+            - "9092"
+        depends_on:
+            - zookeeper
+        restart: unless-stopped
+    postgres:
+        image: postgres:9.6
```
NegativeMjark: Would it make sense to use the alpine linux version to avoid pulling in a full ubuntu install?
postgres:9.6-alpine does the trick for me.

docker/docker-compose.yaml
```diff
@@ -0,0 +1,170 @@
+version: '3'
+services:
+    zookeeper:
+        image: wurstmeister/zookeeper
```
NegativeMjark: Should we be using the "official" zookeeper image here?

docker/docker-compose.yaml
```diff
@@ -0,0 +1,170 @@
...
+        networks:
+            - backend
+        expose:
+            - "2181"
+        restart: unless-stopped
+    kafka:
+        image: wurstmeister/kafka:0.10.2.0
```
NegativeMjark: I couldn't find an "official" kafka image. Are we happy using this guys image here?

docker/docker-compose.yaml
superdump: We could use '2'. It's mostly to encourage using the current version of things. Looking at https://docs.docker.com/compose/compose-file/compose-versioning/#version-2x-to-3x I didn't use anything from 3, but 2 is a minimum for the networks and services.

docker/docker-compose.yaml
superdump: The only reason I publish the port on the host here (which is what the 'ports' directive does) is to be able to use `create-account`. You can change it, but you'll have to adjust the database URIs in `dendrite-config.yaml` appropriately.

docker-build.sh
superdump: Do you have something in your environment that causes non-static go binaries to be built? I cross-compiled on macOS using golang 1.8.3 and got completely static binaries I believe.

docker/generate-keys.sh
superdump: Yeah. Done.

docker/docker-compose.yaml
superdump: We can, though many images are anyway built on the ubuntu base image. I believe the zookeeper image is.

docker/docker-compose.yaml
superdump: Huh. I had somehow missed that there was an official zookeeper image. I'll have to have a play and get that working.

docker/docker-compose.yaml
superdump: We can either use it or build our own in-tree, which I would be fine with but it'll take a bit more time. We could base it on the alpine variants of openjdk.

superdump:
When I get a moment, I'll switch to the official zookeeper image and make a Dockerfile for kafka.

docker/docker-compose.yaml
superdump: Done btw.

docker-build.sh
yangm97: Why don’t you use Docker multi stage builds?

docker/docker-compose.yaml
yangm97: You could treat the `docker-compose.yml` file as “production ready” or something like that, and add (development) niceties like database port forwarding to `docker-compose.override.yml` – this is an override file that is read by default by `docker-compose up`. Lastly, add `docker-compose.custon.yml` to gitignore so people can make their own overrides.

yangm97:
Any updates on this?

erikjohnston:
Closing for now due to lack of progress, though we may want to look at this again when @superdump has time

PR #141, opened 2017-06-16
Add a single config file format for dendrite.

NegativeMjark:
This config file combines all the configuration needed by the
dendrite components into a single file.

The format is mostly inspired by the mediaapi config, but has the following differences:

 1) The file is divided into sections grouped by the subsystem being configured.
 2) The support for paths that begin with "~" has been dropped because it was too difficult to support in the tests.

### DIFF
--- a/src/github.com/matrix-org/dendrite/common/config/config.go
+++ b/src/github.com/matrix-org/dendrite/common/config/config.go
@@ -44,17 +44,17 @@ type Dendrite struct {
 		ServerName gomatrixserverlib.ServerName `yaml:"server_name"`
 		// Path to the private key which will be used to sign requests and events.
 		PrivateKeyPath Path `yaml:"private_key"`
+		// The private key which will be used to sign requests and events.
+		PrivateKey ed25519.PrivateKey `yaml:"-"`
+		// An arbitrary string used to uniquely identify the PrivateKey. Must start with the
+		// prefix "ed25519:".
+		KeyID gomatrixserverlib.KeyID `yaml:"-"`
 		// List of paths to X509 certificates used by the external federation listeners.
 		// These are used to calculate the TLS fingerprints to publish for this server.
 		// Other matrix servers talking to this server will expect the x509 certificate
 		// to match one of these certificates.
 		// The certificates should be in PEM format.
 		FederationCertificatePaths []Path `yaml:"federation_certificates"`
-		// The private key which will be used to sign requests and events.
-		PrivateKey ed25519.PrivateKey `yaml:"-"`
-		// An arbitrary string used to uniquely identify the PrivateKey. Must start with the
-		// prefix "ed25519:".
-		KeyID gomatrixserverlib.KeyID `yaml:"-"`
 		// A list of SHA256 TLS fingerprints for the X509 certificates used by the
 		// federation listener for this server.
 		TLSFingerPrints []gomatrixserverlib.TLSFingerprint `yaml:"-"`
@@ -62,6 +62,7 @@ type Dendrite struct {
 		// Increasing this number will reduce the number of requests made by remote servers
 		// for our key, but increases the period a compromised key will be considered valid
 		// by remote servers.
+		// Defaults to 24 hours.
 		KeyValidityPeriod time.Duration `yaml:"key_validity_period"`
 	} `yaml:"matrix"`
 
@@ -98,12 +99,24 @@ type Dendrite struct {
 
 	// Postgres Config
 	Database struct {
-		MediaServer DataSource `yaml:"media_server"`
-		Account     DataSource `yaml:"account"`
-		Device      DataSource `yaml:"device"`
-		ServerKey   DataSource `yaml:"server_key"`
-		SyncServer  DataSource `yaml:"sync_server"`
-		RoomServer  DataSource `yaml:"room_server"`
+		// The Account database stores the login details and account information
+		// for local users. It is accessed by the ClientAPI.
+		Account DataSource `yaml:"account"`
+		// The Device database stores session information for the devices of logged
+		// in local users. It is accessed by the ClientAPI, the MediaAPI and the SyncAPI.
+		Device DataSource `yaml:"device"`
+		// The MediaAPI database stores information about files uploaded and downloaded
+		// by local users. It is only accessed by the MediaAPI.
+		MediaAPI DataSource `yaml:"media_api"`
+		// The ServerKey database caches the public keys of remote servers.
+		// It may be accessed by the FederationAPI, the ClientAPI, and the MediaAPI.
+		ServerKey DataSource `yaml:"server_key"`
+		// The SyncAPI stores information used by the SyncAPI server.
+		// It is only accessed by the SyncAPI server.
+		SyncAPI DataSource `yaml:"sync_api"`
+		// The RoomServer database stores information about matrix rooms.
+		// It is only accessed by the RoomServer.
+		RoomServer DataSource `yaml:"room_server"`
 	} `yaml:"database"`
 
 	// The internal addresses the components will listen on.
@@ -150,11 +163,13 @@ func Load(configPath string) (*Dendrite, error) {
 	if err != nil {
 		return nil, err
 	}
-	configDirPath, err := filepath.Abs(".")
+	basePath, err := filepath.Abs(".")
 	if err != nil {
 		return nil, err
 	}
-	return loadConfig(configDirPath, configData, ioutil.ReadFile)
+	// Pass the current working directory and ioutil.ReadFile so that they can
+	// be mocked in the tests
+	return loadConfig(basePath, configData, ioutil.ReadFile)
 }
 
 // An Error indicates a problem parsing the config.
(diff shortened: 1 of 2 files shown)


NegativeMjark:
I've also used a PEM format for storing the matrix ed25519 keys, which will hopefully be a bit nicer.

src/github.com/matrix-org/dendrite/common/config/config.go
```diff
@@ -0,0 +1,339 @@
...
+
+// Version is the current version of the config format.
+// This will change whenever we make breaking changes to the config format.
+const Version = "v0"
+
+// Dendrite contains all the config used by a dendrite process.
+// Relative paths are resolved relative to the current working directory
+type Dendrite struct {
+	// The version of the configuration file.
+	Version string `yaml:"version"`
```
superdump: Is `Version` a type or just a member of Dendrite? I'm wondering because maybe we necessarily want these members to be typed so we can use parts of this struct rather than passing around the whole thing.

src/github.com/matrix-org/dendrite/common/config/config.go
```diff
@@ -0,0 +1,339 @@
...
+		// The certificates should be in PEM format.
+		FederationCertificatePaths []Path `yaml:"federation_certificates"`
+		// The private key which will be used to sign requests and events.
+		PrivateKey ed25519.PrivateKey `yaml:"-"`
+		// An arbitrary string used to uniquely identify the PrivateKey. Must start with the
+		// prefix "ed25519:".
+		KeyID gomatrixserverlib.KeyID `yaml:"-"`
+		// A list of SHA256 TLS fingerprints for the X509 certificates used by the
+		// federation listener for this server.
+		TLSFingerPrints []gomatrixserverlib.TLSFingerprint `yaml:"-"`
```
superdump: Very minor: maybe put this next to `FederationCertificatePaths`?

src/github.com/matrix-org/dendrite/common/config/config.go
```diff
@@ -0,0 +1,339 @@
...
+			// Topic for roomserver/api.InputRoomEvent events.
+			InputRoomEvent Topic `yaml:"input_room_event"`
+			// Topic for roomserver/api.OutputRoomEvent events.
+			OutputRoomEvent Topic `yaml:"output_room_event"`
+		}
+	} `yaml:"kafka"`
+
+	// Postgres Config
+	Database struct {
+		MediaServer DataSource `yaml:"media_server"`
```
superdump: The `*Server` seems like inconsistent naming, aside from `RoomServer`. `MediaAPI`, `SyncAPI`?

src/github.com/matrix-org/dendrite/common/config/config.go
```diff
@@ -0,0 +1,339 @@
...
+		Account     DataSource `yaml:"account"`
+		Device      DataSource `yaml:"device"`
+		ServerKey   DataSource `yaml:"server_key"`
```
superdump: Maybe add a comment clarifying what are these three databases (`Account`, `Device`, and `ServerKey`)?

src/github.com/matrix-org/dendrite/common/config/config.go
```diff
@@ -0,0 +1,339 @@
...
+		SyncServer  DataSource `yaml:"sync_server"`
```
superdump: As mentioned above, I think this should be called `SyncAPI` for consistency.

src/github.com/matrix-org/dendrite/common/config/config.go
```diff
@@ -0,0 +1,339 @@
...
+		MediaAPI      Address `yaml:"media_api"`
+		ClientAPI     Address `yaml:"client_api"`
+		FederationAPI Address `yaml:"federation_api"`
+		SyncAPI       Address `yaml:"sync_api"`
+		RoomServer    Address `yaml:"room_server"`
+	} `yaml:"listen"`
+}
+
+// A Path on the filesystem.
+type Path string
```
superdump: Move all these to a `common/types/types.go` perhaps?

src/github.com/matrix-org/dendrite/common/config/config.go
```diff
@@ -0,0 +1,339 @@
...
+	ResizeMethod string `yaml:"method,omitempty"`
+}
+
+// Load a yaml config file
+func Load(configPath string) (*Dendrite, error) {
+	configData, err := ioutil.ReadFile(configPath)
+	if err != nil {
+		return nil, err
+	}
+	configDirPath, err := filepath.Abs(".")
```
superdump: Maybe call this what it is - the current working directory?

src/github.com/matrix-org/dendrite/common/config/config.go
```diff
@@ -0,0 +1,339 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return loadConfig(configDirPath, configData, ioutil.ReadFile)
```
superdump: Why do you pass `ioutil.ReadFile` as an argument?

src/github.com/matrix-org/dendrite/common/config/config.go
```diff
@@ -0,0 +1,339 @@
...
+		// prefix "ed25519:".
+		KeyID gomatrixserverlib.KeyID `yaml:"-"`
+		// A list of SHA256 TLS fingerprints for the X509 certificates used by the
+		// federation listener for this server.
+		TLSFingerPrints []gomatrixserverlib.TLSFingerprint `yaml:"-"`
+		// How long a remote server can cache our server key for before requesting it again.
+		// Increasing this number will reduce the number of requests made by remote servers
+		// for our key, but increases the period a compromised key will be considered valid
+		// by remote servers.
+		KeyValidityPeriod time.Duration `yaml:"key_validity_period"`
```
superdump: Document that the default is 24 hours.

src/github.com/matrix-org/dendrite/common/config/config.go
```diff
@@ -0,0 +1,339 @@
...
+	}
+
+	return nil
+}
+
+func absPath(dir string, path Path) string {
+	if filepath.IsAbs(string(path)) {
+		return filepath.Clean(string(path))
+	}
+	return filepath.Join(dir, string(path))
```
superdump: Maybe comment that `filepath.Join()` calls `filepath.Clean()` on the result so both are cleaned.

src/github.com/matrix-org/dendrite/common/config/config.go
superdump: ...to facilitate testing. Please add a comment to that effect.

src/github.com/matrix-org/dendrite/common/config/config_test.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	return []byte(data), nil
+}
+
+func TestReadKey(t *testing.T) {
+	keyID, _, err := readKeyPEM("path/to/key", []byte(testKey))
+	if err != nil {
+		t.Error("failed to load private key:", err)
+	}
+	wantKeyID := "ed25519:c8NsuQ"
```
superdump: Maybe make this string a constant and use the constant wherever this string appears?

src/github.com/matrix-org/dendrite/common/config/config.go
NegativeMjark: Done 7fa2005

src/github.com/matrix-org/dendrite/common/config/config.go
NegativeMjark: Done 7fa2005

src/github.com/matrix-org/dendrite/common/config/config.go
NegativeMjark: Done 7fa2005

src/github.com/matrix-org/dendrite/common/config/config.go
NegativeMjark: Deon 7fa2005

src/github.com/matrix-org/dendrite/common/config/config.go
NegativeMjark: Done 7fa2005

src/github.com/matrix-org/dendrite/common/config/config.go
NegativeMjark: I've renamed this to `basePath` in 7fa2005.

I'm hedging my bets slightly because I can't make up my mind if the paths should be relative to the config file, relative to the cwd, relative to a configured basepath or if relative paths should be outlawed entirely from the config.

src/github.com/matrix-org/dendrite/common/config/config.go
NegativeMjark: Done 7fa2005

src/github.com/matrix-org/dendrite/common/config/config.go
NegativeMjark: Done c4d34db

src/github.com/matrix-org/dendrite/common/config/config_test.go
NegativeMjark: Done c4d34db

src/github.com/matrix-org/dendrite/common/config/config.go
superdump: This reminds me of an oddity in docker-compose where you could give it a series of yaml files which would overlay each other, with later ones overriding earlier ones. I think the behaviour was that relative paths were relative to the dirname of the first config file, which seems a bit batshit crazy to my mind. I was definitely expecting that they would be relative to cwd.

If you're going to have multiple config files that are combined, I could understand possibly considering having them relative to the config file so that config files can be relative to things they reference and achieve some consistency that way. However, I think relative to cwd is the default expectation unless otherwise specified.

src/github.com/matrix-org/dendrite/common/config/config.go
NegativeMjark: Maybe. At the moment it's kind of nice because any function that takes them ends up looking like

```go
func foo(path config.Path) {}
```

which is kind of nice because you can see that the value is coming from the config.

src/github.com/matrix-org/dendrite/common/config/config.go
NegativeMjark: I've added some documentation for the version string in 14a57a7

src/github.com/matrix-org/dendrite/common/config/config.go
NegativeMjark: Current working directory probably matches peoples expectations the best then. I'll leave it as it is.

src/github.com/matrix-org/dendrite/common/config/config.go
superdump: I wasn't specifically concerned about the `Version` I was rather wondering how one could pass only part of this `Dendrite` struct around when one doesn't need all of it. So I was thinking that maybe the different parts of it should have their own types defined and used here, instead of a monolith struct as the only available type.

NegativeMjark: That would be trivial to do if it turned out to be useful since it would just be a cut+paste job with no other code changes required. I kind of like having all the structure in one place for now until we find a use case.

superdump: We can split this into separate types later, if needed. I'm fine with the nested struct for now.

superdump:
LGTM.

PR #143, opened 2017-06-19
Include the state before the event in roomserver output

NegativeMjark:
We need to know the state of the room before the event in the graph in order to know which servers to send the event to.

This can be different to the current state of the room, for example when there are multiple forward extremities in the event graph.

However since it *should* mostly be the same as the current state we express the state before the event as a delta from the current state.

I *think* the visibility of the event can be determined from the combination current state and and state at the event, so I think it is best to drop the unused VisibilityEventIDs field from the output.

### DIFF
--- a/src/github.com/matrix-org/dendrite/cmd/roomserver-integration-tests/main.go
+++ b/src/github.com/matrix-org/dendrite/cmd/roomserver-integration-tests/main.go
@@ -336,7 +336,8 @@ func main() {
 				"state_key":"@richvdh:matrix.org",
 				"type":"m.room.member"
 			},
-			"VisibilityEventIDs":null,
+			"StateBeforeRemovesEventIDs":["$1463671339126270PnVwC:matrix.org"],
+			"StateBeforeAddsEventIDs":null,
 			"LatestEventIDs":["$1463671339126270PnVwC:matrix.org"],
 			"AddsStateEventIDs":["$1463671337126266wrSBX:matrix.org", "$1463671339126270PnVwC:matrix.org"],
 			"RemovesStateEventIDs":null,
--- a/src/github.com/matrix-org/dendrite/roomserver/api/output.go
+++ b/src/github.com/matrix-org/dendrite/roomserver/api/output.go
@@ -58,7 +58,6 @@ func (ore *OutputRoomEvent) UnmarshalJSON(data []byte) error {
 	// being base64 encoded which is the default for []byte.
 	var content struct {
 		Event                      *json.RawMessage
-		VisibilityEventIDs         []string
 		LatestEventIDs             []string
 		AddsStateEventIDs          []string
 		RemovesStateEventIDs       []string
@@ -90,7 +89,6 @@ func (ore OutputRoomEvent) MarshalJSON() ([]byte, error) {
 	event := json.RawMessage(ore.Event)
 	content := struct {
 		Event                      *json.RawMessage
-		VisibilityEventIDs         []string
 		LatestEventIDs             []string
 		AddsStateEventIDs          []string
 		RemovesStateEventIDs       []string
--- a/src/github.com/matrix-org/dendrite/roomserver/api/output.go
+++ b/src/github.com/matrix-org/dendrite/roomserver/api/output.go
@@ -19,6 +19,14 @@ import (
 )
 
 // An OutputRoomEvent is written when the roomserver receives a new event.
+// It contains the full matrix room event and enough information for a
+// consumer to construct the current state of the room and the state before the
+// event.
+//
+// When we talk about state in a matrix room we are talking about the state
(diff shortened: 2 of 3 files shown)


src/github.com/matrix-org/dendrite/roomserver/api/output.go
```diff
@@ -43,6 +40,14 @@ type OutputRoomEvent struct {
...
 	// they can use the LatestEventIDs to request the full current state.
 	LastSentEventID string
+	// The state event IDs that are part of the state at the event, but not
+	// part of the current state. Together with the StateBeforeRemovesEventIDs
+	// this can be used to construct the state before the event from the
+	// current state.
+	StateBeforeAddsEventIDs []string
+	// The state event IDs that are part of the current state, but not part
+	// of the state at the event.
+	StateBeforeRemovesEventIDs []string
```
Kegsay: When would there ever be something in this slice?

src/github.com/matrix-org/dendrite/roomserver/api/output.go
```diff
@@ -43,6 +40,14 @@ type OutputRoomEvent struct {
 	// If the LastSentEventID doesn't match what they were expecting it to be
 	// they can use the LatestEventIDs to request the full current state.
 	LastSentEventID string
+	// The state event IDs that are part of the state at the event, but not
+	// part of the current state. Together with the StateBeforeRemovesEventIDs
+	// this can be used to construct the state before the event from the
+	// current state.
+	StateBeforeAddsEventIDs []string
```
Kegsay: I'm uncomfortable with this. AIUI, consumers will be incrementally building up the complete room state based on `AddsStateEventIDs` and `RemovesStateEventIDs`. How does `StateBeforeAddsEventIDs` and `StateBeforeRemovesEventIDs` fit into this? You state that:

> this can be used to construct the state before the event from the current state.

But this isn't helpful. My questions are:
 - Why would a consumer ever need to "construct the state before the event from the current state"? I assume this has something to do with routing, but you don't state as much.
 - Should consumers remember the events in this slice, and use it to make the current room state? I imagine not, but you don't say this.
 - What are the clobbering semantics? I assume the order is:
    * Insert `AddsStateEventIDs`
    * Remove `RemovesStateEventIDs`
    * Insert `StateBeforeAddsEventIDs`
    * Remove `StateBeforeRemovesEventIDs`

    But you don't state this.

src/github.com/matrix-org/dendrite/roomserver/input/latest_events.go
```diff
@@ -200,7 +219,18 @@ func writeEvent(
...
 		ore.RemovesStateEventIDs = append(ore.RemovesStateEventIDs, eventIDMap[entry.EventNID])
 	}
-
+	for _, entry := range stateBeforeEventRemoves {
+		ore.StateBeforeRemovesEventIDs = append(ore.StateBeforeRemovesEventIDs, eventIDMap[entry.EventNID])
+	}
+	for _, entry := range stateBeforeEventAdds {
+		ore.StateBeforeAddsEventIDs = append(ore.StateBeforeAddsEventIDs, eventIDMap[entry.EventNID])
+	}
 	// TODO: Fill out VisibilityStateIDs
```
Kegsay: This is no longer true?

src/github.com/matrix-org/dendrite/cmd/roomserver-integration-tests/main.go
```diff
@@ -336,7 +336,8 @@ func main() {
 				"state_key":"@richvdh:matrix.org",
 				"type":"m.room.member"
 			},
-			"VisibilityEventIDs":null,
+			"StateBeforeRemovesEventIDs":["$1463671339126270PnVwC:matrix.org"],
+			"StateBeforeAddsEventIDs":null,
```
Kegsay: So this is effectively saying "to get the state before `$1463671339126270PnVwC:matrix.org` you need to remove `$1463671339126270PnVwC:matrix.org` and add nothing"? Which makes sense I guess.

src/github.com/matrix-org/dendrite/roomserver/api/output.go
NegativeMjark: Done 4f930d2

src/github.com/matrix-org/dendrite/roomserver/input/latest_events.go
NegativeMjark: Removed d609b8d

src/github.com/matrix-org/dendrite/cmd/roomserver-integration-tests/main.go
NegativeMjark: Yep

Kegsay:
LGTM

PR #145, opened 2017-06-23
Set depth of events and whether they need to be federated.

NegativeMjark:
Builds on #143

Set the depth of each new event to one greater than the maximum depth
of it's direct ancestors.

Add a flag to each event passing through the roomserver that tells us
whether the event needs to be sent over federation.

We do this by passing the name of the server to send the event as.
We will need this capability if we want to support vhosting as it is
not possible to tell from the event alone which server to send it as.

(The reason for this is that sometimes a event needs to be sent on
behalf of a different remote matrix server)

### DIFF
--- a/src/github.com/matrix-org/dendrite/federationapi/writers/send.go
+++ b/src/github.com/matrix-org/dendrite/federationapi/writers/send.go
@@ -159,7 +159,7 @@ func (t *txnReq) processEvent(e gomatrixserverlib.Event) error {
 	// TODO: Check that the event is allowed by its auth_events.
 
 	// pass the event to the roomserver
-	if err := t.producer.SendEvents([]gomatrixserverlib.Event{e}); err != nil {
+	if err := t.producer.SendEvents([]gomatrixserverlib.Event{e}, ""); err != nil {
 		return err
 	}
 
--- a/src/github.com/matrix-org/dendrite/federationapi/writers/send.go
+++ b/src/github.com/matrix-org/dendrite/federationapi/writers/send.go
@@ -159,7 +159,7 @@ func (t *txnReq) processEvent(e gomatrixserverlib.Event) error {
 	// TODO: Check that the event is allowed by its auth_events.
 
 	// pass the event to the roomserver
-	if err := t.producer.SendEvents([]gomatrixserverlib.Event{e}, ""); err != nil {
+	if err := t.producer.SendEvents([]gomatrixserverlib.Event{e}, api.DoNotSendToOtherServers); err != nil {
 		return err
 	}
 
--- a/src/github.com/matrix-org/dendrite/roomserver/api/output.go
+++ b/src/github.com/matrix-org/dendrite/roomserver/api/output.go
@@ -70,7 +70,7 @@ type OutputRoomEvent struct {
 	// Or empty if this event shouldn't be pushed to other servers.
 	//
 	// This is used by the federation sender component. We need to tell it what
-	// event is needs to send because it can't tell on its own. Normally if an
+	// event it needs to send because it can't tell on its own. Normally if an
 	// event was created on this server then we are responsible for sending it.
 	// However there are a couple of exceptions. The first is that when the
 	// server joins a remote room through another matrix server, it is the job
--- a/src/github.com/matrix-org/dendrite/roomserver/api/output.go
+++ b/src/github.com/matrix-org/dendrite/roomserver/api/output.go
@@ -70,14 +70,14 @@ type OutputRoomEvent struct {
 	// Or empty if this event shouldn't be pushed to other servers.
 	//
 	// This is used by the federation sender component. We need to tell it what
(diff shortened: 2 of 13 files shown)


src/github.com/matrix-org/dendrite/federationapi/writers/send.go
```diff
@@ -159,7 +159,7 @@ func (t *txnReq) processEvent(e gomatrixserverlib.Event) error {
 	// TODO: Check that the event is allowed by its auth_events.
 
 	// pass the event to the roomserver
-	if err := t.producer.SendEvents([]gomatrixserverlib.Event{e}); err != nil {
+	if err := t.producer.SendEvents([]gomatrixserverlib.Event{e}, ""); err != nil {
```
Kegsay: You know what I'm going to say, don't you?

Kegsay: Albeit the type documentation for `InputRoomEvent` does state this is valid and says what it does, but it's a bit too far divorced from this call-site for it to be found by people. Heck, the only reason why I found it was because it was part of this diff!

Kegsay: Either add a one-liner here or use a self-documenting constant name please. The constant name could be "cute", maybe.

src/github.com/matrix-org/dendrite/roomserver/api/output.go
```diff
@@ -66,6 +66,22 @@ type OutputRoomEvent struct {
 	// The state event IDs that are part of the current state, but not part
 	// of the state at the event.
 	StateBeforeRemovesEventIDs []string
+	// The server name to use to push this event to other servers.
+	// Or empty if this event shouldn't be pushed to other servers.
+	//
+	// This is used by the federation sender component. We need to tell it what
+	// event is needs to send because it can't tell on its own. Normally we
```
Kegsay: > We need to tell it what event ~~is~~ it needs to send

src/github.com/matrix-org/dendrite/roomserver/api/output.go
```diff
@@ -66,6 +66,22 @@ type OutputRoomEvent struct {
...
+	// if an event was created on this server then we are responsible for
```
Kegsay: > Normally we if an event

src/github.com/matrix-org/dendrite/roomserver/api/output.go
```diff
@@ -66,6 +66,22 @@ type OutputRoomEvent struct {
...
+	// sending. However there are a couple of exceptions. The first is that
+	// when the server joins a remote room through another matrix server, it
+	// is the job of the other matrix server to send the event over federation.
+	// The second is the reverse of the first, that is when a remote server
+	// joins a room that we are in over federation using our server it is our
+	// responsibility to send the join event to other matrix server.
+	//
+	// We encode the server name that the event should be sent using here to
+	// future proof the API for virtual hosting.
```
Kegsay: Absolutely excellent explanation, thank you!

src/github.com/matrix-org/dendrite/roomserver/api/output.go
NegativeMjark: Done f705480

src/github.com/matrix-org/dendrite/roomserver/query/query.go
```diff
@@ -34,7 +34,7 @@ type RoomserverQueryAPIDatabase interface {
 	RoomNID(roomID string) (types.RoomNID, error)
 	// Lookup event references for the latest events in the room and the current state snapshot.
 	// Returns an error if there was a problem talking to the database.
-	LatestEventIDs(roomNID types.RoomNID) ([]gomatrixserverlib.EventReference, types.StateSnapshotNID, error)
+	LatestEventIDs(roomNID types.RoomNID) ([]gomatrixserverlib.EventReference, types.StateSnapshotNID, int64, error)
```
Kegsay: Please explain what the `int64` is. I know its depth, but having to look at the implemention rather than the interface to work that out seems somewhat backwards :)

src/github.com/matrix-org/dendrite/roomserver/api/output.go
NegativeMjark: Done 37c8b91

src/github.com/matrix-org/dendrite/federationapi/writers/send.go
NegativeMjark: Done f705480

src/github.com/matrix-org/dendrite/roomserver/query/query.go
NegativeMjark: Done e8a55bb

Kegsay:
LGTM ⭐️

PR #146, opened 2017-06-23
Update the example configs with the new unified config

NegativeMjark:


dendrite-config.yaml
```diff
@@ -0,0 +1,73 @@
...
+    # Where the kafka servers are running.
+    addresses: ["localhost:9092"]
+    # The names of the kafka topics to use.
+    topics:
+        input_room_event: roomserverInput
+        output_room_event: roomserverOutput
+
+# The postgres connection configs for connecting to the databases e.g a postgres:// URI
+database:
+    account: "postgres://dendrite:itsasecret@localhost/account?sslmode=disable"
```
ara4n: Surely the database names should be prefixed, as per your DO script?

dendrite-config.yaml
```diff
@@ -0,0 +1,73 @@
+# The config file version format
+version: v0
+
+# The matrix specific config
+matrix:
+    # The name of the server. This is usually the domain name, e.g 'matrix.org', 'localhost'.
+    server_name: "example.com"
+    # The path to the PEM formatted matrix private key.
+    private_key: "/var/dendrite/matrix_key.pem"
```
ara4n: Wouldn't we be better assuming that people will run in a homedir?  Also, that the keys & certs match those created by something like the DO deployment script?

dendrite-config.yaml
```diff
@@ -0,0 +1,73 @@
+# The config file version format
+version: v0
```
Kegsay: Nit: What's the intent behind this? For Dendrite to whine if you're running a too new/old version of the config file? If so, it'd be nicer if this were an `int` rather than a `string`.

dendrite-config.yaml
4nd3r: homedir? i would asume that dendrite is packaged as deb/rpm and it will use standard places for binaries, config files, certificates (`/etc/ssl/` or similar), whatnot. `/var/dendrite` is wrong.

dendrite-config.yaml
NegativeMjark: Done 3a212a4

dendrite-config.yaml
NegativeMjark: I've moved the keys to /etc 3a212a4.
I was using /var to match the example media-repo config.
(Assuming the docker compose stuff lands in #140 we would probably be better of assuming that the config was in a /etc directory rather than a /home directory)

dendrite-config.yaml
NegativeMjark: I've made it an int and chucked some more comments in there. 9139153
(Saying something is "nicer" is not necessarily helpful here. I'd be interested in what you think you can do with an int that you can't do with a string. In this case I'm assuming it's because you can distinguish between it being too new and being too old when erroring at startup.)

Kegsay: Yes.

Kegsay:
LGTM

PR #147, opened 2017-06-26
Stub APIs needed to let riot join a room

NegativeMjark:
We need to add the directory API and the rooms/initialSync API if we want riot to be able to join a room.

I've also added stubs for some of the other APIs that riot hits to reduce the noise in the network tab on startup.

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/readers/directory.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/readers/directory.go
@@ -44,7 +44,7 @@ func DirectoryRoom(
 	}
 
 	if domain == cfg.Matrix.ServerName {
-		// TODO: Implement joining local room aliases.
+		// TODO: Implement lookup up local room aliases.
 		panic(fmt.Errorf("Looking up local room aliases is not implemented"))
 	} else {
 		resp, err := federation.LookupRoomAlias(domain, roomAlias)
@@ -58,6 +58,8 @@ func DirectoryRoom(
 					}
 				}
 			}
+			// TODO: Return 502 if the remote server errored.
+			// TODO: Return 504 if the remote server timed out.
 			return httputil.LogThenError(req, err)
 		}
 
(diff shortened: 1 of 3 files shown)


src/github.com/matrix-org/dendrite/clientapi/readers/directory.go
```diff
@@ -0,0 +1,85 @@
...
+			switch x := err.(type) {
+			case gomatrix.HTTPError:
+				if x.Code == 404 {
+					return util.JSONResponse{
+						Code: 404,
+						JSON: jsonerror.NotFound("Room alias not found"),
+					}
+				}
+			}
+			return httputil.LogThenError(req, err)
```
Kegsay: `LogThenError` will 500. Should we be proxying back the `gomatrix.HTTPError` where possible?

src/github.com/matrix-org/dendrite/clientapi/routing/routing.go
```diff
@@ -85,6 +105,14 @@ func Setup(
 		return writers.Register(req, accountDB, deviceDB)
 	}))
 
+	r0mux.Handle("/directory/room/{roomAlias}",
+		common.MakeAuthAPI("directory_room", deviceDB, func(req *http.Request, device *authtypes.Device) util.JSONResponse {
+			// TODO: Return the result of looking up the room in the room directory
```
Kegsay: This is spurious surely? From this files point-of-view, it is calling the right thing with the right args.

Kegsay:
Otherwise LGTM - feel free to merge after.

src/github.com/matrix-org/dendrite/clientapi/readers/directory.go
NegativeMjark: We should probably be turning 500 into 502 and timeouts into 504.
We should ignore the status code otherwise.
I've left TODO markers for now.

src/github.com/matrix-org/dendrite/clientapi/routing/routing.go
NegativeMjark: Done

PR #148, opened 2017-06-27
Add a component for sending event to remote matrix servers using fede…

NegativeMjark:
…ration

### DIFF
--- a/src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
+++ b/src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
@@ -160,6 +160,9 @@ func (s *OutputRoomEvent) processMessage(ore api.OutputRoomEvent, ev gomatrixser
 func (s *OutputRoomEvent) joinedHostsAtEvent(
 	ore api.OutputRoomEvent, ev gomatrixserverlib.Event, oldJoinedHosts []types.JoinedHost,
 ) ([]gomatrixserverlib.ServerName, error) {
+	// Combine the delta into a single delta so that the adds and removes can
+	// cancel each other out. This should reduce the number of times we need
+	// to fetch a state event from the room server.
 	combinedAdds, combinedRemoves := combineDeltas(
 		ore.AddsStateEventIDs, ore.RemovesStateEventIDs,
 		ore.StateBeforeAddsEventIDs, ore.StateBeforeRemovesEventIDs,
--- a/src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
+++ b/src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
@@ -185,7 +185,7 @@ func (s *OutputRoomEvent) joinedHostsAtEvent(
 
 	joined := map[gomatrixserverlib.ServerName]bool{}
 	for _, joinedHost := range oldJoinedHosts {
-		if removed[joinedHost.EventID] {
+		if removed[joinedHost.MemberEventID] {
 			// This m.room.member event is part of the current state of the
 			// room, but not part of the state at the event we are processing
 			// Therefore we can't use it to tell whether the server was in
@@ -233,7 +233,7 @@ func joinedHostsFromEvents(evs []gomatrixserverlib.Event) ([]types.JoinedHost, e
 			return nil, err
 		}
 		joinedHosts = append(joinedHosts, types.JoinedHost{
-			EventID: ev.EventID(), ServerName: serverName,
+			MemberEventID: ev.EventID(), ServerName: serverName,
 		})
 	}
 	return joinedHosts, nil
--- a/src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
+++ b/src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
@@ -128,7 +128,7 @@ func (s *OutputRoomEvent) processMessage(ore api.OutputRoomEvent, ev gomatrixser
 		return err
 	}
 
-	if ore.SendAsServer == "" {
+	if ore.SendAsServer == api.DoNotSendToOtherServers {
 		// Ignore event that we don't need to send anywhere.
 		return nil
 	}
--- a/src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
+++ b/src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
@@ -1,0 +1,355 @@
+// Copyright 2017 Vector Creations Ltd
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package consumers
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/matrix-org/dendrite/common"
+	"github.com/matrix-org/dendrite/common/config"
+	"github.com/matrix-org/dendrite/federationsender/queue"
+	"github.com/matrix-org/dendrite/federationsender/storage"
+	"github.com/matrix-org/dendrite/federationsender/types"
+	"github.com/matrix-org/dendrite/roomserver/api"
+	"github.com/matrix-org/gomatrixserverlib"
+	sarama "gopkg.in/Shopify/sarama.v1"
+)
+
+// OutputRoomEvent consumes events that originated in the room server.
+type OutputRoomEvent struct {
+	roomServerConsumer *common.ContinualConsumer
+	db                 *storage.Database
+	queues             *queue.OutgoingQueues
+	query              api.RoomserverQueryAPI
+}
+
+// NewOutputRoomEvent creates a new OutputRoomEvent consumer. Call Start() to begin consuming from room servers.
+func NewOutputRoomEvent(cfg *config.Dendrite, queues *queue.OutgoingQueues, store *storage.Database) (*OutputRoomEvent, error) {
+	kafkaConsumer, err := sarama.NewConsumer(cfg.Kafka.Addresses, nil)
+	if err != nil {
+		return nil, err
+	}
+	roomServerURL := cfg.RoomServerURL()
+
+	consumer := common.ContinualConsumer{
+		Topic:          string(cfg.Kafka.Topics.OutputRoomEvent),
+		Consumer:       kafkaConsumer,
+		PartitionStore: store,
+	}
+	s := &OutputRoomEvent{
+		roomServerConsumer: &consumer,
+		db:                 store,
+		queues:             queues,
+		query:              api.NewRoomserverQueryAPIHTTP(roomServerURL, nil),
+	}
+	consumer.ProcessMessage = s.onMessage
+
+	return s, nil
+}
+
+// Start consuming from room servers
+func (s *OutputRoomEvent) Start() error {
+	return s.roomServerConsumer.Start()
+}
+
+// onMessage is called when the sync server receives a new event from the room server output log.
+// It is not safe for this function to be called from multiple goroutines, or else the
+// sync stream position may race and be incorrectly calculated.
+func (s *OutputRoomEvent) onMessage(msg *sarama.ConsumerMessage) error {
+	// Parse out the event JSON
+	var output api.OutputRoomEvent
+	if err := json.Unmarshal(msg.Value, &output); err != nil {
+		// If the message was invalid, log it and move on to the next message in the stream
+		log.WithError(err).Errorf("roomserver output log: message parse failure")
+		return nil
+	}
+
+	ev, err := gomatrixserverlib.NewEventFromTrustedJSON(output.Event, false)
+	if err != nil {
+		log.WithError(err).Errorf("roomserver output log: event parse failure")
+		return nil
+	}
+	log.WithFields(log.Fields{
+		"event_id":       ev.EventID(),
+		"room_id":        ev.RoomID(),
+		"send_as_server": output.SendAsServer,
+	}).Info("received event from roomserver")
+
+	err = s.processMessage(output, ev)
+
+	if err != nil {
+		// panic rather than continue with an inconsistent database
+		log.WithFields(log.Fields{
+			"event":      string(ev.JSON()),
+			log.ErrorKey: err,
+			"add":        output.AddsStateEventIDs,
+			"del":        output.RemovesStateEventIDs,
+		}).Panicf("roomserver output log: write event failure")
+		return nil
+	}
+
+	return nil
+}
+
+func (s *OutputRoomEvent) processMessage(ore api.OutputRoomEvent, ev gomatrixserverlib.Event) error {
+	addsStateEvents, err := s.lookupStateEvents(ore.AddsStateEventIDs, ev)
+	if err != nil {
+		return err
+	}
+	addsJoinedHosts, err := joinedHostsFromEvents(addsStateEvents)
+	if err != nil {
+		return err
+	}
+	// Update our copy of the current state.
+	// We keep a copy of the current state because the state at each event is
+	// expressed as a delta against the current state.
+	// TODO: handle EventIDMismatchError and recover the current state by talking
+	// to the roomserver
+	oldJoinedHosts, err := s.db.UpdateRoom(
+		ev.RoomID(), ore.LastSentEventID, ev.EventID(),
+		addsJoinedHosts, ore.RemovesStateEventIDs,
+	)
+	if err != nil {
+		return err
+	}
+
+	if ore.SendAsServer == "" {
+		// Ignore event that we don't need to send anywhere.
+		return nil
+	}
+
+	joinedHosts, err := s.joinedHostsAtEvent(ore, ev, oldJoinedHosts)
+	if err != nil {
+		return err
+	}
+
+	if err = s.queues.SendEvent(
+		&ev, gomatrixserverlib.ServerName(ore.SendAsServer), joinedHosts,
+	); err != nil {
+		return err
+	}
+
+	// TODO: Add the event to the transaction queue
+	return nil
+}
+
+// joinedHostsAtEvent works out a list of matrix servers that were joined to
+// the room at the event.
+// It is important to use the state at the event for sending messages because:
+//   1) We shouldn't send messages to servers that weren't in the room.
+//   2) If a server is kicked from the rooms it should still be told about the
+//      kick event,
+// Usually the list can be calculated locally, but sometimes it will need fetch
+// events from the room server.
+// Returns an error if there was a problem talking to the room server.
+func (s *OutputRoomEvent) joinedHostsAtEvent(
+	ore api.OutputRoomEvent, ev gomatrixserverlib.Event, oldJoinedHosts []types.JoinedHost,
+) ([]gomatrixserverlib.ServerName, error) {
+	combinedAdds, combinedRemoves := combineDeltas(
+		ore.AddsStateEventIDs, ore.RemovesStateEventIDs,
+		ore.StateBeforeAddsEventIDs, ore.StateBeforeRemovesEventIDs,
+	)
+	combinedAddsEvents, err := s.lookupStateEvents(combinedAdds, ev)
+	if err != nil {
+		return nil, err
+	}
+
+	combinedAddsJoinedHosts, err := joinedHostsFromEvents(combinedAddsEvents)
+	if err != nil {
+		return nil, err
+	}
+
+	removed := map[string]bool{}
+	for _, eventID := range combinedRemoves {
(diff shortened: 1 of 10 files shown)


src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
```diff
@@ -0,0 +1,358 @@
...
+}
+
+// Start consuming from room servers
+func (s *OutputRoomEvent) Start() error {
+	return s.roomServerConsumer.Start()
+}
+
+// onMessage is called when the sync server receives a new event from the room server output log.
+// It is not safe for this function to be called from multiple goroutines, or else the
+// sync stream position may race and be incorrectly calculated.
```
Kegsay: Lies! All lies!

src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
```diff
@@ -0,0 +1,359 @@
...
+	}
+	log.WithFields(log.Fields{
+		"event_id":       ev.EventID(),
+		"room_id":        ev.RoomID(),
+		"send_as_server": output.SendAsServer,
+	}).Info("received event from roomserver")
+
+	err = s.processMessage(output, ev)
+
+	if err != nil {
```
Kegsay: Nit: You could probably in-line this with `:94`.

src/github.com/matrix-org/dendrite/federationsender/storage/storage.go
```diff
@@ -0,0 +1,124 @@
...
+func (d *Database) PartitionOffsets(topic string) ([]common.PartitionOffset, error) {
+	return d.SelectPartitionOffsets(topic)
+}
+
+// SetPartitionOffset implements common.PartitionStorer
+func (d *Database) SetPartitionOffset(topic string, partition int32, offset int64) error {
+	return d.UpsertPartitionOffset(topic, partition, offset)
+}
+
+// UpdateRoom updates the joined hosts for a room.
```
Kegsay: State what you're returning please. Your named return is unhelpfully named (I can tell it's joined hosts from the type), and the place you call this assigns it to the var `oldJoinedHosts` so I'm interested in knowing if the returned slice contains the hosts in `addHosts` or won't contain hosts in `removeHosts`.

src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
```diff
@@ -0,0 +1,359 @@
...
+		return err
+	}
+
+	if ore.SendAsServer == api.DoNotSendToOtherServers {
+		// Ignore event that we don't need to send anywhere.
+		return nil
+	}
+
+	joinedHosts, err := s.joinedHostsAtEvent(ore, ev, oldJoinedHosts)
```
Kegsay: I don't like how we accumulate joined hosts on a piecemeal basis as it makes it unclear what the joined hosts will actually be. I imagine there is probably a subtle reason behind all of this but I cannot infer this currently.

The steps we perform currently seem to be:
 - We get an event. We work out the delta hosts (`addsJoinedHosts`).
 - We side-effect fetch the current hosts when updating the room (`oldJoinedHosts`).
 - We work out the "real" list of joined hosts for this event, taking into account `StateBeforeAddsEventIDs`  (`joinedHosts`).
 - We send the event to `joinedHosts`.

Right, after working that all out I can tell why we're doing that (so we don't update the room with hosts from `StateBeforeAddsEventIDs`), but this was more difficult to work out than it needs to be. Bear in mind that this function, `processMessage`, is effectively the entry point for all the work for this entire process. It would be a good idea to be clear on the dance you're doing here. A comment along the lines of:

```go

// processMessage updates the list of joined hosts in the room and then sends the event to a potentially different list of joined hosts.
func (s *OutputRoomEvent) processMessage(ore api.OutputRoomEvent, ev gomatrixserverlib.Event) error {
    // ...
}
```

Splitting the function visually into 2 separate "paragraphs" (using line breaks) for each action would be also be nice given how similar their operations are.

src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
```diff
@@ -0,0 +1,359 @@
...
+		removed[eventID] = true
+	}
+
+	joined := map[gomatrixserverlib.ServerName]bool{}
+	for _, joinedHost := range oldJoinedHosts {
+		if removed[joinedHost.EventID] {
+			// This m.room.member event is part of the current state of the
+			// room, but not part of the state at the event we are processing
+			// Therefore we can't use it to tell whether the server was in
+			// the room at the event.
```
Kegsay: This makes no sense in the context of this code. There is no indication that the event ID here is an `m.room.member` event. Please explain.

src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
```diff
@@ -0,0 +1,359 @@
...
+			return nil, err
+		}
+		joinedHosts = append(joinedHosts, types.JoinedHost{
+			EventID: ev.EventID(), ServerName: serverName,
+		})
+	}
+	return joinedHosts, nil
+}
+
+// combineDeltas combines two deltas into a single delta.
```
Kegsay: You're going to need to be clearer on this. Order of operations is important. Do we:
 - Add `adds1`.
 - Remove `removes1`.
 - Add `adds2`.
 - Remove `removes2`.

Or do we:
 - Add `adds1`.
 - Add `adds2`.
 - Remove `removes1`.
 - Remove `removes2`.

Both seem plausible.

src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
```diff
@@ -0,0 +1,359 @@
...
+	var ok bool
+	for _, value := range adds1 {
+		addSet[value] = true
+	}
+	for _, value := range removes1 {
+		removeSet[value] = true
+	}
+	for _, value := range adds2 {
+		if _, ok = removeSet[value]; ok {
+			removeSet[value] = false
```
Kegsay: This doesn't add it to the `addSet` though. If an item was in `removeSet` but not in `adds1`, then `value` will be dropped to the floor.

src/github.com/matrix-org/dendrite/federationsender/queue/queue.go
```diff
@@ -0,0 +1,171 @@
...
+	log "github.com/Sirupsen/logrus"
+	"github.com/matrix-org/gomatrixserverlib"
+	"sync"
+	"time"
+)
+
+// OutgoingQueues is a collection of queues for sending transactions to other
+// matrix servers
+type OutgoingQueues struct {
+	mutex  sync.Mutex
```
Kegsay: I generally name the mutex over what it is supposed to be protecting. I haven't looked at the rest of this file yet, but I imagine this is to protect the `queues` map?

src/github.com/matrix-org/dendrite/federationsender/queue/queue.go
```diff
@@ -0,0 +1,171 @@
...
+
+// SendEvent sends an event to the destinations
+func (oqs *OutgoingQueues) SendEvent(
+	ev *gomatrixserverlib.Event, origin gomatrixserverlib.ServerName,
+	destinations []gomatrixserverlib.ServerName,
+) error {
+	if origin != oqs.origin {
+		return fmt.Errorf(
+			"sendevent: unexpected server to send as: got %q expected %q",
+			origin, oqs.origin,
```
Kegsay: I thought the entire point was that we could send events as another server. You're calling this function with `origin = ore.SendAsServer` in `processMessage`. If this is instead a "not implemented yet" thing, then please adjust the error message accordingly.

src/github.com/matrix-org/dendrite/federationsender/queue/queue.go
```diff
@@ -0,0 +1,171 @@
...
+	destinations = filterDestinations(oqs.origin, destinations)
+
+	log.WithFields(log.Fields{
+		"destinations": destinations, "event": ev.EventID(),
+	}).Info("Sending event")
+
+	oqs.mutex.Lock()
+	defer oqs.mutex.Unlock()
+	for _, destination := range destinations {
+		if destination == oqs.origin {
```
Kegsay: This is surely a pointless additional check? Why do it twice?

src/github.com/matrix-org/dendrite/federationsender/queue/queue.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+	return result
+}
+
+// outgoingQueue is a queue of events for a single destination.
+// It is responsible for sending the events to the destination and
+// ensures that only one request is in flight to a given destination
+// at a time.
+type outgoingQueue struct {
+	mutex              sync.Mutex
```
Kegsay: Again with naming of `mutex`. I have no idea what this is supposed to protect, so I cannot verify you are indeed protecting what you intend to.

src/github.com/matrix-org/dendrite/federationsender/queue/queue.go
```diff
@@ -0,0 +1,171 @@
...
+		result = append(result, destination)
+	}
+	return result
+}
+
+// outgoingQueue is a queue of events for a single destination.
+// It is responsible for sending the events to the destination and
+// ensures that only one request is in flight to a given destination
+// at a time.
+type outgoingQueue struct {
```
Kegsay: This doesn't seem to have anything to do with `OutgoingQueues`, so it is a good candidate for moving to a separate file (same package).

src/github.com/matrix-org/dendrite/federationsender/queue/queue.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+// Send event adds the event to the pending queue for the destination.
+// If the queue is empty then it starts a background goroutine to
+// start sending events to that destination.
+func (oq *outgoingQueue) sendEvent(ev *gomatrixserverlib.Event) {
+	oq.mutex.Lock()
+	defer oq.mutex.Unlock()
+	oq.pendingEvents = append(oq.pendingEvents, ev)
+	if !oq.running {
```
Kegsay: You never set this to `true`.

src/github.com/matrix-org/dendrite/federationsender/types/types.go
```diff
@@ -0,0 +1,44 @@
...
+package types
+
+import (
+	"fmt"
+	"github.com/matrix-org/gomatrixserverlib"
+)
+
+// A JoinedHost is a server that is joined to a matrix room.
+type JoinedHost struct {
+	// THe EventID of a m.room.member event that joins a server to a room.
```
Kegsay: > THe

Kegsay: Oh interesting. The event ID has to reference the `join` membership of a `m.room.member` event. This was not at all clear and makes your comments in another file make a bit more sense. I would be tempted to `s/EventID/MembershipEventID/` for clarity, at the cost of verbosity.

src/github.com/matrix-org/dendrite/federationsender/types/types.go
```diff
@@ -0,0 +1,44 @@
...
+	EventID string
+	// The
```
Kegsay: The what? :)

src/github.com/matrix-org/dendrite/federationsender/types/types.go
```diff
@@ -0,0 +1,44 @@
...
+	ServerName gomatrixserverlib.ServerName
+}
+
+// A EventIDMismatchError indicates that we have got out of sync with the
+// rooms erver.
```
Kegsay: > rooms erver`

src/github.com/matrix-org/dendrite/federationsender/storage/storage.go
```diff
@@ -0,0 +1,124 @@
...
+		}
+		if err = d.deleteJoinedHosts(txn, removeHosts); err != nil {
+			return err
+		}
+		return d.updateRoom(txn, roomID, newEventID)
+	})
+	return
+}
+
+func runTransaction(db *sql.DB, fn func(txn *sql.Tx) error) (err error) {
```
Kegsay: We should really be putting this into `common`...

src/github.com/matrix-org/dendrite/federationsender/storage/room_table.go
```diff
@@ -0,0 +1,84 @@
...
+package storage
+
+import (
+	"database/sql"
+)
+
+const roomSchema = `
+CREATE TABLE IF NOT EXISTS rooms (
+    -- The string ID of the room
+    room_id TEXT NOT NULL CONSTRAINT room_id_unique UNIQUE,
```
Kegsay: This is a long winded way of saying `PRIMARY KEY` isn't it?

src/github.com/matrix-org/dendrite/federationsender/storage/room_table.go
```diff
@@ -0,0 +1,84 @@
...
+    -- We can use this to tell if our view of the room state has become
+    -- desynchronised.
+    last_event_id TEXT NOT NULL
+);`
+
+const insertRoomSQL = "" +
+	"INSERT INTO rooms (room_id, last_event_id)" +
+	" VALUES ($1, '')" +
+	" ON CONFLICT ON CONSTRAINT room_id_unique" +
+	" DO NOTHING"
```
Kegsay: Why aren't we using a proper upsert?

src/github.com/matrix-org/dendrite/federationsender/storage/joined_hosts_table.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/matrix-org/gomatrixserverlib"
+)
+
+const joinedHostsSchema = `
+-- The joined_hosts table stores a list of m.room.member event ids in the
+-- current state for each room where the membership is "join".
+CREATE TABLE IF NOT EXISTS joined_hosts (
+    -- The string ID of the room.
+    room_id TEXT NOT NULL,
+    -- The event ID of the m.room.member
```
Kegsay: specifically the `membership: join` event? Does it matter which member event this is if there is >1?

src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
NegativeMjark: Done

src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
NegativeMjark: Done

src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
NegativeMjark: Fixed

src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
NegativeMjark: Done.

src/github.com/matrix-org/dendrite/federationsender/storage/storage.go
NegativeMjark: Done

src/github.com/matrix-org/dendrite/federationsender/storage/storage.go
NegativeMjark: Sure maybe in another PR though.

src/github.com/matrix-org/dendrite/federationsender/queue/queue.go
NegativeMjark: Done

src/github.com/matrix-org/dendrite/federationsender/queue/queue.go
NegativeMjark: Done

src/github.com/matrix-org/dendrite/federationsender/queue/queue.go
NegativeMjark: I've removed the redundant check.

src/github.com/matrix-org/dendrite/federationsender/queue/queue.go
NegativeMjark: I've moved it to a separate file.

src/github.com/matrix-org/dendrite/federationsender/queue/queue.go
NegativeMjark: Done 903a9d5

src/github.com/matrix-org/dendrite/federationsender/queue/queue.go
NegativeMjark: Fixed

src/github.com/matrix-org/dendrite/federationsender/types/types.go
NegativeMjark: Done

src/github.com/matrix-org/dendrite/federationsender/types/types.go
NegativeMjark: Fixed

src/github.com/matrix-org/dendrite/federationsender/storage/room_table.go
NegativeMjark: Fixed.

src/github.com/matrix-org/dendrite/federationsender/storage/room_table.go
NegativeMjark: Because I want to use an explicit "SELECT FOR UPDATE" to get a row lock.
(I think the INSERT will take get a row lock anyway, but I would prefer to be explicit)

src/github.com/matrix-org/dendrite/federationsender/storage/joined_hosts_table.go
NegativeMjark: Fixed

src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
NegativeMjark: Yep, that is deliberate, I've added some tests fwiw.

src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"testing"
+)
+
+func TestCombineNoOp(t *testing.T) {
+	inputAdd1 := []string{"a", "b", "c"}
+	inputDel1 := []string{"a", "b", "d"}
+	inputAdd2 := []string{"a", "d", "e"}
+	inputDel2 := []string{"a", "c", "e", "e"}
```
Kegsay: This makes me unhappy. I'm reading the processing of this as:
```js
[a,b,c]  // add a b c
[c]  // del a b d
[a,c,d,e]  // add a d e
[d]  // del a c e e
```

This is further backed up by your comment for `combineDeltas`:

> Assumes that the order of operations is add(1), remove(1), add(2), remove(2).

So why are you expecting this to be `[]` and not `[d]`?

src/github.com/matrix-org/dendrite/federationsender/types/types.go
```diff
@@ -0,0 +1,45 @@
...
+// A JoinedHost is a server that is joined to a matrix room.
+type JoinedHost struct {
+	// The MemberEventID of a m.room.member join event.
+	MemberEventID string
+	// The domain part of the state key of the m.room.member join event
+	ServerName gomatrixserverlib.ServerName
+}
+
+// A EventIDMismatchError indicates that we have got out of sync with the
+// rooms server.
```
Kegsay: > room~~s~~ server

src/github.com/matrix-org/dendrite/federationsender/storage/room_table.go
Kegsay: I see. Unfortunately, it's not really clear because the SQL containing `FOR UPDATE` is so far away from the point where we use it. I guess this is fine as it is.

src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver_test.go
NegativeMjark: Because removing d and then adding d is equivalent to a no-op.

src/github.com/matrix-org/dendrite/federationsender/storage/room_table.go
Kegsay: Oh your comments on https://github.com/matrix-org/dendrite/pull/148/commits/5dc645d9f9bc5ae2908a612433fdddd304515643 do help a lot actually, thanks.

Kegsay:
Otherwise LGTM.

src/github.com/matrix-org/dendrite/federationsender/types/types.go
NegativeMjark: Fixed 		94680bf

src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver_test.go
Kegsay: Ohhhh because we just want to work out the add/del deltas. Duh. Okay fine. Thanks.

Kegsay:
LGTM

PR #151, opened 2017-07-07
Profile API

babolivier:
- [x] GET /_matrix/client/r0/profile/{userId}
- [x] GET /_matrix/client/r0/profile/{userId}/avatar_url
- [x] PUT/_matrix/client/r0/profile/{userId}/avatar_url
- [x] GET /_matrix/client/r0/profile/{userId}/displayname
- [x] PUT/_matrix/client/r0/profile/{userId}/displayname
- [ ] Propagate change through a `m.room.member` event
- [ ] Propagate change through the `m.presence` events

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/readers/profile.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/readers/profile.go
@@ -22,7 +22,6 @@ import (
 	"github.com/matrix-org/dendrite/clientapi/auth/storage/accounts"
 	"github.com/matrix-org/dendrite/clientapi/httputil"
 	"github.com/matrix-org/dendrite/clientapi/jsonerror"
-	// "github.com/matrix-org/gomatrixserverlib"
 	"github.com/matrix-org/util"
 )
 
--- a/src/github.com/matrix-org/dendrite/clientapi/readers/profile.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/readers/profile.go
@@ -1,0 +1,70 @@
+// Copyright 2017 Vector Creations Ltd
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package readers
+
+import (
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/matrix-org/dendrite/clientapi/auth/storage/accounts"
+	// "github.com/matrix-org/dendrite/clientapi/httputil"
+	"github.com/matrix-org/dendrite/clientapi/jsonerror"
+	// "github.com/matrix-org/gomatrixserverlib"
+	"github.com/matrix-org/util"
+)
+
+type profileResponse struct {
+	AvatarURL   string `json:"avatar_url"`
+	DisplayName string `json:"displayname"`
+}
+
+func Profile(
+	req *http.Request, accountDB *accounts.Database, userID string,
+) util.JSONResponse {
+	if req.Method == "GET" {
+		localpart := getLocalPart(userID)
+		profile, err := accountDB.GetProfileByLocalpart(localpart)
+		if err == nil {
+			res := profileResponse{
+				AvatarURL:   profile.AvatarURL,
+				DisplayName: profile.DisplayName,
+			}
+			return util.JSONResponse{
+				Code: 200,
+				JSON: res,
+			}
+		}
+		return util.JSONResponse{
+			Code: 500,
+			JSON: jsonerror.Unknown("Failed to load user profile"),
+		}
+	}
+	return util.JSONResponse{
+		Code: 405,
+		JSON: jsonerror.NotFound("Bad method"),
+	}
+}
+
+func getLocalPart(userID string) string {
+	if !strings.HasPrefix(userID, "@") {
+		panic(fmt.Errorf("Invalid user ID"))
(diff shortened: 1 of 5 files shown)


src/github.com/matrix-org/dendrite/clientapi/readers/profile.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// GetProfile implements GET /profile/{userID}
+func GetProfile(
+	req *http.Request, accountDB *accounts.Database, userID string,
+) util.JSONResponse {
+	if req.Method == "GET" {
+		localpart := getLocalPart(userID)
+		profile, err := accountDB.GetProfileByLocalpart(localpart)
+		if err == nil {
```
NegativeMjark: We usually put the `if err !=nil` branch first. That way the code path for the correct case stays un-idented.

src/github.com/matrix-org/dendrite/clientapi/readers/profile.go
```diff
@@ -0,0 +1,178 @@
...
+
+type displayName struct {
+	DisplayName string `json:"displayname"`
+}
+
+// GetProfile implements GET /profile/{userID}
+func GetProfile(
+	req *http.Request, accountDB *accounts.Database, userID string,
+) util.JSONResponse {
+	if req.Method == "GET" {
```
NegativeMjark: We'd usually write this as:
```go
if req.Method != "GET {
    return util.JSONResponse{
￼        Code: 405,
￼ 	JSON: jsonerror.NotFound("Bad method"),
￼    }
}
```

That way the code for valid input stays un-indented.

src/github.com/matrix-org/dendrite/clientapi/readers/profile.go
```diff
@@ -0,0 +1,178 @@
...
+		}
+		return util.JSONResponse{
+			Code: 200,
+			JSON: struct{}{},
+		}
+	}
+	return util.JSONResponse{
+		Code: 405,
+		JSON: jsonerror.NotFound("Bad method"),
+	}
```
NegativeMjark: I'd suggest moving the code for handling "PUT" and "GET" into separate functions. so that this method looks like:

```go
+// AvatarURL implements GET and PUT /profile/{userID}/avatar_url
 func AvatarURL(...) util.JSONResponse {
      if req.Method == "GET" {
            return GetAvatarURL(...)
      }
      if req.Method == "PUT" {
            return PutAvatarURL(...)
      }
      return util.JSONResponse{
￼ 	  Code: 405,
￼ 	  JSON: jsonerror.NotFound("Bad method"),
￼      }
}
```

NegativeMjark:
I think we need to be authenticating the PUT displayName and PUT avatarURL requests. But I suspect that we shouldn't be authenticating the GET versions. @Kegsay do you have any ideas on how to do that using MakeAuthAPI?

Kegsay:
Create a route for the method in `gorilla/mux` for GET which doesn't make an auth API, and create another route for PUT which does make an auth API. 

http://www.gorillatoolkit.org/pkg/mux#Route.Methods

src/github.com/matrix-org/dendrite/clientapi/readers/profile.go
NegativeMjark: Actually since there are different authentication requirements for GET and PUT I'd suggest following Kegan's suggestion in
https://github.com/matrix-org/dendrite/pull/151#issuecomment-313740158

src/github.com/matrix-org/dendrite/clientapi/readers/profile.go
```diff
@@ -0,0 +1,197 @@
...
+	req *http.Request, accountDB *accounts.Database, userID string,
+) util.JSONResponse {
+	if req.Method != "GET" {
+		return util.JSONResponse{
+			Code: 405,
+			JSON: jsonerror.NotFound("Bad method"),
+		}
+	}
+	localpart := getLocalPart(userID)
+	if profile, err := accountDB.GetProfileByLocalpart(localpart); err == nil {
```
NegativeMjark: We generally would write this as:
```go
profile, err := accountDB.GetProfileByLocalpart(localpart)
if err != nil {
    return httputil.LogThenError(req, err) 
}
return util.JSONResponse{
    Code: 200,
    JSON: displayName{ DisplayName: profile.DisplayName},
}
```

babolivier: Better indeed. Changed it.

src/github.com/matrix-org/dendrite/clientapi/readers/profile.go
```diff
@@ -0,0 +1,193 @@
...
```
NegativeMjark: Cool, if you could update GetProfile while you are at it it would be good.

src/github.com/matrix-org/dendrite/clientapi/readers/profile.go
```diff
@@ -0,0 +1,193 @@
...
+			JSON: jsonerror.BadJSON("'avatar_url' must be supplied."),
+		}
+	}
+
+	localpart := getLocalPart(userID)
+	if err := accountDB.SetAvatarURL(localpart, r.AvatarURL); err != nil {
+		return util.JSONResponse{
+			Code: 500,
+			JSON: jsonerror.Unknown("Failed to set avatar URL"),
+		}
```
NegativeMjark: Use httputil.LogThenError(req, err) here?

src/github.com/matrix-org/dendrite/clientapi/readers/profile.go
```diff
@@ -0,0 +1,193 @@
...
+			JSON: jsonerror.BadJSON("'displayname' must be supplied."),
+		}
+	}
+
+	localpart := getLocalPart(userID)
+	if err := accountDB.SetDisplayName(localpart, r.DisplayName); err != nil {
+		return util.JSONResponse{
+			Code: 500,
+			JSON: jsonerror.Unknown("Failed to set display name"),
+		}
```
NegativeMjark: Use httputil.LogThenError(req, err) here?

PR #153, opened 2017-07-10
Notify profile update

babolivier:
Edited the profile update so it sends a message to the kafka topic set in `user_updates` in the config file with the user's ID as key and a JSON matching the following example as a value:

```json
{
    "updated": "displayname",
    "oldValue": "",
    "newValue": "John Doe"
}
```

Values for the `updated` attribute can be either `displayname` or `avatar_url`.

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/readers/profile.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/readers/profile.go
@@ -1,0 +1,70 @@
+// Copyright 2017 Vector Creations Ltd
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package readers
+
+import (
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/matrix-org/dendrite/clientapi/auth/storage/accounts"
+	// "github.com/matrix-org/dendrite/clientapi/httputil"
+	"github.com/matrix-org/dendrite/clientapi/jsonerror"
+	// "github.com/matrix-org/gomatrixserverlib"
+	"github.com/matrix-org/util"
+)
+
+type profileResponse struct {
+	AvatarURL   string `json:"avatar_url"`
+	DisplayName string `json:"displayname"`
+}
+
+func Profile(
+	req *http.Request, accountDB *accounts.Database, userID string,
+) util.JSONResponse {
+	if req.Method == "GET" {
+		localpart := getLocalPart(userID)
+		profile, err := accountDB.GetProfileByLocalpart(localpart)
+		if err == nil {
+			res := profileResponse{
+				AvatarURL:   profile.AvatarURL,
+				DisplayName: profile.DisplayName,
+			}
+			return util.JSONResponse{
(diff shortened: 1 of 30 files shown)


src/github.com/matrix-org/dendrite/clientapi/readers/profile.go
```diff
@@ -38,6 +43,12 @@ type displayName struct {
 	DisplayName string `json:"displayname"`
 }
 
+type profileUpdate struct {
+	Updated  string `json:"updated"`
```
NegativeMjark: Could you add some comments to explain what "updated", "old_value", and "new_value" mean please?

src/github.com/matrix-org/dendrite/clientapi/readers/profile.go
```diff
@@ -38,6 +43,12 @@ type displayName struct {
 	DisplayName string `json:"displayname"`
 }
 
+type profileUpdate struct {
```
NegativeMjark: This should really be somewhere in common that other packages can import.
That way the components that consume the topic can use the same structure when parsing incoming messages.
I'm not certain where we want to put that sort of thing in dendrite yet, so there's no need to do so in this PR.
But it'd be good to have a TODO marker here so that we don't forget.

src/github.com/matrix-org/dendrite/clientapi/readers/profile.go
```diff
@@ -159,3 +196,49 @@ func getLocalPart(userID string) string {
...
+		panic(fmt.Errorf("No update to send"))
+	}
+
+	value, err := json.Marshal(update)
+	if err != nil {
+		return err
+	}
+	m.Value = sarama.ByteEncoder(value)
+
+	producer, err := sarama.NewSyncProducer(cfg.Kafka.Addresses, nil)
```
NegativeMjark: We probably don't want to create a new producer for each message.
I'd suggest making a UserUpdateProducer type in the clientapi/producers package along the same lines as the room server producer.

src/github.com/matrix-org/dendrite/clientapi/readers/profile.go
```diff
@@ -159,3 +196,49 @@ func getLocalPart(userID string) string {
 	// Return the part after the "@"
 	return strings.Split(username, "@")[1]
 }
+
+// Send an update using kafka to notify the roomserver of the profile update
+// Returns an error if the update failed to send
+func sendUpdate(userID string, oldProfile *authtypes.Profile, newAvatarURL string,
+	newDisplayName string, accountDB *accounts.Database, cfg config.Dendrite,
```
NegativeMjark: Why does this take an accountDB?

babolivier: That comes from an older version of this function, I forgot to remove it in the parameters.

src/github.com/matrix-org/dendrite/clientapi/readers/profile.go
babolivier: > We probably don't want to create a new producer for each message.

That's what I thought, but I didn't see a best way to do it. Thanks for the suggestion.

src/github.com/matrix-org/dendrite/clientapi/readers/profile.go
```diff
@@ -159,3 +196,49 @@ func getLocalPart(userID string) string {
...
+) error {
```
NegativeMjark: Would it make sense to write this as ```sendUpdate(userID, updates, oldValue, newValue string, .,,)```
Then the call sites can be written as:

```go
sendUpdate(userID, "avatar_url", oldProfile.AvatarURL, r.AvatarURL,...)
```

```go
sendUpdate(userID, "displayName", oldProfile.DisplayName, r.DisplayName,...)
```

babolivier: Would make more sense, indeed.

PR #155, opened 2017-07-11
Make the roomserver output format more flexible

NegativeMjark:
Add a `"type"` key so that we can include different kinds of updates within the same stream.
(We need this so that we can support invites)
The value for each type is stored under a top-level key with the same name as the type.
So updates with type `"new_room_event"` go under a `"new_room_event"` top level key.
This makes it easier to parse in type safe languages.


src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
```diff
@@ -72,31 +72,32 @@ func (s *OutputRoomEvent) Start() error {
...
 		log.WithError(err).Errorf("roomserver output log: message parse failure")
 		return nil
 	}
-
-	ev, err := gomatrixserverlib.NewEventFromTrustedJSON(output.Event, false)
-	if err != nil {
-		log.WithError(err).Errorf("roomserver output log: event parse failure")
+	if output.Type != api.OutputTypeNewRoomEvent {
+		log.WithField("type", output.Type).Debug(
+			"roomserver output logignoring unknown output type",
```
Kegsay: > logignoring

Kegsay:
LGTM

PR #156, opened 2017-07-12
Make the room server RPC API formats more like our other APIs

NegativeMjark:
Use "snake_case" for JSON key names, and use "lowerCamelCase" for the HTTP API paths.


Kegsay:
LGTM

PR #158, opened 2017-07-12
Add a HTTP API for roomserver input

NegativeMjark:
This does 2 things:
  1) neaten up the format of input room events.
  2) Adds an HTTP API for inputting room events to the room server.

It doesn't get rid of the kafka consumer and it doesn't update the clientapi and federation api to use the new HTTP API.

### DIFF
--- a/src/github.com/matrix-org/dendrite/roomserver/api/input.go
+++ b/src/github.com/matrix-org/dendrite/roomserver/api/input.go
@@ -16,7 +16,9 @@
 package api
 
 import (
-	"encoding/json"
+	"net/http"
+
+	"github.com/matrix-org/gomatrixserverlib"
 )
 
 const (
@@ -25,19 +27,14 @@ const (
 	// These events are state events used to authenticate other events.
 	// They can become part of the contiguous event graph via backfill.
 	KindOutlier = 1
-	// KindJoin event start a new contiguous event graph. The event must be a
-	// m.room.member event joining this server to the room. This must come with
-	// the state at the event. If the event is contiguous with the existing
-	// graph for the room then it is treated as a normal new event.
-	KindJoin = 2
 	// KindNew event extend the contiguous graph going forwards.
 	// They usually don't need state, but may include state if the
 	// there was a new event that references an event that we don't
 	// have a copy of.
-	KindNew = 3
+	KindNew = 2
 	// KindBackfill event extend the contiguous graph going backwards.
 	// They always have state.
-	KindBackfill = 4
+	KindBackfill = 3
 )
 
 // DoNotSendToOtherServers tells us not to send the event to other matrix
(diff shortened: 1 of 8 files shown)


src/github.com/matrix-org/dendrite/roomserver/api/input.go
```diff
@@ -25,19 +27,14 @@ const (
 	// These events are state events used to authenticate other events.
 	// They can become part of the contiguous event graph via backfill.
 	KindOutlier = 1
-	// KindJoin event start a new contiguous event graph. The event must be a
-	// m.room.member event joining this server to the room. This must come with
-	// the state at the event. If the event is contiguous with the existing
-	// graph for the room then it is treated as a normal new event.
-	KindJoin = 2
```
Kegsay: Why are you removing this?

src/github.com/matrix-org/dendrite/roomserver/api/input.go
```diff
@@ -49,77 +46,66 @@ const DoNotSendToOtherServers = ""
...
-		StateEventIDs: ire.StateEventIDs,
-		Event:         &event,
-		HasState:      ire.HasState,
-		SendAsServer:  ire.SendAsServer,
+// RoomserverInputRoomEventsPath is the HTTP path for the InputRoomEvents API.
+const RoomserverInputRoomEventsPath = "/api/roomserver/inputRoomEvents"
+
+// NewRoomserverInputAPIHTTP creates a RoomserverInputAPI implemented by talking to a HTTP POST API.
+// If httpClient is nil then it uses the http.DefaultClient
+func NewRoomserverInputAPIHTTP(roomserverURL string, httpClient *http.Client) RoomserverInputAPI {
```
Kegsay: I was under the impression we were going to use protocol buffers and grpc? What changed?

src/github.com/matrix-org/dendrite/roomserver/api/query.go
```diff
@@ -135,12 +136,12 @@ func NewRoomserverQueryAPIHTTP(roomserverURL string, httpClient *http.Client) Ro
...
 		httpClient = http.DefaultClient
 	}
-	return &httpRoomserverQueryAPI{roomserverURL, *httpClient}
+	return &httpRoomserverQueryAPI{roomserverURL, httpClient}
 }
 
 type httpRoomserverQueryAPI struct {
 	roomserverURL string
-	httpClient    http.Client
+	httpClient    *http.Client
```
Kegsay: Why are you changing this? I guess this means we've been using the default client all this time?

src/github.com/matrix-org/dendrite/roomserver/input/events.go
```diff
@@ -79,8 +76,8 @@ func processRoomEvent(db RoomEventDatabase, ow OutputRoomEventWriter, input api.
 		if input.HasState {
 			// We've been told what the state at the event is so we don't need to calculate it.
 			// Check that those state events are in the database and store the state.
-			entries, err := db.StateEntriesForEventIDs(input.StateEventIDs)
-			if err != nil {
+			var entries []types.StateEntry
+			if entries, err = db.StateEntriesForEventIDs(input.StateEventIDs); err != nil {
```
Kegsay: Humm. I'm not sure I like this. Semantically this diff is the same but style-wise I dislike it. I dislike it because, at a glance, I thought `entries` was scoped to just the `if` block due to the inlining. It's not, since there is a `=` and not `:=`, but looking for a single `:` on this long if statement is harder than just having it what it was before.

I don't mind us inlining `expression; conditional` provided the expression just returns an `err` since that scopes things nicely. WDYT?

src/github.com/matrix-org/dendrite/roomserver/input/input.go
```diff
@@ -0,0 +1,107 @@
...
+	return err
+}
+
+// InputRoomEvents implements api.RoomserverInputAPI
+func (r *RoomserverInputAPI) InputRoomEvents(
+	request *api.InputRoomEventsRequest,
+	response *api.InputRoomEventsResponse,
+) error {
+	for i := range request.InputRoomEvents {
+		if err := processRoomEvent(r.DB, r, request.InputRoomEvents[i]); err != nil {
```
Kegsay: Right so stuff in this function below this line is just copypasta from the kafka input version right? Just want to confirm.

src/github.com/matrix-org/dendrite/roomserver/api/input.go
NegativeMjark: It looks like nobody is setting this correctly and I can't actually see how to use it correctly in the room-server.

src/github.com/matrix-org/dendrite/roomserver/input/input.go
NegativeMjark: Well, I had to change the `c` to `r` and the shutdown handler is *slightly* different. But yes it is taken from a copypasta.

src/github.com/matrix-org/dendrite/roomserver/api/query.go
NegativeMjark: I'm changing it because otherwise we are copying the http.Client by value. Passing an http.Client around by value is mosty safe because the http.Client is actually just a pointer to a http.RoundTripper and some config. However if the user changes the config *after* passing the http.Client to NewRoomserverQueryAPI then we'll end up in a weird state.

(We have been using the default client everywhere, but that's just because we haven't got round to setting up a custom client in any of the components yet.)

src/github.com/matrix-org/dendrite/roomserver/api/input.go
NegativeMjark: It started becoming too much effort to get working and I'd like to focus on getting the matrix-specific logic written down. So for now it's much easier to stick with simple HTTP APIs.

It still should be fairly easy to switch out when we need to, I think the work is just:

1) Write protocol buffer schema for all the Query and Input RPC functions.
2) Set up a build environment with correct versions of protoc and grpc to do the codegen.
3) Setup prometheus style metrics for grpc.

But getting all that working and debugged could easily eat a couple of days, so I'd rather keep making progress on the core matrix stuff

Kegsay: SGTM

src/github.com/matrix-org/dendrite/roomserver/api/query.go
Kegsay: SGTM

src/github.com/matrix-org/dendrite/roomserver/input/input.go
Kegsay: :)

src/github.com/matrix-org/dendrite/roomserver/api/input.go
Kegsay: SGTM

src/github.com/matrix-org/dendrite/roomserver/input/events.go
NegativeMjark: Hmm, I dunno I think it's fairly safe here TBH. You know the scope of entries from the `var entries` line and the variable shadowing checks prevent us from using `:=` if there is a `var entries` line.

Kegsay: okay.

Kegsay:
LGTM

PR #159, opened 2017-07-12
Keep track of membership

babolivier:
Allow the client API server to save a user membership to rooms by consuming the messages sent from the roomserver via kafka, detecting membership updates (`m.room.member`) and:

* saving them in DB in case of `join`
* removing them from the DB in case of `leave` or `ban`

That way the client API server can know which users are members of which rooms (e.g. useful for profile update)


babolivier:
FTR I don't see why the tests are failing. Log is

```
roomserver.go:71: undefined: api.OutputRoomEvent
roomserver.go:71: undefined: api.OutputRoomEvent
```

even though `github.com/matrix-org/dendrite/roomserver/api` is imported at https://github.com/matrix-org/dendrite/blob/013a0a3732517030b4b1aa3fce09168806946d3e/src/github.com/matrix-org/dendrite/clientapi/consumers/roomserver.go#L26

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/membership_table.go
```diff
@@ -0,0 +1,81 @@
...
+// limitations under the License.
+
+package accounts
+
+import (
+	"database/sql"
+)
+
+const membershipSchema = `
+-- Stores data about accounts profiles.
```
NegativeMjark: The comment is probably out of date.

babolivier: Whoops, indeed, copy/pasted a bit too quickly

src/github.com/matrix-org/dendrite/clientapi/consumers/roomserver.go
```diff
@@ -0,0 +1,121 @@
...
+		return nil
+	}
+	log.WithFields(log.Fields{
+		"event_id": ev.EventID(),
+		"room_id":  ev.RoomID(),
+		"type":     ev.Type(),
+	}).Info("received event from roomserver")
+
+	if ev.Type() == "m.room.member" && ev.StateKey() != nil {
+		localpart := getLocalPart(*ev.StateKey())
```
NegativeMjark: A couple of points here:

1) I've added a function to gomatrixserverlib for splitting IDs which might be useful https://godoc.org/github.com/matrix-org/gomatrixserverlib#SplitID

2) Shouldn't this be checking which server the user is on? We don't want to be clobbering our local users with remote users with the same local part.

babolivier: > I've added a function to gomatrixserverlib for splitting IDs which might be useful https://godoc.org/github.com/matrix-org/gomatrixserverlib#SplitID

Awesome, copy/pasting this `getLocalPart` function all around dendrite was starting to annoy me.

> Shouldn't this be checking which server the user is on? We don't want to be clobbering our local users with remote users with the same local part.

Good point, didn't think about that.

src/github.com/matrix-org/dendrite/clientapi/consumers/roomserver.go
```diff
@@ -0,0 +1,121 @@
...
+func (s *OutputRoomEvent) onMessage(msg *sarama.ConsumerMessage) error {
+	// Parse out the event JSON
+	var output api.OutputRoomEvent
+	if err := json.Unmarshal(msg.Value, &output); err != nil {
+		// If the message was invalid, log it and move on to the next message in the stream
+		log.WithError(err).Errorf("roomserver output log: message parse failure")
+		return nil
+	}
+
+	ev, err := gomatrixserverlib.NewEventFromTrustedJSON(output.Event, false)
```
NegativeMjark: I kind of changed the roomserver output format slightly in #155. The change needed here is fairly trivial. You don't need to parse the event JSON anymore, but you do need to check the event type.

babolivier: Ooooh and you also renamed the Ouput[Room]Event structure, that's why the tests are failing!

src/github.com/matrix-org/dendrite/clientapi/consumers/roomserver.go
```diff
@@ -0,0 +1,121 @@
...
+		log.WithError(err).Errorf("roomserver output log: event parse failure")
+		return nil
+	}
+	log.WithFields(log.Fields{
+		"event_id": ev.EventID(),
+		"room_id":  ev.RoomID(),
+		"type":     ev.Type(),
+	}).Info("received event from roomserver")
+
+	if ev.Type() == "m.room.member" && ev.StateKey() != nil {
```
NegativeMjark: Unfortunately the matrix protocol makes this a bit more complicated.
You need to look at the "AddsStateEventIDs" and "RemovesStateEventIDs" rather than looking at the event itself.

This is necessary because the state resolution process means that even message events can introduce new state events to the room, and those state events might be ones that we've never seen before.

The way the sync server handles this is that it checks the "AddsStateEventIDs" and the "RemovesStateEventIDs".
It removes all the entries from current state with an event ID in "RemovesStateEventIDs". It looks up the events for the "AddsStateEventIDs". It does this by calling "QueryEventsByID" on the RoomserverQueryAPI. As an optimisation if the "AddsStateEventIDs" list is a single entry which matches the "Event", then it uses the copy of the event it read from the kafka log rather than fetching it from the Roomserver.

Sorry that this is so complicated. I hope it makes some small amount of sense.

NegativeMjark: Note that to implement this correctly  you'll need to store the event_ids for the "join" membership events so that you know which events to remove using "RemovesStateEventIDs".

babolivier:
So I've tried to add the changes you were talking about (tell me if there's somthing wrong about it). I might also need some help on fixing the tests (seems like a PG error, but I don't think I know PG well enough to fix it all by myself, or at least I don't understand where the error comes from atm).

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/membership_table.go
```diff
@@ -0,0 +1,110 @@
...
+CREATE TABLE IF NOT EXISTS memberships (
+    -- The Matrix user ID localpart for the member
+    localpart TEXT NOT NULL,
+    -- The room this user is a member of
+    room_id TEXT NOT NULL,
+	-- The ID of the join membership event
+	event_id TEXT NOT NULL,
+
+	-- A user can only be member of a room once
+	PRIMARY KEY (localpart, room_id)
```
NegativeMjark: The indentation here is inconsistent. The first lines are indented with spaces, the last lines are indented with tabs.

babolivier: Seems like my editor messed it up. Fixed.

NegativeMjark:
Ah it looks like it is trying the create a table with the same name twice in the syncserver integration tests. This should be fine because the tables are created with "IF NOT EXISTS" but I bet that postgresql will fail if the same table is created by different connections at the same time.

I think the race is being cause by:
https://github.com/matrix-org/dendrite/blob/master/src/github.com/matrix-org/dendrite/cmd/syncserver-integration-tests/main.go#L150-L160

```go
	cmd, cmdChan := test.StartServer(
		"sync-api",
		serverArgs,
		postgresDatabase,
		postgresContainerName,
		databases,
	)

	if err := createTestUser(testDatabase, "alice", "@alice:localhost"); err != nil {
		panic(err)
	}
```

`StartServer` will start running the `dendrite-sync-api-server`, and `createTestUser` will run `create-account`. These will be running in parallel and both will attempt to create their respective database tables.

I think the fix will be to move the bit of `StartServer(...)` that calls `cmd.Start()` to be after the `createTestUser(...)`. Unfortunately we can't move the entire call to `StartServer(...)` to be after `createTestUser(...)` because `StartServer` creates the databases that `createTestUser(...)` tries to talk to.

src/github.com/matrix-org/dendrite/clientapi/consumers/roomserver.go
```diff
@@ -0,0 +1,192 @@
...
+		}
+
+		switch membership {
+		case "join":
+			if err := s.db.SaveMembership(localpart, roomID, eventID); err != nil {
+				return err
+			}
+		case "leave":
+		case "ban":
+			if err := s.db.RemoveMembership(localpart, roomID); err != nil {
```
NegativeMjark: This shouldn't be necessary. Any "join" memberships will have already been removed by RemovesStateEventIDs.

src/github.com/matrix-org/dendrite/clientapi/consumers/roomserver.go
```diff
@@ -0,0 +1,192 @@
...
+	// Fast path if the only state event added is the event itself.
+	if len(addsStateEventIDs) == 1 && addsStateEventIDs[0] == event.EventID() {
+		return []gomatrixserverlib.Event{event}, nil
+	}
+
+	result := []gomatrixserverlib.Event{}
+	missing := []string{}
+	for _, id := range addsStateEventIDs {
+		// Check if the event is already known
+		localpart, server, err := s.db.GetMembershipByEventID(id)
```
NegativeMjark: This shouldn't be necessary. The roomserver should never include events in "AddsStateEventIDs" that are already in the database.

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/membership_table.go
```diff
@@ -0,0 +1,110 @@
...
+    -- The Matrix user ID localpart for the member
+    localpart TEXT NOT NULL,
+    -- The room this user is a member of
+    room_id TEXT NOT NULL,
+    -- The ID of the join membership event
+    event_id TEXT NOT NULL,
+
+    -- A user can only be member of a room once
+    PRIMARY KEY (localpart, room_id)
+);
```
NegativeMjark: It'd be good to have an index on the event_id column so that deleteMembershipByEventID can work efficiently.

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/membership_table.go
```diff
@@ -0,0 +1,110 @@
...
+	"SELECT localpart, room_id FROM memberships WHERE event_id = $1"
+
+const selectMembershipsByLocalpartSQL = "" +
+	"SELECT room_id FROM memberships WHERE localpart = $1"
+
+const deleteMembershipSQL = "" +
+	"DELETE FROM memberships WHERE localpart = $1 AND room_id = $2"
+
+const deleteMembershipByEventIDSQL = "" +
+	"DELETE FROM memberships WHERE event_id = $1"
```
NegativeMjark: You can do multiple deletes in a single SQL statement on postgresql by using `ANY` and passing an array of strings.
There is an example of this in the federation sender https://github.com/matrix-org/dendrite/blob/master/src/github.com/matrix-org/dendrite/federationsender/storage/joined_hosts_table.go#L50

src/github.com/matrix-org/dendrite/clientapi/consumers/roomserver.go
```diff
@@ -0,0 +1,192 @@
...
+		if err := s.updateMembership(event); err != nil {
+			return err
+		}
+	}
+
+	for _, id := range output.NewRoomEvent.RemovesStateEventIDs {
+		if err := s.db.RemoveMembershipByEventID(id); err != nil {
+			return err
+		}
+	}
```
NegativeMjark: I think we need to remove the RemovesStateEventIDs before adding the AddsStateEventIDs. Otherwise if a user changes their display name and resends a join event we'll get a constraint violation if we add the new membership event before removing the old membership event.

NegativeMjark:
It might be a good idea to process the AddsStateEventID and RemovesStateEventID inside a transaction so that the update either succeeds or fails as a single unit.

src/github.com/matrix-org/dendrite/clientapi/consumers/roomserver.go
babolivier: Either I didn't understand your comment correctly, or I need you to explain me why the "check the database and retrieve missing events" thing is done in the sync server here: https://github.com/matrix-org/dendrite/blob/master/src/github.com/matrix-org/dendrite/syncapi/consumers/roomserver.go#L137-L144

NegativeMjark: The syncserver has a separate table containing all the events it ever received. It's possible that some of the events that became part of the room state where ones that the sync server already had received.

However the `membership` table here just contains the current state and if the events were already part of the current state then the room server wouldn't be asking us to add them.

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/membership_table.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (s *membershipStatements) insertMembership(localpart string, roomID string, eventID string, txn *sql.Tx) (err error) {
+	_, err = txn.Stmt(s.insertMembershipStmt).Exec(localpart, roomID, eventID)
+	return
+}
+
+func (s *membershipStatements) deleteMembership(localpart string, roomID string, txn *sql.Tx) (err error) {
+	_, err = txn.Stmt(s.deleteMembershipStmt).Exec(localpart, roomID)
+	return
+}
```
NegativeMjark: You can probably get rid of the `deleteMembership` method.

src/github.com/matrix-org/dendrite/clientapi/consumers/roomserver.go
babolivier: OK, I understand now, thanks.

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/storage.go
```diff
@@ -93,6 +104,56 @@ func (d *Database) CreateAccount(localpart, plaintextPassword string) (*authtype
...
+	return txn.Commit()
+}
+
+// GetMembershipByEventID returns the membership (as a user localpart and a room ID)
+// for which the `join` membership event ID matches a given event ID
+// If no membership match this event ID, the localpart and room ID will be empty strings
+// If an error happens during the retrieval, returns the SQL error
+func (d *Database) GetMembershipByEventID(eventID string) (string, string, error) {
+	return d.memberships.selectMembershipByEventID(eventID)
+}
```
NegativeMjark: I don't think GetMembershipByEventID is used anywhere.  So we can probably get rid of it now.

src/github.com/matrix-org/dendrite/clientapi/consumers/roomserver.go
```diff
@@ -0,0 +1,191 @@
...
+
+	events, err := s.lookupStateEvents(output.NewRoomEvent.AddsStateEventIDs, ev)
+	if err != nil {
+		return err
+	}
+
+	txn, err := s.db.StartTransaction()
+	if err != nil {
+		return err
+	}
```
NegativeMjark: We've generally tried to avoid leaking SQL related stuff outside of the Database abstraction layer.
So I'm not super happy about exposing the SQL transaction here.
It might be simpler to combine adding and removing the memberships into a single method that takes a list of event ID strings to remove and a list event objects to add.

babolivier: Didn't think of this solution (the SQL in the consumer was also bugging me), I'm on it.

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/storage.go
babolivier: Indeed. Not sure it'd be relevant for future improvements either. I'll remove it.

src/github.com/matrix-org/dendrite/clientapi/consumers/roomserver.go
NegativeMjark: The WithTransaction function added by #162 might be useful here.

PR #161, opened 2017-07-17
Use HTTP API for roomserver input.

NegativeMjark:


src/github.com/matrix-org/dendrite/cmd/dendrite-room-server/main.go
```diff
@@ -33,9 +31,8 @@ import (
 )
 
 var (
-	logDir              = os.Getenv("LOG_DIR")
-	configPath          = flag.String("config", "", "The path to the config file. For more information, see the config file in this repository.")
-	stopProcessingAfter = os.Getenv("STOP_AFTER")
+	logDir     = os.Getenv("LOG_DIR")
+	configPath = flag.String("config", "", "The path to the config file. For more information, see the config file in this repository.")
```
NegativeMjark: I've removed the STOP_AFTER option. Previously it was necessary for benchmarking because otherwise there was no easy way to tell when the roomserver had finished processing the test input.
However now that the roomserver uses synchronous HTTP it's much easier to tell when it has finished processing input since we can just wait for our HTTP requests to finish.

Kegsay:
LGTM

PR #163, opened 2017-07-19
Propagate profile update through rooms

babolivier:
Upon profile update, the client API server will now:

* Retrieve all rooms memberships for the user
* Build one event per room which to update the `m.room.member` event for this room (which ID is saved as part of the membership, see #159)
* Update the database with the new `m.room.member` event ID
* Send the events to the room server

This PR completes the implementation of the profile API (as long as user presence isn't implemented).

Edit: In this PR, profile data is also sent in membership events upon room join or creation.


src/github.com/matrix-org/dendrite/clientapi/readers/profile.go
```diff
@@ -175,13 +238,96 @@ func SetDisplayName(
...
+		if err != nil {
+			return nil, err
+		}
+		builder.AuthEvents = refs
+
+		// Generate a new event ID and set it in the database
+		eventID := fmt.Sprintf("$%s:%s", util.RandomString(16), cfg.Matrix.ServerName)
+		if err = db.UpdateMembership(membership.EventID, eventID); err != nil {
+			return nil, err
+		}
```
NegativeMjark: We shouldn't be updating the membership here, we should only be updating the membership using the output from the roomserver.

The idea is that the roomserver can act as a single source of truth for the state of the room. If  we update the state here then we could end up in with a view of the membership that is inconsistent with the room server thinks the state is.

babolivier: Makes more sense, indeed.

src/github.com/matrix-org/dendrite/clientapi/readers/profile.go
```diff
@@ -175,13 +238,96 @@ func SetDisplayName(
...
 
-func getLocalPart(userID string) string {
-	if !strings.HasPrefix(userID, "@") {
-		panic(fmt.Errorf("Invalid user ID"))
+func buildMembershipEvents(
+	memberships []authtypes.Membership, db *accounts.Database,
+	oldProfile *authtypes.Profile, changedKey string, newValue string,
+	userID string, cfg *config.Dendrite, queryAPI api.RoomserverQueryAPI,
+) ([]gomatrixserverlib.Event, error) {
+	ev := []gomatrixserverlib.Event{}
```
NegativeMjark: I'd call "ev" something like "events" or "evs". Using plurals is a good way to hint to the reader that the variable is a collection rather than a single object.

src/github.com/matrix-org/dendrite/clientapi/readers/profile.go
```diff
@@ -175,13 +238,96 @@ func SetDisplayName(
...
+			Membership: "join",
+		}
+
+		if changedKey == "displayname" {
+			content.DisplayName = newValue
+			content.AvatarURL = oldProfile.AvatarURL
+		} else if changedKey == "avatar_url" {
+			content.DisplayName = oldProfile.DisplayName
+			content.AvatarURL = newValue
+		}
```
NegativeMjark: This might be simpler if this function was passed the new profile as a authtypes.Profile

src/github.com/matrix-org/dendrite/clientapi/readers/profile.go
```diff
@@ -175,13 +238,96 @@ func SetDisplayName(
...
+			return nil, err
+		}
+
+		authEvents := gomatrixserverlib.NewAuthEvents(nil)
+
+		// Iterating the old way because range seems to mess things up. Might be
+		// worth investigating.
+		for i := 0; i < len(queryRes.StateEvents); i++ {
+			authEvents.AddEvent(&queryRes.StateEvents[i])
+		}
```
NegativeMjark: Ah yes this can be a bit subtle.

This should work:

```go
for i := range(queryRes.StateEvents) {
    authEvents.AddEvent(&queryRes.StateEvents[i])
}
```

However this will have problems:

```go
for _, event := range(queryRes.StateEvents) {
    authEvents.AddEvent(&event)
}
````

The reason is that authEvents will be passed an address to the iteration variable "events" that is clobbered on each iteration of the loop.

File: src/github.com/matrix-org/dendrite/clientapi/readers/profile.go [hunk 5]
```diff
@@ -175,13 +238,96 @@ func SetDisplayName(
...
+		builder := gomatrixserverlib.EventBuilder{
+			Sender:   userID,
+			RoomID:   membership.RoomID,
+			Type:     "m.room.member",
+			StateKey: &userID,
+		}
+
+		if err := builder.SetUnsigned(prev); err != nil {
+			return nil, err
+		}
```
NegativeMjark: I don't think this is the right place to set the prevContent. Unfortunately the concept of a "prev_content" doesn't really make sense until the events have been linearised by the roomserver. So I think that in general this is something that has to be synthesised in the syncserver.

src/github.com/matrix-org/dendrite/clientapi/readers/profile.go
babolivier: Ah, yeah, totally forgot `range` also gives you access to the index.

src/github.com/matrix-org/dendrite/clientapi/readers/profile.go
babolivier: Erf, roger that. I'll get on diving into the syncserver code then.

babolivier:
My latest commits address the issues @NegativeMjark raised yesterday.

FTR, I think it'd be good to move `clientapi/events/eventcontent.go` to `common` as we now also use it in the sync API server.

babolivier:
FTR, I just added a missing check on profile update routes to prevent a user from updating another user's profile.

src/github.com/matrix-org/dendrite/syncapi/storage/syncserver.go
```diff
@@ -100,9 +100,18 @@ func (d *SyncServerDatabase) WriteEvent(
...
+			// If the event is a m.room.member event, and has unsigned content,
+			// we need to save it as it is very likely to be a membership update
+			// (e.g. if a user updates their profile)
+			if ev.Type() == "m.room.member" && len(ev.Unsigned()) > 0 {
+				addStateEvents = append(addStateEvents, *ev)
+			} else if len(removeStateEventIDs) == 0 {
+				// Nothing to do, the event may have just been a message event.
+				return nil
+			}
 		}
```
NegativeMjark: This seems like a *very* odd thing to be doing? Why are we trying to force an event that isn't part of the current state into being part of the current state?

babolivier: We need to save the event in DB (so we can reference it in future updates) as an update of the user's membership in the room. However, the event isn't marked as being part of the current state. Adding it to the current state array seemed to me like the simplest way to have our event processed by the upsert.

NegativeMjark: Hmm, it seems odd that it isn't being marked as part of the current state. I wonder if there is a bug in the room server.

src/github.com/matrix-org/dendrite/syncapi/consumers/roomserver.go
```diff
@@ -177,6 +202,63 @@ func (s *OutputRoomEvent) lookupStateEvents(
...
+
+	prev := prevMembership{
+		PrevContent: content,
+		PrevID:      prevEvent.EventID(),
+		UserID:      prevEvent.Sender(),
+	}
+
+	if err = builder.SetUnsigned(prev); err != nil {
+		return event, err
+	}
```
NegativeMjark: I'm adding a way to set the unsigned field of an event without building a new one in matrix-org/gomatrixserverlib#55. You will want to use that when it lands.

src/github.com/matrix-org/dendrite/syncapi/storage/syncserver.go
babolivier: If this is a bug, it must come from the roomserver because the empty array is created from the `AddsStateEventIDs` array provided by the roomserver.

src/github.com/matrix-org/dendrite/syncapi/consumers/roomserver.go
babolivier: Indeed! This will make the process a lot simpler 😃

NegativeMjark: I've matrix-org/gomatrixserverlib#55 has landed and I've updated the dependency on the dendrite master branch FWIW

src/github.com/matrix-org/dendrite/clientapi/readers/profile.go
```diff
@@ -175,13 +258,68 @@ func SetDisplayName(
...
+		builder.AuthEvents = refs
+
+		eventID := fmt.Sprintf("$%s:%s", util.RandomString(16), cfg.Matrix.ServerName)
+		now := time.Now()
+		event, err := builder.Build(eventID, now, cfg.Matrix.ServerName, cfg.Matrix.KeyID, cfg.Matrix.PrivateKey)
+		if err != nil {
+			return nil, err
+		}
```
NegativeMjark: I might be missing something here, but are you setting the PrevEvents or the Depth on the builder?

NegativeMjark: If we aren't then it might explain the problems that resulted in https://github.com/matrix-org/dendrite/pull/163#discussion_r128984277 . (If we aren't setting the depth and prev_events then the membership events will conflict with the existing membership events and because the depth defaults to 0 it will lose in state conflict resolution.

babolivier: Oh, OK, I missed that, thanks.
I'm already working on setting those, should push something soon.

src/github.com/matrix-org/dendrite/syncapi/consumers/roomserver.go
```diff
@@ -177,6 +202,44 @@ func (s *OutputRoomEvent) lookupStateEvents(
...
+	if event.Type() != "m.room.member" {
+		return event, nil
+	}
+	membership, err := event.Membership()
+	if err != nil {
+		return event, err
+	}
+	if membership != "join" {
+		return event, nil
+	}
```
NegativeMjark: Should we be settting the "prev_content", "replaces_state" and "prev_sender" for every state event? not just the join m.room.member events?

babolivier: Makes sense, I didn't think about that.

babolivier: Ah, there would be an issue with that. The reason I use the `event.MemberContent` here: https://github.com/matrix-org/dendrite/blob/37a9db5e6ac4d9b8fa7e9457eeceb4903e794fd8/src/github.com/matrix-org/dendrite/syncapi/consumers/roomserver.go#L45 is that, because `event.Content()` returns `[]byte`, when I directly set it in "prev_content" without unmarshalling it, it will be processed as a string ("prev_content" will just be an useless string in the actual request).
To fix this, I'm unmarshalling it into a structure specific to the event. So, unless we have a structure for the content of each event type the roomserver will receive, and a way to automatically use the right one depending on the type, it won't work :/

src/github.com/matrix-org/dendrite/syncapi/consumers/roomserver.go
```diff
@@ -35,6 +36,15 @@ type OutputRoomEvent struct {
 	db                 *storage.SyncServerDatabase
 	notifier           *sync.Notifier
 	query              api.RoomserverQueryAPI
+	serverName         gomatrixserverlib.ServerName
+	keyID              gomatrixserverlib.KeyID
+	privateKey         []byte
+}
+
+type prevMembership struct {
+	PrevContent events.MemberContent `json:"prev_content"`
```
NegativeMjark: You can use `*json.RawMessage` here if you don't need to parse the old content.

NegativeMjark: ( `json.RawMessage` is a named type for a `[]byte` containing raw JSON which implements `json.Marshaller` by returning the raw JSON.)

babolivier: Wasn't aware this type existed, thanks.

babolivier:
OK, so tests are still failing. That's because, since we update every state event, we also update non-join membership events.

Specifically, the failing test is expecting this:

```json
{
    "content": {
        "membership": "leave"
    },
    "event_id": "$om1F4AI8tCYlHUSp:localhost",
    "origin_server_ts": 1494411855278,
    "sender": "@alice:localhost",
    "state_key": "@charlie:localhost",
    "type": "m.room.member"
}
```
and get this:
```json
{
    "content": {
        "membership": "leave"
    },
    "event_id": "$om1F4AI8tCYlHUSp:localhost",
    "origin_server_ts": 1494411855278,
    "sender": "@alice:localhost",
    "state_key": "@charlie:localhost",
    "type": "m.room.member",
    "unsigned": {
        "prev_content": {
            "membership": "join"
        },
        "prev_sender": "@charlie:localhost",
        "replaces_state": "$uJVKyzZi8ZX0kOd9:localhost"
    }
}
```

@NegativeMjark What should we do? Update the test or add a specific exception to membership updates for `leave` events?

PR #165, opened 2017-07-24
Add tables for tracking the state of invites to the room server.

NegativeMjark:
This adds tables to the roomserver for tracking which invites are active for a given user.
An invite is removed when the user joins the room or sends a leave event to reject the invite.
The invites are tracked separately from the usual room events because you can receive invites for rooms that the server isn't currently in.

Invites for rooms the server is not in will be handled by adding additional input APIs on the roomserver that the federationapi component will hit. Those APIs are not implemented in this pull request, and will be implemented later.


src/github.com/matrix-org/dendrite/roomserver/input/latest_events.go
```diff
@@ -66,69 +66,76 @@ func updateLatestEvents(
...
-	sendAsServer string,
-) error {
+type latestEventsUpdater struct {
+	db                      RoomEventDatabase
+	updater                 types.RoomRecentEventsUpdater
+	ow                      OutputRoomEventWriter
+	roomNID                 types.RoomNID
+	stateAtEvent            types.StateAtEvent
+	event                   gomatrixserverlib.Event
+	sendAsServer            string
```
Kegsay: Shouldn't this be a server domain type?

src/github.com/matrix-org/dendrite/roomserver/input/latest_events.go
```diff
@@ -66,69 +66,76 @@ func updateLatestEvents(
...
-func doUpdateLatestEvents(
-	db RoomEventDatabase,
-	updater types.RoomRecentEventsUpdater,
-	ow OutputRoomEventWriter,
-	roomNID types.RoomNID,
-	stateAtEvent types.StateAtEvent,
-	event gomatrixserverlib.Event,
-	sendAsServer string,
-) error {
+type latestEventsUpdater struct {
```
Kegsay: I'm struggling to identify what the purpose is of this struct. Can you please add some comments to the type definition.

src/github.com/matrix-org/dendrite/roomserver/input/membership.go
```diff
@@ -0,0 +1,223 @@
...
+	"fmt"
+
+	"github.com/matrix-org/dendrite/roomserver/api"
+	"github.com/matrix-org/dendrite/roomserver/types"
+	"github.com/matrix-org/gomatrixserverlib"
+)
+
+func updateMemberships(
+	db RoomEventDatabase, updater types.RoomRecentEventsUpdater, removed, added []types.StateEntry,
+) ([]api.OutputEvent, error) {
```
Kegsay: It's not clear what the output is from this function. Please comment.

src/github.com/matrix-org/dendrite/roomserver/input/membership.go
```diff
@@ -0,0 +1,223 @@
...
+				re = &ev.Event
+			}
+			targetNID = change.removed.EventStateKeyNID
+		}
+		if change.added.EventNID != 0 {
+			ev, _ := eventMap(events).lookup(change.added.EventNID)
+			if ev != nil {
+				ae = &ev.Event
+			}
+			targetNID = change.added.EventStateKeyNID
```
Kegsay: So I was wondering why I didn't like this. I think it's because of the assumptions we're making here. We seem to clobber `targetNID` without checking if it was set on `:54` by the `removed` if statement. Why? I assume it's because we're assuming that they are always the same, but that's not clear from the code.

src/github.com/matrix-org/dendrite/roomserver/input/membership.go
```diff
@@ -0,0 +1,223 @@
...
+		if updates, err = updateMembership(updater, targetNID, re, ae, updates); err != nil {
+			return nil, err
+		}
+	}
+	return nil, nil
+}
+
+func updateMembership(
+	updater types.RoomRecentEventsUpdater, targetNID types.EventStateKeyNID,
+	remove *gomatrixserverlib.Event, add *gomatrixserverlib.Event,
```
Kegsay: aka:
```go
remove, add *gomatrixserverlib.Event
```

src/github.com/matrix-org/dendrite/roomserver/input/membership.go
```diff
@@ -0,0 +1,223 @@
...
+		}
+		if updates, err = updateMembership(updater, targetNID, re, ae, updates); err != nil {
+			return nil, err
+		}
+	}
+	return nil, nil
+}
+
+func updateMembership(
+	updater types.RoomRecentEventsUpdater, targetNID types.EventStateKeyNID,
```
Kegsay: Nit: I dislike the var name `targetNID` as I translate that as "target id" and that has no context to it... is it an event ID? A state key ID? A user ID? Maybe `stateKeyNID`?

src/github.com/matrix-org/dendrite/roomserver/input/membership.go
```diff
@@ -0,0 +1,223 @@
...
+	if old == new {
+		return updates, nil
+	}
+
+	mu, err := updater.MembershipUpdater(targetNID)
+	if err != nil {
+		return nil, err
+	}
+
+	return updateWithNewMembership(mu, new, add, updates)
```
Kegsay: You call this function in this place only and you just made a `MembershipUpdater` in order to pass it to this function. Why don't you put `updateWithNewMembership` as a function on `MembershipUpdater`?

src/github.com/matrix-org/dendrite/roomserver/input/membership.go
```diff
@@ -0,0 +1,223 @@
...
+	updates []api.OutputEvent,
+) ([]api.OutputEvent, error) {
+	switch newMembership {
+	case "invite":
+		needsSending, err := mu.SetToInvite(*add)
+		if err != nil {
+			return nil, err
+		}
+		if needsSending {
+			updates = appendInviteUpdate(updates, add)
```
Kegsay: I'd inline this tbh

src/github.com/matrix-org/dendrite/roomserver/input/membership.go
```diff
@@ -0,0 +1,223 @@
...
+		return updates, nil
+	case "join":
+		if mu.IsJoin() {
+			return updates, nil
+		}
+		retired, err := mu.SetToJoin(add.Sender())
+		if err != nil {
+			return nil, err
+		}
+		return appendRetireUpdates(updates, add, newMembership, retired), nil
```
Kegsay: We spoke about this yesterday: it'd be great if we could have a function per case statement rather than factoring out at this point.

src/github.com/matrix-org/dendrite/roomserver/input/membership.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+	return updates
+}
+
+type stateChange struct {
+	removed types.StateEntry
+	added   types.StateEntry
+}
+
+func pairUpChanges(removed, added []types.StateEntry) []stateChange {
```
Kegsay: Not a great function name. This is grouping together state entries based on their state key tuple. I'd prefer `groupByStateKey(removed, added)` personally.

Kegsay: Also, it looks like we're relying on `removed` and `added` having a particular sort order, can you please state what that is.

src/github.com/matrix-org/dendrite/roomserver/input/membership.go
```diff
@@ -0,0 +1,223 @@
...
+			result = append(result, stateChange{
+				removed: removed[ri],
+				added:   added[ai],
+			})
+			ai++
+			ri++
+		case added[ai].StateKeyTuple.LessThan(removed[ri].StateKeyTuple):
+			result = append(result, stateChange{added: added[ai]})
+			ai++
+		default:
```
Kegsay: I don't like the fact that you've represented this as a switch/case. You really have 2 separate operations:
 - If we've reached the end, dump the remaining from the other slice into `result` then `return`.
 - Check if the `StateKeyTuple` is >, = or < and then append add/rm/both.

Separately, I don't understand why "if the state key tuple of added is less than the state key tuple of removed, append the added only". Why do we bias for lower entries? There's some significance I'm missing here, which you need to comment on.

src/github.com/matrix-org/dendrite/roomserver/input/latest_events.go
```diff
@@ -66,69 +66,76 @@ func updateLatestEvents(
...
+	updater                 types.RoomRecentEventsUpdater
+	ow                      OutputRoomEventWriter
+	roomNID                 types.RoomNID
+	stateAtEvent            types.StateAtEvent
+	event                   gomatrixserverlib.Event
+	sendAsServer            string
+	lastEventIDSent         string
+	latest                  []types.StateAtEventAndReference
+	removed                 []types.StateEntry
+	added                   []types.StateEntry
```
Kegsay: You need to comment on the semantics/relationship between `removed` and `added`. They can be of unequal length (which I have derived from a random function), and we're going to compare the state-key-tuples of each element left-to-right and keep the 'lesser'. I don't know why we do this.

src/github.com/matrix-org/dendrite/roomserver/input/membership.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+}
+
+// membershipChanges pairs up the membership state changes from a sorted list
+// of state removed and a sorted list of state added.
+func membershipChanges(removed, added []types.StateEntry) []stateChange {
+	changes := pairUpChanges(removed, added)
+	var result []stateChange
+	for _, c := range changes {
+		if c.added.EventTypeNID == types.MRoomMemberNID ||
```
Kegsay: It's rather curious you've chosen to design this code like this. So, as I understand it:
 - We pair up all state changes in the removed/added lists.
 - We then filter out all state changes except for the `m.room.member` ones.
 - We return this filtered list of state changes.

Why do we bother working out the list of all state changes if we're just going to discard (most of?) them immediately? We never call `pairUpChanges` anywhere else.

src/github.com/matrix-org/dendrite/roomserver/storage/invite_table.go
```diff
@@ -0,0 +1,149 @@
...
+	room_nid BIGINT NOT NULL,
+	-- The numeric ID for the state key of the invite m.room.member event.
+	-- This tells us who the invite is for.
+	-- This is used to query the active invites for a user.
+	target_nid BIGINT NOT NULL,
+	-- The numeric ID for the sender of the invite m.room.member event.
+	-- This tells us who sent the invite.
+	-- This is used to work out which matrix server we should talk to when
+	-- we try to join the room.
+	sender_nid BIGINT NOT NULL DEFAULT 0,
```
Kegsay: Why does this have `DEFAULT 0` but the `target_nid` doesn't?

src/github.com/matrix-org/dendrite/roomserver/storage/invite_table.go
```diff
@@ -0,0 +1,149 @@
...
+	-- This is used to track whether the invite is still active.
+	-- This is set implicitly when processing KIND_NEW events and explicitly
```
Kegsay: > This is set implicitly when processing KIND_NEW events

In English please? You mean, this is set implicitly if we receive an explicit join/leave event?

src/github.com/matrix-org/dendrite/roomserver/storage/invite_table.go
```diff
@@ -0,0 +1,149 @@
...
+	"INSERT INTO invites (invite_event_id, room_nid, target_nid," +
+	" sender_nid, invite_event_json) VALUES ($1, $2, $3, $4, $5)" +
+	" ON CONFLICT DO NOTHING"
+
+const selectInviteActiveForUserInRoomSQL = "" +
+	"SELECT invite_event_id, sender_nid FROM invites" +
+	" WHERE target_nid = $1 AND room_nid = $2" +
+	" AND NOT retired"
+
+// Retire every active invite.
```
Kegsay: > Retire every active invite **between a user and a room**.

Right? Since if 5 people invite me to room X, this will retire all 5 invites.

src/github.com/matrix-org/dendrite/roomserver/storage/invite_table.go
```diff
@@ -0,0 +1,149 @@
...
+		targetNID, roomNID,
+	)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+	var result []types.EventStateKeyNID
+	for rows.Next() {
+		var senderNID int64
+		if err := rows.Scan(&senderNID); err != nil {
```
Kegsay: The SQL statement is:
```
SELECT invite_event_id, sender_nid FROM invites
```
But here we extract the `senderNID` only (and more importantly, first). This is presumably wrong.

src/github.com/matrix-org/dendrite/roomserver/storage/membership_table.go
```diff
@@ -0,0 +1,106 @@
...
+import (
+	"database/sql"
+
+	"github.com/matrix-org/dendrite/roomserver/types"
+)
+
+type membershipState int64
+
+const (
+	membershipStateLeaveOrBan membershipState = 0
```
Kegsay: I don't like `0` enums because it is the zero-value for ints. It's too easy to make a mistake here, when we have usages like:
```go
return u.membership == membershipStateLeaveOrBan
```
A zero-value `membershipUpdater` would return `true` here, whereas really it should be failing (fast preferably).

src/github.com/matrix-org/dendrite/roomserver/storage/membership_table.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	return statementList{
+		{&s.insertMembershipStmt, insertMembershipSQL},
+		{&s.selectMembershipForUpdateStmt, selectMembershipForUpdateSQL},
+		{&s.updateMembershipStmt, updateMembershipSQL},
+	}.prepare(db)
+}
+
+func (s *membershipStatements) insertMembership(
```
Kegsay: It's not really inserting a membership though, is it? It's inserting a row for the given member which can later have their membership inserting into it.

src/github.com/matrix-org/dendrite/roomserver/types/types.go
```diff
@@ -163,10 +171,33 @@ type RoomRecentEventsUpdater interface {
...
+	Transaction
+}
+
+// A MembershipUpdater is used to update the membership of a user in a room.
+// (On postgresql this wraps a database transaction that holds a "FOR UPDATE"
+//  lock on the row in the membership table for this user in the room)
+type MembershipUpdater interface {
+	// True if the target user is invited to the room.
+	IsInvite() bool
+	// True if the target user is joined to the room.
```
Kegsay: Is this the value before or after updating? Or the current value?

src/github.com/matrix-org/dendrite/roomserver/types/types.go
```diff
@@ -163,10 +171,33 @@ type RoomRecentEventsUpdater interface {
...
+	IsJoin() bool
+	// True if the target user is not invited or joined to the room.
+	IsLeave() bool
+	// Set the state to invite.
+	// Returns whether this invite needs to be sent
+	SetToInvite(event gomatrixserverlib.Event) (needsSending bool, err error)
```
Kegsay: Does this hold?
```go
if !updater.IsInvite() {
    updater.SetToInvite(event)
}
assert(updater.IsInvite())
```
If so, you need to state that. Likewise for the rest of them.

src/github.com/matrix-org/dendrite/roomserver/types/types.go
```diff
@@ -163,10 +171,33 @@ type RoomRecentEventsUpdater interface {
...
+	// Set the state to join.
+	SetToJoin(senderID string) (inviteIDs []string, err error)
```
Kegsay: What is `senderID`? I assume the user whose membership is being updated? Weird var name though, given we have the idea of "sender" and "target" for updating memberships.

Kegsay: What is `inviteIDs`? I guess event IDs but only because I have read the impl first...

src/github.com/matrix-org/dendrite/roomserver/types/types.go
```diff
@@ -163,10 +171,33 @@ type RoomRecentEventsUpdater interface {
...
+	// Set the state to leave.
+	// Returns a list of invite event IDs that this state change retired.
+	SetToLeave(senderID string) (inviteIDs []string, err error)
+	// Implements Transaction so it can be committed or rolledback.
+	Transaction
```
Kegsay: Interesting design idea, let's see how well it is in practice..

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
```diff
@@ -381,3 +379,128 @@ func (d *Database) StateEntriesForTuples(
...
+
+// IsLeave implements types.MembershipUpdater
+func (u *membershipUpdater) IsLeave() bool {
+	return u.membership == membershipStateLeaveOrBan
+}
+
+// SetToInvite implements types.MembershipUpdater
+func (u *membershipUpdater) SetToInvite(event gomatrixserverlib.Event) (bool, error) {
+	// TODO: assign the state key inside the transaction.
+	senderNID, err := u.d.assignStateKeyNID(event.Sender())
```
Kegsay: Why can't you just do this? Pass the `txn` to `assignStateKeyNID`?

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
```diff
@@ -381,3 +379,128 @@ func (d *Database) StateEntriesForTuples(
...
+	if err != nil {
+		return false, err
+	}
+	eventID := event.EventID()
```
Kegsay: Why aren't you inlining this? You never use this variable anywhere else. In this form, it just adds to my cognitive load as I have to parse each function and consider where you might update it / use it elsewhere. Inlining it removes all those considerations.

src/github.com/matrix-org/dendrite/roomserver/types/types.go
```diff
@@ -163,10 +171,33 @@ type RoomRecentEventsUpdater interface {
...
+	// True if the target user is not invited or joined to the room.
+	IsLeave() bool
+	// Set the state to invite.
+	// Returns whether this invite needs to be sent
+	SetToInvite(event gomatrixserverlib.Event) (needsSending bool, err error)
+	// Set the state to join.
+	SetToJoin(senderID string) (inviteIDs []string, err error)
+	// Set the state to leave.
+	// Returns a list of invite event IDs that this state change retired.
+	SetToLeave(senderID string) (inviteIDs []string, err error)
```
Kegsay: Are these calls idempotent? Can I call `SetToLeave` twice and get the same IDs back without any errors? You need to either way in your interface.

src/github.com/matrix-org/dendrite/roomserver/input/membership.go
Kegsay: On reflection, `targetUserNID` is probably the most desirable form.

src/github.com/matrix-org/dendrite/roomserver/input/membership.go
NegativeMjark: I think MembershipUpdater is an interface?

src/github.com/matrix-org/dendrite/roomserver/input/membership.go
NegativeMjark: Done db3f3a9

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
NegativeMjark: Done 89ee9da

src/github.com/matrix-org/dendrite/roomserver/input/membership.go
NegativeMjark: Done.

src/github.com/matrix-org/dendrite/roomserver/input/latest_events.go
NegativeMjark: Maybe. I might fix that in a separate PR given how big this PR already is.

src/github.com/matrix-org/dendrite/roomserver/input/latest_events.go
NegativeMjark: Done 3931af3

src/github.com/matrix-org/dendrite/roomserver/input/membership.go
Kegsay: So? You can have additional private functions on a concrete impl.

src/github.com/matrix-org/dendrite/roomserver/input/membership.go
NegativeMjark: Done 6f5cf19

src/github.com/matrix-org/dendrite/roomserver/input/membership.go
NegativeMjark: Done

src/github.com/matrix-org/dendrite/roomserver/input/membership.go
NegativeMjark: Done

src/github.com/matrix-org/dendrite/roomserver/input/membership.go
NegativeMjark: Changed

src/github.com/matrix-org/dendrite/roomserver/input/membership.go
NegativeMjark: MembershipUpdater is implemented in the storage package.

src/github.com/matrix-org/dendrite/roomserver/input/membership.go
NegativeMjark: It pairs up the adds and the removes? I don't see a problem with the name tbh.

src/github.com/matrix-org/dendrite/roomserver/input/membership.go
NegativeMjark: You've reviewed this before :) in https://github.com/matrix-org/dendrite/blob/master/src/github.com/matrix-org/dendrite/roomserver/state/state.go#L197-L222.

I've added some comments in e25ad65 that might help :)

src/github.com/matrix-org/dendrite/roomserver/storage/invite_table.go
NegativeMjark: I've left a comment.

src/github.com/matrix-org/dendrite/roomserver/input/membership.go
NegativeMjark: It used to be one function, but that triggered gocyclo :)

src/github.com/matrix-org/dendrite/roomserver/storage/invite_table.go
NegativeMjark: I've attempted to translate.

src/github.com/matrix-org/dendrite/roomserver/storage/invite_table.go
NegativeMjark: Fixed dc18fa9

src/github.com/matrix-org/dendrite/roomserver/storage/invite_table.go
NegativeMjark: Fixed.

src/github.com/matrix-org/dendrite/roomserver/storage/membership_table.go
NegativeMjark: Fixed

src/github.com/matrix-org/dendrite/roomserver/storage/membership_table.go
NegativeMjark: Commented 0768166

src/github.com/matrix-org/dendrite/roomserver/types/types.go
NegativeMjark: I've updated the comment.

src/github.com/matrix-org/dendrite/roomserver/types/types.go
NegativeMjark: Nope. Commented

src/github.com/matrix-org/dendrite/roomserver/types/types.go
NegativeMjark: I've changed the names.

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
NegativeMjark: I've inlined it now.

src/github.com/matrix-org/dendrite/roomserver/types/types.go
NegativeMjark: Nope, Nope, commented.

src/github.com/matrix-org/dendrite/roomserver/input/latest_events.go
NegativeMjark: commented 4ffc180

Kegsay:
Probably LGTM. The sheer scale of this PR makes me have less confidence in the quality of my own review. Please keep them smaller in future.

PR #166, opened 2017-07-25
Save and retrieve account data

babolivier:
* Save account data upon request on:
    * `PUT /_matrix/client/r0/user/{userId}/rooms/{roomId}/account_data/{type}`
    * `PUT /_matrix/client/r0/user/{userId}/account_data/{type}`
* Send global account data and room-specific account data upon request on `GET /_matrix/client/r0/sync`


src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/account_data_table.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/matrix-org/gomatrixserverlib"
+)
+
+const accountDataSchema = `
+-- Stores data about accounts data.
+CREATE TABLE IF NOT EXISTS account_data (
+    -- The Matrix user ID localpart for this account
+    localpart TEXT NOT NULL,
+    -- The room ID for this data (null if not specific to a room)
+    room_id TEXT,
```
NegativeMjark: Are you using null or are you using the empty string here?

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/account_data_table.go
```diff
@@ -0,0 +1,131 @@
...
+);
+`
+
+const insertAccountDataSQL = `
+	INSERT INTO account_data(localpart, room_id, type, content) VALUES($1, $2, $3, $4)
+	ON CONFLICT (localpart, room_id, type) DO UPDATE SET content = EXCLUDED.content
+`
+
+const selectGlobalAccountDataSQL = "" +
+	"SELECT type, content FROM account_data WHERE localpart = $1 AND room_id = ''"
```
NegativeMjark: If we are using the empty string for the global account data then we probably can use the same SQL for selecting global account data as we are using for local account data.

src/github.com/matrix-org/dendrite/syncapi/sync/requestpool.go
```diff
@@ -77,9 +80,14 @@ func (rp *RequestPool) OnIncomingSyncRequest(req *http.Request, device *authtype
 		if err != nil {
 			res = httputil.LogThenError(req, err)
 		} else {
-			res = util.JSONResponse{
-				Code: 200,
-				JSON: syncData,
+			syncData, err = rp.appendAccountData(syncData, device.UserID)
```
NegativeMjark: Are we including all the account data to every sync response? It might be worth adding a TODO marker to say that we should include the data that has changed for incremental sync responses.

babolivier: Makes sense, indeed.

src/github.com/matrix-org/dendrite/syncapi/sync/requestpool.go
```diff
@@ -104,3 +112,27 @@ func (rp *RequestPool) currentSyncForUser(req syncRequest, currentPos types.Stre
...
+	}
+
+	events, err := rp.accountDB.GetAccountData(localpart, "")
+	if err != nil {
+		return nil, err
+	}
+	data.AccountData.Events = events
+
+	for r, j := range data.Rooms.Join {
+		events, err := rp.accountDB.GetAccountData(localpart, r)
```
NegativeMjark: It might be nice to do this as a single bulk lookup that returns all the account data for a single user. That way we would have fewer round trips to the database.

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/account_data_table.go
NegativeMjark: It looks like you are using the empty string, you might want to update the comment.

PR #167, opened 2017-07-27
Implement room aliases

babolivier:
Implements:

* `PUT /_matrix/client/r0/directory/room/{roomAlias}`
* `GET /_matrix/client/r0/directory/room/{roomAlias}`
* `DELETE /_matrix/client/r0/directory/room/{roomAlias}`
* Sending a `m.room.aliases` state event to the room upon change in the aliases list

Note: I did all the processing in the `query` package of the room server. However, I'm not sure it's the right place for it to be, let me know if you want it move elsewhere.

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/readers/directory.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/readers/directory.go
@@ -22,10 +22,15 @@ import (
 	"github.com/matrix-org/dendrite/clientapi/httputil"
 	"github.com/matrix-org/dendrite/clientapi/jsonerror"
 	"github.com/matrix-org/dendrite/common/config"
+	"github.com/matrix-org/dendrite/roomserver/api"
 	"github.com/matrix-org/gomatrix"
 	"github.com/matrix-org/gomatrixserverlib"
 	"github.com/matrix-org/util"
 )
+
+type roomID struct {
+	RoomID string `json:"room_id"`
+}
 
 // DirectoryRoom looks up a room alias
 func DirectoryRoom(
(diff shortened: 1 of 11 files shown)


babolivier:
FTR build is failing in travis, however I don't undersand the message and I can't reproduce it locally, so I'll need some help on this one.

NegativeMjark:
Hmm it looks like the pre-commit hook is failing. Have you tried running `./hooks/pre-commit` on your local copy?

(You can get git to run the hooks automatically before committing by running `./hooks/install.sh`)

babolivier:
I always commit with the pre-commit hook installed, and these commits were no exception :/
The pre-commit hook worked just fine, and running `./travis-test.sh` locally didn't even fail...

NegativeMjark:
I wonder if you have a different version of gocyclo then. I'll try updating my copy and see if it makes a difference.

NegativeMjark:
So when I run `gocyclo` using the latest version I get:

```bash
$ go cyclo -over 12 src/
13 query (*RoomserverQueryAPI).SetupHTTP src/github.com/matrix-org/dendrite/roomserver/query/query.go:348:1
```

babolivier:
Oh, I just updated mine and got the error. Weird, I installed my laptop's current distro way after gocyclo's latest commit on GitHub.

I'll work on it then.

src/github.com/matrix-org/dendrite/roomserver/storage/room_aliases_table.go
```diff
@@ -0,0 +1,98 @@
...
+`
+
+const insertRoomAliasSQL = "" +
+	"INSERT INTO room_aliases (alias, room_id) VALUES ($1, $2)"
+
+const selectRoomIDFromAliasSQL = "" +
+	"SELECT room_id FROM room_aliases WHERE alias = $1"
+
+const selectAliasesFromRoomIDSQL = "" +
+	"SELECT alias FROM room_aliases WHERE room_id = $1"
```
NegativeMjark: I'd be happier if there was an index on `room_aliases(room_id)` for this query.

src/github.com/matrix-org/dendrite/clientapi/readers/directory.go
```diff
@@ -62,10 +86,86 @@ func DirectoryRoom(
...
 	}
+
+	if domain != cfg.Matrix.ServerName {
+		return util.JSONResponse{
+			Code: 403,
+			JSON: jsonerror.Forbidden("Alias must be on local homeserver"),
+		}
+	}
+
+	var r roomID
```
NegativeMjark: I'd inline the struct definition here so just write:
```go
var r struct {
    RoomID `json:"room_id"`
}
```

That way I'd avoid having to think of a good name for the struct type :)
(calling it "roomID" is confusing because it's a struct containing a room ID rather than being a room ID)
If we really want it to be a named type then I'd call it something like "setAliasRequest".

PR #171, opened 2017-08-02
Implement membership APIs

babolivier:
This PR:

* Implements `POST /_matrix/client/r0/rooms/{roomID}/(join|kick|ban|unban|leave|invite)`: build and send a `m.room.member` event when an authenticated user hits one of these endpoints
* Adds a `FillBuilder` function in the client API server because some steps of building an events were becoming large copy/pastes from one file to another, and edits said files to call this function when possible

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/writers/membership.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/writers/membership.go
@@ -77,7 +77,7 @@ func SendMembership(
 		return httputil.LogThenError(req, err)
 	}
 
-	if err = events.FillBuilder(&builder, queryAPI); err == events.ErrRoomNoExists {
+	if err = events.FillBuilder(&builder, queryAPI, nil); err == events.ErrRoomNoExists {
 		return util.JSONResponse{
 			Code: 404,
 			JSON: jsonerror.NotFound(err.Error()),
--- a/src/github.com/matrix-org/dendrite/clientapi/writers/membership.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/writers/membership.go
@@ -95,6 +95,11 @@ func SendMembership(
 	}
 }
 
+// getMembershipStateKey extracts the target user ID of a membership change.
+// For "join" and "leave" this will be the ID of the user making the change.
+// For "ban", "unban", "kick" and "invite" the target user ID will be in the JSON request body.
+// In the latter case, if there was an issue retrieving the user ID from the request body,
+// returns a JSONResponse with a corresponding error code and message.
 func getMembershipStateKey(
 	req *http.Request, device *authtypes.Device, membership string,
 ) (stateKey string, reason string, response *util.JSONResponse) {
--- a/src/github.com/matrix-org/dendrite/clientapi/writers/membership.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/writers/membership.go
@@ -42,14 +42,19 @@ func SendMembership(
 		return *reqErr
 	}
 
-	localpart, _, err := gomatrixserverlib.SplitID('@', stateKey)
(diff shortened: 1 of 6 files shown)


src/github.com/matrix-org/dendrite/clientapi/writers/membership.go
```diff
@@ -0,0 +1,137 @@
...
+
+	// "unban" or "kick" isn't a valid membership value, change it to "leave"
+	if membership == "unban" || membership == "kick" {
+		membership = "leave"
+	}
+
+	content := events.MemberContent{
+		Membership:  membership,
+		DisplayName: profile.DisplayName,
+		AvatarURL:   profile.AvatarURL,
```
NegativeMjark: It looks like the profile information here is always the event sender. I don't think that is correct for kicks, bans and invites where the profile information should be for the user ID in the state key rather than the sender.

babolivier: Indeed, didn't catch this one

src/github.com/matrix-org/dendrite/clientapi/writers/membership.go
```diff
@@ -0,0 +1,137 @@
...
+	} else if err != nil {
+		return httputil.LogThenError(req, err)
+	}
+
+	eventID := fmt.Sprintf("$%s:%s", util.RandomString(16), cfg.Matrix.ServerName)
+	now := time.Now()
+	event, err := builder.Build(eventID, now, cfg.Matrix.ServerName, cfg.Matrix.KeyID, cfg.Matrix.PrivateKey)
+	if err != nil {
+		return httputil.LogThenError(req, err)
+	}
```
NegativeMjark: It looks like every call site for FillBuilder immediately calls builder.Build afterwards. I wonder if we should have a "FillBuildAndSendEvent()" utility function.

babolivier: Adding the builder.Build would be relevant (adding it rn), but I'm not sure about sending the event in the same function though, as some function do stuff between building and sending the event (e.g. `profile.go` aggregates all events into an array before sending all the event in one call, `sendevent.go` does some additional checks between building and sending the event, etc).

NegativeMjark: Okay seems reasonable.

src/github.com/matrix-org/dendrite/clientapi/writers/membership.go
```diff
@@ -0,0 +1,137 @@
...
+
+	return util.JSONResponse{
+		Code: 200,
+		JSON: struct{}{},
+	}
+}
+
+func getMembershipStateKey(
```
NegativeMjark: Maybe add a bit of doc string to say what this does, something like:
```go
// getMembershipStateKey extracts the target user ID of a membership change.
// For "join" and "leave" this will be the ID of the user making the change.
// For "ban", "unban", "kick" and "invite" the target user ID will be in the JSON request body.
```

src/github.com/matrix-org/dendrite/clientapi/writers/membership.go
```diff
@@ -0,0 +1,134 @@
...
+	if reqErr != nil {
+		return *reqErr
+	}
+
+	localpart, _, err := gomatrixserverlib.SplitID('@', stateKey)
+	if err != nil {
+		return httputil.LogThenError(req, err)
+	}
+
+	profile, err := accountDB.GetProfileByLocalpart(localpart)
```
NegativeMjark: What happens if you try to ban a user who isn't on the same server as you? Presumably we shouldn't be looking up profile information for users on remote servers in our own database. (Ideally you'd hit the profile information over federation, but it might be best to leave it for this PR.)

PR #172, opened 2017-08-02
Simplify preparing the roomserver SQL

NegativeMjark:
(This is a bit nicer to write and it has simpler control flow from the perspective of gocyclo)


Kegsay:
LGTM - though it's sad that it makes it a bit harder for a human to grok. In this case it isn't too difficult to muddle through though.

PR #173, opened 2017-08-03
Move setting up the api mux to outside the routing.Setup functions.

NegativeMjark:
This makes it possible to setup all the component APIs on a single http
listener which is necessary if we want to combine all the components
into a single monolith.


Kegsay:
LGTM - thanks for this, I was planning on doing exactly this at some point.

PR #174, opened 2017-08-03
Membership viewing API

babolivier:
This PR implements `GET /_matrix/client/r0/rooms/{roomID}/members`

Still left to do:
* If the user was previously in the room but has left since, send the memberships list as it was before they left instead of the current one
* If the room isn't local, retrieve the list of memberships through federation

But I'm not sure either of them are currently doable.

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/membership_table.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/membership_table.go
@@ -30,6 +30,11 @@ CREATE TABLE IF NOT EXISTS memberships (
     room_id TEXT NOT NULL,
     -- The ID of the join membership event
     event_id TEXT NOT NULL,
+	-- Is the user still in the room? We use a boolean instead of just deleting
+	-- the row to know if the user has been in the room at some point
+	-- If set to false, event_id would be the event ID of the leave/kick/ban
+	-- membership event
+	still_in_room BOOLEAN NOT NULL,
 
     -- A user can only be member of a room once
     PRIMARY KEY (localpart, room_id)
@@ -40,24 +45,26 @@ CREATE UNIQUE INDEX IF NOT EXISTS membership_event_id ON memberships(event_id);
 `
 
 const insertMembershipSQL = `
-	INSERT INTO memberships(localpart, room_id, event_id) VALUES ($1, $2, $3)
-	ON CONFLICT (localpart, room_id) DO UPDATE SET event_id = EXCLUDED.event_id
+	INSERT INTO memberships(localpart, room_id, event_id, still_in_room)
+	VALUES ($1, $2, $3, $4)
+	ON CONFLICT (localpart, room_id) DO
+	UPDATE SET event_id = EXCLUDED.event_id, still_in_room = EXCLUDED.still_in_room
 `
 
 const selectMembershipSQL = "" +
 	"SELECT * from memberships WHERE localpart = $1 AND room_id = $2"
 
 const selectMembershipsByLocalpartSQL = "" +
-	"SELECT room_id, event_id FROM memberships WHERE localpart = $1"
+	"SELECT room_id, event_id FROM memberships WHERE localpart = $1 AND still_in_room = true"
 
 const selectMembershipsByRoomIDSQL = "" +
-	"SELECT localpart, event_id FROM memberships WHERE room_id = $1"
+	"SELECT localpart, event_id FROM memberships WHERE room_id = $1 AND still_in_room = true"
 
 const deleteMembershipsByEventIDsSQL = "" +
 	"DELETE FROM memberships WHERE event_id = ANY($1)"
 
 const updateMembershipByEventIDSQL = "" +
-	"UPDATE memberships SET event_id = $2 WHERE event_id = $1"
+	"UPDATE memberships SET event_id = $2, still_in_room = $3 WHERE event_id = $1"
 
 type membershipStatements struct {
 	deleteMembershipsByEventIDsStmt  *sql.Stmt
@@ -91,8 +98,8 @@ func (s *membershipStatements) prepare(db *sql.DB) (err error) {
 	return
 }
 
-func (s *membershipStatements) insertMembership(localpart string, roomID string, eventID string, txn *sql.Tx) (err error) {
-	_, err = txn.Stmt(s.insertMembershipStmt).Exec(localpart, roomID, eventID)
+func (s *membershipStatements) insertMembership(localpart string, roomID string, eventID string, stillInRoom bool, txn *sql.Tx) (err error) {
+	_, err = txn.Stmt(s.insertMembershipStmt).Exec(localpart, roomID, eventID, stillInRoom)
 	return
 }
 
--- a/src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/membership_table.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/membership_table.go
@@ -52,7 +52,7 @@ const insertMembershipSQL = `
 `
 
 const selectMembershipSQL = "" +
-	"SELECT * from memberships WHERE localpart = $1 AND room_id = $2"
+	"SELECT event_id, still_in_room from memberships WHERE localpart = $1 AND room_id = $2"
 
 const selectMembershipsByLocalpartSQL = "" +
 	"SELECT room_id, event_id FROM memberships WHERE localpart = $1 AND still_in_room = true"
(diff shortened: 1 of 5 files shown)


src/github.com/matrix-org/dendrite/clientapi/readers/memberships.go
```diff
@@ -0,0 +1,87 @@
...
+	accountDB *accounts.Database, cfg config.Dendrite,
+	queryAPI api.RoomserverQueryAPI,
+) util.JSONResponse {
+	localpart, server, err := gomatrixserverlib.SplitID('@', device.UserID)
+	if err != nil {
+		return httputil.LogThenError(req, err)
+	}
+
+	events := []gomatrixserverlib.ClientEvent{}
+	if server == cfg.Matrix.ServerName {
```
NegativeMjark: This check seems weird to me. If this API is authenticated then the UserID will always be a local user on the server.

babolivier: Indeed. Meant to do the check on the room ID, but messed it up somehow.

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/membership_table.go
```diff
@@ -30,6 +30,11 @@ CREATE TABLE IF NOT EXISTS account_memberships (
     room_id TEXT NOT NULL,
     -- The ID of the join membership event
     event_id TEXT NOT NULL,
+	-- Is the user still in the room? We use a boolean instead of just deleting
+	-- the row to know if the user has been in the room at some point
+	-- If set to false, event_id would be the event ID of the leave/kick/ban
+	-- membership event
+	still_in_room BOOLEAN NOT NULL,
```
NegativeMjark: Is this just whether the user is "joined" or not? It might be easier to call it "joined"?

babolivier: Joined can be misleading imho. I'd interpret it as "the user has joined at some point" without regard to whether he since left or not.

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/membership_table.go
```diff
@@ -40,27 +45,34 @@ CREATE UNIQUE INDEX IF NOT EXISTS account_membership_event_id ON account_members
...
+	UPDATE SET event_id = EXCLUDED.event_id, still_in_room = EXCLUDED.still_in_room
 `
 
 const selectMembershipSQL = "" +
-	"SELECT * from account_memberships WHERE localpart = $1 AND room_id = $2"
+	"SELECT event_id, still_in_room from account_memberships WHERE localpart = $1 AND room_id = $2"
 
 const selectMembershipsByLocalpartSQL = "" +
-	"SELECT room_id, event_id FROM account_memberships WHERE localpart = $1"
+	"SELECT room_id, event_id FROM account_memberships WHERE localpart = $1 AND still_in_room = true"
```
NegativeMjark: I think you can drop the `= true` from the SQL.

babolivier: Don't we want to only return users that are still in the room?

src/github.com/matrix-org/dendrite/clientapi/readers/memberships.go
```diff
@@ -0,0 +1,91 @@
...
+		if err := queryAPI.QueryEventsByID(&queryReq, &queryRes); err != nil {
+			return httputil.LogThenError(req, err)
+		}
+
+		for _, event := range queryRes.Events {
+			ev := gomatrixserverlib.ToClientEvent(event, gomatrixserverlib.FormatAll)
+			events = append(events, ev)
+		}
+	} else {
+		// TODO: Get memberships from federation
```
NegativeMjark: I don't think you ever need to do this? If the user is allowed to view the room then they are joined to the room, which implies that the server is in the room, which implies that the server has a copy of the room state.

So it should always be possible to return the list of membership events by just by looking in the local database.

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/membership_table.go
```diff
@@ -40,27 +45,34 @@ CREATE UNIQUE INDEX IF NOT EXISTS account_membership_event_id ON account_members
...
+
+const selectMembershipsByRoomIDSQL = "" +
+	"SELECT localpart, event_id FROM account_memberships WHERE room_id = $1 AND still_in_room = true"
```
NegativeMjark: I'm not sure this is the best table to do this sort of query over. It doesn't have an index on the `room_id` so it won't be particularly efficient.

It might be better to use the tables in the roomsever instead?

NegativeMjark: although getting the current state from the room server tables is a bit of an adventure at the moment
I think you'd need to add a new query API in order to do so.

babolivier: Yep, the roomserver's memberships table wasn't there when I opened this PR. It would indeed be the best place to do this (with a call to the client API server's memberships table to get the join event ID and another trip to the roomserver API to get the actual events). I'll work on that.

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/membership_table.go
NegativeMjark: Sorry I mean that you can  write "AND still_in_room" instead of "AND still_in_room = true" since still_in_room is a boolean.

babolivier: Oh, right.
This isn't needed anymore, though, since we do no longer retrieve the info from this table. I'll have a cleaning pass on the client API storage this afternoon.

src/github.com/matrix-org/dendrite/roomserver/input/membership.go
```diff
@@ -152,16 +151,20 @@ func updateToInviteMembership(
 func updateToJoinMembership(
 	mu types.MembershipUpdater, add *gomatrixserverlib.Event, updates []api.OutputEvent,
 ) ([]api.OutputEvent, error) {
-	// If the user is already marked as being joined then we can return immediately.
-	// TODO: Is this code reachable given the "old != new" guard in updateMembership?
+	// If the user is already marked as being joined, we call SetToJoin to update
+	// the event ID then we can return immediately. Retired is ignored as
 	if mu.IsJoin() {
```
NegativeMjark: The comment seems to stop mid sentence after the "Retired is ignored as".

NegativeMjark:
The roomserver query API is looking pretty good :+1: 

Handling the case were the user has left the room shouldn't be too bad if there is an entry for the state at the leave event in the events table. We'd just need to load the appropriate state possibly using LoadCombinedStateAfterEvents https://github.com/matrix-org/dendrite/blob/master/src/github.com/matrix-org/dendrite/roomserver/state/state.go#L101 and then filtering down to m.room.member events.

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/membership_table.go
```diff
@@ -41,14 +41,11 @@ CREATE UNIQUE INDEX IF NOT EXISTS account_membership_event_id ON account_members
 
 const insertMembershipSQL = `
 	INSERT INTO account_memberships(localpart, room_id, event_id) VALUES ($1, $2, $3)
-	ON CONFLICT (localpart, room_id) DO UPDATE SET event_id = EXCLUDED.event_id
+	ON CONFLICT (localpart, room_id) DO	UPDATE SET event_id = EXCLUDED.event_id
```
NegativeMjark: A spurious tab character seems to have appeared between DO and UPDATE.

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/membership_table.go
```diff
@@ -41,14 +41,11 @@ CREATE UNIQUE INDEX IF NOT EXISTS account_membership_event_id ON account_members
...
 `
 
-const selectMembershipSQL = "" +
-	"SELECT * from account_memberships WHERE localpart = $1 AND room_id = $2"
-
 const selectMembershipsByLocalpartSQL = "" +
-	"SELECT room_id, event_id FROM account_memberships WHERE localpart = $1"
+	"SELECT * FROM account_memberships WHERE localpart = $1"
```
NegativeMjark: I'd prefer that we were explicit about the fields we are returning. Otherwise it will break in odd ways if we add a new column.

babolivier: Right. I was mechanically reverting the changes from the diff without thinking much about that.

PR #175, opened 2017-08-03
Add a monolithic server that combines all the components into one

NegativeMjark:


src/github.com/matrix-org/dendrite/cmd/dendrite-monolith-server/main.go
```diff
@@ -0,0 +1,278 @@
...
+
+	if *configPath == "" {
+		log.Fatal("--config must be supplied")
+	}
+	cfg, err := config.Load(*configPath)
+	if err != nil {
+		log.Fatalf("Invalid config file: %s", err)
+	}
+
+	m := monolith{cfg: cfg, api: mux.NewRouter()}
```
Kegsay: Even though this is a private type, it's still important to tell the reader which fields need to be supplied to `monolith` and which it will set when you call `setupXXX`, without having to read 200 lines of monolith functions. I'd be happier with one of:
 - Add a small `newMonolith(cfg, api)` unexported function.
 - Re-jig the order and grouping of fields in the type definition and add a one-liner on the required fields.

src/github.com/matrix-org/dendrite/cmd/dendrite-monolith-server/main.go
```diff
@@ -0,0 +1,278 @@
...
+	m.setupDatabases()
+	m.setupRoomServer()
+	m.setupProducers()
```
Kegsay: I like this overall, but I don't like the fact that there are subtle ordering issues here.

We're effectively constructing a dependency graph here. We need to setup databases first before setting up other things, etc. Just a one-liner at the start of this entire setup block to say "ordering matters" would be nice.

Kegsay:
Otherwise LGTM! Does this mean we an `rm -rf client-api-proxy`? 👼

PR #176, opened 2017-08-04
Fix bad initialisation of sync API notifier in monolith

babolivier:


Kegsay:
LGTM, I should've caught that. Thank you.

PR #179, opened 2017-08-05
Add .editorconfig

jplatte:
By default GitHub renders tabs as 8 characters wide, which is really annoying. Adding this `.editorconfig` changes that to 4 chars wide.

`.editorconfig` is not only recognized by GitHub, but also basically every editor (hence the name, and all the settings that aren't the indent size) although many require an extra plugin / package. Details can be found on http://editorconfig.org/.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,18 @@
+root = true
+
+[*]
+charset = utf-8
+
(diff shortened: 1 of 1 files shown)


ara4n:
that is super-useful - thanks!

PR #181, opened 2017-08-07
[WIP] Published room APIs

babolivier:
I'm creating this PR early so I can more easily keep track of what's left for me to do (I do it because it's a pretty big one). When I consider it done, I'll remove "[WIP]" from the title and I'll assign someone for a review.

TODO:

* [x] Add database schema and functions to keep track of published rooms on the room server
* [x] Add database functions to retrieve room aliases in batch
* [x] Create a published rooms API on the room server which handles updates and retrieval of published rooms
* [x] Implement `PUT /_matrix/client/r0/directory/list/room/{roomID}`
* [x] Implement `GET /_matrix/client/r0/directory/list/room/{roomID}`
* [x] Basic implementation of `POST /_matrix/client/r0/publicRooms`
* [x] Basic implementation of `GET /_matrix/client/r0/publicRooms`
* [x] Compute the number of members in each room (using the new `roomserver_membership` table)
* [ ] Handle pagination (`prev_batch`, `next_batch`)
* [ ] Dynamically edit `world_readable` and `guest_can_join` in response to `[...] /_matrix/client/r0/publicRooms`

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/routing/routing.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/routing/routing.go
@@ -41,7 +41,7 @@ const pathPrefixUnstable = "/_matrix/client/unstable"
 func Setup(
 	apiMux *mux.Router, httpClient *http.Client, cfg config.Dendrite,
 	producer *producers.RoomserverProducer, queryAPI api.RoomserverQueryAPI,
-	aliasAPI api.RoomserverAliasAPI,
+	aliasAPI api.RoomserverAliasAPI, publicRoomAPI api.RoomserverPublicRoomAPI,
 	accountDB *accounts.Database,
 	deviceDB *devices.Database,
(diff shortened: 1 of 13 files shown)


NegativeMjark:
I'm not sure that we should be handling the public room calculations inside the roomserver.
There are two problems with doing so:
1) We want to shard the roomservers by room ID at some point. This means that queries which touch more than one room would have to iterate over all the shards.
2) We want the public room API to be fast. This means that ideally the response is pre-calculated and served from a single database table, in particular it shouldn't need to calculate room membership counts each time it is queried.

I was envisioning that the public room list would be handled by creating a new, separate component from the room-server that consumed the event stream generated by the room server to generate a pre-calculated public room response.

Then we'd either let it handle the relevant client API paths directly, or proxy the requests through the clientAPI component.

babolivier:
Closing this PR in order to implement the process in a separate component (as suggested) and to commit this in a new one.

PR #182, opened 2017-08-08
Add a route matching the trailing slash on the state event sending route

babolivier:
I know it looks like a nasty solution (code duplication and stuff), but after many tries and a bit of research it looks like it's [the official way of solving this problem](https://github.com/gorilla/mux/issues/30#issuecomment-21254225).

FTR, the issue here is that `PUT` requests on `/_matrix/client/r0/rooms/{roomID}/state/{eventType}/` aren't supported by Dendrite because of the trailing slash, despite being implicitly allowed in the Matrix spec. This causes some functionalities in Riot to not work with Dendrite (such as changing a room's avatar) because it's using the route with the trailing slash.

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/routing/routing.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/routing/routing.go
@@ -100,6 +100,13 @@ func Setup(
 			return writers.SendEvent(req, device, vars["roomID"], vars["eventType"], vars["txnID"], &emptyString, cfg, queryAPI, producer)
 		}),
 	)
+	r0mux.Handle("/rooms/{roomID}/state/{eventType}/",
+		common.MakeAuthAPI("send_message", deviceDB, func(req *http.Request, device *authtypes.Device) util.JSONResponse {
+			vars := mux.Vars(req)
+			emptyString := ""
+			return writers.SendEvent(req, device, vars["roomID"], vars["eventType"], vars["txnID"], &emptyString, cfg, queryAPI, producer)
+		}),
+	)
 	r0mux.Handle("/rooms/{roomID}/state/{eventType}/{stateKey}",
 		common.MakeAuthAPI("send_message", deviceDB, func(req *http.Request, device *authtypes.Device) util.JSONResponse {
 			vars := mux.Vars(req)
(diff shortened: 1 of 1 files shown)


src/github.com/matrix-org/dendrite/clientapi/routing/routing.go
```diff
@@ -100,6 +100,13 @@ func Setup(
 			return writers.SendEvent(req, device, vars["roomID"], vars["eventType"], vars["txnID"], &emptyString, cfg, queryAPI, producer)
 		}),
 	)
+	r0mux.Handle("/rooms/{roomID}/state/{eventType}/",
```
m90: Picking up from the discussion in the gorilla repo this could probably also be `"/rooms/{roomID}/state/{eventType:[a-z]+(?:\\/)?}"` (making assumptions on what `eventType` looks like, but I guess you should be able to adjust that).

babolivier: Indeed I could use a regular expression instead of just a variable name. I'm still exchanging a bit in the gorilla repo to see if there's a better way, but if not I'll go for that.

src/github.com/matrix-org/dendrite/clientapi/routing/routing.go
```diff
@@ -93,11 +94,16 @@ func Setup(
 			return writers.SendEvent(req, device, vars["roomID"], vars["eventType"], vars["txnID"], nil, cfg, queryAPI, producer)
 		}),
 	)
-	r0mux.Handle("/rooms/{roomID}/state/{eventType}",
+	r0mux.Handle("/rooms/{roomID}/state/{eventType:[a-z._-]+(?:\\/)?}",
```
NegativeMjark: I don't think you need the group. I'm pretty sure you can write this as:

```go
"{eventType:[a-z._-]+/?}"
```

Hmm, this regexp is a tad more restrictive that what synapse does. I'm not sure it that is a problem though. Think the synapse regexp is:

```go
"{event_type:[^/]+/?}"
```

based on https://github.com/matrix-org/synapse/blob/master/synapse/rest/client/v1/room.py#L190

PR #183, opened 2017-08-15
Optionally use naffka in the monolithic server

NegativeMjark:
This PR chases down all the places where we create a kafka consumer or producer and replaces them with dependency injected versions, this means that we can optionally replace the kafka consumer or producers objects with a naffka versions in the monolithic server.

### DIFF
--- a/src/github.com/matrix-org/dendrite/cmd/dendrite-client-api-server/main.go
+++ b/src/github.com/matrix-org/dendrite/cmd/dendrite-client-api-server/main.go
@@ -33,6 +33,7 @@ import (
 	"github.com/matrix-org/gomatrixserverlib"
 
 	log "github.com/Sirupsen/logrus"
+	sarama "gopkg.in/Shopify/sarama.v1"
 )
 
 var (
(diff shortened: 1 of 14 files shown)


src/github.com/matrix-org/dendrite/cmd/dendrite-client-api-server/main.go
```diff
@@ -50,24 +51,28 @@ func main() {
 		log.Fatalf("Invalid config file: %s", err)
 	}
 
-	log.Info("config: ", cfg)
```
NegativeMjark: This isn't strictly related to the PR, but it was bugging me since it logs the ed25519 private key of the server,

src/github.com/matrix-org/dendrite/common/config/config.go
```diff
@@ -297,21 +326,32 @@ func (config *Dendrite) check() error {
...
+			problems = append(problems, fmt.Sprintf("naffka can only be used in a monolithic server"))
+		}
+	} else {
+		// If we aren't using naffka then we need to have at least one kafka
+		// server to talk to.
+		checkNotZero("kafka.addresses", int64(len(config.Kafka.Addresses)))
+	}
 	checkNotEmpty("kafka.topics.output_room_event", string(config.Kafka.Topics.OutputRoomEvent))
 	checkNotEmpty("kafka.topics.output_client_data", string(config.Kafka.Topics.OutputClientData))
+	checkNotEmpty("kafka.topics.user_updates", string(config.Kafka.Topics.UserUpdates))
```
NegativeMjark: Not strictly related to this PR, but I thought I might as well add the check since I was in the neighbourhood.

src/github.com/matrix-org/dendrite/common/config/config_test.go
```diff
@@ -42,9 +43,9 @@ media:
 kafka:
   addresses: ["localhost:9092"]
   topics:
-    input_room_event: input.room
     output_room_event: output.room
     output_client_data: output.client
+    user_updates: output.user
```
NegativeMjark: Not strictly related to this PR, but the "input_room_event" is no longer a thing, but "user_updates" is a thing now.

PR #185, opened 2017-08-16
Implement public rooms APIs

babolivier:
This PR adds a new component to Dendrite which implements:

* `PUT /_matrix/client/r0/directory/list/room/{roomID}`
* `GET /_matrix/client/r0/directory/list/room/{roomID}`
* `POST /_matrix/client/r0/publicRooms`
* `GET /_matrix/client/r0/publicRooms`

Like all other components in Dendrite, it can be run separated from the others, or as part of the monolith.

I also moved the `eventcontent.go` file from the client API server to `common` as the public rooms API server is also using it.

Sorry for the length or this PR. I should have thought about how I could have split it in smaller ones beforehand.

### DIFF
--- a/src/github.com/matrix-org/dendrite/publicroomsapi/consumers/roomserver.go
+++ b/src/github.com/matrix-org/dendrite/publicroomsapi/consumers/roomserver.go
@@ -37,8 +37,8 @@ func NewOutputRoomEvent(
 	cfg *config.Dendrite,
 	kafkaConsumer sarama.Consumer,
 	store *storage.PublicRoomsServerDatabase,
+	queryAPI api.RoomserverQueryAPI,
 ) *OutputRoomEvent {
-
 	consumer := common.ContinualConsumer{
 		Topic:          string(cfg.Kafka.Topics.OutputRoomEvent),
 		Consumer:       kafkaConsumer,
@@ -47,6 +47,7 @@ func NewOutputRoomEvent(
 	s := &OutputRoomEvent{
 		roomServerConsumer: &consumer,
 		db:                 store,
+		query:              queryAPI,
 	}
 	consumer.ProcessMessage = s.onMessage
 
@@ -82,5 +83,12 @@ func (s *OutputRoomEvent) onMessage(msg *sarama.ConsumerMessage) error {
 		"type":     ev.Type(),
 	}).Info("received event from roomserver")
 
-	return s.db.UpdateRoomFromEvent(ev)
+	queryReq := api.QueryEventsByIDRequest{output.NewRoomEvent.AddsStateEventIDs}
+	var queryRes api.QueryEventsByIDResponse
+	if err := s.query.QueryEventsByID(&queryReq, &queryRes); err != nil {
+		log.Warn(err)
+		return err
+	}
+
+	return s.db.UpdateRoomFromEvents(queryRes.Events, output.NewRoomEvent.RemovesStateEventIDs)
 }
--- a/src/github.com/matrix-org/dendrite/publicroomsapi/consumers/roomserver.go
+++ b/src/github.com/matrix-org/dendrite/publicroomsapi/consumers/roomserver.go
@@ -33,12 +33,11 @@ type OutputRoomEvent struct {
 }
 
 // NewOutputRoomEvent creates a new OutputRoomEvent consumer. Call Start() to begin consuming from room servers.
-func NewOutputRoomEvent(cfg *config.Dendrite, store *storage.PublicRoomsServerDatabase) (*OutputRoomEvent, error) {
-	kafkaConsumer, err := sarama.NewConsumer(cfg.Kafka.Addresses, nil)
-	if err != nil {
-		return nil, err
-	}
-	roomServerURL := cfg.RoomServerURL()
+func NewOutputRoomEvent(
+	cfg *config.Dendrite,
+	kafkaConsumer sarama.Consumer,
+	store *storage.PublicRoomsServerDatabase,
+) *OutputRoomEvent {
 
 	consumer := common.ContinualConsumer{
 		Topic:          string(cfg.Kafka.Topics.OutputRoomEvent),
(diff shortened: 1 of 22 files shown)


babolivier:
FTR 10c2bb3 (I'm talking about `room_aliases_table.go`, the other file was added by mistake) is unrelated to the PR itself, but I stumbled upon this bug preventing the creation of more than one alias in a room and fixed it while I was there.

src/github.com/matrix-org/dendrite/publicroomsapi/consumers/roomserver.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	ev := output.NewRoomEvent.Event
+	log.WithFields(log.Fields{
+		"event_id": ev.EventID(),
+		"room_id":  ev.RoomID(),
+		"type":     ev.Type(),
+	}).Info("received event from roomserver")
+
+	return s.db.UpdateRoomFromEvent(ev)
```
NegativeMjark: Hmm, technically this should be using the "AddsStateEventIDs" and "RemovesStateEventIDs" fields, since we are interested in the current state, rather than the events themselves.

I'm beginning to wonder if the output stream from the roomserver has the format for doing this sort of thing. It's currently possible to write a consumer that does the right thing, but it seems to easy to write one that does the wrong thing as well.

Maybe we could have a separate output topic for changes to current state for a room? That way the consumers don't have to reconstruct the changes themselves.

Alternatively we could make a CurrentStateConsumer in common that handled getting the changes to current state from the AddsStateEventIDs and RemovesStateEventIDs. That might make it easier to make this sort of consumer do the right thing?

Also are we doing the right thing be giving lists of event IDs for the changes in state rather than sending the full events? I wanted to save space on the wire but it makes the consumers a lot more complicated that they would otherwise be so it might be premature optimisation.

src/github.com/matrix-org/dendrite/publicroomsapi/storage/storage.go
```diff
@@ -0,0 +1,211 @@
...
+// If the event's membership is "join", increments the value, if not, decrements it.
+// Returns an error if the update failed.
+func (d *PublicRoomsServerDatabase) updateNumJoinedUsers(membershipEvent gomatrixserverlib.Event, roomID string) error {
+	membership, err := membershipEvent.Membership()
+	if err != nil {
+		return err
+	}
+
+	if membership == "join" {
+		return d.statements.incrementJoinedMembersInRoom(roomID)
```
NegativeMjark: I think this will break if a user changes their display name since they will send a new join event to the rooms they are already joined to.

babolivier: Indeed, didn't think of this use case. I'll look into it.

src/github.com/matrix-org/dendrite/publicroomsapi/storage/storage.go
```diff
@@ -0,0 +1,211 @@
...
+// the database representation of the room identified by a given ID, using a given
+// string data field from content of the Matrix event triggering the update.
+// It determines the attribute's type by checking the length of the given strForTrue
+// parameter: If it's a non-empty string, it means that the attribute is a boolean
+// which will be set to true if the given data field matches it, and to false in
+// any other case. If the string is empty, it means that the attribute is a string
+// and will be updated with the field's value.
+// Returns an error if decoding the Matrix event's content or updating the attribute
+// failed.
+func (d *PublicRoomsServerDatabase) updateStringOrBooleanAttribute(
```
NegativeMjark: This might be simpler to do as two functions, a updateStringAttribute, and an updateBooleanAttribute.

src/github.com/matrix-org/dendrite/publicroomsapi/consumers/roomserver.go
babolivier: > Hmm, technically this should be using the "AddsStateEventIDs" and "RemovesStateEventIDs" fields, since we are interested in the current state, rather than the events themselves.

Right, I forgot about that.

> Maybe we could have a separate output topic for changes to current state for a room? That way the consumers don't have to reconstruct the changes themselves.
>
> Alternatively we could make a CurrentStateConsumer in common that handled getting the changes to current state from the AddsStateEventIDs and RemovesStateEventIDs. That might make it easier to make this sort of consumer do the right thing?
>
> Also are we doing the right thing be giving lists of event IDs for the changes in state rather than sending the full events? I wanted to save space on the wire but it makes the consumers a lot more complicated that they would otherwise be so it might be premature optimisation.

Since, IIRC, a roomserver consumer either ignores the AddsStateEventIDs or ends up querying the roomserver to fetch the events (I don't recall seeing a consumer storing only the events IDs), sending the whole events seems better to me. Not sure about the RemovesStateEventIDs though (since the events to remove are usually already stored and retrievable with their ID).

src/github.com/matrix-org/dendrite/publicroomsapi/storage/storage.go
babolivier: I'm not entirely sure. I thought that since the process was very similar (and simple), it would just be better to have one single function to avoid duplication of code.

But if it might be difficult to understand for someone else, I'll just split it then.

src/github.com/matrix-org/dendrite/publicroomsapi/storage/storage.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	if membership == "join" {
+		for _, event := range eventsToRemove {
+			if event.Type() == "m.room.member" {
+				// This is an update for an user who already joined the room, most
+				// likely triggered by a profile update. We don't increment the
+				// number of joined member in this case because we already counted
+				// this user.
+				return nil
```
NegativeMjark: Does this work if there was more that one join event in the state added?
Would it be better to increment the count for each join in the state added and decrement the count for each join in the state removed?

PR #186, opened 2017-08-17
Remove StopProcessingAfter from the roomserver consumer as it is unused

NegativeMjark:


NegativeMjark:
It used to be used by the roomserver binary for load testing. E.g. tell it to stop after 1000 messages and time how long it took to stop. However now that the input API is synchronous it isn't needed to do load testing. We can now load test by seeing how long the requests to the input API take to complete.

PR #187, opened 2017-08-18
Add input API for adding invites to the roomserver.

NegativeMjark:
This API handles invites received over federation that occur outside of a room.

### DIFF
--- a/src/github.com/matrix-org/dendrite/common/sql.go
+++ b/src/github.com/matrix-org/dendrite/common/sql.go
@@ -18,6 +18,24 @@ import (
 	"database/sql"
 )
 
+// A Transaction is something that can be committed or rolledback.
+type Transaction interface {
+	// Commit the transaction
+	Commit() error
+	// Rollback the transaction.
+	Rollback() error
+}
+
+// EndTransaction ends a transaction.
+// If the transaction succeeded then it is committed, otherwise it is rolledback.
+func EndTransaction(txn Transaction, succeeded *bool) {
+	if *succeeded {
+		txn.Commit()
+	} else {
+		txn.Rollback()
+	}
+}
+
 // WithTransaction runs a block of code passing in an SQL transaction
 // If the code returns an error or panics then the transactions is rolledback
 // Otherwise the transaction is committed.
@@ -26,16 +44,14 @@ func WithTransaction(db *sql.DB, fn func(txn *sql.Tx) error) (err error) {
 	if err != nil {
 		return
 	}
-	defer func() {
-		if r := recover(); r != nil {
-			txn.Rollback()
-			panic(r)
-		} else if err != nil {
-			txn.Rollback()
-		} else {
-			err = txn.Commit()
-		}
-	}()
+	succeeded := false
+	defer EndTransaction(txn, &succeeded)
+
 	err = fn(txn)
+	if err != nil {
+		return
+	}
+
+	succeeded = true
 	return
 }
--- a/src/github.com/matrix-org/dendrite/roomserver/input/events.go
+++ b/src/github.com/matrix-org/dendrite/roomserver/input/events.go
@@ -15,6 +15,8 @@
 package input
 
 import (
+	"fmt"
+
 	"github.com/matrix-org/dendrite/roomserver/api"
 	"github.com/matrix-org/dendrite/roomserver/state"
 	"github.com/matrix-org/dendrite/roomserver/types"
@@ -39,6 +41,8 @@ type RoomEventDatabase interface {
 	GetLatestEventsForUpdate(roomNID types.RoomNID) (updater types.RoomRecentEventsUpdater, err error)
 	// Lookup the string event IDs for a list of numeric event IDs
 	EventIDs(eventNIDs []types.EventNID) (map[types.EventNID]string, error)
+	// Build a membership updater for the target user in a room.
+	MembershipUpdater(roomID, targerUserID string) (types.MembershipUpdater, error)
 }
 
 // OutputRoomEventWriter has the APIs needed to write an event to the output logs.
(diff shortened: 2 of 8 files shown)


src/github.com/matrix-org/dendrite/roomserver/input/events.go
```diff
@@ -103,13 +107,84 @@ func processRoomEvent(db RoomEventDatabase, ow OutputRoomEventWriter, input api.
...
 	}
 
-	// TODO:
-	//  * Caculate the new current state for the room if the forward extremities have changed.
-	//  * Work out the delta between the new current state and the previous current state.
-	//  * Work out the visibility of the event.
-	//  * Write a message to the output logs containing:
-	//      - The event itself
-	//      - The visiblity of the event, i.e. who is allowed to see the event.
-	//      - The changes to the current state of the room.
```
NegativeMjark: This isn't strictly related to this PR, but I think all of that TODO is out of date, so it might as well be deleted.

src/github.com/matrix-org/dendrite/roomserver/input/events.go
```diff
@@ -103,13 +107,84 @@ func processRoomEvent(db RoomEventDatabase, ow OutputRoomEventWriter, input api.
...
+		}
+
+		return ow.WriteOutputEvents(roomID, outputUpdates)
+	})
+}
+
+// withTransaction runs a function inside a transaction.
+// Commits the transaction if the function succeeds.
+// Rollsback the transaction if the function fails or panics.
+func withTransaction(t types.Transaction, f func() error) (err error) {
```
NegativeMjark: It occurs to me that this is nearly exactly the same code as common.WithTransaction, except that it takes an interface. I think `sql.Tx` implements the `types.Transaction` interface so in theory we could make a single implementation of WithTransaction that worked everywhere. Do you think that would make sense?

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
```diff
@@ -104,15 +106,15 @@ func (d *Database) StoreEvent(event gomatrixserverlib.Event, authEventNIDs []typ
 	}, nil
 }
 
-func (d *Database) assignRoomNID(roomID string) (types.RoomNID, error) {
+func (d *Database) assignRoomNID(txn *sql.Tx, roomID string) (types.RoomNID, error) {
```
richvdh: I wonder if we should make it clear that `txn` can be nil. Or is that a general convention?

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
```diff
@@ -380,6 +382,27 @@ func (d *Database) StateEntriesForTuples(
...
 }
 
+// MembershipUpdater implements input.RoomEventDatabase
+func (d *Database) MembershipUpdater(roomID, targetUserID string) (updater types.MembershipUpdater, err error) {
+	err = common.WithTransaction(d.db, func(txn *sql.Tx) error {
+		var (
+			roomNID       types.RoomNID
+			targetUserNID types.EventStateKeyNID
+		)
+		roomNID, err = d.assignRoomNID(txn, roomID)
```
richvdh: ok, I'll bite: why does this need to happen in a transaction, whereas other calls to assignRoomNID do not?

src/github.com/matrix-org/dendrite/roomserver/input/events.go
richvdh: well, I'm certainly not in favour of having two copies of the code, nor am I in favour of having two functions which do similar but different things whose names differ only by case. There's nothing specific to the roomserver here; it belongs in `common` just as much as the other `WithTransaction` does.

The main difference seems to be that `common.WithTransaction` starts the transaction, whereas this one assumes you already have a Transaction. I think this might be more appropriately named `common.CompleteTransaction` or something; and you could implement `common.WithTransaction` in terms of it?

src/github.com/matrix-org/dendrite/roomserver/input/latest_events.go
```diff
@@ -52,25 +52,12 @@ func updateLatestEvents(
...
-			err = updater.Commit()
-		} else {
-			// Ignore any error we get rolling back since we don't want to
-			// clobber the current error
-			// TODO: log the error here.
-			updater.Rollback()
-		}
-	}()
 
 	u := latestEventsUpdater{
```
richvdh: I think the answer's no, but: is there any chance of this panicking (and thus wanting to be wrapped in the withTransaction call)?

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
```diff
@@ -380,6 +382,27 @@ func (d *Database) StateEntriesForTuples(
 	return d.statements.bulkSelectFilteredStateBlockEntries(stateBlockNIDs, stateKeyTuples)
 }
 
+// MembershipUpdater implements input.RoomEventDatabase
+func (d *Database) MembershipUpdater(roomID, targetUserID string) (updater types.MembershipUpdater, err error) {
+	err = common.WithTransaction(d.db, func(txn *sql.Tx) error {
```
richvdh: erm; will this not commit the transaction being returned in `updater.txn`, this making it a bit pointless?

NegativeMjark: Umm, yes, good point :)

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
NegativeMjark: Umm, good question, I think because we already have a transaction lying around at this point and it's a bit of a waste to snaffle a free database connection to run the stmt outside the transaction.

src/github.com/matrix-org/dendrite/roomserver/input/latest_events.go
NegativeMjark: I think the answer is no as well. I guess I could move it inside the withTransaction call anyway...

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
NegativeMjark: General convention, I guess we could go back and document the other places we do this?

src/github.com/matrix-org/dendrite/roomserver/input/events.go
NegativeMjark: Done 6d69180

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
NegativeMjark: Fixed 76c7f17

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
richvdh: I guess I don't mind, provided *all* the functions that take a `*sql.Tx` handle it being `nil`. The thing that would really annoy me is if I have to go and chase through 15 layers of function calls to check if I need to fix anything before passing `nil`.

src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
richvdh: ok. Just feels like a shame that you had to go and change `assignRoomNID` and the things it calls just so that you could pass in a transaction that you happened to have...

src/github.com/matrix-org/dendrite/common/sql.go
```diff
@@ -18,6 +18,24 @@ import (
 	"database/sql"
 )
 
+// A Transaction is something that can be committed or rolledback.
+type Transaction interface {
```
NegativeMjark: Done 5d785f0

PR #191, opened 2017-08-21
Make txn *sql.Tx arguments optional everywhere using a utility function

NegativeMjark:


src/github.com/matrix-org/dendrite/common/sql.go
```diff
@@ -55,3 +55,13 @@ func WithTransaction(db *sql.DB, fn func(txn *sql.Tx) error) (err error) {
 	succeeded = true
 	return
 }
+
+// TxStmt wraps an SQL stmt inside an optional transaction.
+// If the transaction is nil then it returns the original statement.
```
richvdh: mention that this means it will be executed without a transaction?

PR #193, opened 2017-08-21
notes on running monolith server

richvdh:


richvdh:
@NegativeMjark any idea what the test failure means?

PR #194, opened 2017-08-21
Membership viewing API when user left the room

babolivier:
Following-up #174, if the user left the room, retrieves its state and filters it to keep only the `m.room.member` events of `join` membership, then send it to the user.

### DIFF
--- a/src/github.com/matrix-org/dendrite/roomserver/query/query.go
+++ b/src/github.com/matrix-org/dendrite/roomserver/query/query.go
@@ -49,10 +49,11 @@ type RoomserverQueryAPIDatabase interface {
 	// false if not.
 	// Returns an error if there was a problem talking to the database.
 	GetMembership(roomNID types.RoomNID, requestSenderUserID string) (membershipEventNID types.EventNID, stillInRoom bool, err error)
-	// Lookup the "join" membership event numeric IDs for all user that are
-	// currently members of a given room.
+	// Lookup the membership event numeric IDs for all user that are or have
+	// been members of a given room. Only lookup events of "join" membership if
+	// joinOnly is set to true.
 	// Returns an error if there was a problem talking to the database.
-	GetJoinMembershipEventNIDsForRoom(roomNID types.RoomNID) ([]types.EventNID, error)
+	GetMembershipEventNIDsForRoom(roomNID types.RoomNID, joinOnly bool) ([]types.EventNID, error)
 	// Look up the active invites targeting a user in a room and return the
 	// numeric state key IDs for the user IDs who sent them.
 	// Returns an error if there was a problem talking to the database.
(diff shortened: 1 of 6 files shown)


src/github.com/matrix-org/dendrite/roomserver/storage/storage.go
```diff
@@ -539,20 +540,66 @@ func (d *Database) GetMembershipEvents(roomNID types.RoomNID, requestSenderUserI
...
+
+	// Fetch the state as it was when this event was fired
+	stateEntries, err := state.LoadCombinedStateAfterEvents(d, prevState)
+	if err != nil {
+		return nil, err
+	}
+
+	var eventNIDs []types.EventNID
+	for _, entry := range stateEntries {
+		eventNIDs = append(eventNIDs, entry.EventNID)
```
NegativeMjark: I think you can filter for membership events here by checking for `entry.EventTypeNID == types.MRoomMemberNID`

NegativeMjark:
I'm slightly concerned at the amount of logic that is sprouting in the database layer for the membership query API. I suspect that most of the logic should be in the query API implementation.

babolivier:
Agreed. I'll move it to the API.

src/github.com/matrix-org/dendrite/roomserver/query/query.go
```diff
@@ -40,13 +40,31 @@ type RoomserverQueryAPIDatabase interface {
...
+	// Lookup all aliases referring to a given room ID.
+	// Returns an error if there was a problem talking to the database.
+	GetAliasesFromRoomID(roomID string) ([]string, error)
+	// Remove a given room alias.
+	// Returns an error if there was a problem talking to the database.
+	RemoveRoomAlias(alias string) error
+	// Lookup the membership of a given user in a given room.
+	// Returns the numeric ID of the latest membership event sent from this user
+	// in this room, along a boolean set to true if the user is still in this room,
+	// false if not.
```
NegativeMjark: It looks like you've accidentally re-added the ".*Alias.*" methods here. I don't think they are needed here anymore since the aliases are handled in a separate package.

babolivier: Done https://github.com/matrix-org/dendrite/pull/194/commits/d2f284173644d9c83f7b620cf8d4c5726a178067

src/github.com/matrix-org/dendrite/roomserver/query/query.go
```diff
@@ -215,6 +239,59 @@ func (r *RoomserverQueryAPI) QueryMembershipsForRoom(
...
+	// Filter the events to only keep the "join" membership events
+	for _, event := range stateEvents {
+		membership, err := event.Membership()
+		if err != nil {
+			return nil, err
+		}
+
+		if membership == "join" {
+			events = append(events, event)
+		}
```
NegativeMjark: Do we want to always filter down to the joined members? It looks like synapse returns all the membership events for `/members` and filters down to just the joined members for `/joined_members`.

https://github.com/matrix-org/synapse/blob/v0.22.1/synapse/rest/client/v1/room.py#L375-L393

babolivier: I didn't really look into Synapse's behavior, as I only worked with the spec. There's only the `/members` endpoint in it (https://matrix.org/speculator/spec/HEAD/client_server/unstable.html#get-matrix-client-r0-rooms-roomid-members) and it seemed to me from the examples that only `join` events were returned.

I'll follow Synapse's implementation then.

babolivier: Done https://github.com/matrix-org/dendrite/pull/194/commits/f946702dd54bfd13d9601f4ea6e4ad1ea7c70f8b

PR #198, opened 2017-08-23
Prefix logs with component name

babolivier:
With the monolith and all logs coming out of a single process, it's difficult to know which component emits which logs, and can get really confusing when logs look similar (e.g. almost all consumers output the same logs). This PR aims at fixing this issue by prefixing each line of log with the name of the component emitting it.

An example of this can be seen in [Travis logs](https://travis-ci.org/matrix-org/dendrite/builds/267536917).

This PR also fixes the import paths for the `logrus` package, which had the wrong casing.

The `Incoming request` and `Responding (XXX bytes)` logs aren't prefixed as they are emitted from https://github.com/matrix-org/util and I didn't think it was relevant to PR there with prefixes (plus it would have to impact all other go projects from Matrix).

The `main.go` files for each individual component don't use prefix since the component emitting the log in a given terminal is already implied when started in non-monolithic mode.

### DIFF
--- a/src/github.com/matrix-org/dendrite/common/log.go
+++ b/src/github.com/matrix-org/dendrite/common/log.go
@@ -18,8 +18,8 @@ import (
 	"os"
 	"path/filepath"
 
-	"github.com/Sirupsen/logrus"
 	"github.com/matrix-org/dugong"
+	"github.com/sirupsen/logrus"
 )
 
 type utcFormatter struct {
--- a/src/github.com/matrix-org/dendrite/common/log.go
+++ b/src/github.com/matrix-org/dendrite/common/log.go
@@ -15,26 +15,45 @@
 package common
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/matrix-org/dugong"
+	"github.com/mgutz/ansi"
 	"github.com/sirupsen/logrus"
 )
 
-type utcFormatter struct {
-	logrus.Formatter
+type dendriteFormatter struct {
+	logrus.TextFormatter
 }
 
-func (f utcFormatter) Format(entry *logrus.Entry) ([]byte, error) {
+func (f dendriteFormatter) Format(entry *logrus.Entry) ([]byte, error) {
+	if _, ok := entry.Data["prefix"]; ok {
+		prefix, ok := entry.Data["prefix"].(string)
+		if !ok {
+			return f.TextFormatter.Format(entry)
+		}
+
+		prefix = strings.ToUpper(prefix)
+
+		if !f.TextFormatter.DisableColors {
+			prefix = ansi.Color(prefix, "white+b")
+		}
+
+		entry.Message = fmt.Sprintf("%s: %s", prefix, entry.Message)
+		delete(entry.Data, "prefix")
+	}
+
 	entry.Time = entry.Time.UTC()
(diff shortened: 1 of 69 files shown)


src/github.com/matrix-org/dendrite/common/log.go
```diff
@@ -15,26 +15,62 @@
...
-func (f utcFormatter) Format(entry *logrus.Entry) ([]byte, error) {
+func (f dendriteFormatter) Format(entry *logrus.Entry) (format []byte, err error) {
 	entry.Time = entry.Time.UTC()
-	return f.Formatter.Format(entry)
+
+	if _, ok := entry.Data["prefix"]; ok {
+		prefix, ok := entry.Data["prefix"].(string)
+		if !ok {
+			return f.TextFormatter.Format(entry)
+		}
```
NegativeMjark: You don't need to check if the key is in the map. `entry.Data["prefix"].(string)` is safe to even if the key is in the map.

This is because `entry.Data["prefix"]` will be `nil` and `prefix, ok := interface{}(nil).(string)` will be `"", false`.

https://play.golang.org/p/blNqSLNe7Y

I'd write this as:

```go
prefix, ok := entry.Data["prefix"].(string)
￼if !ok {
￼    return f.TextFormatter.Format(entry)
￼}
```

and unindent the rest of the function.

babolivier: Yep, re-reading these lines makes me think I messed this up (might have been to tired to think correctly yesterday)

NegativeMjark:
Does this get set in time for the http logging in https://github.com/matrix-org/util/blob/master/json.go#L111 ?

babolivier:
The `common.SetupLogging` function is the first thing called in the `main()`, so it should be.

NegativeMjark:
I guess it'll be set to the global prefix, rather than the prefix for the component the HTTP API is part of.

babolivier:
Well there's no global prefix, the log coming from util will appear as a normal log without prefix like:

```
INFO[2017-08-23T14:10:41.146898434Z] Incoming request                              req.id=0z8crsUB0hLO req.method=GET req.path="/_matrix/client/r0/sync"
```
And having it to specify the specific component would require changing the prototype of util's `NewJSONRequestHandler` function, which would likely cause trouble in the other projects using it.

Or maybe I didn't understand your comment and you were suggesting having a global prefix is non is provided?

src/github.com/matrix-org/dendrite/common/log.go
babolivier: Done https://github.com/matrix-org/dendrite/pull/198/commits/b10cf01ae449d882f416441a08fd9686654028e3

NegativeMjark:
Ah right so they aren't prefixed with anything. How hard would it be to prefix the HTTP log lines with the correct component?

babolivier:
Not so much (I think), it would require to pass the prefix along `common.Make(Auth)API` and `util.NewJSONRequestHandler`. It would mean that the latter's prototype have to change, and the hardest part would be tracking down every Matrix project using it and update it so it use the new prototype.

NegativeMjark:
I guess this is a bit stale now.

> Oh, I think I sat on it long enough that it no longer merges. I liked the idea. but I'd rather we pushed it further down the stack to where we start the logging for the http request.
> (Which would be simpler to implement and would catch more of the logging)

PR #202, opened 2017-08-29
Check validity of keys for an incoming m.room.third_party_invite event

babolivier:


src/github.com/matrix-org/dendrite/roomserver/input/events.go
```diff
@@ -56,6 +58,15 @@ func processRoomEvent(db RoomEventDatabase, ow OutputRoomEventWriter, input api.
...
 	event := input.Event
 
+	if err := checkThirdPartyKeys(input.Event); err != nil {
+		// At least one of the key could not be validated. The specification
+		// says that it should not cause a rejection of the event.
+		// https://matrix.org/docs/spec/client_server/r0.2.0.html#id58
+		// There is no direct way to warn the users that the member's
+		// membership is questionable, so log the error as a warning.
+		logrus.Warn(err)
+	}
```
NegativeMjark: I don't think we want to have this check here at all. The room server should not need to call out to external servers to do it's job.

It's unclear as to whether we should have this check at all. The key validity check is performed by the server exchanging the "m.room.thirdpartyinvite" with the "m.room.member" event. Since the key validity checks depend on contacting an external server we can't use them as part of the normal event authentication since different servers might get different answers to the HTTP GET.

This means that we don't really have anything useful we can do with the result of the check other than warn about it.

It might be nice to know if a third-party-invite public key was no longer valid in the client sync API but we don't have a specification for that and since that state can change over time without the homeserver being notified it's unclear how we would implement that efficiently.

babolivier:
Right, seems like I misunderstood the spec. I'll close this PR then.

PR #206, opened 2017-09-01
Add config and checks for trusted ID servers

babolivier:
This PR:

* adds a section in the configuration file in order to configure trusted identity servers (defaults to no trusted servers)
* adds checks on all operation proxying requests to identity servers, and responds with `M_SERVER_NOT_TRUSTED` if the server isn't in the list of trusted servers


NegativeMjark:
I don't like how the threepid package has sprouted a mixture of functions that return an `util.JSONResponse` and functions that return an `error` that you call `ProcessIDServerError` to convert a JSON response.

How hard would it be to convert `CheckAndProcessInvite` to return an `error` rather than a `util.JSONResponse` ?

src/github.com/matrix-org/dendrite/clientapi/threepid/invites.go
```diff
@@ -66,6 +66,23 @@ type idServerStoreInviteResponse struct {
...
 
+// ProcessIDServerError differentiate errors caused by the absence of the identity
+// server in the list of trusted servers in the configuration file from normal
+// errors caused by a failure in a process.
+// Returns the representation of a 400 error in the former case (without logging),
+// and the one of a 500 error in the latter (with logging of the error).
+func ProcessIDServerError(req *http.Request, err error) util.JSONResponse {
+	// Check whether this is an
+	mErr, ok := err.(*jsonerror.MatrixError)
+	if !ok || mErr.ErrCode != "M_SERVER_NOT_TRUSTED" {
```
NegativeMjark: I think this check would be a lot cleaner if we used a custom error type rather than jsonerror.MatrixError here.

PR #209, opened 2017-09-05
Add get event federation API

erikjohnston:


src/github.com/matrix-org/dendrite/roomserver/query/query.go
```diff
@@ -330,6 +330,62 @@ func (r *RoomserverQueryAPI) QueryInvitesForUser(
...
+) error {
+	stateEntries, err := state.LoadStateAtEvent(r.DB, request.EventID)
+	if err != nil {
+		return err
+	}
+
+	stateAtEvent, err := r.loadStateEvents(stateEntries)
+	if err != nil {
+		return err
+	}
```
NegativeMjark: It would be nice if this could be done without loading *all* the events.
I suspect we could improve the performance here by storing numeric IDs for the membership and the domain part of the state key in the events table.

src/github.com/matrix-org/dendrite/roomserver/query/query.go
```diff
@@ -330,6 +330,62 @@ func (r *RoomserverQueryAPI) QueryInvitesForUser(
...
+			break
+		}
+	}
+
+	if isInRoom {
+		response.AllowedToSeeEvent = true
+		return nil
+	}
+
+	// TODO: Check if history visibility is shared and if the server is currently in the room
```
NegativeMjark: FWIW I think synapse always checks if the server is currently in the room.

https://github.com/matrix-org/synapse/blob/master/synapse/handlers/federation.py#L1385-L1394

src/github.com/matrix-org/dendrite/roomserver/query/query.go
NegativeMjark: Hmm it sounds like adding the necessary indexes is going to be a lot of effort. So I'm happy if we leave a TODO and add the index in another PR.

src/github.com/matrix-org/dendrite/federationapi/readers/events.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	if !authResponse.AllowedToSeeEvent {
+		return util.MessageResponse(403, "server not allowed to see event")
+	}
+
+	var eventsResponse api.QueryEventsByIDResponse
+	err = query.QueryEventsByID(&api.QueryEventsByIDRequest{
+		EventIDs: []string{eventID},
+	}, &eventsResponse)
```
NegativeMjark: I wonder if this would look nicer if it was written as:

```go
err = query.QueryEventsByID(
    &api.QueryEventsByIDRequest{EventIDs: []string{eventID}},
￼    &eventsResponse,
)
```

What do you think?

erikjohnston: wfm

PR #222, opened 2017-09-11
3PID invite exchange over federation

babolivier:
Follows https://github.com/matrix-org/gomatrixserverlib/pull/61

This PR address the case where the server receives a 3PID invite from an identity server for a room it's not in. It will then send to the room's server all the necessary data for building a `m.room.member` event so the room's server can authenticate it.

### DIFF
--- a/src/github.com/matrix-org/dendrite/federationapi/writers/threepid.go
+++ b/src/github.com/matrix-org/dendrite/federationapi/writers/threepid.go
@@ -129,27 +129,27 @@ func createInviteFrom3PIDInvite(
 		}
 		err = federation.ExchangeThirdPartyInvite(remoteServer, *builder)
 		return nil, err
-	} else {
-		// Auth the event locally
-		builder.Depth = queryRes.Depth
-		builder.PrevEvents = queryRes.LatestEvents
-
-		authEvents := gomatrixserverlib.NewAuthEvents(nil)
-
-		for i := range queryRes.StateEvents {
-			authEvents.AddEvent(&queryRes.StateEvents[i])
-		}
-
-		if err = fillDisplayName(builder, content, authEvents); err != nil {
-			return nil, err
-		}
-
-		refs, err := eventsNeeded.AuthEventReferences(&authEvents)
-		if err != nil {
-			return nil, err
-		}
-		builder.AuthEvents = refs
 	}
+
+	// Auth the event locally
+	builder.Depth = queryRes.Depth
+	builder.PrevEvents = queryRes.LatestEvents
+
+	authEvents := gomatrixserverlib.NewAuthEvents(nil)
+
+	for i := range queryRes.StateEvents {
+		authEvents.AddEvent(&queryRes.StateEvents[i])
+	}
+
+	if err = fillDisplayName(builder, content, authEvents); err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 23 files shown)


src/github.com/matrix-org/dendrite/federationapi/writers/threepid.go
```diff
@@ -120,11 +121,17 @@ func createInviteFrom3PIDInvite(
 	}
 
 	if !queryRes.RoomExists {
-		// TODO: Use federation to auth the event
-		return nil, nil
+		// Use federation to auth the event
+		var remoteServer gomatrixserverlib.ServerName
+		_, remoteServer, err = gomatrixserverlib.SplitID('!', inv.RoomID)
```
NegativeMjark: I think synapse also falls back to the server who sent the "m.room.third_party_invite" by using the user ID of the sender of the invite, or am I imagining things?

babolivier: Actually it first checks on the sender's server and then the room's server. Fixed in https://github.com/matrix-org/dendrite/pull/222/commits/0e356e2d272afae7640022d92c7ad68ca730e91a

src/github.com/matrix-org/dendrite/federationapi/writers/threepid.go
```diff
@@ -120,11 +121,30 @@ func createInviteFrom3PIDInvite(
 	}
 
 	if !queryRes.RoomExists {
-		// TODO: Use federation to auth the event
-		return nil, nil
+		// Use federation to auth the event
+		var remoteServers []gomatrixserverlib.ServerName
+		_, remoteServers[0], err = gomatrixserverlib.SplitID('@', inv.Sender)
```
NegativeMjark: This will have problems because the remoteServers slice will have length 0.
```go
remoteServers := make([]gomatrixserverlib.ServerName, 2)
```
might be better here.

src/github.com/matrix-org/dendrite/federationapi/writers/threepid.go
```diff
@@ -120,11 +121,30 @@ func createInviteFrom3PIDInvite(
...
+		if err != nil {
+			return nil, err
+		}
+
+		for _, server := range remoteServers {
+			if server == cfg.Matrix.ServerName {
+				continue
+			}
+
+			err = federation.ExchangeThirdPartyInvite(server, *builder)
```
NegativeMjark: If there is an error you may want to continue and try the next server.

src/github.com/matrix-org/dendrite/federationapi/writers/threepid.go
babolivier: Indeed. Fixed in https://github.com/matrix-org/dendrite/pull/222/commits/ebe91ac251f4f963625cf2fd3d9142d95407cc30

src/github.com/matrix-org/dendrite/federationapi/writers/threepid.go
babolivier: Fixed in https://github.com/matrix-org/dendrite/pull/222/commits/e1c8861c4ab96463fcc1aa1c759544e48aae7adc

PR #233, opened 2017-09-18
Add contexts to device database

NegativeMjark:


hooks/pre-commit
```diff
@@ -7,6 +7,16 @@ export GOGC=400
...
 
+echo "Checking that it builds"
+gb build
+
+# Check that all the packages can build.
+# When `go build` is given multiple packages it won't output anything, and just
+# checks that everything builds. This seems to do a better job of handling
+# missing imports than `gb build` does.
+echo "Double checking it builds..."
+go build github.com/matrix-org/dendrite/cmd/...
```
NegativeMjark: I've moved the build steps to run before the linter since it they are much faster than the linters and I'm a bit sick of seeing half the linters fail in parallel with the same build failure.

PR #235, opened 2017-09-19
Make media repo use error rather than jsonErrorResponse

erikjohnston:
This also includes changes to `.gitignore` so that we only ignore folders at the top level.

### DIFF
--- a/src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
+++ b/src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
@@ -36,6 +36,7 @@ import (
 	"github.com/matrix-org/dendrite/mediaapi/types"
 	"github.com/matrix-org/gomatrixserverlib"
 	"github.com/matrix-org/util"
+	"github.com/pkg/errors"
 )
 
 const mediaIDCharacters = "A-Za-z0-9_=-"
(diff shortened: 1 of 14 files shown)


src/github.com/matrix-org/dendrite/mediaapi/writers/download.go
```diff
@@ -167,26 +189,27 @@ func (r *downloadRequest) Validate() *util.JSONResponse {
...
+	activeRemoteRequests *types.ActiveRemoteRequests,
+	activeThumbnailGeneration *types.ActiveThumbnailGeneration,
+) (*types.MediaMetadata, error) {
 	// check if we have a record of the media in our database
 	mediaMetadata, err := db.GetMediaMetadata(r.MediaMetadata.MediaID, r.MediaMetadata.Origin)
 	if err != nil {
-		r.Logger.WithError(err).Error("Error querying the database.")
-		resErr := jsonerror.InternalServerError()
-		return &resErr
+		return nil, errors.Wrap(err, "Error querying the database.")
```
NegativeMjark: I'd leave out the full stop since this will be formatted as ```"Error querying the database." + ": " + err.Error()```.
Generally people seem to start error messages with lowercase letters so that they scan better in that format.
(I thought we had an automatic linter that checked the error format, but it might just have been kegan.)

Issue #212, opened 2017-09-06
Current sync server needs to be rewritten to support different streams


erikjohnston:
Currently its a hack to just support messages. While things like account data can be hacked on top, we probably want to fix it sooner rather than later.

See: https://github.com/matrix-org/dendrite/issues/211

NegativeMjark:
The sync server now happily supports multiple streams by using a shared postgres sequence to generate IDs. This will have scaling issues if we need to run multiple sync server instances, but that is tracked by  #211 for now.

PR #237, opened 2017-09-19
Use a shared PostgreSQL sequence to generate ids.

NegativeMjark:
Share an auto-incrementing sequence between the account data and the room event table.
This means that account data updates can be received independently of room events updates.

This should give some basic support for fixing #212

### DIFF
--- a/src/github.com/matrix-org/dendrite/syncapi/storage/account_data_table.go
+++ b/src/github.com/matrix-org/dendrite/syncapi/storage/account_data_table.go
@@ -18,14 +18,19 @@ import (
 	"context"
 	"database/sql"
 
+	"github.com/matrix-org/dendrite/common"
+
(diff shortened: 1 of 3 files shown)


src/github.com/matrix-org/dendrite/syncapi/storage/account_data_table.go
```diff
@@ -18,23 +18,26 @@ import (
...
+	"github.com/matrix-org/dendrite/common"
+
 	"github.com/matrix-org/dendrite/syncapi/types"
 )
 
 const accountDataSchema = `
+-- This sequence is shared between all the tables generated from kafka logs.
+CREATE SEQUENCE IF NOT EXISTS syncapi_stream_id;
+
 -- Stores the users account data
```
erikjohnston: LIES

PR #241, opened 2017-09-20
Add 'unused' lint

erikjohnston:


src/github.com/matrix-org/dendrite/cmd/syncserver-integration-tests/testdata.go
```diff
@@ -35,15 +35,6 @@ const (
...
 	i18StateAliceRoomName
-	i19BobMsg
-	i20StateBobLeave
-	i21AliceMsg
-	i22StateAliceInviteBob
-	i23StateBobRejectInvite
-	i24AliceMsg
-	i25StateAliceRoomName
-	i26StateCharlieJoin
-	i27CharlieMsg
```
NegativeMjark: Can we keep these somehow? I suspect they might be the only nice way to identify the test data here.

richvdh:
ok to test

src/github.com/matrix-org/dendrite/cmd/syncserver-integration-tests/testdata.go
erikjohnston: I've kept them, though I find the entire file deeply confusing.

NegativeMjark:
LGTM

PR #255, opened 2017-09-21
Update the install instructions

richvdh:
To include some of the newer components.


INSTALL.md
```diff
@@ -108,23 +108,50 @@ The following contains scripts which will run all the required processes in orde
...
+                                   |    |     +--------------------------+                 ||
+Matrix      +------------------+   |    |   :7773    |                                     ||
+Clients --->| client-api-proxy |---+    +<-----------+                                     ||
+            +------------------+   |    |                                                  ||
+          :8008                    | CS API   +----------------------------+               ||
+                                   +--------->| dendrite-client-api-server |==++           ||
+                                        |     +----------------------------+  ||           ||
+                                        |   :7771    |                        ||           ||
+                                        |            |            client_data ||           ||
+                                        +<-----------+         ++=============++           ||
```
NegativeMjark: I think the client_data stream goes to the sync-api-server not the room-server

INSTALL.md
```diff
@@ -108,23 +108,50 @@ The following contains scripts which will run all the required processes in orde
...
+                                        |     +-----------------------------------+              Matrix
+                                        |     | dendrite-federation-sender-server |------------> Servers
+                                        |     +-----------------------------------+
+                                        |   :7776
+                                        |
+                                        +<-----------+
+                                                     |
+Matrix      +----------------------+  SS API  +--------------------------------+
+Servers --->| federation-api-proxy |--------->| dendrite-federation-api-server |
+            +----------------------+          +--------------------------------+
```
NegativeMjark: Technically the federation-api-proxy talks to the media-api-server as well...

INSTALL.md
richvdh: So it does. fixed in 46e0a66

INSTALL.md
richvdh: Are you sure? how?

NegativeMjark: Oh, I guess no one implemented that yet. It SHOULD talk to the media-api-server as well :)

NegativeMjark: I'll fix that in a separate PR I guess.

PR #257, opened 2017-09-22
Implement shared secret registration

erikjohnston:


src/github.com/matrix-org/dendrite/clientapi/writers/register.go
```diff
@@ -133,6 +157,79 @@ func Register(req *http.Request, accountDB *accounts.Database, deviceDB *devices
...
+	switch r.Type {
+	case authtypes.LoginTypeSharedSecret:
+		if cfg.Matrix.RegistrationSharedSecret == "" {
+			return util.MessageResponse(400, "Shared secret registration is disabled")
+		}
+
+		valid, err := isValidMacLogin(r.Username, r.Password, r.Admin, r.Mac, cfg.Matrix.RegistrationSharedSecret)
+		if err != nil {
+			return httputil.LogThenError(req, err)
+		}
```
NegativeMjark: Most of the errors here are the result of invalid input.
At the very least we should have a TODO to suggest converting them to 4xx series errors.

erikjohnston: Ugh, true. I wish there was a nicer way of dealing with this

PR #258, opened 2017-09-22
Add missing GET for /login

erikjohnston:


erikjohnston:
The login function already handles GET method.

babolivier:
Indeed, sorry I missed it.

PR #271, opened 2017-09-28
Add config for setting up a jaeger opentracing reporter

NegativeMjark:


dendrite-config.yaml
```diff
@@ -91,3 +91,12 @@ listen:
...
     public_rooms_api: "localhost:7775"
     federation_sender: "localhost:7776"
+
+# The configuration for tracing the dendrite components.
+tracing:
+    # Config for the jaeger opentracing reporter.
+    # See https://godoc.org/github.com/uber/jaeger-client-go/config#Configuration
+    # for documtation.
+    jaeger:
+        # Remove this line to enable the reporter.
```
erikjohnston: Surely just set `disabled: false`, as otherwise its not valid yaml.

And `enabled: false` is possibly better?

PR #272, opened 2017-09-29
WIP: Implement Filter POSTing and GETting

jcgruenhage:
This depends on matrix-org/gomatrix#42.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,8 +24,8 @@ before_script:
  - openssl req -x509 -newkey rsa:4096 -keyout server.key -out server.crt -days 365 -nodes -subj /CN=localhost
 
 script:
- - ./travis-install-kafka.sh
- - ./travis-test.sh
+ - ./scripts/install-local-kafka.sh
+ - ./scripts/travis-test.sh
 
 notifications:
     webhooks:
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -17,7 +17,7 @@ Dendrite can be run in one of two configurations:
  - For Kafka (optional if using the monolith server):
    - Unix-based system (https://kafka.apache.org/documentation/#os)
    - JDK 1.8+ / OpenJDK 1.8+
-   - Apache Kafka 0.10.2+ (see [travis-install-kafka.sh](travis-install-kafka.sh) for up-to-date version numbers)
+   - Apache Kafka 0.10.2+ (see [scripts/install-local-kafka.sh](scripts/install-local-kafka.sh) for up-to-date version numbers)
 
 
 ## Setting up a development environment
@@ -34,7 +34,7 @@ go get github.com/constabulary/gb/...
 gb build
 ```
 
-If using Kafka, install and start it (c.f. [travis-install-kafka.sh](travis-install-kafka.sh)):
+If using Kafka, install and start it (c.f. [scripts/install-local-kafka.sh](scripts/install-local-kafka.sh)):
 ```bash
 MIRROR=http://apache.mirror.anlx.net/kafka/0.10.2.0/kafka_2.11-0.10.2.0.tgz
 
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -2,33 +2,4 @@
 
 set -eu
 
-# Tune the GC to use more memory to reduce the number of garbage collections
-export GOGC=400
-export GOPATH="$(pwd):$(pwd)/vendor"
-export PATH="$PATH:$(pwd)/vendor/bin:$(pwd)/bin"
-
-echo "Checking that it builds"
-gb build
-
-# Check that all the packages can build.
-# When `go build` is given multiple packages it won't output anything, and just
-# checks that everything builds. This seems to do a better job of handling
-# missing imports than `gb build` does.
-echo "Double checking it builds..."
-go build github.com/matrix-org/dendrite/cmd/...
-
-echo "Installing lint search engine..."
-go install github.com/alecthomas/gometalinter/
-gometalinter --config=linter.json ./... --install
-
-echo "Looking for lint..."
-gometalinter --config=linter.json ./... --enable-gc
-
-echo "Double checking spelling..."
-misspell -error src *.md
-
-echo "Testing..."
-gb test
-
-
-echo "Done!"
+./scripts/build-test-lint.sh
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -1,0 +1,7 @@
+# Dev Scripts
+
+These are a collection of scripts that should be helpful for those developing
+on dendrite.
+
+See `find-lint.sh` for environment variables that control linter resource
+usage.
--- a/scripts/build-test-lint.sh
+++ b/scripts/build-test-lint.sh
@@ -1,0 +1,26 @@
+#! /bin/bash
+
+# Builds, tests and lints dendrite, and should be run before pushing commits
+
+set -eu
+
+export GOPATH="$(pwd):$(pwd)/vendor"
+export PATH="$PATH:$(pwd)/vendor/bin:$(pwd)/bin"
+
+echo "Checking that it builds"
+gb build
+
+# Check that all the packages can build.
+# When `go build` is given multiple packages it won't output anything, and just
+# checks that everything builds. This seems to do a better job of handling
+# missing imports than `gb build` does.
+echo "Double checking it builds..."
+go build github.com/matrix-org/dendrite/cmd/...
+
+./scripts/find-lint.sh
+
+echo "Double checking spelling..."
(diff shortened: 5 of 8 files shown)


src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/filter_table.go
```diff
@@ -0,0 +1,92 @@
...
+	"fmt"
+)
+
+const filterSchema = `
+-- Stores data about filters
+CREATE TABLE IF NOT EXISTS account_filter (
+	-- The filter
+	filter TEXT NOT NULL,
+	-- The ID
+	id SERIAL UNIQUE,
```
jcgruenhage: SERIAL means that postgres will just increment that forever, which leaks a tiny bit of data to users creating filters, namely how many filters have been created on the server. Synapse does something similar, but per user.

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/storage.go
```diff
@@ -314,3 +319,21 @@ func (d *Database) GetThreePIDsForLocalpart(
...
+}
+
+func (d *Database) PutFilter(
+	ctx context.Context, localpart, filter string,
+) (string, error) {
+	err := d.filter.insertFilter(ctx, filter, localpart)
+	if err != nil {
+		return "", err
+	}
+	return d.filter.findMaxID(ctx, localpart)
```
jcgruenhage: This is somewhat ugly, but I haven't found another way to get the integer created by postgres.

src/github.com/matrix-org/dendrite/clientapi/routing/routing.go
```diff
@@ -185,23 +185,17 @@ func Setup(
...
-			return util.JSONResponse{
-				Code: 200,
-				JSON: struct{}{},
-			}
+	r0mux.Handle("/user/{userId}/filter",
+		common.MakeAuthAPI("put_filter", deviceDB, func(req *http.Request, device *authtypes.Device) util.JSONResponse {
+			vars := mux.Vars(req)
+			return writers.PutFilter(req, device, accountDB, vars["userId"])
 		}),
-	).Methods("POST", "OPTIONS")
```
jcgruenhage: What was this needed for? I haven't found that in the spec.

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/storage.go
NegativeMjark: You can use ```RETURNING``` to get the ID chosen by the ```INSERT``` https://www.postgresql.org/docs/9.5/static/dml-returning.html
There's an example of use doing this in dendrite at https://github.com/matrix-org/dendrite/blob/master/src/github.com/matrix-org/dendrite/syncapi/storage/output_room_events_table.go#L56

src/github.com/matrix-org/dendrite/clientapi/routing/routing.go
NegativeMjark: We need the OPTIONS to satisfy CORS request pre-flighting https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request otherwise it's hard to hit the API from web clients.

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/filter_table.go
NegativeMjark: Seems reasonable. (If we wanted to be clever here we could use a SHA-256 hash of the filter as the filter ID. That way we wouldn't get duplicates if the clients uploaded the same filter multiple times.)

jcgruenhage: We would still get duplicates for multiple users though.. That this is at all user specific is somewhat displeasing, it would be nicer if there was just an ID and the associated filter, completely leaving out the users. @NegativeMjark Do you think I should use SHA256 as the ID or keep it as it is currently?

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/storage.go
jcgruenhage: I'll definitely do this, if we stick to using "SERIAL" here. If we use the hash as the ID, this is would not be needed anymore.

src/github.com/matrix-org/dendrite/clientapi/routing/routing.go
jcgruenhage: oh, okay, so I'll readd this. Should I go ahead and edit the current commit, or are multiple commits maybe better? The first would make the git history nicer, the latter would make the review and evolution inside this PR look nicer.

erikjohnston: We squash merge so it doesn't really matter :)

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/filter_table.go
erikjohnston: I think ID is fine for now

src/github.com/matrix-org/dendrite/clientapi/routing/routing.go
jcgruenhage: okay, nice 👍

jcgruenhage: About the OPTIONS thingy there, I do not need to add more than this additional method, do I? I couldn't find more handling of OPTIONS at some other points, so I guess the go http package just handles that on his own?

erikjohnston: No, you don't need to do anything more. It gets handled by the wrapper i think

erikjohnston:
The travis test failure is now due to linting errors:

```
src/github.com/matrix-org/dendrite/clientapi/writers/filter.go:30:1:warning: filterResponse is unused (deadcode)
src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/storage.go:331:1:warning: exported method Database.PutFilter should have comment or be unexported (golint)
src/github.com/matrix-org/dendrite/clientapi/readers/filter.go:30:1:warning: exported function GetFilter should have comment or be unexported (golint)
src/github.com/matrix-org/dendrite/clientapi/writers/filter.go:34:1:warning: exported function PutFilter should have comment or be unexported (golint)
src/github.com/matrix-org/dendrite/clientapi/readers/filter.go:33:19:warning: 2 other occurrence(s) of "GET" found in: src/github.com/matrix-org/dendrite/clientapi/readers/login.go:63:19 src/github.com/matrix-org/dendrite/clientapi/readers/profile.go:52:19 (goconst)
src/github.com/matrix-org/dendrite/clientapi/readers/login.go:63:19:warning: 2 other occurrence(s) of "GET" found in: src/github.com/matrix-org/dendrite/clientapi/readers/filter.go:33:19 src/github.com/matrix-org/dendrite/clientapi/readers/profile.go:52:19 (goconst)
src/github.com/matrix-org/dendrite/clientapi/readers/profile.go:52:19:warning: 2 other occurrence(s) of "GET" found in: src/github.com/matrix-org/dendrite/clientapi/readers/filter.go:33:19 src/github.com/matrix-org/dendrite/clientapi/readers/login.go:63:19 (goconst)
src/github.com/matrix-org/dendrite/clientapi/writers/filter.go:30:6:warning: type filterResponse is unused (U1000) (megacheck)
src/github.com/matrix-org/dendrite/clientapi/writers/filter.go:62:2:warning: empty branch (SA9003) (megacheck)
src/github.com/matrix-org/dendrite/clientapi/writers/filter.go:67:2:warning: empty branch (SA9003) (megacheck)
```

jcgruenhage:
I'll fix those after fixing the missing TODOs above.

erikjohnston:
I think so long as you handle the two errors of the form:

```
	if err != nil {
		//TODO
	}
```

I'm happy to merge as is (and adding issues to track the TODOs)

jcgruenhage:
I've already fixed that locally (at home, I'm at work currently), and I've fixed everything the linter complains about. I'll rebase it once I'm home and then push the changes here, so that you can merge them.

jcgruenhage:
ehm....
what happened?!

jcgruenhage:
Apparently github autocloses PRs, when you acidentally force push the destination branch to the source branch...

erikjohnston:
> Apparently github autocloses PRs, when you acidentally force push the destination branch to the source branch...

Sigh, guess what I just did...

PR #275, opened 2017-10-03
Add a very basic /initialSync implementation

erikjohnston:
This is done by coercing a full sync response into the right shape.

Some notable TODOs:

- Support specifying a limt
- Support invites
- Correctly format events
- Add presence, account data, etc

The motivation for including such a bare bones implementation is to allow more sytests to run.

### DIFF
--- a/src/github.com/matrix-org/dendrite/syncapi/sync/request.go
+++ b/src/github.com/matrix-org/dendrite/syncapi/sync/request.go
@@ -26,7 +26,10 @@ import (
 )
 
 const defaultSyncTimeout = time.Duration(30) * time.Second
-const defaultTimelineLimit = 20
+
+// DefaultTimelineLimit is the default number limit of timeline entries if no
+// limit was given.
+const DefaultTimelineLimit = 20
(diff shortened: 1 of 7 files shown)


erikjohnston:
(I'll add a tracking issue for the TODOs and add the number to the comments when/if this gets approved)

src/github.com/matrix-org/dendrite/syncapi/sync/request.go
```diff
@@ -26,7 +26,10 @@ import (
 )
 
 const defaultSyncTimeout = time.Duration(30) * time.Second
-const defaultTimelineLimit = 20
+
+// DefaultTimelineLimit is the default number limit of timeline entries if no
+// limit was given.
+const DefaultTimelineLimit = 20
```
richvdh: no particular objection to this, but it sees unnecessary given you haven't added any references to it outside the package?

richvdh:
I think we've decided to change sytest to use /sync rather than do this, ftr

PR #276, opened 2017-10-05
Implement GET /rooms/{roomID}/state/{type}/{stateKey}

erikjohnston:


erikjohnston:
Hmm. We should probably auth these endpoints.

erikjohnston:
@richvdh agreed about the FIXME's, I've been trying to add issues for this sort of thing. Do you have an opinion on TODO vs FIXME?

richvdh:
I'd have gone with FIXME in this case, since failing to do so has more unfortunate effects than things not working. But shrug

PR #281, opened 2017-10-05
Generate new devices for each new /login

erikjohnston:


src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
```diff
@@ -62,7 +62,7 @@ func VerifyAccessToken(req *http.Request, deviceDB DeviceDatabase) (device *auth
 	if err != nil {
 		if err == sql.ErrNoRows {
 			resErr = &util.JSONResponse{
-				Code: 403,
+				Code: 401,
```
erikjohnston: This matches synapse behaviour, and is what sytest expects

richvdh: Indeed: https://matrix.org/speculator/spec/HEAD/client_server/unstable.html#client-authentication

"When credentials are required but missing or invalid, the HTTP call will return with a status of 401 and the error code, M_MISSING_TOKEN or M_UNKNOWN_TOKEN respectively."

so, er, shouldn't we be returning `M_UNKNOWN_TOKEN` ?

src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
```diff
@@ -86,6 +86,18 @@ func GenerateAccessToken() (string, error) {
 	return base64.RawURLEncoding.EncodeToString(b), nil
 }
 
+// GenerateDeviceID creates a new device id. Returns an error if failed to generate
+// random bytes.
+func GenerateDeviceID() (string, error) {
+	b := make([]byte, tokenByteLength)
```
richvdh: let's not have 42-character device IDs. The current ones are 8 characters. Sorry, but you probably need to do this the hard way, with reference to the database.

src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
```diff
@@ -62,7 +65,7 @@ func VerifyAccessToken(req *http.Request, deviceDB DeviceDatabase) (device *auth
 	if err != nil {
 		if err == sql.ErrNoRows {
 			resErr = &util.JSONResponse{
-				Code: 403,
+				Code: 401,
```
richvdh: still think we need `M_UNKNOWN_TOKEN` here

src/github.com/matrix-org/dendrite/clientapi/readers/login.go
```diff
@@ -113,15 +114,12 @@ func Login(
...
-			return util.JSONResponse{
-				Code: 500,
-				JSON: jsonerror.Unknown("Failed to generate access token"),
-			}
+			httputil.LogThenError(req, err)
 		}
 
 		// TODO: Use the device ID in the request
 		dev, err := deviceDB.CreateDevice(
-			req.Context(), acc.Localpart, auth.UnknownDeviceID, token,
```
richvdh: can we get rid of `auth.UnknownDeviceID` now?

PR #288, opened 2017-10-06
Untangle precommit and travis test scripts

erikjohnston:
The main motivation for this is that we want to be able to configure the linters depending on if they're run by travis (use all the CPU and RAM) or locally (avoid locking up the computer)


scripts/README.md
```diff
@@ -0,0 +1,21 @@
+# Dev Scripts
+
+These are a collection of scripts that should be helpful for those developing
+on dendrite.
+
+- `./scripts/find-lint.sh` runs the linters against dendrite,
```
richvdh: this looks like a list that is going to get out of date in about 30 seconds.

I'd suggest sticking these descriptions in the comments at the top of each script. And possibly duplicating the env var descriptions to each one?

scripts/find-lint.sh
```diff
@@ -0,0 +1,32 @@
...
+go install github.com/alecthomas/gometalinter/
+gometalinter --config=linter.json ./... --install
+
+echo "Looking for lint..."
+gometalinter ./... $args
+
+echo "Double checking spelling..."
+misspell -error src *.md
+
+echo "Done!"
```
richvdh: this is going to look silly when it's called from another script

scripts/travis-test.sh
```diff
@@ -13,7 +16,7 @@ gb build github.com/matrix-org/dendrite/cmd/mediaapi-integration-tests
 gb build github.com/matrix-org/dendrite/cmd/client-api-proxy
 
 # Run the pre commit hooks
```
richvdh: no longer the pre-commit hooks

scripts/README.md
```diff
@@ -0,0 +1,21 @@
...
+  `./scripts/find-lint.sh fast` runs a subset of faster lints
+- `./scripts/build-test-lint.sh` builds, tests and lints dendrite, and
+  should be run before pushing commits
+- `./scripts/install-local-kafka.sh` downloads, installs and runs a
+  kafka instance
+- `./scripts/travis-test.sh` is what travis runs
+
+
+The linters can take a lot of resources and are slow, so they can be
+configured using two enviroment variables:
```
richvdh: enviroment

scripts/travis-test.sh
```diff
@@ -2,6 +2,9 @@
 
 set -eu
 
+# Tune the GC to use more memory to reduce the number of garbage collections
+export GOGC=400
```
richvdh: do we need this as well as DENDRITE_LINT_DISABLE_GC ?

.travis.yml
```diff
@@ -24,8 +24,8 @@ before_script:
  - openssl req -x509 -newkey rsa:4096 -keyout server.key -out server.crt -days 365 -nodes -subj /CN=localhost
 
 script:
- - ./travis-install-kafka.sh
- - ./travis-test.sh
+ - ./scripts/install-local-kafka.sh
```
richvdh: this is linked to as travis-install-kafka from the INSTALL instructions iirc, pls could you have a grep and fix

scripts/travis-test.sh
erikjohnston: `DENDRITE_LINT_DISABLE_GC` won't disable GC for the unit/integration tests or anything, which I guess is why its there?

PR #289, opened 2017-10-06
Clean up README a bit

erikjohnston:


README.md
```diff
@@ -2,153 +2,28 @@
 
 Dendrite will be a matrix homeserver written in go.
 
-# Install
+Its still very much a work in progress, but installation instructions can
```
richvdh: It's

README.md
```diff
@@ -2,153 +2,28 @@
...
-The server would then decompose into two categories: writers that add new
-entries to the logs and readers that read those entries.
+For general questions about Dendrite we have a dedicated room on Matrix
+[#dendrite:matrix.org](https://riot.im/develop/#/room/#dendrite:matrix.org).
 
-The event logs then serve to decouple the two components, the writers and
-readers need only agree on the format of the entries in the event log.
-This format could be largely derived from the wire format of the events used
-in the client and federation protocols:
+For questions related to developing there's also [#dendrite-dev:matrix.org](https://riot.im/develop/#/room/#dendrite-dev:matrix.org)
```
richvdh: we've already got this in CONTRIBUTING; don't think it needs to be here under 'support' too. (in fact, I rather wonder why we are offering support on dendrite at all: if you aren't contributing, you don't want to be running it for now).

Issue #284, opened 2017-10-05
Implement /register/available API


erikjohnston:
This is as simple `GET /register/available?username=....` request that checks if a user name is available. It returns `200 OK {"available": True}` if the username is available and valid, and a `400` response if its not available/invalid. 

The synapse code is: https://github.com/matrix-org/synapse/blob/erikj/groups_merged/synapse/rest/client/v2_alpha/register.py#L144-L155

Rate limiting can be ignored for now and done separately.

PR #291, opened 2017-10-06
Implement /register/available API

MTRNord:
My first PR so if there is anything wrong with it please tell me ;) I hope I didn't miss anything.

Adds `GET /register/available?username=....`
I think it does what is expected by #284


src/github.com/matrix-org/dendrite/clientapi/routing/routing.go
```diff
@@ -131,6 +131,10 @@ func Setup(
 		return writers.LegacyRegister(req, accountDB, deviceDB, &cfg)
 	})).Methods("POST", "OPTIONS")
 
+	r0mux.Handle("/register/available", common.MakeExternalAPI("registerAvailable", func(req *http.Request) util.JSONResponse {
+		return readers.RegisterAvailable(req, accountDB)
+	})).Methods("GET", "OPTIONS")
```
babolivier: Handling `OPTIONS` here isn't needed as CORS pre-flighted requests aren't sent for a `GET` request: https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Preflighted_requests

src/github.com/matrix-org/dendrite/clientapi/readers/register.go
```diff
@@ -0,0 +1,66 @@
...
+			JSON: jsonerror.InvalidUsername("User ID can only contain characters a-z, 0-9, or '_-./'"),
+		}
+	} else if username[0] == '_' { // Regex checks its not a zero length string
+		return &util.JSONResponse{
+			Code: 400,
+			JSON: jsonerror.InvalidUsername("User ID can't start with a '_'"),
+		}
+	}
+	return nil
+}
```
erikjohnston: This seems to be duplicated from `writers.register`? Either we should make he validation function in writers public, or we should move the `RegisterAvailable` func to `writers.register` (which is probably my preferred approach)

erikjohnston: This seems to be duplicated from writers.register? Either we should make he validation function in writers public, or we should move the RegisterAvailable func to writers.register (which is probably my preferred approach)

MTRNord: The function on writers.register is basicly te same except that one also checks the password. So either way I would atleast need to split the validate function in username and password

erikjohnston: I think splitting up into `validateUserName` and `validatePassword` makes sense

PR #292, opened 2017-10-07
Fix initial / fullSync request

MTRNord:
The initial/fullSync request sometimes crashed as the timeout triggered before it finished.

Compare Synpase which got the same if: https://github.com/matrix-org/synapse/blob/b23cb8fba8c783bf7a267bfbe33b50e010f17787/synapse/handlers/sync.py#L192

Signed-Off-By: Marcel Radzio <mtrnord1@gmail.com>


MTRNord:
It might need to rethinked. Doing this keeps Riot from doing the very first request after creation for some reason

erikjohnston:
So I think the code should work something like:

1. Calculate sync response
2. If initial sync, timeout=0 or response is non empty always return
3. Start timer and wait to be notified
4. If woken by notify, calculate sync response and return if response non empty else wait to be notified
5. When timeout hits return

Closing for now.

erikjohnston:
https://github.com/matrix-org/dendrite/issues/293

PR #295, opened 2017-10-10
Merge readers/writers/routing pacakges

### LINKED ISSUE #273: Remove reader/writer split
The HTTP handlers in the components are split into reader and writer directories. This was a fairly arbitrary distinction, and turns out to not be so helpful. Most read APIs have a corresponding write API, and i would be more natural for them to be in the same file rather than in different directories.

erikjohnston:
Fixes #273


erikjohnston:
(The big changes to the filter.go are simply me merging readers/filter.go and writers/filter.go, which breaks the diffing algorithm)

PR #299, opened 2017-10-10
Update dendrite-config.yaml to use server.crt, not .pem

erikjohnston:


richvdh:
you need three things:
* pem-formatted private ed25519 signing key (currently called `matrix_key.pem` everywhere?)
* pem-formatted private RSA key for TLS (currently called `server.key` everywhere?)
* pem-formatted X.509 cert for the above for TLS (currently called `server.crt` in the install and `server.pem` in dendrite-config.yaml)?

I can't help feel that we could do quite a lot better than this.

richvdh:
suppose this PR is an improvement anyway :-S

ara4n:
I'd strongly suggest we call the signing key signing_key.pem (perhaps prefixed by the hostname) - surely this is something we actually got right in synapse, and should continue to get right in dendrite?

ara4n:
but synapse calls them `sierra.signing.key`, `sierra.tls.crt`, `sierra.tls.dh` and `sierra.tls.key` which seems completely unambiguous and clear?

Issue #300, opened 2017-10-10
Handle duplicate kafka/naffka messages


erikjohnston:
Since we only persist the partition offset after we've processed the message it's possible for dendrite services to receive the last message from kafka again if it died at the right/wrong point.

Currently e.g. federation sender doesn't handle this (causing #290)

Two options:
1. Ensure we handle duplicate messages
2. Persist the results of processing the messages atomically with the offsets

erikjohnston:
> Persist the results of processing the messages atomically with the offsets

This is probably the better option, but its probably going to be quite faffy to actually code due to having to figure out how to atomically persist things without pushing processing into the database layer and using DB transactions

neilalexander:
We're NATS now.

PR #301, opened 2017-10-10
Handle duplicate kafka messages

erikjohnston:
The way we store the partition offsets for kafka streams means that when we start after a crash we may get the last message we processed again. This means that we have to be careful to ensure that the processing handles consecutive duplicates correctly.

This may or may not be a temporary fix depending on how we more generally address #300

### DIFF
--- a/src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
+++ b/src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
@@ -146,7 +146,7 @@ func (s *OutputRoomEvent) processMessage(ore api.OutputNewRoomEvent) error {
 	}
 
 	// Work out which hosts were joined at the event itself.
-	joinedHostsAtEvent, err := s.joinedHostsAtEvent(ore, *oldJoinedHosts)
+	joinedHostsAtEvent, err := s.joinedHostsAtEvent(ore, oldJoinedHosts)
 	if err != nil {
 		return err
 	}
--- a/src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
+++ b/src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
@@ -137,6 +137,8 @@ func (s *OutputRoomEvent) processMessage(ore api.OutputNewRoomEvent) error {
 	if oldJoinedHosts == nil {
 		// This means that there is nothing to update as this is a duplicate
 		// message.
+		// This can happen if dendrite crashed between reading the message and
+		// persisting the stream position.
 		return nil
 	}
 
--- a/src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
+++ b/src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
@@ -134,25 +134,27 @@ func (s *OutputRoomEvent) processMessage(ore api.OutputNewRoomEvent) error {
 		return err
 	}
 
+	if oldJoinedHosts == nil {
+		// This means that there is nothing to update as this is a duplicate
+		// message.
+		return nil
+	}
+
 	if ore.SendAsServer == api.DoNotSendToOtherServers {
 		// Ignore event that we don't need to send anywhere.
 		return nil
 	}
 
 	// Work out which hosts were joined at the event itself.
-	joinedHostsAtEvent, err := s.joinedHostsAtEvent(ore, oldJoinedHosts)
+	joinedHostsAtEvent, err := s.joinedHostsAtEvent(ore, *oldJoinedHosts)
 	if err != nil {
 		return err
 	}
 
 	// Send the event.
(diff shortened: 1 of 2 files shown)


src/github.com/matrix-org/dendrite/federationsender/storage/storage.go
```diff
@@ -62,30 +62,47 @@ func (d *Database) prepare() error {
...
 			return err
 		}
+
 		lastSentEventID, err := d.selectRoomForUpdate(ctx, txn, roomID)
 		if err != nil {
 			return err
 		}
+
+		if lastSentEventID == newEventID {
+			// We've handled this messages before, so lets just ignore it
```
richvdh: this messages

src/github.com/matrix-org/dendrite/federationsender/storage/storage.go
```diff
@@ -62,30 +62,47 @@ func (d *Database) prepare() error {
...
+		if err != nil {
 			return err
 		}
+
 		lastSentEventID, err := d.selectRoomForUpdate(ctx, txn, roomID)
 		if err != nil {
 			return err
 		}
+
+		if lastSentEventID == newEventID {
```
richvdh: why only a strict equality check? what if we processed more than one message?

src/github.com/matrix-org/dendrite/federationsender/storage/storage.go
richvdh: let's

src/github.com/matrix-org/dendrite/federationsender/storage/storage.go
```diff
@@ -62,30 +62,47 @@ func (d *Database) prepare() error {
...
+// this isn't a duplicate message.
 func (d *Database) UpdateRoom(
 	ctx context.Context,
 	roomID, oldEventID, newEventID string,
 	addHosts []types.JoinedHost,
 	removeHosts []string,
-) (joinedHosts []types.JoinedHost, err error) {
-	err = common.WithTransaction(d.db, func(txn *sql.Tx) error {
-		if err = d.insertRoom(ctx, txn, roomID); err != nil {
+) (*[]types.JoinedHost, error) {
```
richvdh: does this need to be a pointer? slices can be `nil`, no?

src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
```diff
@@ -134,25 +134,27 @@ func (s *OutputRoomEvent) processMessage(ore api.OutputNewRoomEvent) error {
 		return err
 	}
 
+	if oldJoinedHosts == nil {
```
richvdh: can we remove the TODO at  line 123?

src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
```diff
@@ -134,25 +134,27 @@ func (s *OutputRoomEvent) processMessage(ore api.OutputNewRoomEvent) error {
...
+		// This means that there is nothing to update as this is a duplicate
+		// message.
```
richvdh: I think it's worth commenting on why this situation arises. If it's only because dendrite crashed between reading the message and persisting the stream position, say so.

src/github.com/matrix-org/dendrite/federationsender/storage/storage.go
erikjohnston: We can only get consecutive duplicate messages from kafka (when we crash before persisting our stream position), and I'm not sure we really want to be storing more event IDs.

erikjohnston: (I've added to the comment beneath it to explain)

src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
erikjohnston: TBH, I don't know to what extent that TODO was:
- just for paranoia if upstream sends down a buggy message
- its intended that nodes can throw away state for a room
- there's a situation where upstream wants to send down incomplete stream of state

Basically, I'm inclined to leave it for now as I don't think this PR addresses the issue that the TODO was trying to cover

src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
erikjohnston: (Done)

src/github.com/matrix-org/dendrite/federationsender/consumers/roomserver.go
richvdh: mmkay. I'm inclined to think that a TODO we don't understand doesn't have much value; possibly it could do with clarifying.

PR #302, opened 2017-10-11
Fix some edge cases with /sync

### LINKED ISSUE #293: Fix /sync edge cases
The correct way of dealing with sync requests is:

1. Calculate sync response
2. If initial sync, timeout=0 or response is non empty always return
3. Start timer and wait to be notified
4. If woken by notify, calculate sync response and return if response non empty else wait to be notified
5. When timeout hits return

The current behaviour a) doesn't respect `timeout=0`, b) doesn't check if there is anything to immediately return before waiting on notifier (including when doing initial sync)

erikjohnston:
Including:
- Handle timeout=0 correctly
- Always return immediately on initial sync
- Handle spurious wake ups from the notifier

Fixes #293

### DIFF
--- a/src/github.com/matrix-org/dendrite/syncapi/sync/notifier.go
+++ b/src/github.com/matrix-org/dendrite/syncapi/sync/notifier.go
@@ -100,7 +100,9 @@ func (n *Notifier) OnNewEvent(ev *gomatrixserverlib.Event, userID string, pos ty
 	}
 }
 
-// WaitForEvents blocks until there are new events for this request.
+// WaitForEvents blocks until there are events for this request after sincePos.
+// In particular, it will return immediately if there are already events after
+// sincePos for the request, but otherwise blocks waiting for new events.
 func (n *Notifier) WaitForEvents(req syncRequest, sincePos types.StreamPosition) types.StreamPosition {
 	// Do what synapse does: https://github.com/matrix-org/synapse/blob/v0.20.0/synapse/notifier.py#L298
 	// - Bucket request into a lookup map keyed off a list of joined room IDs and separately a user ID
--- a/src/github.com/matrix-org/dendrite/syncapi/sync/notifier.go
+++ b/src/github.com/matrix-org/dendrite/syncapi/sync/notifier.go
@@ -101,7 +101,7 @@ func (n *Notifier) OnNewEvent(ev *gomatrixserverlib.Event, userID string, pos ty
 }
 
 // WaitForEvents blocks until there are new events for this request.
-func (n *Notifier) WaitForEvents(req syncRequest) types.StreamPosition {
+func (n *Notifier) WaitForEvents(req syncRequest, sincePos types.StreamPosition) types.StreamPosition {
 	// Do what synapse does: https://github.com/matrix-org/synapse/blob/v0.20.0/synapse/notifier.py#L298
 	// - Bucket request into a lookup map keyed off a list of joined room IDs and separately a user ID
 	// - Incoming events wake requests for a matching room ID
@@ -117,7 +117,7 @@ func (n *Notifier) WaitForEvents(req syncRequest) types.StreamPosition {
 	// TODO: We increment the stream position for any event, so it's possible that we return immediately
 	//       with a pos which contains no new events for this user. We should probably re-wait for events
 	//       automatically in this case.
-	if req.since != currentPos {
+	if sincePos != currentPos {
 		n.streamLock.Unlock()
 		return currentPos
 	}
@@ -139,6 +139,11 @@ func (n *Notifier) Load(ctx context.Context, db *storage.SyncServerDatabase) err
 	}
 	n.setUsersJoinedToRooms(roomToUsers)
 	return nil
+}
+
+// CurrentPosition returns the current stream position
+func (n *Notifier) CurrentPosition() types.StreamPosition {
+	return n.currPos
 }
 
 // setUsersJoinedToRooms marks the given users as 'joined' to the given rooms, such that new events from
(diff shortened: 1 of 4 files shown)


src/github.com/matrix-org/dendrite/syncapi/types/types.go
```diff
@@ -74,6 +74,16 @@ func NewResponse(pos StreamPosition) *Response {
...
 
+// IsEmpty returns true if the response is empty, i.e. used to decided whether
+// to return the response immediately to the client or to wait for more data.
+func (r *Response) IsEmpty() bool {
+	return len(r.Rooms.Join) > 0 ||
+		len(r.Rooms.Invite) > 0 ||
+		len(r.Rooms.Leave) > 0 ||
+		len(r.AccountData.Events) > 0 ||
+		len(r.Presence.Events) > 0
+}
```
krombel: function returns `true` if there is data although `false` is expected.

erikjohnston: D'oh

src/github.com/matrix-org/dendrite/syncapi/sync/notifier.go
```diff
@@ -101,7 +101,7 @@ func (n *Notifier) OnNewEvent(ev *gomatrixserverlib.Event, userID string, pos ty
 }
 
 // WaitForEvents blocks until there are new events for this request.
-func (n *Notifier) WaitForEvents(req syncRequest) types.StreamPosition {
+func (n *Notifier) WaitForEvents(req syncRequest, sincePos types.StreamPosition) types.StreamPosition {
```
richvdh: what's `sincePos`? document it please

src/github.com/matrix-org/dendrite/syncapi/sync/requestpool.go
```diff
@@ -62,48 +62,71 @@ func (rp *RequestPool) OnIncomingSyncRequest(req *http.Request, device *authtype
...
+			JSON: syncData,
 		}
-	case res := <-done: // received a response
-		return res
 	}
+
+	// Otherwise, we wait for the notifier to tell us if something *may* have
+	// happened. We loop in case it turns out that nothing did happen.
+
+	timeoutChan := time.After(syncReq.timeout) // case of timeout=0 is handled above
```
richvdh: is it safe to return without reading the time value from this? There isn't going to be some memory leak or something?

src/github.com/matrix-org/dendrite/syncapi/sync/requestpool.go
```diff
@@ -62,48 +62,71 @@ func (rp *RequestPool) OnIncomingSyncRequest(req *http.Request, device *authtype
...
+}
+
+// makeNotifyChannel returns a channel that produces the current stream position
+// when there *may* be something to return to the client. Only produces a single
+// value and then closes the channel.
+func (rp *RequestPool) makeNotifyChannel(syncReq syncRequest, sincePos types.StreamPosition) chan types.StreamPosition {
+	notified := make(chan types.StreamPosition)
+
+	go (func() {
+		currentPos := rp.notifier.WaitForEvents(syncReq, sincePos)
```
richvdh: do we not need to arrange for this to be cancelled if we time out (or the request goes away)?

src/github.com/matrix-org/dendrite/syncapi/sync/requestpool.go
erikjohnston: Sigh, I thought the GC would do something useful, since pretty much everywhere seems to suggest doing timeouts like this. But I've found how to do it with explicit clean up.

src/github.com/matrix-org/dendrite/syncapi/sync/requestpool.go
erikjohnston: Woops, was going to add a TODO/issue. This is no worse than before, and is going to be fiddly to pass cancellations into the notifier given how its implemented

erikjohnston: (... so I was going to punt it to a separate PR/round of work)

src/github.com/matrix-org/dendrite/syncapi/sync/requestpool.go
richvdh: "everywhere suggests doing it like X" doesn't mean that X is a sane solution, in my bitter experience :/

src/github.com/matrix-org/dendrite/syncapi/sync/requestpool.go
richvdh: +1 for punting with a todo

richvdh: though: I'm surprised it's that fiddly. Doesn't the request context help you with this?

src/github.com/matrix-org/dendrite/syncapi/sync/requestpool.go
erikjohnston: Yeah, but when its gobyexample.com, golang blog and the golang wiki, I was inclined to trust them :/

src/github.com/matrix-org/dendrite/syncapi/sync/requestpool.go
erikjohnston: Ideally yes, the context should help, but I think we're using sufficiently primitives that don't work nicely with contexts (condition variables and mutexes). My hunch is that there is a saner way of doing it, but I haven't dug into it basically.

PR #307, opened 2017-10-15
Implement /logout/all

### LINKED ISSUE #282: Implement /logout/all
This should be a simple extension of the `/logout` API, but that logs out all devices rather than just the current one.

rreuvekamp:
Closes #282


rreuvekamp:
Oops, format to run golint for some reason.

erikjohnston:
Awesome, this looks perfect! Thanks!

> Maybe shouldn't have made a new route for it, but just made Logout take an optional "/all"....

I think this is one of those cases where trying to combine the two code paths will make it more complicated, so I'm happy with it like this.

Issue #280, opened 2017-10-05
Basic GET /devices endpoint


erikjohnston:
Add `GET /devices` and `GET /device/{deviceID}` endpoint.

The response for `/devices` should look something like:

```
{ "devices": [
    { "device_id": ..., }.
    { "device_id": ..., }.
] }
```

In the future we'll also want to add other information, such as device display name, last seen, etc, but that can be added at a later time

PR #308, opened 2017-10-15
Add GET /device/{deviceID}

ptman:
First attempt at #280 

How do I write tests?
What should I copyright to? Me? The Matrix foundation (or what is it called)?

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/auth/storage/devices/devices_table.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/auth/storage/devices/devices_table.go
@@ -154,14 +154,14 @@ func (s *devicesStatements) selectDeviceByToken(
 }
 
 func (s *devicesStatements) selectDeviceByID(
-	ctx context.Context, localpart, device_id string,
+	ctx context.Context, localpart, deviceID string,
 ) (*authtypes.Device, error) {
 	var dev authtypes.Device
 	var created int64
 	stmt := s.selectDeviceByIDStmt
-	err := stmt.QueryRowContext(ctx, localpart, device_id).Scan(&created)
+	err := stmt.QueryRowContext(ctx, localpart, deviceID).Scan(&created)
 	if err == nil {
-		dev.ID = device_id
+		dev.ID = deviceID
 		dev.UserID = makeUserID(localpart, s.serverName)
 	}
 	return &dev, err
--- a/src/github.com/matrix-org/dendrite/clientapi/auth/storage/devices/devices_table.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/auth/storage/devices/devices_table.go
@@ -54,6 +54,12 @@ const insertDeviceSQL = "" +
 const selectDeviceByTokenSQL = "" +
 	"SELECT device_id, localpart FROM device_devices WHERE access_token = $1"
 
+const selectDeviceByIDSQL = "" +
+	"SELECT created_ts FROM device_devices WHERE localpart = $1 and device_id = $2"
+
+const selectDevicesByLocalpartSQL = "" +
+	"SELECT device_id FROM device_devices WHERE localpart = $1"
+
 const deleteDeviceSQL = "" +
 	"DELETE FROM device_devices WHERE device_id = $1 AND localpart = $2"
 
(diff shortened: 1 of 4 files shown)


src/github.com/matrix-org/dendrite/clientapi/auth/storage/devices/storage.go
```diff
@@ -52,6 +52,21 @@ func (d *Database) GetDeviceByAccessToken(
 	return d.devices.selectDeviceByToken(ctx, token)
 }
 
+// GetDeviceByID returns the device matching the given ID.
+// Returns sql.ErrNoRows if no matching device was found.
+func (d *Database) GetDeviceByID(
+	ctx context.Context, id string,
+) (*authtypes.Device, error) {
```
erikjohnston: (Add localpart and rename id -> device_id as per previous comment)

src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
```diff
@@ -40,6 +40,8 @@ var deviceIDByteLength = 6
 type DeviceDatabase interface {
 	// Look up the device matching the given access token.
 	GetDeviceByAccessToken(ctx context.Context, token string) (*authtypes.Device, error)
+	// Look up the device matching the given id.
+	GetDeviceByID(ctx context.Context, id string) (*authtypes.Device, error)
```
erikjohnston: I don't think we need it in this interface.

src/github.com/matrix-org/dendrite/clientapi/auth/storage/devices/devices_table.go
```diff
@@ -140,6 +153,44 @@ func (s *devicesStatements) selectDeviceByToken(
 	return &dev, err
 }
 
+func (s *devicesStatements) selectDeviceByID(
+	ctx context.Context, id string,
```
erikjohnston: I think device IDs are scoped to a user, so this should probably take a localpart as an argument.

I'd also rename `id` to `device_id` just to make it a bit more obvious which sort of ID you're talking about

src/github.com/matrix-org/dendrite/clientapi/routing/device.go
```diff
@@ -0,0 +1,91 @@
...
+// GetDeviceByID handles /device/{deviceID}
+func GetDeviceByID(
+	req *http.Request, deviceDB *devices.Database, id string,
+) util.JSONResponse {
+	ctx := req.Context()
+	device, err := deviceDB.GetDeviceByID(ctx, id)
+	if err == sql.ErrNoRows {
+		return util.JSONResponse{
+			Code: 404,
+			JSON: jsonerror.NotFound("Not Found"),
```
erikjohnston: Something like `"Unknown device"` might be a more helpful comment

erikjohnston:
> How do I write tests?

For the most part we rely on integration tests using [SyTest](https://github.com/matrix-org/sytest), so currently we're only really adding unit tests when there's more complicated logic.

> What should I copyright to? Me? The Matrix foundation (or what is it called)?

The copyright is still yours, so assign it to you :)

src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
ptman: Right, should've figured that since GetDevicesByLocalpart works nicely without being included.

ptman:
Thanks for the feedback. I squashed to wipe the copied copyright (funny, that). Not sure if you normally prefer to squash or not.

src/github.com/matrix-org/dendrite/clientapi/auth/storage/devices/devices_table.go
```diff
@@ -140,6 +153,44 @@ func (s *devicesStatements) selectDeviceByToken(
 	return &dev, err
 }
 
+func (s *devicesStatements) selectDeviceByID(
+	ctx context.Context, localpart, device_id string,
```
erikjohnston: Sorry, but I'm a crank and we use `deviceID` rather than `device_id`... (I've been writing too much python).

This will fix the lints:

>  don't use underscores in Go names; method parameter device_id should be deviceID (golint)

erikjohnston:
We squash on merge, so we generally don't bother squashing when updating PRs.

erikjohnston:
Woo! This is great, thanks!

Closes #280

ptman:
I noticed I can't see this merged to master yet. Am I missing something?

erikjohnston:
Did I press close instead of merge? :confused:

erikjohnston:
Yes, yes I did. I'm an idiot. I don't seem to be able to reopen the PR either

PR #311, opened 2017-10-17
Refactor Notifier to return channel

erikjohnston:
This has two benefits:

1. Using channels makes it easier to time out while waiting
2. Allows us to clean up goroutines that were waiting if we timeout the
   request

### DIFF
--- a/src/github.com/matrix-org/dendrite/syncapi/sync/requestpool.go
+++ b/src/github.com/matrix-org/dendrite/syncapi/sync/requestpool.go
@@ -19,7 +19,6 @@ import (
 	"net/http"
 	"time"
 
-	log "github.com/sirupsen/logrus"
 	"github.com/matrix-org/dendrite/clientapi/auth/authtypes"
 	"github.com/matrix-org/dendrite/clientapi/auth/storage/accounts"
 	"github.com/matrix-org/dendrite/clientapi/httputil"
@@ -28,6 +27,7 @@ import (
 	"github.com/matrix-org/dendrite/syncapi/types"
 	"github.com/matrix-org/gomatrixserverlib"
 	"github.com/matrix-org/util"
+	log "github.com/sirupsen/logrus"
 )
 
 // RequestPool manages HTTP long-poll connections for /sync
@@ -85,7 +85,7 @@ func (rp *RequestPool) OnIncomingSyncRequest(req *http.Request, device *authtype
 	for {
 		select {
 		// Wait for notifier to wake us up
-		case currPos = <-rp.makeNotifyChannel(*syncReq, currPos):
+		case currPos = <-rp.notifier.WaitForEvents(*syncReq, currPos):
 		// Or for timeout to expire
 		case <-timer.C:
 			return util.JSONResponse{
@@ -114,24 +114,6 @@ func (rp *RequestPool) OnIncomingSyncRequest(req *http.Request, device *authtype
 		}
 
 	}
-}
-
-// makeNotifyChannel returns a channel that produces the current stream position
-// when there *may* be something to return to the client. Only produces a single
-// value and then closes the channel.
-func (rp *RequestPool) makeNotifyChannel(syncReq syncRequest, sincePos types.StreamPosition) chan types.StreamPosition {
-	notified := make(chan types.StreamPosition)
-
-	// TODO(#303): We need to ensure that WaitForEvents gets properly cancelled
-	// when the request is finished, or use some other mechanism to ensure we
-	// don't leak goroutines here
-	go (func() {
-		currentPos := rp.notifier.WaitForEvents(syncReq, sincePos)
-		notified <- currentPos
-		close(notified)
-	})()
-
-	return notified
 }
 
 type stateEventInStateResp struct {
--- a/src/github.com/matrix-org/dendrite/syncapi/sync/requestpool.go
+++ b/src/github.com/matrix-org/dendrite/syncapi/sync/requestpool.go
@@ -85,7 +85,7 @@ func (rp *RequestPool) OnIncomingSyncRequest(req *http.Request, device *authtype
 	for {
 		select {
 		// Wait for notifier to wake us up
-		case currPos = <-rp.notifier.WaitForEvents(*syncReq, currPos):
+		case currPos = <-rp.notifier.GetNotifyChannel(*syncReq, currPos):
 		// Or for timeout to expire
 		case <-timer.C:
 			return util.JSONResponse{
--- a/src/github.com/matrix-org/dendrite/syncapi/sync/requestpool.go
+++ b/src/github.com/matrix-org/dendrite/syncapi/sync/requestpool.go
@@ -82,14 +82,14 @@ func (rp *RequestPool) OnIncomingSyncRequest(req *http.Request, device *authtype
 	timer := time.NewTimer(syncReq.timeout) // case of timeout=0 is handled above
 	defer timer.Stop()
 
-	userStream := rp.notifier.GetListener(*syncReq)
-	defer userStream.Close()
+	userStreamListener := rp.notifier.GetListener(*syncReq)
+	defer userStreamListener.Close()
 
 	for {
 		select {
 		// Wait for notifier to wake us up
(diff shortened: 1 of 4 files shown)


src/github.com/matrix-org/dendrite/syncapi/sync/userstream.go
```diff
@@ -25,55 +27,108 @@ import (
...
+		timeOfLastChannel: time.Now(),
+		pos:               currPos,
 	}
 }
 
 // Wait blocks until there is a new stream position for this user, which is then returned.
 // waitAtPos should be the position the stream thinks it should be waiting at.
-func (s *UserStream) Wait(waitAtPos types.StreamPosition) (pos types.StreamPosition) {
-	s.cond.L.Lock()
+func (s *UserStream) Wait(ctx context.Context, waitAtPos types.StreamPosition) <-chan types.StreamPosition {
```
erikjohnston: Woops, should update name and docstring

src/github.com/matrix-org/dendrite/syncapi/sync/userstream.go
```diff
@@ -25,55 +27,108 @@ import (
...
+	go func() {
+		<-ctx.Done()
+
+		s.lock.Lock()
+		defer s.lock.Unlock()
+
+		// Icky but efficient way of filtering out the given channel
+		for idx, ch := range s.waitingChannels {
+			if posChannel == ch {
+				lastIdx := len(s.waitingChannels)
```
richvdh: needs a -1, surely? I think this is why the tests are failing.

src/github.com/matrix-org/dendrite/syncapi/sync/userstream.go
```diff
@@ -25,55 +27,108 @@ import (
...
-	s.numWaiting--
-	s.cond.L.Unlock()
-	return
+
+	s.waitingChannels = append(s.waitingChannels, posChannel)
+
+	// We spawn off a goroutine that waits for the request to finish and removes the
+	// channel from waitingChannels
+	go func() {
+		<-ctx.Done()
```
richvdh: I don't suppose there's any elegant way to cancel this goroutine when the wait completes, so that we don't several piling up if there are are repeated calls in a single request?

richvdh: I guess another channel which is (just) closed when the wait completes, which you can select on here, would do. You might have to make `waitingChannels` a list of `(chan, chan)` tuples. Not sure if it's worth the additional complexity. WDYT?

src/github.com/matrix-org/dendrite/syncapi/sync/userstream.go
```diff
@@ -25,55 +27,108 @@ import (
...
+
+		s.lock.Lock()
+		defer s.lock.Unlock()
+
+		// Icky but efficient way of filtering out the given channel
+		for idx, ch := range s.waitingChannels {
```
richvdh: can we be sure that this is going to be a short list? I'm a bit concerned about having to iterate it looking for a request which probably isn't there any more.

src/github.com/matrix-org/dendrite/syncapi/sync/userstream.go
```diff
@@ -25,55 +27,108 @@ import (
...
 	}
 }
 
-// Wait blocks until there is a new stream position for this user, which is then returned.
-// waitAtPos should be the position the stream thinks it should be waiting at.
-func (s *UserStream) Wait(waitAtPos types.StreamPosition) (pos types.StreamPosition) {
-	s.cond.L.Lock()
+// Wait returns a channel that produces a single stream position when
+// a new event *may* be available to return to the client.
+func (s *UserStream) Wait(ctx context.Context, waitAtPos types.StreamPosition) <-chan types.StreamPosition {
```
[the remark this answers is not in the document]
erikjohnston: I've gone for `GetNotifyChannel`, which seems a bit better?

src/github.com/matrix-org/dendrite/syncapi/sync/userstream.go
erikjohnston: It should equate to the number of active syncs for a user, which should be small enough that iterating should be fast.

src/github.com/matrix-org/dendrite/syncapi/sync/userstream.go
erikjohnston: Why doesn't golang have builtin functions to do this sort of thing for you :'(

src/github.com/matrix-org/dendrite/syncapi/sync/userstream.go
erikjohnston: I guess we could actually have it so that the channels are kept alive so that a single sync requests only has one notification channel it reuses, and then we can close that channel when `ctx.Done()` returns. Still means an extra go routing per sync call, but is probably fine?

(Why are go channels so awful?)

src/github.com/matrix-org/dendrite/syncapi/sync/userstream.go
```diff
@@ -25,55 +27,108 @@ import (
...
-		UserID: userID,
-		cond:   sync.NewCond(&sync.Mutex{}),
+		UserID:            userID,
+		timeOfLastChannel: time.Now(),
+		pos:               currPos,
 	}
 }
 
-// Wait blocks until there is a new stream position for this user, which is then returned.
-// waitAtPos should be the position the stream thinks it should be waiting at.
```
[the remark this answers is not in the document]
erikjohnston: (Added it back)

erikjohnston:
Ugh, wait, go channels will block the sender if nothing is read, which doesn't seem ideal in `Broadcast`

erikjohnston:
@richvdh sorry, I've changed everything again but hopefully this will *actually* work

src/github.com/matrix-org/dendrite/syncapi/sync/userstream.go
```diff
@@ -15,65 +15,143 @@
...
-	numWaiting int
+	// The last time when we had some listeners waiting
+	timeOfLastChannel time.Time
+	// The number of listeners waiting
+	numWaiting uint
+}
+
+// UserStreamListener allows a sync request to wait for updates for a user.
+type UserStreamListener struct {
+	*UserStream
```
richvdh: could we make this an explicit field? I find the implicit references to the UserStream members confusing.

src/github.com/matrix-org/dendrite/syncapi/sync/requestpool.go
```diff
@@ -82,10 +82,14 @@ func (rp *RequestPool) OnIncomingSyncRequest(req *http.Request, device *authtype
 	timer := time.NewTimer(syncReq.timeout) // case of timeout=0 is handled above
 	defer timer.Stop()
 
+	userStream := rp.notifier.GetListener(*syncReq)
```
richvdh: this is a `UserStreamListener` - please can we name it accordingly?

richvdh:
Modulo some quibbling this seems quite nice and elegant now. Even if it does seem to require more hacking around with channels than you might have hoped for :/.

src/github.com/matrix-org/dendrite/syncapi/sync/userstream.go
```diff
@@ -120,37 +120,40 @@ func (s *UserStream) TimeOfLastNonEmpty() time.Time {
...
+	s.userStream.lock.Lock()
+	defer s.userStream.lock.Unlock()
 
-	return s.pos
+	return s.userStream.pos
 }
 
 // GetNotifyChannel returns a channel that is closed when there may be an
 // update for the user.
 // sincePos specifies from which point we want to be notified about
```
richvdh: right, thanks for adding the comment at line 138; this makes more sense now. Please could you also add something about it here - "if we're already past that point in the stream, we'll return a closed channel immediately. otherwise we'll return a channel which will be closed once there is an update". Only with better words and formatting

richvdh:
(at last, sorry this has been an epic)

erikjohnston:
Thanks for doing it!

PR #312, opened 2017-10-17
Don't add empty room entries to sync responses

erikjohnston:


erikjohnston:
Its tested by sytest, though we should start looking more at unit tests for dendrite. This is in the storage layer so would need to rejig things to be able to test this, so won't do it as part of this PR

PR #316, opened 2017-10-26
Implement /make_join and /send_join

erikjohnston:
This PR is mainly just to track whats happening on this front. It hasn't been tested as sytest doesn't like dendrite federation atm, and we don't have the alias lookup APIs to test it using Riot, so this is blocked for now.

(This will probably end up being split into separate PRs)

### DIFF
--- a/src/github.com/matrix-org/dendrite/federationapi/routing/join.go
+++ b/src/github.com/matrix-org/dendrite/federationapi/routing/join.go
@@ -90,7 +90,7 @@ func MakeJoin(
 
 	return util.JSONResponse{
 		Code: 200,
-		JSON: event,
+		JSON: map[string]interface{}{"event": builder},
 	}
 }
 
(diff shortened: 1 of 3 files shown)


src/github.com/matrix-org/dendrite/federationapi/routing/join.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+	}
+
+	// Check that the event is signed by the server sending the request.
+	verifyRequests := []gomatrixserverlib.VerifyJSONRequest{{
+		ServerName: event.Origin(),
+		Message:    event.Redact().JSON(),
+		AtTS:       event.OriginServerTS(),
+	}}
+	verifyResults, err := keys.VerifyJSONs(httpReq.Context(), verifyRequests)
```
krombel: Why do you use `httpReq.Context()` instead of `ctx` here? I know it is the same internally but a little bit confusing :smirk:

erikjohnston: Good catch!

erikjohnston:
This is now rebased on top of https://github.com/matrix-org/dendrite/pull/352

PR #317, opened 2017-10-31
Federation: Implement Query Profile API

### LINKED ISSUE #278: Federation: Implement Query profile API
Implement the server portion of: `GET /_matrix/federation/v1/query/profile?user_id=...&field=...`

See e.g. https://github.com/matrix-org/dendrite/blob/bcf58fad84bf53b6a6cf13a8d11bd46d82cc3bab/src/github.com/matrix-org/dendrite/clientapi/readers/profile.go#L170-L189 for how to query the room server for profile info

anoadragon453:
Implements the server portion of: `GET
/_matrix/federation/v1/query/profile?user_id=...&field=...`

Closes #278

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/jsonerror/jsonerror.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/jsonerror/jsonerror.go
@@ -73,10 +73,10 @@ func MissingArgument(msg string) *MatrixError {
 	return &MatrixError{"M_MISSING_ARGUMENT", msg}
 }
 
-// InvalidArgumentBody is an error when the client tries to provide an
-// invalid body under a valid argument
-func InvalidArgumentBody(msg string) *MatrixError {
-	return &MatrixError{"M_INVALID_ARGUMENT_BODY", msg}
+// InvalidArgumentValue is an error when the client tries to provide an
+// invalid value for a valid argument
+func InvalidArgumentValue(msg string) *MatrixError {
+	return &MatrixError{"M_INVALID_ARGUMENT_VALUE", msg}
 }
 
 // MissingToken is an error when the client tries to access a resource which
--- a/src/github.com/matrix-org/dendrite/clientapi/jsonerror/jsonerror.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/jsonerror/jsonerror.go
@@ -67,6 +67,18 @@ func NotFound(msg string) *MatrixError {
 	return &MatrixError{"M_NOT_FOUND", msg}
 }
 
+// MissingArgument is an error when the client tries to access a resource
+// without providing an argument that is required.
+func MissingArgument(msg string) *MatrixError {
+	return &MatrixError{"M_MISSING_ARGUMENT", msg}
+}
+
+// InvalidArgumentBody is an error when the client tries to provide an
+// invalid body under a valid argument
+func InvalidArgumentBody(msg string) *MatrixError {
+	return &MatrixError{"M_INVALID_ARGUMENT_BODY", msg}
+}
+
 // MissingToken is an error when the client tries to access a resource which
 // requires authentication without supplying credentials.
 func MissingToken(msg string) *MatrixError {
--- a/src/github.com/matrix-org/dendrite/federationapi/routing/profile.go
+++ b/src/github.com/matrix-org/dendrite/federationapi/routing/profile.go
@@ -1,0 +1,93 @@
+// Copyright 2017 New Vector Ltd
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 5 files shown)


src/github.com/matrix-org/dendrite/federationapi/routing/profile.go
```diff
@@ -0,0 +1,93 @@
...
+	AvatarURL   string `json:"avatar_url"`
+	DisplayName string `json:"displayname"`
+}
+
+type avatarURL struct {
+	AvatarURL string `json:"avatar_url"`
+}
+
+type displayName struct {
+	DisplayName string `json:"displayname"`
```
babolivier: These three structures look like duplication from https://github.com/matrix-org/dendrite/blob/master/src/github.com/matrix-org/dendrite/clientapi/routing/profile.go#L34-L45 we could easily avoid. I would have exported the ones in the client API (by capitilising the first letter of their name and updating references to them in `GetProfile`, `GetAvatarURL` and `GetDisplayName`, and also doc'ing them in comments otherwise linters will go crazy) to use them here instead.

src/github.com/matrix-org/dendrite/federationapi/routing/profile.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// GetProfile implements /_matrix/federation/v1/query/profile
```
babolivier: It might be interesting to also specify the HTTP method associated with this endpoint.

src/github.com/matrix-org/dendrite/federationapi/routing/profile.go
```diff
@@ -0,0 +1,93 @@
...
+			res = displayName{
+				profile.DisplayName,
+			}
+		case "avatar_url":
+			res = avatarURL{
+				profile.AvatarURL,
+			}
+		default:
+			code = 400
+			res = jsonerror.InvalidArgumentBody("The request body did not contain allowed values of argument 'field'. Allowed: 'avatar_url', 'displayname'.")
```
babolivier: This is purely a detail but I feel like this error might be a bit misleading (can be understood as "several/all of the allowed value").

src/github.com/matrix-org/dendrite/clientapi/jsonerror/jsonerror.go
```diff
@@ -67,6 +67,18 @@ func NotFound(msg string) *MatrixError {
...
 }
 
+// MissingArgument is an error when the client tries to access a resource
+// without providing an argument that is required.
+func MissingArgument(msg string) *MatrixError {
+	return &MatrixError{"M_MISSING_ARGUMENT", msg}
+}
+
+// InvalidArgumentBody is an error when the client tries to provide an
+// invalid body under a valid argument
```
babolivier: Again, that's more a detail, I'm not sure of the phrasing ("body" makes me think of the whole body of a request). fwiw I'd write it as "to provide an invalid value for a valid argument".

anoadragon453:
@babolivier Should be fixed up now, thanks!

src/github.com/matrix-org/dendrite/federationapi/routing/profile.go
```diff
@@ -0,0 +1,81 @@
...
+
+package routing
+
+import (
+	"net/http"
+
+	"github.com/matrix-org/dendrite/clientapi/auth/storage/accounts"
+	"github.com/matrix-org/dendrite/clientapi/httputil"
+	"github.com/matrix-org/dendrite/clientapi/jsonerror"
+	clientRouting "github.com/matrix-org/dendrite/clientapi/routing"
```
erikjohnston: It feels weird that we're importing the routing for another component here. I think the best bet would be to move the structs to `src/github.com/matrix-org/dendrite/common/types.go` and then you can import common.

src/github.com/matrix-org/dendrite/federationapi/routing/profile.go
```diff
@@ -0,0 +1,81 @@
...
+
+	// httpReq.FormValue will return an empty string if value is not found
+	if userID == "" {
+		return util.JSONResponse{
+			Code: 400,
+			JSON: jsonerror.MissingArgument("The request body did not contain required argument 'user_id'."),
+		}
+	}
+
+	localpart, _, err := gomatrixserverlib.SplitID('@', userID)
```
erikjohnston: We probably want to check that the domain of the userID matches our server.

anoadragon453: Should that be done here or in `accountDB.GetProfileByLocalpart`?

anoadragon453: Should be done, check if my provided solution was correct though, not entirely sure myself...

src/github.com/matrix-org/dendrite/federationapi/routing/profile.go
anoadragon453: This was fixed as well.

PR #318, opened 2017-11-05
Check for existing filter before inserting a new one

### LINKED ISSUE #298: Deduplicate identical filters
i.e. porting functionality of https://github.com/matrix-org/synapse/pull/2219 to dendrite.

This can either be done as in the PR by comparing the filter dicts, or by using the hash of a filter as the ID

anoadragon453:
Closes #298

This is my naive solution to the issue. Let me know if it should be done another way and I can switch things around.

We check whether a row with the given filter and user localpart already exists in the database, and just returns that filterID if so.

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/filter_table.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/filter_table.go
@@ -17,6 +17,8 @@ package accounts
 import (
 	"context"
 	"database/sql"
+
+	"github.com/matrix-org/gomatrixserverlib"
 )
 
 const filterSchema = `
@@ -70,7 +72,7 @@ func (s *filterStatements) prepare(db *sql.DB) (err error) {
 func (s *filterStatements) selectFilter(
 	ctx context.Context, localpart string, filterID string,
 ) (filter string, err error) {
-	err = s.selectFilterStmt.QueryRowContext(ctx, localpart, filterID).Scan(&filter)
+	err = s.selectFilterStmt.QueryRowContext(ctx, localpart, filterID).Scan(&filterJSON)
 	return
 }
 
@@ -79,13 +81,21 @@ func (s *filterStatements) insertFilter(
 ) (pos string, err error) {
 	var existingFilter string
 
+	// This can result in a race condition when two clients try to insert the
+	// same filter and localpart at the same time, however this is not a
+	// problem as both calls will result in the same filterID
+	filterJSON, errN := gomatrixserverlib.CanonicalJSON(filter)
+	if err {
+		return "", err
+	}
+
 	// Check if filter already exists in the database
 	err = s.selectFilterByContentStmt.QueryRowContext(ctx,
-		localpart, filter).Scan(&existingFilter)
+		localpart, filterJSON).Scan(&existingFilter)
 	if existingFilter != "" {
 		return existingFilter, err
 	}
 
-	err = s.insertFilterStmt.QueryRowContext(ctx, filter, localpart).Scan(&pos)
+	err = s.insertFilterStmt.QueryRowContext(ctx, filterJSON, localpart).Scan(&pos)
 	return
 }
--- a/src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/filter_table.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/filter_table.go
@@ -38,12 +38,16 @@ CREATE INDEX IF NOT EXISTS account_filter_localpart ON account_filter(localpart)
 const selectFilterSQL = "" +
 	"SELECT filter FROM account_filter WHERE localpart = $1 AND id = $2"
 
+const selectFilterByContentSQL = "" +
+	"SELECT filter FROM account_filter WHERE localpart = $1 AND filter = $2"
+
 const insertFilterSQL = "" +
 	"INSERT INTO account_filter (filter, id, localpart) VALUES ($1, DEFAULT, $2) RETURNING id"
 
 type filterStatements struct {
-	selectFilterStmt *sql.Stmt
-	insertFilterStmt *sql.Stmt
+	selectFilterStmt          *sql.Stmt
+	selectFilterByContentStmt *sql.Stmt
+	insertFilterStmt          *sql.Stmt
 }
 
 func (s *filterStatements) prepare(db *sql.DB) (err error) {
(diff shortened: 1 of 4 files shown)


erikjohnston:
This is a good first cut!

A couple of notes:
1. The same data structure won't necessarily serialise to the same JSON output (e.g. keys could be reordered), if that happens we could still get duplicate filters in the DB. This isn't a problem per se, and is certainly better than we have now. There is a function `gomatrixserverlib.CanonicalJSON` that ensures the same data structure serialises to the same JSON bytes
2. This only deduplicates filters for the same user, whereas actually a lot users will have the same filter. We need to be careful here as although the filter IDs are unique we don't want to let anybody see any filter. We could use the hash of the filter as the filter ID, which would deduplicate automatically and make it difficult/impossible to guess valid IDs, but has the drawback that they would produce very long IDs.

I think it should be easy to add fix the first point, but thoughts welcome on the second

anoadragon453:
>There is a function `gomatrixserverlib.CanonicalJSON` that ensures the same data structure serialises to the same JSON bytes

Ah, very useful! I'll make sure to stick this in.

>We could use the hash of the filter as the filter ID, which would deduplicate automatically and make it difficult/impossible to guess valid IDs, but has the drawback that they would produce very long IDs.

We could still use hashes, but perhaps truncate them just as Signal does for contacts without revealing the full hash and thus a possible link back to the identity. Only question is how much to truncate it by to where it's still enough bits to ensure uniqueness.

erikjohnston:
> Edit: And yes, no need to have the same filter stored multiple times for different users (I assume?). Do we even need to store the user at all? I assume if they have the filter ID then they likely know what filter they're after?

Yup, *if* the filter ID space if sparse, i.e. if we just use an incrementing integer then people can just cycle through all filters ever, which is probably not ideal.

anoadragon453:
@erikjohnston So did we decide on using truncated hashes as the filter ID? We can still store the user so that people requesting a filter that they don't have access to still won't be able to access it even if they do have the correct hash. The only issue with that is that we would have duplicated hashes.

We *could* have two tables: one for filters with a link between filter and filter ID (hash), and another table with usernames and filter IDs. That way you could have many users with the same filter, but only have to store that filter once, while also being able to check whether a user should have access to that filter or not.

anoadragon453:
Also what did you think `gomatrixserverlib.CanonicalJSON(...)` should be wrapped around? Around `filter` on [this line](https://github.com/anoadragon453/dendrite/blob/5a09d5123d7b304556b9170557be9dea70b17653/src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/filter_table.go#L89)?

erikjohnston:
To be honest i think it sounds like using hashes is a bit over kill for now.

> Also what did you think gomatrixserverlib.CanonicalJSON(...) should be wrapped around? Around filter on this line?

Looks plausible, you'd need to also canonicalise when checking if there is an existing filter.

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/filter_table.go
```diff
@@ -70,6 +77,15 @@ func (s *filterStatements) selectFilter(
...
 ) (pos string, err error) {
+	var existingFilter string
+
+	// Check if filter already exists in the database
+	err = s.selectFilterByContentStmt.QueryRowContext(ctx,
+		localpart, filter).Scan(&existingFilter)
+	if existingFilter != "" {
+		return existingFilter, err
+	}
+
```
erikjohnston: We probably want to add a comment to say that there is a race here if two clients try to insert the same filter at the same time, which is fine.

anoadragon453:
@erikjohnston Finished the above changes. So want to stick with the current checking method or play around with filter IDs some more?

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/filter_table.go
```diff
@@ -62,14 +71,31 @@ func (s *filterStatements) prepare(db *sql.DB) (err error) {
...
 	return
 }
 
 func (s *filterStatements) insertFilter(
-	ctx context.Context, filter string, localpart string,
-) (pos string, err error) {
-	err = s.insertFilterStmt.QueryRowContext(ctx, filter, localpart).Scan(&pos)
+	ctx context.Context, filter []byte, localpart string,
+) (pos []byte, err error) {
+	var existingFilter []byte
```
erikjohnston: Hang on, are we returning the filter ID or the actual filter? Might be worth documenting the return type of this function

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/storage.go
```diff
@@ -324,15 +324,15 @@ func (d *Database) GetThreePIDsForLocalpart(
 // Returns an error if no such filter exists or if there was an error taling to the database.
 func (d *Database) GetFilter(
 	ctx context.Context, localpart string, filterID string,
-) (string, error) {
+) ([]byte, error) {
 	return d.filter.selectFilter(ctx, localpart, filterID)
 }
 
 // PutFilter puts the passed filter into the database.
 // Returns an error if something goes wrong.
```
erikjohnston: Update docs about what it returns

src/github.com/matrix-org/dendrite/clientapi/routing/filter.go
```diff
@@ -111,13 +111,13 @@ func PutFilter(
...
-	filterID, err := accountDB.PutFilter(req.Context(), localpart, string(filterArray))
+	filterID, err := accountDB.PutFilter(req.Context(), localpart, filterArray)
 	if err != nil {
 		return httputil.LogThenError(req, err)
 	}
 
 	return util.JSONResponse{
 		Code: 200,
-		JSON: filterResponse{FilterID: filterID},
+		JSON: filterResponse{FilterID: string(filterID)},
```
erikjohnston: This looks to be due to confusion between if PutFilter returns the filter json or the filterID

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/filter_table.go
anoadragon453: Ah, your correct, we should be returning the ID from this method, not the filter itself. Good catch!

erikjohnston:
\o/  Sorry about the amount of back and forth

PR #322, opened 2017-11-07
Split travis into multiple jobs

erikjohnston:
The motivation for this is to make it easier to see whether a travis failure is due to linting, unit tests or integration test failures, without having to look in the logs.

It also means that each job is independent, so if e.g. the linting fails then the unit tests will still be run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,8 +27,8 @@ install:
 
 # Generate a self-signed X.509 certificate for TLS.
 before_script:
- - if [[ "$TEST_SUITE" -eq "integ-test" ]]; then openssl req -x509 -newkey rsa:512 -keyout server.key -out server.crt -days 365 -nodes -subj /CN=localhost; fi
- - if [[ "$TEST_SUITE" -eq "integ-test" ]]; then ./scripts/install-local-kafka.sh; fi
+ - if [[ "$TEST_SUITE" == "integ-test" ]]; then openssl req -x509 -newkey rsa:512 -keyout server.key -out server.crt -days 365 -nodes -subj /CN=localhost; fi
+ - if [[ "$TEST_SUITE" == "integ-test" ]]; then ./scripts/install-local-kafka.sh; fi
 
 script:
  - ./scripts/travis-test.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,13 +8,6 @@ env:
 - TEST_SUITE="unit-test"
 - TEST_SUITE="integ-test"
 
-matrix:
-  include:
-    - env: TEST_SUITE="integ-test"
-      apt:
-        packages:
-          - openssl
-
 sudo: false
 
 # Use trusty for postgres 9.5 support
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,8 +24,7 @@ install:
 
 # Generate a self-signed X.509 certificate for TLS.
 before_script:
- - if [[ "$TEST_SUITE" == "integ-test" ]]; then openssl req -x509 -newkey rsa:512 -keyout server.key -out server.crt -days 365 -nodes -subj /CN=localhost; fi
- - if [[ "$TEST_SUITE" == "integ-test" ]]; then ./scripts/install-local-kafka.sh; fi
+ - if [[ "${TEST_SUITE:-integ-test}" == "integ-test" ]]; then ./scripts/install-local-kafka.sh; fi
 
 script:
  - ./scripts/travis-test.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,10 +22,6 @@ services:
 install:
  - go get github.com/constabulary/gb/...
 
-# Generate a self-signed X.509 certificate for TLS.
-before_script:
- - if [[ "${TEST_SUITE:-integ-test}" == "integ-test" ]]; then ./scripts/install-local-kafka.sh; fi
-
 script:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -21,10 +24,10 @@ install:
 
 # Generate a self-signed X.509 certificate for TLS.
 before_script:
- - openssl req -x509 -newkey rsa:4096 -keyout server.key -out server.crt -days 365 -nodes -subj /CN=localhost
+ - if [[ "$TEST_SUITE" == "integ-test" ]]; then openssl req -x509 -newkey rsa:512 -keyout server.key -out server.crt -days 365 -nodes -subj /CN=localhost; fi
```
richvdh: is there a reason not to do these in `travis-test.sh`?

scripts/travis-test.sh
```diff
@@ -8,20 +8,25 @@ set -eu
...
-gb build github.com/matrix-org/dendrite/cmd/dendrite-room-server
-gb build github.com/matrix-org/dendrite/cmd/roomserver-integration-tests
-gb build github.com/matrix-org/dendrite/cmd/dendrite-sync-api-server
-gb build github.com/matrix-org/dendrite/cmd/syncserver-integration-tests
-gb build github.com/matrix-org/dendrite/cmd/create-account
-gb build github.com/matrix-org/dendrite/cmd/dendrite-media-api-server
-gb build github.com/matrix-org/dendrite/cmd/mediaapi-integration-tests
-gb build github.com/matrix-org/dendrite/cmd/client-api-proxy
+# We don't bother to build if we're only checking the linting
+if [ "${TEST_SUITE-}" != "lint" ]; then
```
richvdh: why `"${TEST_SUITE-}"` rather than `"${TEST_SUITE}"` ?

scripts/travis-test.sh
```diff
@@ -8,20 +8,25 @@ set -eu
...
+fi
 
 # Run unit tests and linters
 ./scripts/build-test-lint.sh
 
-# Run the integration tests
-bin/roomserver-integration-tests
-bin/syncserver-integration-tests
-bin/mediaapi-integration-tests
+if [ "${TEST_SUITE-unit-test}" == "integ-test" ]; then
```
richvdh: why `"${TEST_SUITE-unit-test}"` rather than `"${TEST_SUITE}"` ?

scripts/travis-test.sh
```diff
@@ -8,20 +8,25 @@ set -eu
...
+    gb build github.com/matrix-org/dendrite/cmd/dendrite-sync-api-server
+    gb build github.com/matrix-org/dendrite/cmd/syncserver-integration-tests
+    gb build github.com/matrix-org/dendrite/cmd/create-account
+    gb build github.com/matrix-org/dendrite/cmd/dendrite-media-api-server
+    gb build github.com/matrix-org/dendrite/cmd/mediaapi-integration-tests
+    gb build github.com/matrix-org/dendrite/cmd/client-api-proxy
+fi
 
 # Run unit tests and linters
 ./scripts/build-test-lint.sh
```
richvdh: it feels odd that this ends up reading the TEST_SUITE var. it might be clearer just to inline bits of it here.

scripts/travis-test.sh
```diff
@@ -8,20 +8,25 @@ set -eu
 export GOGC=400
```
richvdh: I think it might be nice if you documented the TEST_SUITE env var (its potential settings) in the comment at the top.

scripts/travis-test.sh
erikjohnston: Its to correctly handle unset variables without bash being sad (since we have set -eu)

scripts/travis-test.sh
erikjohnston: (Done)

.travis.yml
erikjohnston: We can move the openssl, but I find it nice to install the dependencies before we start the tests. It isn't the best reason in the world.

richvdh: I feel like I'd rather have all the magic in travis-test rather than split between that script and this config file.

scripts/find-lint.sh
```diff
@@ -31,7 +31,7 @@ then args="$args --enable-gc"
 fi
 
 echo "Installing lint search engine..."
-go install github.com/alecthomas/gometalinter/
+gb build github.com/alecthomas/gometalinter/
```
richvdh: apparently this installs gometalinter into `bin` rather than `vendor/bin`; I'm not sure if that is necessarily a bad thing but it certainly surprised me.

erikjohnston: Given nothing else gets built into `vendor/bin`, i think it makes more sense like that?

richvdh: I think you're probably right, but a couple of thoughts:

1: It's a bit of a gotcha as you can end up with different versions in `vendor/bin` and `bin`, which contributed to my hell yesterday until I figured out what was going on.

2. It's a shame that `bin` fills up with all sorts of crap rather than just our own build products. If I want to build dendrite and tar it up to transfer to another machine, then I have to pick and choose what I want from `bin` rather than just taking the whole dir. It would be nice to separate our own build products from the lint scripts. I'm not sure I have a good way of doing that, but I feel like perhaps this is a backwards step.

erikjohnston:
I've added an issue to track that, though I think you're right.

Issue #279, opened 2017-10-05
Add /voip/turnServer endpoint


erikjohnston:
See: https://github.com/matrix-org/synapse/blob/6c1bb1601e43c89637ae5bd8720c255646ca8141/synapse/rest/client/v1/voip.py#L26-L64

This involves adding some config options and doing some HMACing

anoadragon453:
Will take this if nobody else wants it.

t3chguy:
This has now been merged

PR #323, opened 2017-11-07
implement voip/turnServer API endpoint

t3chguy:
Implements #279

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/routing/voip.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/routing/voip.go
@@ -1,0 +1,85 @@
+// Copyright 2017 Michael Telatysnki <7t3chguy@gmail.com>
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package routing
+
+import (
+	"net/http"
+
+	"crypto/hmac"
+	"crypto/sha1"
+	"encoding/base64"
+	"fmt"
+	"time"
+
+	"github.com/matrix-org/dendrite/clientapi/auth/authtypes"
+	"github.com/matrix-org/dendrite/common/config"
+	"github.com/matrix-org/util"
+)
+
+type turnServerResponse struct {
+	Username string `json:"username"`
+	Password string `json:"password"`
+	URIs []string `json:"uris"`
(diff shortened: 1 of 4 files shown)


src/github.com/matrix-org/dendrite/clientapi/routing/voip.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	duration, err := time.ParseDuration(turnConfig.UserLifetime)
+	if err != nil {
+		util.GetLogger(req.Context()).WithError(err).Warn("Invalid configuration value turn.turn_user_lifetime")
+		return util.JSONResponse{
+			Code: 200,
+			JSON: struct{}{},
+		}
+	}
```
erikjohnston: I would move this into `common/config.go` so that we fail at start up

src/github.com/matrix-org/dendrite/clientapi/routing/voip.go
```diff
@@ -0,0 +1,85 @@
...
+	Password string `json:"password"`
+	URIs []string `json:"uris"`
+	TTL int `json:"ttl"`
+}
+
+func RequestTurnServer(req *http.Request, device *authtypes.Device, cfg config.Dendrite) util.JSONResponse {
+	turnConfig := cfg.TURN
+
+	// TODO Guest Support
+	if len(turnConfig.URIs) == 0 /* || (isGuest && !turnConfig.AllowGuests) */ {
```
erikjohnston: I'd probably remove all the guest stuff, we can add it back in if/when we actually implement guests.

src/github.com/matrix-org/dendrite/clientapi/routing/voip.go
```diff
@@ -0,0 +1,85 @@
...
+
+	if turnConfig.SharedSecret != "" {
+		expiry := time.Now().Add(duration).Unix()
+		resp.Username = fmt.Sprintf("%d:%s", expiry, device.UserID)
+
+		mac := hmac.New(sha1.New, []byte(turnConfig.SharedSecret))
+		mac.Write([]byte(resp.Username))
+		resp.Password = base64.StdEncoding.EncodeToString(mac.Sum(nil))
+	} else if turnConfig.Username != "" && turnConfig.Password != "" {
+		// Already have turnConfig.Username and turnConfig.Password in resp
```
erikjohnston: I think I would prefer to write this as:

```golang
resp := turnServerResponse {
    URIs: turnConfig.URIs,
    TTL: int(duration.Seconds()),
}

if turnConfig.SharedSecret != "" {
    ...
} else if turnConfig.Username != "" && turnConfig.Password != "" {
    resp.Username = turnConfig.Username
    resp.Password = turnConfig.Password
}

This makes it clearer that `turnConfig.Username` etc is optional

**Comment by erikjohnston** [comment]
The linting is failing with:

```
clientapi/routing/voip.go:70:12:warning: error return value not checked (mac.Write([]byte(resp.Username))) (errcheck)
clientapi/routing/voip.go:38:1:warning: exported function RequestTurnServer should have comment or be unexported (golint)
```

erikjohnston:
Woo, thanks!

PR #328, opened 2017-11-12
Disallow capital letters in usernames

### LINKED ISSUE #325: In registration and login APIs replace any capitals with lower case

anoadragon453:
Fixes #325

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/routing/register.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/routing/register.go
@@ -28,7 +28,6 @@ import (
 
 	"github.com/matrix-org/dendrite/common/config"
 
-	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


ara4n:
We just did this in synapse (https://github.com/matrix-org/synapse/pull/2561) and it caused havoc, as it turns out that many clients (e.g. Riot/iOS & Riot/Android) handle the UX of rejected usernames very badly - and in general relying on server-side validation is always going to provide bad UX.  So instead we've literally just shifted to squashing the case on registration & login instead: https://github.com/matrix-org/synapse/pull/2662.  This looks to be the better approach - sorry for the confusion :/

anoadragon453:
@ara4n No problem! As you can see from the commit it was a very simple change.

Presumably this pull and the related issue should thus then be closed :)

erikjohnston:
Yeah, have updated the issue, thanks!

PR #329, opened 2017-11-13
Factor out keyring initialisation

richvdh:
Take the keyring init from the client-api-server and the monolith out to a
shared file


src/github.com/matrix-org/dendrite/common/keydb/keyring.go
```diff
@@ -0,0 +1,32 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package keydb
+
+import "github.com/matrix-org/gomatrixserverlib"
+
+// CreateKeyRing creates and configures a KeyRing object.
+//
+// It creats the necessary key fetchers and collects them into a KeyRing
```
erikjohnston: creats

erikjohnston:
Otherwise LGTM

PR #330, opened 2017-11-13
Store keys rather than json in the keydatabase

richvdh:
Application-level changes corresponding to https://github.com/matrix-org/gomatrixserverlib/pull/71.

Rather than storing the raw JSON returned from a /keys/v1/query call in the table, store the key itself. This makes keydb.Database implement the updated KeyDatabase interface.

Since it includes changng the db schema, it will be necessary to drop a table on any existing databases:

     DROP TABLE keydb_server_keys

(The table is only a cache really so that shouldn't otherwise be a problem).

~NB this PR is built on top of #329~.


erikjohnston:
If we don't store the JSON will we still be able to serve up responses for other servers? Which I think we might want to do if we are to act as a perspective server?

richvdh:
> If we don't store the JSON will we still be able to serve up responses for other servers? Which I think we might want to do if we are to act as a perspective server?

I thought about this, but I'm basically of the opinion that being a perspective server is a separate job: I would implement it as a separate binary with a separate DB.

erikjohnston:
wfm

PR #333, opened 2017-11-14
Move /state request handling out of RequestPool

erikjohnston:
We should probably move the handling out from the syncapi, but that
requires the clientapi to stream the current state which it currently
doesn't. This at least stops the sync and state handling being done in
one file.


richvdh:
> This at least stops the sync and state handling being done in
one file.

Ok but there are easier ways of doing that than this. Why shouldn't this be done by the RequestPool?

erikjohnston:
> Ok but there are easier ways of doing that than this. Why shouldn't this be done by the RequestPool?

Oh? This moves the state handling stuff in such a way to be consistent with the rest of the codebase. The `RequestPool` is used to do all the sync stuff, so I find it weird the state handling would be tacked on to it. As you can see state only has a single dependency from `RequestPool` anyway.

erikjohnston:
(I just rebased onto master to see if that fixes the linting)

PR #334, opened 2017-11-14
Just run the linter in the commit hook

richvdh:
The pre-commit hook took 45 seconds to run on my machine, which was more than
enough time for me to get distracted by a swordfight in the corridor.

Let's just run the linters (which still take 6 seconds). It's not the place of
a commit hook to run every test we can think of - that is what CI is for.

### DIFF
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -2,4 +2,4 @@
 
 set -eu
 
-./scripts/build-test-lint.sh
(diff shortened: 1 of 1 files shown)


richvdh:
(https://xkcd.com/303/, in case you didn't get the reference)

erikjohnston:
(... and even 45 seconds is quite bad)

PR #338, opened 2017-11-16
Implement room_alias federation end point

rschulman:
This PR attempts to implement the room_alias federation query end point. It borrows heavily from the similar code in `clientapi/readers/directory.go`.

Adding this code required introducing a new parameter in the `Setup` function in `routing.go` and therefore new syntax for calling `Setup` in `dendrite-federation-api-server` and `dendrite-monolith-server`.


src/github.com/matrix-org/dendrite/federationapi/routing/query.go
```diff
@@ -0,0 +1,90 @@
...
+var resp gomatrixserverlib.RespDirectory
+
+	if domain == cfg.Matrix.ServerName {
+		queryReq := api.GetAliasRoomIDRequest{Alias: roomAlias}
+		var queryRes api.GetAliasRoomIDResponse
+		if err = aliasAPI.GetAliasRoomID(httpReq.Context(), &queryReq, &queryRes); err != nil {
+			return httputil.LogThenError(httpReq, err)
+		}
+
+		if len(queryRes.RoomID) > 0 {
```
erikjohnston: Generally we prefer `if queryRes.RoomID != ""`

src/github.com/matrix-org/dendrite/federationapi/routing/query.go
```diff
@@ -0,0 +1,90 @@
...
+			// TODO: List servers that are aware of this room alias
+			resp = gomatrixserverlib.RespDirectory{
+				RoomID:  queryRes.RoomID,
+				Servers: []gomatrixserverlib.ServerName{},
+			}
+		} else {
+			// If the response doesn't contain a non-empty string, return an error
+			return util.JSONResponse{
+				Code: 404,
+				JSON: jsonerror.NotFound("Room alias " + roomAlias + " not found."),
```
erikjohnston: I know you've copied and pasted this, but string concatenation should pretty much always be don via `fmt.Springf`, i.e.:

```golang
jsonerror.NotFound(fmt.Sprintf("Room alias %s not found", roomAlias))
```

src/github.com/matrix-org/dendrite/federationapi/routing/routing.go
```diff
@@ -111,4 +112,14 @@ func Setup(
 			return Version()
 		},
 	)).Methods("GET")
+
+	v1fedmux.Handle("/query/directory/?room_alias={alias}", common.MakeFedAPI(
```
erikjohnston: Unfortunately you can't get query params out in this way. See how the query profile API does it:
https://github.com/matrix-org/dendrite/blob/561315e1d616fd490c1ec720e7a825e3ed916249/src/github.com/matrix-org/dendrite/federationapi/routing/routing.go#L108-L115
https://github.com/matrix-org/dendrite/blob/561315e1d616fd490c1ec720e7a825e3ed916249/src/github.com/matrix-org/dendrite/federationapi/routing/profile.go#L35

src/github.com/matrix-org/dendrite/federationapi/routing/routing.go
```diff
@@ -105,6 +106,16 @@ func Setup(
 		},
 	)).Methods("GET")
 
+	v1fedmux.Handle("/query/directory/?room_alias={alias}", common.MakeFedAPI(
```
erikjohnston: This should just be `/query/directory/`

src/github.com/matrix-org/dendrite/federationapi/routing/query.go
```diff
@@ -0,0 +1,90 @@
...
+
+// RoomAliasToID converts the queried alias into a room ID and returns it
+func RoomAliasToID(
+	httpReq *http.Request,
+	federation *gomatrixserverlib.FederationClient,
+	cfg config.Dendrite,
+	aliasAPI api.RoomserverAliasAPI,
+	roomAlias string,
+) util.JSONResponse {
+	_, domain, err := gomatrixserverlib.SplitID('#', httpReq.FormValue("alias"))
```
erikjohnston: We shouldn't be taking `roomAlias` as a function parameter and instead just set `roomAlias := httpReq.FormValue("alias")`.

We should probably also check that `roomAlias != ""` explicitly so we can error with something like `"must supply alias param"`

src/github.com/matrix-org/dendrite/federationapi/routing/routing.go
```diff
@@ -105,6 +106,23 @@ func Setup(
...
 
+	v1fedmux.Handle("/query/directory/", common.MakeFedAPI(
+		"federation_query_room_alias", cfg.Matrix.ServerName, keys,
+		func(httpReq *http.Request, request *gomatrixserverlib.FederationRequest) util.JSONResponse {
+			return RoomAliasToID(
+				httpReq, federation, cfg, aliasAPI,
+			)
+		},
+	)).Methods("Get")
+	v1fedmux.Handle("/query/directory/?room_alias={alias}", common.MakeFedAPI(
```
erikjohnston: I think you may have forgotten to delete this copy

src/github.com/matrix-org/dendrite/federationapi/routing/routing.go
```diff
@@ -105,6 +106,23 @@ func Setup(
...
 	)).Methods("GET")
 
+	v1fedmux.Handle("/query/directory/", common.MakeFedAPI(
+		"federation_query_room_alias", cfg.Matrix.ServerName, keys,
+		func(httpReq *http.Request, request *gomatrixserverlib.FederationRequest) util.JSONResponse {
+			return RoomAliasToID(
+				httpReq, federation, cfg, aliasAPI,
+			)
+		},
+	)).Methods("Get")
```
erikjohnston: Woops, needs to be `GET` I think

erikjohnston:
Yay! Thank you!

erikjohnston:
Oh, test failures look like we're failing to download kafka... But everything else looks happy so merging.

PR #339, opened 2017-11-16
Add gofmt linter

erikjohnston:


linter-fast.json
```diff
@@ -11,6 +11,7 @@
         "misspell",
         "errcheck",
         "vet",
+	"gofmt",
```
richvdh: indentation looks off

linter.json
```diff
@@ -17,6 +17,7 @@
         "errcheck",
         "vet",
         "megacheck",
+	"gofmt",
```
richvdh: ditto

PR #341, opened 2017-11-16
Fix /sync when we have no events

erikjohnston:
We used a since token of 0 to mean that no token was given. However, naffka
streams start at 0. This causes clients to get stuck spinning forever until an
event is sent.

This changes it so that we pass around pointers instead, with nil meaning a
since token wasn't given.

### DIFF
--- a/src/github.com/matrix-org/dendrite/syncapi/sync/request.go
+++ b/src/github.com/matrix-org/dendrite/syncapi/sync/request.go
@@ -34,7 +34,7 @@ type syncRequest struct {
 	userID        string
 	limit         int
 	timeout       time.Duration
-	since         types.StreamPosition
+	since         *types.StreamPosition
 	wantFullState bool
 	log           *log.Entry
 }
@@ -70,13 +70,14 @@ func getTimeout(timeoutMS string) time.Duration {
 	return time.Duration(i) * time.Millisecond
 }
 
-func getSyncStreamPosition(since string) (types.StreamPosition, error) {
+func getSyncStreamPosition(since string) (*types.StreamPosition, error) {
 	if since == "" {
-		return types.StreamPosition(0), nil
+		return nil, nil
 	}
 	i, err := strconv.Atoi(since)
 	if err != nil {
-		return types.StreamPosition(0), err
+		return nil, err
 	}
-	return types.StreamPosition(i), nil
+	token := types.StreamPosition(i)
+	return &token, nil
 }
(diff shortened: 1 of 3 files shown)


src/github.com/matrix-org/dendrite/syncapi/sync/request.go
```diff
@@ -34,7 +34,7 @@ type syncRequest struct {
 	userID        string
 	limit         int
 	timeout       time.Duration
-	since         types.StreamPosition
+	since         *types.StreamPosition
```
richvdh: worth  having a comment to explain what `nil` means?

src/github.com/matrix-org/dendrite/syncapi/sync/request.go
```diff
@@ -70,13 +70,16 @@ func getTimeout(timeoutMS string) time.Duration {
 	return time.Duration(i) * time.Millisecond
 }
 
-func getSyncStreamPosition(since string) (types.StreamPosition, error) {
+// getSyncStreamPosition tries to pass a string taken from the API to a stream
```
richvdh: "a string" seems a bit vague. "a 'since' token" ?

src/github.com/matrix-org/dendrite/syncapi/sync/requestpool.go
```diff
@@ -93,7 +93,7 @@ func (rp *RequestPool) OnIncomingSyncRequest(req *http.Request, device *authtype
 		case <-timer.C:
 			return util.JSONResponse{
 				Code: 200,
-				JSON: types.NewResponse(syncReq.since),
+				JSON: types.NewResponse(currPos),
```
richvdh: is this an unrelated fix? looks ok, just wondering.

erikjohnston: I otherwise would need to dereference `syncReq.since`, though in practice it probably doesn't really matter which one we chose─other than using `currPos` might be a bit faster.

src/github.com/matrix-org/dendrite/syncapi/sync/request.go
```diff
@@ -70,13 +70,16 @@ func getTimeout(timeoutMS string) time.Duration {
 	return time.Duration(i) * time.Millisecond
 }
 
-func getSyncStreamPosition(since string) (types.StreamPosition, error) {
+// getSyncStreamPosition tries to pass a 'since' token taken from the API to a
```
richvdh: s/pass/parse/

PR #343, opened 2017-11-17
Implemented ReCaptcha registration method

### LINKED ISSUE #285: Implement RECAPTCHA registration flow.
This should be the case of simply adding RECAPTCHA to the returned registration flows and implementing the check in [synapse](https://github.com/matrix-org/synapse/blob/c72058b/synapse/handlers/auth.py#L275-L319) at the appropriate point: https://github.com/matrix-org/dendrite/blob/bdfade26ed832a3519016ef278cf4605b9f13ccc/src/github.com/matrix-org/dendrite/clientapi/writers/register.go#L230-L254

anoadragon453:
Closes #285.

Currently just the registration flow. I'm not sure of how to signal to Riot that the homeserver requests ReCaptcha just yet.


anoadragon453:
...github is being weird with the merge conflict.

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -147,6 +160,64 @@ func validatePassword(password string) *util.JSONResponse {
...
+
+	defer resp.Body.Close()
+
+	// Grab the body of the response from the captcha server
+	var r recaptchaResponse
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return &util.JSONResponse{
+			Code: 500,
+			JSON: jsonerror.BadJSON("Error in contacting captcha server" + err.Error()),
```
anoadragon453: Ideally these should be `InternalServerError` rather than a `BadJSON`, however jsonerror's `InternalServerError` does not accept any error string.

Should we create another type in jsonerror that accepts error text? Like `InteralServerErrorWithDesc` or similar or do something else?

anoadragon453:
Not sure why the tests are failing :)

t3chguy:
> Not sure why the tests are failing :)

Looks like your call to `completeRegistration` is missing its final param, device name: https://github.com/matrix-org/dendrite/blob/bad701c703cb97c8352fea671e506f47c063a91c/src/github.com/matrix-org/dendrite/clientapi/routing/register.go#L292

t3chguy:
> src/github.com/matrix-org/dendrite/clientapi/routing/register.go:222::warning: cyclomatic complexity 14 of function Register() is high (> 12) (gocyclo)

just means that ::Register should be split into submethods

t3chguy:
> Currently just the registration flow. I'm not sure of how to signal to Riot that the homeserver requests ReCaptcha just yet.

You intercept `/register`, return 401 to trigger UI auth and pass the flows&stages in its response e.g below

image

anoadragon453:
> Looks like your call to completeRegistration is missing its final param, display name: 

Ah, didn't have that locally, helps to update your branch to master.

anoadragon453:
@t3chguy Alright, I believe the correct flow is already returned with the following, but correct me if I'm wrong:

https://github.com/matrix-org/dendrite/pull/343/files#diff-789859953fb6c45e704d547fce33b9eeR240

t3chguy:
@anoadragon453 looking at that line it looks like people can circumvent the captcha by just using dummy auth instead, which seems counterproductive?

t3chguy:
https://github.com/matrix-org/dendrite/pull/343/files#diff-789859953fb6c45e704d547fce33b9eeR237

anoadragon453:
As @t3chguy mentioned, we're going to have to rethink the registration flow to more than just a large switch statement. Either to [the way synapse does it](https://github.com/matrix-org/synapse/blob/c72058b/synapse/handlers/auth.py) or something better if possible.

```
src/github.com/matrix-org/dendrite/clientapi/routing/register.go:192:23:warning: error return value not checked (defer resp.Body.Close()) (errcheck)
```

Also no idea about that, defer doesn't return anything...

t3chguy:
defer doesn't return anything but `Close()` returns an error interface-type

anoadragon453:
Ah ok, I'll just wrap it in a deferred function then.

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
erikjohnston: We use `return httputil.LogThenError(req, err)` for this generally.

erikjohnston:
@anoadragon453 If you don't care about the error on close (e.g. if its a http request) you can just do:

```golang
defer req.Close()  // nolint: errcheck
```

But if its something like closing a file after writing, then you should care.

erikjohnston:
I think this whole registration bit is going to need a refactor, as we only want to complete registration once the user has gone through the full flow rather than after each stage. The current stuff cheats by only implementing flows that have a single stage.

erikjohnston:
I hadn't realised that registration hadn't been built out properly already :/

anoadragon453:
Well it's quite simple to have each step of a flow as its own method and then run them in succession. My question is if flows are hardcoded or generated from available methods.

If they were to be generated, we'd need rules on which flows are necessary (does position in the flow matter as well?). One rule could be that if ReCaptcha is enabled, it must be added into every existing flow besides sharedsecret. What other flows would be necessary? Ideally it would be very easy to add new ones.

I'll look at how Synapse does it in the meantime.

anoadragon453:
Alright, so after looking through the spec, it looks like this should be the basic outline for how the authentication handler should be designed:

```
authHandler() {
    if authdict does not contain "auth" key or empty, return all possible flows.
    otherwise, run stage given by the request with included client parameters
    now, check if this session has completed enough stages for a full flow (flows are ordered, so should be a quick check.)
    (Should we handle a flow that was completed, but completed out-of-order?)
        if a flow hasn't been completed, return flows again along with completed stages so far
    yay, you did it! register complete!
```

This is all assuming I understand the spec correctly. Thoughts? Otherwise I believe I can implement the above in this PR as well.

anoadragon453:
Also heads up that we're gonna have to change a bunch of the spec links to stable now after the r3.0.0 release :P

anoadragon453:
Relevant Synapse code for building flow building:

https://github.com/matrix-org/synapse/blob/552f123bea1014680ab798b7e34cd1b23424a189/synapse/rest/client/v2_alpha/register.py#L307-L330

erikjohnston:
Sounds sane to me. I don't know much about registration to be honest, so I'll defer to any thoughts @t3chguy or @dbkr have

t3chguy:
all /me knows about registration is what the spec says, /me is real new to this matrix thingimajig

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -376,6 +496,62 @@ func isValidMacLogin(
...
+
+	a_copy := make([]string, len(a))
+	b_copy := make([]string, len(b))
+
+	for loginType := range a {
+		a_copy = append(a_copy, string(loginType))
+	}
+	for loginType := range b {
+		b_copy = append(b_copy, string(loginType))
+	}
```
anoadragon453: Ideally we would just write `copy(a_copy, a)`, but since one is a slice of `string` and the other is a slice of `authtypes.LoginType` (which is *typed* as a string), we can't :P

anoadragon453:
Registration carries through with or without captcha registration enabled, but then gets stuck complaining about filters:

```
ERRO[2017-11-23T01:53:20.830430627Z] request failed                                error=
"sql: no rows in result set" req.id=UXRHohIdqXeH req.method=POST req.path="/_matrix/client/r0/user/@recaptchaguest:localhost/filter"
INFO[2017-11-23T01:53:20.830497493Z] Responding (55 bytes)                         code=500 req.id=UXRHohIdqXeH req.method=POST req.path="/_matrix/client/r0/user/@recaptchaguest:
localhost/filter"
```

Unsure if related to this PR but does stop the user from getting to the home page. Any idea @erikjohnston? @richvdh

erikjohnston:
@anoadragon453 you probably want to merge in master

anoadragon453:
Done now, thanks!

anoadragon453:
Rebased on top of the latest master.

Let me know if you still want to change the logging @erikjohnston.

If anyone needs Recaptcha pub/priv keys to test I can PM them to you.

anoadragon453:
Ready to review!

ara4n:
you haz conflicts

anoadragon453:
img_20171204_123924

erikjohnston:
You know what, let's just disable the maligned lint.

erikjohnston:
https://github.com/matrix-org/dendrite/pull/365

PR #347, opened 2017-11-20
Update version of kafka

erikjohnston:


scripts/install-local-kafka.sh
```diff
@@ -7,7 +7,7 @@ set -eu
 # The mirror to download kafka from is picked from the list of mirrors at
 # https://www.apache.org/dyn/closer.cgi?path=/kafka/0.10.2.0/kafka_2.11-0.10.2.0.tgz
```
krombel: This URL should be updated as well

richvdh:
what @krombel said. lgtm otherwise

Issue #269, opened 2017-09-27
Check room alias isn't taken when creating new room


erikjohnston:
This is slightly tricky as it interacts with #267. We'll probably want to have an API that allows us to "reserve" an alias and later "claim" it once the room is actually created, or something

spaetz:
just checked whether this is still relevant, and yes as hinted by https://github.com/matrix-org/dendrite/blob/5424b88f3061dddbeae53df8b67bd064e8cbc400/clientapi/routing/createroom.go#L473
we first create a room and then try to create the alias, bailing out when it is already taken. So this is issue is still relevant.

PR #348, opened 2017-11-20
Check roomid doesnt exist when creating

### LINKED ISSUE #267: Check room ID doesn't exist when creating a new room

rreuvekamp:
Closes #267

I've chosen this approach after the discussion in the issue item.

This approach could also be used for #269, and in that case we might want to use the same database table (add a field alias_name or something and make room_id and alias_name the primary key together).

##Things that could be not nice

Nothing stops a malicious clientAPI from not reserving a room ID.

Currently, reserved room IDs are not removed.
There could be a garbage collector that removed items if they are older than, let's say, five minutes or if the room ID exists now.
The clientAPI could also remove the reservation after creating the room, but if an error occures while/after creating the room, the clientAPI will not fire the 'unreserve' request.

Signed-off-by: Remi Reuvekamp <git@remireuvekamp.nl>


rreuvekamp:
... Will fix when I'm on a proper internet connection.

erikjohnston:
(The broken tests have been fixed now)

src/github.com/matrix-org/dendrite/clientapi/routing/createroom.go
```diff
@@ -90,10 +90,27 @@ type fledglingEvent struct {
...
+		queryReq := api.QueryReserveRoomIDRequest{RoomID: checkRoomID}
+		var queryResp api.QueryReserveRoomIDResponse
+		err := queryAPI.QueryReserveRoomID(req.Context(), &queryReq, &queryResp)
+		if err != nil {
+			return httputil.LogThenError(req, err)
+		}
+
+		if queryResp.Success {
+			roomID = checkRoomID
+		}
```
erikjohnston: I would break here and have an unconditional loop.

We probably also should have a counter so that we only try this e.g. 10 times.

src/github.com/matrix-org/dendrite/roomserver/api/query.go
```diff
@@ -155,6 +154,19 @@ type QueryServerAllowedToSeeEventResponse struct {
 	AllowedToSeeEvent bool `json:"can_see_event"`
 }
 
+// QueryReserveRoomIDRequest is a request to QueryServerAllowedToSeeEvent
```
erikjohnston: `QueryServerAllowedToSeeEvent` should be replaced

src/github.com/matrix-org/dendrite/roomserver/api/query.go
```diff
@@ -155,6 +154,19 @@ type QueryServerAllowedToSeeEventResponse struct {
...
+type QueryReserveRoomIDRequest struct {
+	// The room ID to check and reserve.
+	RoomID string `json:"room_id"`
+}
+
+// QueryReserveRoomIDResponse is a response to QueryServerAllowedToSeeEvent
```
erikjohnston: Ditto, `QueryServerAllowedToSeeEvent` should be replaced

src/github.com/matrix-org/dendrite/roomserver/query/query.go
```diff
@@ -418,6 +424,43 @@ func (r *RoomserverQueryAPI) QueryServerAllowedToSeeEvent(
...
+	}
+
+	err = r.DB.ReserveRoomID(ctx, request.RoomID)
+	if err != nil {
+		return err
+	}
+
+	response.Success = true
+
+	return nil
```
erikjohnston: All of this needs to happen in a single database transaction to avoid races where two reserve room requests come in at the same time. In fact, I would probably have a single table which was just a list of reserved/used roomIDs and do:

`INSERT INTO roomserver_reserved_rooms (room_id) VALUES ($1) ON CONFLICT (room_id) DO NOTHING`

and then check if that returns any rows updated or not. If no rows were updated then we know that the roomID has already been reserved.

richvdh:
closing this pending updates as erik requests

rreuvekamp:
F. I still had this open. I am so sorry. I really hope I didn't make any other development stagnate. I am so sorry.

(Also please Github let me comment without reopening the thread)

richvdh:
no worries at all, thank you for contributing!

I should have checked if you're still interesting on working on it - if you are I'd be glad to reopen!

PR #349, opened 2017-11-21
Code Style: Prefer WithFields over Infof

erikjohnston:


erikjohnston:
Can also add more caveats if its still felt the recommendation is too strong?

PR #350, opened 2017-11-22
Add database.naffka to example config

erikjohnston:


dendrite-config.yaml
```diff
@@ -97,6 +98,8 @@ database:
     server_key: "postgres://dendrite:itsasecret@localhost/dendrite_serverkey?sslmode=disable"
     federation_sender: "postgres://dendrite:itsasecret@localhost/dendrite_federationsender?sslmode=disable"
     public_rooms_api: "postgres://dendrite:itsasecret@localhost/dendrite_publicroomsapi?sslmode=disable"
+    # If using naffka you need to specify a naffka database
+    # naffka: "postgres://dendrite:itsasecret@localhost/dendrite_naffk?sslmode=disable"
```
anoadragon453: `dendrite_naffk`? :)

erikjohnston:
> It's starting to feel a bit grim that you have to configure all these separate databases - perhaps for the monolith we should just use one?

Yeah, I'm not really sure why we don't

PR #351, opened 2017-11-22
Fix panic if upload name wasn't supplied

erikjohnston:


src/github.com/matrix-org/dendrite/mediaapi/routing/upload.go
```diff
@@ -188,7 +189,7 @@ func (r *uploadRequest) Validate(maxFileSizeBytes config.FileSizeBytes) *util.JS
 			JSON: jsonerror.Unknown("HTTP Content-Type request header must be set."),
 		}
 	}
-	if r.MediaMetadata.UploadName[0] == '~' {
+	if strings.HasPrefix(string(r.MediaMetadata.UploadName), "~") {
```
richvdh: is the cast really needed? `UploadName` is a `Filename` which is already a `string`?

erikjohnston: Yup, as it doesn't implicitly cast types like that. 

(Personally I think we overuse new types all over the code base; it seems a little pointless unless we have functions that only accept the new type which doesn't seem to be the case most of the time)

PR #352, opened 2017-11-22
Implement query to get state and auth chain

erikjohnston:
This will be used when handling income join requests from federation.

### DIFF
--- a/src/github.com/matrix-org/dendrite/roomserver/api/query.go
+++ b/src/github.com/matrix-org/dendrite/roomserver/api/query.go
@@ -155,6 +155,32 @@ type QueryServerAllowedToSeeEventResponse struct {
 	AllowedToSeeEvent bool `json:"can_see_event"`
 }
 
+// QueryStateAndAuthChainRequest is a request to QueryStateAndAuthChain
+type QueryStateAndAuthChainRequest struct {
+	// The room ID to query the state in.
+	RoomID string `json:"room_id"`
+	// The list of events to return state after.
+	PrevEventIDs []string `json:"prev_event_ids"`
+	// The list of auth events to get the auth chain for.
+	AuthEventIDs []string `json:"auth_event_ids"`
+}
+
+// QueryStateAndAuthChainResponse is a response to QueryStateAndAuthChain
+type QueryStateAndAuthChainResponse struct {
+	// Copy of the request for debugging.
+	QueryStateAndAuthChainRequest
+	// Does the room exist on this roomserver?
+	// If the room doesn't exist this will be false and StateEvents will be empty.
+	RoomExists bool `json:"room_exists"`
+	// Do all the previous events exist on this roomserver?
+	// If some of previous events do not exist this will be false and StateEvents will be empty.
+	PrevEventsExist bool `json:"prev_events_exist"`
+	// The state and auth chain events that were requested.
+	// The lists will be in an arbitrary order.
+	StateEvents     []gomatrixserverlib.Event `json:"state_events"`
+	AuthChainEvents []gomatrixserverlib.Event `json:"auth_chain_events"`
+}
+
 // RoomserverQueryAPI is used to query information from the room server.
 type RoomserverQueryAPI interface {
 	// Query the latest events and state for a room from the room server.
@@ -198,6 +224,13 @@ type RoomserverQueryAPI interface {
 		request *QueryServerAllowedToSeeEventRequest,
 		response *QueryServerAllowedToSeeEventResponse,
 	) error
+
+	// Query to get state and auth chain
+	QueryStateAndAuthChain(
+		ctx context.Context,
+		request *QueryStateAndAuthChainRequest,
+		response *QueryStateAndAuthChainResponse,
+	) error
 }
 
 // RoomserverQueryLatestEventsAndStatePath is the HTTP path for the QueryLatestEventsAndState API.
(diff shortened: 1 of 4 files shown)


src/github.com/matrix-org/dendrite/roomserver/query/query.go
```diff
@@ -418,6 +418,108 @@ func (r *RoomserverQueryAPI) QueryServerAllowedToSeeEvent(
 	return nil
 }
 
+// QueryStateAndAuthChain implements api.RoomserverQueryAPI
+func (r *RoomserverQueryAPI) QueryStateAndAuthChain(
```
richvdh: doesn't this need calling from the router or something?

richvdh: There's a lot of code here for there to be no tests at all. Let's at least have some simple test cases so that we're more inclined to add tests in the future.

src/github.com/matrix-org/dendrite/roomserver/query/query.go
```diff
@@ -418,6 +418,108 @@ func (r *RoomserverQueryAPI) QueryServerAllowedToSeeEvent(
...
+
+		events, err := r.DB.Events(ctx, nids)
+		if err != nil {
+			return nil, err
+		}
+
+		// Work out a) which events we should add to the returned list of
+		// events and b) which of the auth events we haven't seen yet and
+		// add them to the list of events to fetch.
+		eventsToFetch = eventsToFetch[:0]
```
richvdh: is this an obscure way of writing `eventsToFetch = []string{}` ? And should it instead just be `eventsToFetch = nil` ?

src/github.com/matrix-org/dendrite/roomserver/query/query.go
```diff
@@ -418,6 +418,108 @@ func (r *RoomserverQueryAPI) QueryServerAllowedToSeeEvent(
...
+	var authEvents []gomatrixserverlib.Event
+
+	// List of event ids to fetch. These will be added to the result and
+	// their auth events will be fetched (if they haven't been previously)
+	eventsToFetch := authEventIDs
+
+	// Set of events we've already fetched.
+	fetchedEventMap := make(map[string]bool)
+	for _, eventID := range authEventIDs {
+		fetchedEventMap[eventID] = true
```
richvdh: could we populate this after we've done the fetching, which (a) feels more intuitive, (b) saves duplication between here and line 511

src/github.com/matrix-org/dendrite/roomserver/query/query.go
```diff
@@ -418,6 +418,108 @@ func (r *RoomserverQueryAPI) QueryServerAllowedToSeeEvent(
...
+	}
+
+	// Check if there's anything left to do
+	for len(eventsToFetch) > 0 {
+		// Convert eventIDs to events. First need to fetch NIDs
+		nidMap, err := r.DB.EventNIDs(ctx, authEventIDs)
```
richvdh: s/authEventIDs/eventsToFetch/ ? Or am I misunderstanding what's going on here?

PS: this is the sort of thing I'd expect to see tested...

src/github.com/matrix-org/dendrite/roomserver/query/query.go
```diff
@@ -418,6 +418,108 @@ func (r *RoomserverQueryAPI) QueryServerAllowedToSeeEvent(
...
+		if err != nil {
+			return nil, err
+		}
+
+		// Work out a) which events we should add to the returned list of
+		// events and b) which of the auth events we haven't seen yet and
+		// add them to the list of events to fetch.
+		eventsToFetch = eventsToFetch[:0]
+		for _, event := range events {
+			for _, authEventID := range event.AuthEventIDs() {
```
richvdh: could do with a comment here: "recursively fetch the auth events for each of the fetched events" or sth

src/github.com/matrix-org/dendrite/roomserver/api/query.go
```diff
@@ -155,6 +155,32 @@ type QueryServerAllowedToSeeEventResponse struct {
...
+	// The list of events to return state after.
+	PrevEventIDs []string `json:"prev_event_ids"`
+	// The list of auth events to get the auth chain for.
+	AuthEventIDs []string `json:"auth_event_ids"`
+}
+
+// QueryStateAndAuthChainResponse is a response to QueryStateAndAuthChain
+type QueryStateAndAuthChainResponse struct {
+	// Copy of the request for debugging.
+	QueryStateAndAuthChainRequest
```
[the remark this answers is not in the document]
erikjohnston: Its what everything else does.

src/github.com/matrix-org/dendrite/roomserver/query/query.go
erikjohnston: Woops, fixed

erikjohnston: Added tests.

src/github.com/matrix-org/dendrite/roomserver/query/query.go
erikjohnston: It's a standard way of clearing a slice. (The difference is both the empty and nil slice results in a new allocation, whereas reuses the slice backing storage, aiui)

src/github.com/matrix-org/dendrite/roomserver/api/query.go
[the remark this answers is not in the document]
richvdh: No it's not: https://github.com/matrix-org/dendrite/pull/352/files#diff-4f5dd12212a836aae3a6bb015d9936f0R121

src/github.com/matrix-org/dendrite/roomserver/query/query.go
richvdh: bletch. is reusing the backing store actually useful here, I wonder.

erikjohnston:
> I'm a bit concerned about when this is going to be called: are we going to end up calling it for every new event, or what?

It gets called when another server joins a room via dendrite, and we need to respond with the full state and auth chain.

There's definitely scope for this getting pulled out of the room server and/or have some of the auth chain calculations cached in DB.

PR #354, opened 2017-11-23
Set up proper registration flow

anoadragon453:
* We now keep track of sessions and their completed registration stages.
* We only complete registration if the client has completed a full flow.
* New Derived section in config for data derived from config options.
* New config options for captcha.
* Send params back to client for each registration stage.

Required for #343

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/auth/authtypes/flow.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/auth/authtypes/flow.go
@@ -15,7 +15,7 @@
 package authtypes
 
 // Flow represents one possible way that the client can authenticate a request.
-// http://matrix.org/docs/spec/HEAD/client_server/r0.3.0.html#user-interactive-authentication-api
+// https://matrix.org/docs/spec/client_server/r0.3.0.html#user-interactive-authentication-api
 type Flow struct {
 	Stages []LoginType `json:"stages"`
 }
--- a/src/github.com/matrix-org/dendrite/clientapi/auth/authtypes/flow.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/auth/authtypes/flow.go
@@ -1,0 +1,21 @@
+// Copyright 2017 Vector Creations Ltd
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package authtypes
+
+// Flow represents one possible way that the client can authenticate a request.
+// http://matrix.org/docs/spec/HEAD/client_server/r0.3.0.html#user-interactive-authentication-api
+type Flow struct {
+	Stages []LoginType `json:"stages"`
+}
--- a/src/github.com/matrix-org/dendrite/clientapi/auth/authtypes/flow.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/auth/authtypes/flow.go
@@ -1,4 +1,4 @@
-// Copyright 2017 Vector Creations Ltd
+// Copyright Andrew Morgan <andrew@amorgan.xyz>
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/src/github.com/matrix-org/dendrite/clientapi/routing/register.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/routing/register.go
@@ -251,21 +251,19 @@ func handleRegistrationFlow(
 		}
 	}
 
-	// Check if a registration flow has been completed successfully
-	for _, flow := range cfg.Derived.Registration.Flows {
-		if checkFlowsEqual(flow, authtypes.Flow{sessions[sessionID]}) {
-			return completeRegistration(req.Context(), accountDB, deviceDB,
-				r.Username, r.Password, r.InitialDisplayName)
+	// Check if the user's registration flow has been completed successfully
+	if !checkFlowCompleted(authtypes.Flow{sessions[sessionID]}, cfg.Derived.Registration.Flows) {
+		// There are still more stages to complete.
+		// Return the flows and those that have been completed.
+		return util.JSONResponse{
+			Code: 401,
+			JSON: newUserInteractiveResponse(sessionID,
+				cfg.Derived.Registration.Flows, cfg.Derived.Registration.Params),
 		}
 	}
 
-	// There are still more stages to complete.
-	// Return the flows and those that have been completed.
-	return util.JSONResponse{
-		Code: 401,
-		JSON: newUserInteractiveResponse(sessionID,
-			cfg.Derived.Registration.Flows, cfg.Derived.Registration.Params),
-	}
+	return completeRegistration(req.Context(), accountDB, deviceDB,
+		r.Username, r.Password, r.InitialDisplayName)
 }
 
 // LegacyRegister process register requests from the legacy v1 API
@@ -420,24 +418,22 @@ func isValidMacLogin(
 	return hmac.Equal(givenMac, expectedMAC), nil
 }
 
-// checkFlowsEqual checks if two registration flows have the same stages
-// within them. Order of stages does not matter.
-func checkFlowsEqual(aFlow, bFlow authtypes.Flow) bool {
-	a := aFlow.Stages
-	b := bFlow.Stages
-	if len(a) != len(b) {
-		return false
-	}
+// checkFlows checks a single flow a against another, b. If a contains at least
+// all of the stages that b does, checkFlows returns true.
+func checkFlows(a []authtypes.LoginType, b []authtypes.LoginType) bool {
+	// Sort the slices for simple comparison
 	sort.Slice(a, func(i, j int) bool { return a[i] < a[j] })
 	sort.Slice(b, func(i, j int) bool { return b[i] < b[j] })
 
-	// Account for any extra stages a user may do unnecessarily
-	extraStages := len(b) - len(a)
+	// Account for any extra stages a user may unnecessarily do.
+	extraStages := len(a) - len(b)
 	for i := range b {
 		if extraStages < 0 {
+			// The provided flow has run out of possible extraneous stages.
 			return false
 		}
 		if a[i] != b[i] {
+			// Wasn't a match, drop an extraneous stage.
 			extraStages--
 			continue
 		}
@@ -445,11 +441,24 @@ func checkFlowsEqual(aFlow, bFlow authtypes.Flow) bool {
 	return true
 }
 
+// checkFlowCompleted checks if a registration flow complies with any flow
+// dictated by the server. Order of stages does not matter. A user may complete
+// extra stages as long as the required stages of at least one flow is met.
+func checkFlowCompleted(flow authtypes.Flow, derivedFlows []authtypes.Flow) bool {
+	// Iterate through possible flows to check whether any have been fully completed.
+	for _, derivedFlow := range derivedFlows {
+		if checkFlows(flow.Stages, derivedFlow.Stages) {
+			return true
+		}
+	}
+	return false
+}
+
 type availableResponse struct {
 	Available bool `json:"available"`
 }
 
-// RegisterAvailable checks if the username is already taken or invalid
+// RegisterAvailable checks if the username is already taken or invalid.
 func RegisterAvailable(
 	req *http.Request,
 	accountDB *accounts.Database,
--- a/src/github.com/matrix-org/dendrite/clientapi/routing/register.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/routing/register.go
@@ -23,8 +23,8 @@ import (
 	"fmt"
 	"net/http"
 	"regexp"
+	"sort"
 	"strings"
-	"time"
 
 	"github.com/matrix-org/dendrite/common/config"
 
@@ -43,9 +43,14 @@ const (
 	minPasswordLength = 8   // http://matrix.org/docs/spec/client_server/r0.2.0.html#password-based
 	maxPasswordLength = 512 // https://github.com/matrix-org/synapse/blob/v0.20.0/synapse/rest/client/v2_alpha/register.py#L161
 	maxUsernameLength = 254 // http://matrix.org/speculator/spec/HEAD/intro.html#user-identifiers TODO account for domain
+	sessionIDLength   = 24
 )
 
-var validUsernameRegex = regexp.MustCompile(`^[0-9a-zA-Z_\-./]+$`)
+var (
+	// TODO: Remove old sessions. Need to do so on a session-specific timeout.
+	sessions           = make(map[string][]authtypes.LoginType) // Sessions and completed flow stages
+	validUsernameRegex = regexp.MustCompile(`^[0-9a-zA-Z_\-./]+$`)
+)
 
 // registerRequest represents the submitted registration request.
 // It can be broken down into 2 sections: the auth dictionary and registration parameters.
@@ -68,21 +73,16 @@ type authDict struct {
 	Type    authtypes.LoginType         `json:"type"`
 	Session string                      `json:"session"`
 	Mac     gomatrixserverlib.HexString `json:"mac"`
+
 	// TODO: Lots of custom keys depending on the type
 }
 
 // http://matrix.org/speculator/spec/HEAD/client_server/unstable.html#user-interactive-authentication-api
 type userInteractiveResponse struct {
-	Flows     []authFlow             `json:"flows"`
+	Flows     []authtypes.Flow       `json:"flows"`
 	Completed []authtypes.LoginType  `json:"completed"`
 	Params    map[string]interface{} `json:"params"`
 	Session   string                 `json:"session"`
-}
-
-// authFlow represents one possible way that the client can authenticate a request.
-// http://matrix.org/speculator/spec/HEAD/client_server/unstable.html#user-interactive-authentication-api
-type authFlow struct {
-	Stages []authtypes.LoginType `json:"stages"`
 }
 
 // legacyRegisterRequest represents the submitted registration request for v1 API.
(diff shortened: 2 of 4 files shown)


src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -212,16 +238,36 @@ func Register(
...
+	for _, flow := range cfg.Derived.Flows {
+		if checkFlowsEqual(flow, authtypes.Flow{sessions[sessionID]}) {
+			return completeRegistration(req.Context(), accountDB, deviceDB,
+				r.Username, r.Password, r.InitialDisplayName)
+		}
+	}
+
+	// There are still more stages to complete.
+	// Return the flows and those that have been completed.
+	return util.JSONResponse{
```
anoadragon453: @erikjohnston Did you mention that you typically use `return httputil.LogThenError(req, err)` for this rather than creating a `util.JSONResponse` on the spot?

erikjohnston: Only for when returning 500s. There may be other helper functions for doing 401

erikjohnston: But I think this is fine as is

anoadragon453: Cool, alright.

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -376,6 +422,62 @@ func isValidMacLogin(
...
+	letterIdxMask = 1<<letterIdxBits - 1 // All 1-bits, as many as letterIdxBits
+	letterIdxMax  = 63 / letterIdxBits   // # of letter indices fitting in 63 bits
+)
+
+var src = rand.NewSource(time.Now().UnixNano())
+
+// RandString returns a random string of characters with a given length.
+// Do note that it is not thread-safe in its current form.
+// https://stackoverflow.com/a/31832326
+func RandString(n int) string {
```
erikjohnston: `util.RandomString` is already a thing

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -376,6 +422,62 @@ func isValidMacLogin(
...
+		aCopy = append(aCopy, string(loginType))
+	}
+	for loginType := range b {
+		bCopy = append(bCopy, string(loginType))
+	}
+
+	sort.Strings(aCopy)
+	sort.Strings(bCopy)
+
+	return reflect.DeepEqual(aCopy, bCopy)
```
erikjohnston: Since these are just slices of strings I'd just do a for loop. I don't trust the reflect package not to be weird or slow.

```golang
for i := range aCopy {
    if a[i] != b[i] {
        return false
    }
}
return true
```

erikjohnston: It might actually be nicer to just do, instead of creating copies:

```golang
sort.Slice(aFlow, func(x, y) { return x < y })
```

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -43,9 +46,13 @@ const (
 	minPasswordLength = 8   // http://matrix.org/docs/spec/client_server/r0.2.0.html#password-based
 	maxPasswordLength = 512 // https://github.com/matrix-org/synapse/blob/v0.20.0/synapse/rest/client/v2_alpha/register.py#L161
 	maxUsernameLength = 254 // http://matrix.org/speculator/spec/HEAD/intro.html#user-identifiers TODO account for domain
+	sessionIDLength   = 24
 )
 
-var validUsernameRegex = regexp.MustCompile(`^[0-9a-zA-Z_\-./]+$`)
+var (
+	sessions           = make(map[string][]authtypes.LoginType) // Sessions and completed flow stages
```
erikjohnston: We'll want to prune these sessions so that we don't just slowly but surely fill up with old sessions.

Happy for you to just add a TODO and can deal with it in another PR.

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
anoadragon453: Cool, was hoping we'd have something already :)

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
anoadragon453: > It might actually be nicer to just do, instead of creating copies:
>     `sort.Slice(aFlow, func(x, y) { return x < y })`

Oh, much cleaner. I'll give that a whirl.

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
anoadragon453: Yeah, good catch. Obviously removing when registering is completed is fine and dandy, but will need to do something about a session length for non-completed registrations.

erikjohnston: Actually, we'll probably want to keep around completed registrations for a while in case the client retries the last register request.

anoadragon453: You mean even after `completeRegistration()` has been carried out?

In that case we can just have a session length be the only way to prune sessions.

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
anoadragon453: Nice, works: https://play.golang.com/p/wLPjs2gfWE

erikjohnston:
(Assigning to @richvdh to double check that this is how registration actually works)

src/github.com/matrix-org/dendrite/common/config/config.go
```diff
@@ -189,6 +190,17 @@ type Dendrite struct {
...
+	// Any information derived from the configuration options for later use.
+	Derived struct {
+		// Flows for registration. As long as they given flows only relies on config file options,
+		// we can generate them on startup and store them until needed
+		Flows []authtypes.Flow `json:"flows"`
+
+		// Params for registration. Data that is returned to the client while registering and
+		// that which is necessary to complete certain registration flow stages
+		Params map[string]interface{} `json:"params"`
+	}
```
erikjohnston: We may want to stick this in a `Registration` sub structure, otherwise its not clear that `config.Derived.Flows` is a registration thing

src/github.com/matrix-org/dendrite/clientapi/auth/authtypes/flow.go
```diff
@@ -0,0 +1,21 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package authtypes
+
+// Flow represents one possible way that the client can authenticate a request.
+// http://matrix.org/speculator/spec/HEAD/client_server/unstable.html#user-interactive-authentication-api
```
richvdh: suggest linking to https://matrix.org/docs/spec/client_server/r0.3.0.html. Even if you need the unstable version (which you don't here), https://matrix.org/docs/spec/client_server/unstable.html is more reliable than the speculator.

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -376,6 +420,25 @@ func isValidMacLogin(
 	return hmac.Equal(givenMac, expectedMAC), nil
 }
 
+// checkFlowsEqual checks if two registration flows have the same stages
+// within them. Order of stages does not matter.
+func checkFlowsEqual(aFlow, bFlow authtypes.Flow) bool {
+	a := aFlow.Stages
+	b := bFlow.Stages
+	if len(a) != len(b) {
```
richvdh: it's valid for the client to complete *extra* stages.

src/github.com/matrix-org/dendrite/common/config/config.go
richvdh: +1.

TBH I'm not entirely sure this belongs in the Config object at all. Shouldn't it be up to the register.go module to figure out the right flows based on the config?

anoadragon453: The intention was to generate these structures once on startup, then use them over and over again during each registration.

We could place the actual calculation in register.go, though we would simply be calling it from config.go after the config file was read.

anoadragon453:
Updated to fix spec links, allow client to complete extra flow stages and still register and create a Registration sub-struct under Derived.

src/github.com/matrix-org/dendrite/clientapi/auth/authtypes/flow.go
```diff
@@ -0,0 +1,21 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package authtypes
+
+// Flow represents one possible way that the client can authenticate a request.
+// http://matrix.org/docs/spec/HEAD/client_server/r0.3.0.html#user-interactive-authentication-api
```
richvdh: broken link

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
richvdh: This is still going to fail if the client completes extra stages.

You're also going to have to rename the function and change its comment.

richvdh:
Some other points:

* please could you sign off with your full name? Sorry but those are the rules :/
* it would make reviewing changes easier if you made changes in new commits rather than amending the previous one
* it feels like we should have some tests on this stuff

anoadragon453:
I've cleaned up the flow-checking code, switched to my real name (thanks for poking me to do that finally), and added some simple tests for the flow-checking.

We should probably have some integration tests for registration in general though. Do you think those should be bundled in this PR or put in another one?

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -251,21 +251,19 @@ func handleRegistrationFlow(
 		}
 	}
 
-	// Check if a registration flow has been completed successfully
-	for _, flow := range cfg.Derived.Registration.Flows {
-		if checkFlowsEqual(flow, authtypes.Flow{sessions[sessionID]}) {
-			return completeRegistration(req.Context(), accountDB, deviceDB,
-				r.Username, r.Password, r.InitialDisplayName)
+	// Check if the user's registration flow has been completed successfully
+	if !checkFlowCompleted(authtypes.Flow{sessions[sessionID]}, cfg.Derived.Registration.Flows) {
```
richvdh: it feels odd that we are constructing a `Flow` here and then immediately unpacking it in `checkFlowCompleted`. Why not just pass in the `[]LoginType` ?

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -420,36 +418,47 @@ func isValidMacLogin(
...
+func checkFlows(a []authtypes.LoginType, b []authtypes.LoginType) bool {
+	// Sort the slices for simple comparison
 	sort.Slice(a, func(i, j int) bool { return a[i] < a[j] })
 	sort.Slice(b, func(i, j int) bool { return b[i] < b[j] })
 
-	// Account for any extra stages a user may do unnecessarily
-	extraStages := len(b) - len(a)
+	// Account for any extra stages a user may unnecessarily do.
+	extraStages := len(a) - len(b)
 	for i := range b {
```
richvdh: this doesn't look right.

Given:
```
a = { 'A', 'B' }
b = { 'Z' }
```

does this do the right thing (ie, return false)? Please add a test for it.

Also I think you're requiring that flows are completed in the same order as they are configured, which isn't necessary.

src/github.com/matrix-org/dendrite/clientapi/routing/register_test.go
```diff
@@ -0,0 +1,86 @@
...
+	testFlow5 := authtypes.Flow{
+		[]authtypes.LoginType{
+			authtypes.LoginType("type3"),
+			authtypes.LoginType("type2"),
+			authtypes.LoginType("type1"),
+		},
+	}
+
+	if !checkFlowCompleted(testFlow1, derivedFlows) {
+		t.Error(fmt.Sprint("Failed to verify registration flow: ", testFlow1, ", from derived flows: ", derivedFlows, ". Should be true."))
```
richvdh: `fmt.Sprint` is unnecessary here

src/github.com/matrix-org/dendrite/clientapi/routing/register_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/matrix-org/dendrite/clientapi/auth/authtypes"
+)
+
+func TestFlowChecking(t *testing.T) {
+	derivedFlows := []authtypes.Flow{
```
richvdh: `derivedFlows` seems an odd name here. Should it be `allowedFlows` or something?

src/github.com/matrix-org/dendrite/clientapi/routing/register_test.go
```diff
@@ -0,0 +1,86 @@
...
+			authtypes.LoginType("type2"),
+			authtypes.LoginType("type1"),
+		},
+	}
+
+	if !checkFlowCompleted(testFlow1, derivedFlows) {
+		t.Error(fmt.Sprint("Failed to verify registration flow: ", testFlow1, ", from derived flows: ", derivedFlows, ". Should be true."))
+	}
+	if checkFlowCompleted(testFlow2, derivedFlows) {
+		t.Error(fmt.Sprint("Failed to verify registration flow: ", testFlow2, ", from derived flows: ", derivedFlows, ". Should be false."))
```
richvdh: given you're testing that `testFlow2` can *not* be verified against `derivedFlows`, I think this error message is misleading.

src/github.com/matrix-org/dendrite/clientapi/routing/register_test.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	testFlow5 := authtypes.Flow{
+		[]authtypes.LoginType{
+			authtypes.LoginType("type3"),
+			authtypes.LoginType("type2"),
+			authtypes.LoginType("type1"),
+		},
+	}
+
+	if !checkFlowCompleted(testFlow1, derivedFlows) {
```
richvdh: I think it would be nice to split each of these tests out into its own function, which is named according to what you're trying to test. You can make `derivedFlows` a global const to make that easier.

src/github.com/matrix-org/dendrite/clientapi/routing/register_test.go
```diff
@@ -0,0 +1,86 @@
+// Copyright 2017 Vector Creations Ltd
+// Copyright 2017 New Vector Ltd
```
richvdh: We don't ask you to assign copyright, and neither Vector Creations nor New Vector wrote this code, so the copyright belongs to you. You can either put your own name here or leave out the Copyright lines.

richvdh: (likewise you're free to add an assertion of your own copyright to any files you make non-trivial changes to. Not that I think it matters much from the legal point of view)

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
anoadragon453: Since they're sorted, the order in fact does not matter. You can play with the algo here:

https://play.golang.org/p/CHGHI03OBO

I've added the test case to both the playground and the .go file, both work :)

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
anoadragon453: My inner OCD wanted to only accept Flow's into checkFlowCompleted, but yes I can change it :)

src/github.com/matrix-org/dendrite/clientapi/routing/register_test.go
anoadragon453: That is nicer sounding, yes.

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
richvdh:
```
	derivedFlows = []Flow{Flow{[]LoginType{LoginType("Z")}}}
	fmt.Println("Derived Flows:", derivedFlows)

	userFlow2 := Flow{[]LoginType{LoginType("A"), LoginType("B")}}
	fmt.Println("userFlow2:", userFlow2, checkFlowCompleted(userFlow2))
```

that gives true for me. I don't call that "working".

anoadragon453: Ahh, misread your original comment and switched the allowed and input flows. 

Yes, that input does result in the incorrect output!

anoadragon453:
Should be good to go™

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -425,18 +425,21 @@ func checkFlows(a []authtypes.LoginType, b []authtypes.LoginType) bool {
 	sort.Slice(a, func(i, j int) bool { return a[i] < a[j] })
```
richvdh: hrm. This is going to modify the inputs to `checkFlows` - ie, it's going to write to `cfg.Derived.Registration.Flows` and `sessions[sessionID]`.

I guess it's harmless enough, but it's certainly unexpected. It also highlights the redundancy here in sorting for every pairing of session/flow.

My suggestion would be that you do one of:
* sort the lists much earlier and require sorted lists be passed into `checkFlowCompleted`
* rather than sorting the lists, construct a `map[LoginType]bool` for the completed stages and iterate over the steps in `b`

While I'm here, can we come up with better names than `a` and `b`? `completedStages` / `requiredStages` ?

src/github.com/matrix-org/dendrite/clientapi/routing/register_test.go
```diff
@@ -109,3 +109,26 @@ func TestFlowCheckingUnorderedAndExtraneous(t *testing.T) {
...
+func TestFlowCheckingShortIncorrectInput(t *testing.T) {
+	testFlow := []authtypes.LoginType{
+		authtypes.LoginType("stage8"),
+	}
+	if checkFlowCompleted(testFlow, allowedFlows) {
+		t.Error("Incorrect registration flow verification: ", testFlow, ", from allowed flows: ", allowedFlows, ". Should be false.")
+	}
+}
+
+// Should return false as we're providing less stages than are required.
```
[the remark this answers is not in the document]
anoadragon453: Whoops, copy/paste error.

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
anoadragon453: Would simply creating a copy of the stages before sorting them in-place be sufficient?

anoadragon453:
We no longer unexpectedly sort required and allow slices. Also switched out the `a` and `b` var names for more descriptive ones.

anoadragon453:
Heh, thanks for your own!

PR #355, opened 2017-11-24
Add OPTIONS handlers everywhere

richvdh:
Everything gets pre-flighted now, because of the Authorization header, so we
need to support OPTIONS everywhere.


richvdh:
empirically :)

PR #359, opened 2017-11-28
Refactor BuildEvents to split out AddPrevEventsToEvent

erikjohnston:
This is so that we can add prev events to the builder explicitly,
without having to actually build the event.

### DIFF
--- a/src/github.com/matrix-org/dendrite/common/events.go
+++ b/src/github.com/matrix-org/dendrite/common/events.go
@@ -42,9 +42,30 @@ func BuildEvent(
 	builder *gomatrixserverlib.EventBuilder, cfg config.Dendrite,
 	queryAPI api.RoomserverQueryAPI, queryRes *api.QueryLatestEventsAndStateResponse,
 ) (*gomatrixserverlib.Event, error) {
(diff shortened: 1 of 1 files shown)


erikjohnston:
Used by https://github.com/matrix-org/dendrite/pull/316

src/github.com/matrix-org/dendrite/common/events.go
```diff
@@ -42,11 +42,32 @@ func BuildEvent(
...
+	now := time.Now()
+	event, err := builder.Build(eventID, now, cfg.Matrix.ServerName, cfg.Matrix.KeyID, cfg.Matrix.PrivateKey)
 	if err != nil {
 		return nil, err
 	}
 
+	return &event, nil
+}
+
+// AddPrevEventsToEvent fills out the prev_events and auth_events fields in EventBuilder
```
richvdh: s/EventBuilder/builder/ ?

erikjohnston: (whoops, I got confused and merged things in a weird order and generally was a bit silly. I've fixed this directly on master)

PR #360, opened 2017-11-29
Add config option to disable registration

erikjohnston:


erikjohnston:
Sigh, and now `LegacyRegister` is above the gocyclo limit

richvdh:
> Sigh, and now LegacyRegister is above the gocyclo limit

It doesn't fit on my screen, and tbh makes my eyes hurt. I think splitting it up would be a good thing.

src/github.com/matrix-org/dendrite/common/config/config.go
```diff
@@ -83,6 +83,9 @@ type Dendrite struct {
 		// If set, allows registration by anyone who also has the shared
 		// secret, even if registration is otherwise disabled.
 		RegistrationSharedSecret string `yaml:"registration_shared_secret"`
+		// If set disables new users from registering (excpet via shared
```
richvdh: excpet

erikjohnston:
I've split out the parsing and validation into a separate function, which seemed the sanest way of splitting it.

PR #361, opened 2017-11-30
Squash username to all lowercase letters upon registration

### LINKED ISSUE #325: In registration and login APIs replace any capitals with lower case

anoadragon453:
Closes #325.


erikjohnston:
Might be worth adding a check to validateUsername to enforce lower case only too?

anoadragon453:
@erikjohnston Added the squashing to some new spots, not sure if we need to do it in Legacy Register or not. validateUsername now also enforces all lowercase letters.

PR #363, opened 2017-11-30
Pass opentracing span contexts through kafka streams

erikjohnston:
This will allow us to see how what downstream effects the request has.

This change produces the following (on a monolith):

screenshot from 2017-11-30 16-45-21

We'll probably want to figure out how to differentiate the different components better, but we need to do that for the logging anyway.

The integration test is failing because we do a bytewise comparison of the output of the roomserver, but the `opentracing_carrier` is non-deterministic. Not sure how best to fix that, thoughts? Maybe try and do a JSON subset comparison or something?

### DIFF
--- a/src/github.com/matrix-org/dendrite/roomserver/api/output.go
+++ b/src/github.com/matrix-org/dendrite/roomserver/api/output.go
@@ -15,7 +15,11 @@
 package api
 
 import (
+	"context"
+
 	"github.com/matrix-org/gomatrixserverlib"
+	opentracing "github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/ext"
 )
 
 // An OutputType is a type of roomserver output.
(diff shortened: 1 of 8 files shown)


src/github.com/matrix-org/dendrite/roomserver/api/output.go
```diff
@@ -41,6 +45,45 @@ type OutputEvent struct {
...
+	}
+
+	o.OpentracingCarrier = carrier
+
+	return nil
+}
+
+// StartSpanAndReplaceContext produces a context and opentracing span from the
+// info embedded in OutputEvent
+func (o *OutputEvent) StartSpanAndReplaceContext(
```
richvdh: I'm not sure it's fair to expect people to realise that "Span" refers to opentracing; ideally this name would make it clear, but I appreciate `StartOpentracingSpanAndReplaceContext` is not a catchy name. I'd like us to think about better names, but perhaps I need to wait for some clarity on how this fits together

erikjohnston:
FTR, thinking more about this I think that that is the wrong layer to be handling this, and we should move it.

PR #364, opened 2017-12-01
Add CORS headers to all responses including errors

erikjohnston:
While util.MakeJSONAPI handles adding CORS headers and OPTIONS requests,
that only gets hit for known APIs. This means that clients get CORS
errors for unkown APIs, rather than seeing the 404, which causes them to
retry.


richvdh:
wait, now I'm not so sure

erikjohnston:
> Does it mean that we can remove some redundant code elsewhere?

Its all done inside of the `github.com/matrix-org/util` package, so not sure what else is relying on it...

richvdh:
> Its all done inside of the github.com/matrix-org/util package, so not sure what else is relying on it...

I think having CORS handling in the util package is disasterous anyway tbh.

PR #366, opened 2017-12-05
Add some basic docs about opentracing

erikjohnston:


docs/opentracing.md
```diff
@@ -0,0 +1,112 @@
...
+```
+Component 1       |<─────────────────── HTTP ────────────────────>|
+                     |<──────────────── RPC ─────────────────>|
+Component 2                |<─ SQL ─>|     |<── RPC ───>|
+Component 3                                  |<─ SQL ─>|
+```
+
+This is achieved by serializing span information during all communication
+between components. For HTTP requests, this is achieved by the sender
+serializing the sapn into a HTTP header, and the receiver deserializing the span
```
richvdh: sapn

PR #367, opened 2017-12-05
Write and read transaction id from sync DB

erikjohnston:


erikjohnston:
I'm sorely tempted to rename `TransactionID` to something like `DeviceTransactionID` as its a bit confusing to have `TransactionID.TransactionID` as a thing. It should probably happen after this stuff lands though, otherwise the merge commits are going to be a bit horrific.

src/github.com/matrix-org/dendrite/syncapi/storage/output_room_events_table.go
```diff
@@ -44,22 +46,24 @@ CREATE TABLE IF NOT EXISTS syncapi_output_room_events (
     -- A list of event IDs which represent a delta of added/removed room state. This can be NULL
     -- if there is no delta.
     add_state_ids TEXT[],
-    remove_state_ids TEXT[]
+	remove_state_ids TEXT[],
```
richvdh: what's happened with the indentation here?

erikjohnston: ... turns out the rest of the SQL using spaces...

PR #368, opened 2017-12-05
 Add transaction ID to events if sending device

erikjohnston:
Include #367


src/github.com/matrix-org/dendrite/syncapi/storage/syncserver.go
```diff
@@ -594,18 +597,29 @@ func (d *SyncServerDatabase) getStateDeltas(
...
+			stateEvents: streamEventsToEvents(device, state[joinedRoomID]),
 			roomID:      joinedRoomID,
 		})
 	}
 
 	return deltas, nil
 }
 
-func streamEventsToEvents(in []streamEvent) []gomatrixserverlib.Event {
+func streamEventsToEvents(device *authtypes.Device, in []streamEvent) []gomatrixserverlib.Event {
```
richvdh: please could you add a comment here? what does it do with `device`?

src/github.com/matrix-org/dendrite/syncapi/storage/syncserver.go
```diff
@@ -594,18 +597,29 @@ func (d *SyncServerDatabase) getStateDeltas(
...
 	out := make([]gomatrixserverlib.Event, len(in))
 	for i := 0; i < len(in); i++ {
 		out[i] = in[i].Event
+		if device != nil && in[i].transactionID != nil {
+			if device.ID == in[i].transactionID.DeviceID {
```
richvdh: do we not need to check the userid, given you've told me that DeviceIDs are not unique?

src/github.com/matrix-org/dendrite/syncapi/storage/syncserver.go
```diff
@@ -292,7 +293,7 @@ func (d *SyncServerDatabase) CompleteSync(
 		if err != nil {
 			return nil, err
 		}
-		recentEvents := streamEventsToEvents(recentStreamEvents)
+		recentEvents := streamEventsToEvents(nil, recentStreamEvents)
```
richvdh: why doesn't this get a device?

src/github.com/matrix-org/dendrite/syncapi/storage/syncserver.go
```diff
@@ -211,7 +212,7 @@ func (d *SyncServerDatabase) syncStreamPositionTx(
 // IncrementalSync returns all the data needed in order to create an incremental sync response.
```
richvdh: this is a public function, so definitely needs documentation of what `device` does.

src/github.com/matrix-org/dendrite/syncapi/storage/syncserver.go
```diff
@@ -209,7 +215,11 @@ func (d *SyncServerDatabase) syncStreamPositionTx(
 	return types.StreamPosition(maxID), nil
 }
 
-// IncrementalSync returns all the data needed in order to create an incremental sync response.
+// IncrementalSync returns all the data needed in order to create an incremental
+// sync response for the given device. If the device has a deviceID events
```
richvdh: "... for the given device". But device can be `nil`?

richvdh: "If the device has a deviceID": when does a device not have a deviceID?

erikjohnston: I think I was getting a bit confused with synapse mainly

src/github.com/matrix-org/dendrite/syncapi/storage/syncserver.go
```diff
@@ -216,10 +216,10 @@ func (d *SyncServerDatabase) syncStreamPositionTx(
 }
 
 // IncrementalSync returns all the data needed in order to create an incremental
-// sync response for the given device. If the device has a deviceID events
-// returned will include any client transaction IDs associated with the device.
-// These transaction IDs come from when the device sent the event via an API
-// that included a transaction ID.
+// sync response for tshe given user. Events returned will include any client
+// transaction IDs associated with the given device. These transaction IDs come
```
richvdh: ... but the device can be nil, in which case...

erikjohnston: I don't think the device can be nil here, since we use it to get at the `UserID` and it has come (almost) directly from the rest layer?

erikjohnston: Hmm, should we be passing it by pointer at this point?

richvdh: > Hmm, should we be passing it by pointer at this point?

well, that is the obvious question. I'd say not if it can't be nil.

erikjohnston: Yeah, fixed. Though in general I'm becoming cautious with how often we seem to be passing things by value as some of the structs seem quite big, but Device is tiny so shouldn't matter.

PR #372, opened 2017-12-10
User registration return M_USER_IN_USE when username is already taken

CromFr:
When registering a new user using POST `/_matrix/client/r0/register`, the server was returning a 500 error when user name was already taken.

I added a check in `completeRegistration` to verify if the username is available before inserting it, and return a 400 `M_USER_IN_USE` error if there is a conflict, as [defined in matrix-doc](https://matrix.org/speculator/spec/HEAD/client_server/unstable.html#post-matrix-client-r0-register)

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/storage.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/storage.go
@@ -24,7 +24,7 @@ import (
 	"github.com/matrix-org/gomatrixserverlib"
 	"golang.org/x/crypto/bcrypt"
 	// Import the postgres database driver.
-	_ "github.com/lib/pq"
+	"github.com/lib/pq"
 )
 
 // Database represents an account database
@@ -118,7 +118,8 @@ func (d *Database) SetDisplayName(
 }
 
 // CreateAccount makes a new account with the given login name and password, and creates an empty profile
-// for this account. If no password is supplied, the account will be a passwordless account.
+// for this account. If no password is supplied, the account will be a passwordless account. If the
+// account already exists, it will return nil, nil.
 func (d *Database) CreateAccount(
 	ctx context.Context, localpart, plaintextPassword string,
 ) (*authtypes.Account, error) {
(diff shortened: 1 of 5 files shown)


src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/storage.go
```diff
@@ -127,6 +128,12 @@ func (d *Database) CreateAccount(
 		return nil, err
 	}
 	if err := d.profiles.insertProfile(ctx, localpart); err != nil {
+		if err, ok := err.(*pq.Error); ok {
+			if err.Code.Class() == "23" {
+				// 23 => unique_violation => Account already exists
+				return nil, nil
+			}
+		}
```
erikjohnston: I would be tempted to pull this out into a function in `common/sql.go`, something like `IsConstraintViolationErr(err error)`

CromFr: I just realized `err.Code.Class() == 23` represents any constraint violation.

In this case, I think we'd rather have a `IsUniqueConstraintViolationErr(err error)` that checks if the error code is `23505`, since we want to return `nil,nil` only if the account already exists

src/github.com/matrix-org/dendrite/common/sql.go
```diff
@@ -66,3 +68,9 @@ func TxStmt(transaction *sql.Tx, statement *sql.Stmt) *sql.Stmt {
 	}
 	return statement
 }
+
+// IsUniqueConstraintViolationErr returns true if the error is a postgresql unique_violation error
+func IsUniqueConstraintViolationErr(err error) bool {
+	pqErr, ok := err.(*pq.Error)
+	return ok && pqErr.Code == "23505"
```
erikjohnston: This should be `ok && err.Code.Class() == "23"`, no?

CromFr: `23505` is the specific error code for `unique_violation`. I commented on this [here](https://github.com/matrix-org/dendrite/pull/372#discussion_r156749720) but it got hidden by the following commit

erikjohnston: Hmm, having dug around it does look like only that error code gets used for all the various ways of defining primary keys/unique constraints.

erikjohnston:
Thanks for this!

(The integ tests fail because of a flakiness of travis, so merging)

erikjohnston:
Oh actually, can you sign off as per https://github.com/matrix-org/dendrite/blob/master/CONTRIBUTING.md#sign-off

CromFr:
My bad, it's ok now :)

PR #374, opened 2017-12-14
Implement some createRoom parameters

CromFr:
I'm sure there is plenty to discuss about this PR, since I'm still learning how Dendrite is put together :)

#### createRoom implementation changes:
- Set room join rules & history visibility according to `Preset` parameter (+ request preset value validation)
- Added `GuestCanJoin` parameter (see matrix-doc PR #1093)
- Process state events defined in the `InitialState` parameter
- Do not set room name if empty/not provided
- Do not set room topic if empty/not provided

#### Thoughts
- In Synapse implementation, initial events are overridden by those in the `InitialState` parameter (if they have the same event type), to reduce the number of sent events. It may be good to do the same here.
- Made `fledglingEvent` convertible from json. It may be good to rename it to something more generic like `stateEvent`
- I'm appending events to be sent to the room server in a slice. There may be a nicer/cleaner/more efficient way? (ie pre-allocate the slice)

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/routing/createroom.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/routing/createroom.go
@@ -55,13 +55,11 @@ const (
 
 const (
 	joinRulePublic = "public"
-	// joinRuleCanJoin = "can_join"
 	joinRuleInvite = "invite"
-	// joinRuleKnock   = "knock"
-	// joinRulePrivate = "private"
 )
 const (
 	historyVisibilityShared = "shared"
+	// TODO: These should be implemented once history visibility is implemented
 	// historyVisibilityWorldReadable = "world_readable"
 	// historyVisibilityInvited       = "invited"
 )
@@ -222,11 +220,9 @@ func createRoom(req *http.Request, device *authtypes.Device,
 	if r.Topic != "" {
 		eventsToMake = append(eventsToMake, fledglingEvent{"m.room.topic", "", common.TopicContent{Topic: r.Topic}})
 	}
-	// eventsToMake = append(eventsToMake,
-	// 	// TODO: invite events
-	// 	// TODO: 3pid invite events
-	// 	// TODO: m.room.aliases
-	// )
+	// TODO: invite events
+	// TODO: 3pid invite events
+	// TODO: m.room.aliases
 
 	authEvents := gomatrixserverlib.NewAuthEvents(nil)
 	for i, e := range eventsToMake {
--- a/src/github.com/matrix-org/dendrite/clientapi/routing/createroom.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/routing/createroom.go
@@ -15,7 +15,6 @@
 package routing
 
 import (
-	"encoding/json"
 	"fmt"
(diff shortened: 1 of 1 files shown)


src/github.com/matrix-org/dendrite/clientapi/routing/createroom.go
```diff
@@ -43,10 +42,30 @@ type createRoomRequest struct {
...
+	presetTrustedPrivateChat = "trusted_private_chat"
+	presetPublicChat         = "public_chat"
+)
+
+const (
+	joinRulePublic = "public"
+	// joinRuleCanJoin = "can_join"
+	joinRuleInvite = "invite"
+	// joinRuleKnock   = "knock"
+	// joinRulePrivate = "private"
```
erikjohnston: I believe `private` is implemented so we can uncomment it, and I wouldn't bother with keeping knock or can_join there so just delete them.

src/github.com/matrix-org/dendrite/clientapi/routing/createroom.go
```diff
@@ -43,10 +42,30 @@ type createRoomRequest struct {
...
+)
+const (
+	historyVisibilityShared = "shared"
+	// historyVisibilityWorldReadable = "world_readable"
+	// historyVisibilityInvited       = "invited"
```
erikjohnston: Would be good to put a `// TODO: These should be implemented once history visibility is implemented`

src/github.com/matrix-org/dendrite/clientapi/routing/createroom.go
```diff
@@ -166,16 +209,24 @@ func createRoom(req *http.Request, device *authtypes.Device,
...
+		eventsToMake = append(eventsToMake, fledglingEvent{"m.room.name", "", common.NameContent{Name: r.Name}})
+	}
+	if r.Topic != "" {
+		eventsToMake = append(eventsToMake, fledglingEvent{"m.room.topic", "", common.TopicContent{Topic: r.Topic}})
 	}
+	// eventsToMake = append(eventsToMake,
+	// 	// TODO: invite events
+	// 	// TODO: 3pid invite events
+	// 	// TODO: m.room.aliases
+	// )
```
erikjohnston: Just leave these as plain `// TODO:` rather than part of commented out code

src/github.com/matrix-org/dendrite/clientapi/routing/createroom.go
CromFr: `private` isn't used in this file, and I haven't found any other occurrence across other files.
Also [the doc](https://matrix.org/speculator/spec/HEAD/client_server/unstable.html#m-room-join-rules) states that `private` is an unimplemented keyword (and I found no implementation in Synapse)

I think there should be only `public` and `invite` here.

Btw `can_join` was just a terrible mix up/copy paste error (it's a `guest_access` value)

erikjohnston: Oh, hmm, empirically `private` does work on synapse I think, but if its not specced then yeah, may as well leave it out for now.

PR #377, opened 2017-12-17
Load Application Service Configuration Files

anoadragon453:
WIP for all of application services. This'll load the config of each AS specified in `dendrite.yaml` and unmarshal the values out into `config.Dendrite.Derived.ApplicationServices`, which is a slice of type `ApplicationService`.

`yaml.v2` was updated for its new `UnmarshalStrict()` method, which will error out if any types don't match from the YAML to a struct.

### DIFF
--- a/dendrite-config.yaml
+++ b/dendrite-config.yaml
@@ -120,3 +120,7 @@ tracing:
     # for documtation.
     jaeger:
         disabled: true
+
+# A list of application service config files to use
+applicationservice:
+    app_service_config_files:
--- a/dendrite-config.yaml
+++ b/dendrite-config.yaml
@@ -123,4 +123,4 @@ tracing:
 
 # A list of application service config files to use
 applicationservice:
-    app_service_config_files:
+    app_service_config_files: []
--- a/dendrite-config.yaml
+++ b/dendrite-config.yaml
@@ -122,5 +122,5 @@ tracing:
         disabled: true
 
 # A list of application service config files to use
-applicationservice:
-    app_service_config_files: []
+application_services:
+    config_files: []
--- a/src/github.com/matrix-org/dendrite/common/config/appservice.go
+++ b/src/github.com/matrix-org/dendrite/common/config/appservice.go
@@ -1,0 +1,79 @@
+// Copyright 2017 Andrew Morgan <andrew@amorgan.xyz>
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config
+
+import (
+	"io/ioutil"
+	"path/filepath"
+
+	"gopkg.in/yaml.v2"
+)
+
+// ApplicationServiceNamespace is the namespace that a specific application
+// service has management over.
+type ApplicationServiceNamespace struct {
+	// Whether or not the namespace is managed solely by this application service
+	Exclusive bool `yaml:"exclusive"`
+	// A regex pattern that represents the namespace
+	Regex string `yaml:"regex"`
+}
+
+// ApplicationService represents a Matrix application service.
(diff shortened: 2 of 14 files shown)


anoadragon453:
Travis error seems to be unrelated to this PR.

dendrite-config.yaml
```diff
@@ -120,3 +120,7 @@ tracing:
     # for documtation.
     jaeger:
         disabled: true
+
+# A list of application service config files to use
+applicationservice:
+    app_service_config_files:
```
erikjohnston: I think you need to specify an empty list in yaml? I.e.:

```yaml
applicationservice:
    app_service_config_files: []
```

src/github.com/matrix-org/dendrite/common/config/appservice.go
```diff
@@ -0,0 +1,79 @@
...
+	// Base URL of the application service
+	URL string `yaml:"url"`
+	// Application service token provided in requests to a homeserver
+	ASToken string `yaml:"as_token"`
+	// Homeserver token provided in requests to an application service
+	HSToken string `yaml:"hs_token"`
+	// Localpart of application service user
+	SenderLocalpart string `yaml:"sender_localpart"`
+	// Information about an application service's namespaces
+	Namespaces map[string][]interface{} `yaml:"namespaces"`
```
erikjohnston: Should really use something other than `interface{}` here, config type checking at start up ftw

src/github.com/matrix-org/dendrite/common/config/config.go
```diff
@@ -360,6 +371,14 @@ func (config *Dendrite) derive() {
 		config.Derived.Registration.Flows = append(config.Derived.Registration.Flows,
 			authtypes.Flow{Stages: []authtypes.LoginType{authtypes.LoginTypeDummy}})
 	}
+
+	// Load application service configuration files
+	fmt.Println(config.ApplicationService.ConfigFiles)
+	if len(config.ApplicationService.ConfigFiles) > 0 {
+		if err := loadAppservices(config); err != nil {
+			fmt.Println("Error loading AS config: ", err)
```
erikjohnston: should return an error rather than printing and continuing

src/github.com/matrix-org/dendrite/common/config/config.go
```diff
@@ -360,6 +371,14 @@ func (config *Dendrite) derive() {
 		config.Derived.Registration.Flows = append(config.Derived.Registration.Flows,
 			authtypes.Flow{Stages: []authtypes.LoginType{authtypes.LoginTypeDummy}})
 	}
+
+	// Load application service configuration files
+	fmt.Println(config.ApplicationService.ConfigFiles)
```
erikjohnston: This should be a `logrus` thing, no? Or do we not have logging set up at this point yet?

erikjohnston:
> Travis error seems to be unrelated to this PR.

Travis is failing the integration test runs even though they pass for reasons unclear, so just ignore them for now :(

src/github.com/matrix-org/dendrite/common/config/config.go
anoadragon453: Oops, knew I left debugging in there somewhere.

anoadragon453:
Fixed up!

src/github.com/matrix-org/dendrite/common/config/config.go
```diff
@@ -206,6 +206,13 @@ type Dendrite struct {
 		Jaeger jaegerconfig.Configuration `yaml:"jaeger"`
 	}
 
+	// Application Services
+	// https://matrix.org/docs/spec/application_service/unstable.html
+	ApplicationService struct {
+		// Configuration files for various application services
+		ConfigFiles []string `yaml:"app_service_config_files"`
+	}
```
erikjohnston: This needs a `yaml:"application_services"`  or whatever

src/github.com/matrix-org/dendrite/common/config/config.go
```diff
@@ -360,6 +375,15 @@ func (config *Dendrite) derive() {
 		config.Derived.Registration.Flows = append(config.Derived.Registration.Flows,
 			authtypes.Flow{Stages: []authtypes.LoginType{authtypes.LoginTypeDummy}})
 	}
+
+	// Load application service configuration files
+	if len(config.ApplicationService.ConfigFiles) > 0 {
+		if err := loadAppservices(config); err != nil {
+			return err
+		}
+	}
```
erikjohnston: I wouldn't bother putting the non-empty guard there, the `loadAppservices` function is happy with getting a config without any appservices I think.

PR #378, opened 2017-12-17
Update go-yaml/yaml

anoadragon453:
Update YAML library for new features such as `UnmarshalStrict()`. Used by #377.


erikjohnston:
Generally we're fine with bumping the dependency versions as part of the PRs that use them (as long as they're in separate commits still).

Issue #346, opened 2017-11-20
Need a way to enable debug logging


richvdh:
We could do with a way of configuring dendrite to enable debug logs. Presumably via an env var, like LOG_DIR?

While we're at it, perhaps we could get rid of the separate error/warn/info logs. I'm not a fan of having three logfiles to check.

anoadragon453:
The way I've seen it done in other projects is you'll have different levels of logging, where each has its own place in a hierarchy, and in a config file you can specify which level and above that you'd like printed to the logs.

Then each line of logging would be prefixed with its level. Developers may want maximum logging, whereas users may only want errors and above.

An example of levels from [matrix-appservice-discord](https://github.com/Half-Shot/matrix-appservice-discord/blob/7a1c27a3f47962b46468190fabbf8ab09e9ef5a2/config/config.sample.yaml#L14):

```
logging:
   level: "warn" #silly, verbose, info, http, warn, error
```

erikjohnston:
As a tangential thought: it would be nice to figure out how we can have different [logrus hooks](https://github.com/sirupsen/logrus#hooks), e.g. I was playing with putting dendrite logs into logstash. Right now I need to edit the code.

ElyKar:
My thoughts on this are the same as anoa: the logging level is set in a config file, individual lines are prefixed by the level.
In the case of dendrite, the possibilities should match logrus's levels: debug, info, warn, error, fatal and panic.

Breaking this down a little bit (though there's not much) :
 - Add a logging.level to the configuration
 - Also, change the configuration of the directory from the LOG_DIR env var to a logging.path parameter (mixing two ways of configuring logs seems off, and this is the only env var required for configuring dendrite) 
- Modify the library github.com/matrix-org/dugong to accomodate the changes

For supporting various logrus hooks, I have no idea how to do it apart from implementing them one by one. Though, a few of them may be interesting to have (journald, syslog, Logstash, ...).

Kegsay:
I agree with @deedubs on this. We don't have any real users on Dendrite yet, so it's easy to remove, and I really don't want to be supporting log stuff. Spin it up in docker and export stdout to a central logging location.

The ability to configure logging landed in https://github.com/matrix-org/dendrite/pull/379 - this issue is no longer relevant.

PR #379, opened 2017-12-18
Enable debug logs using a configuration parameter and put all logs in a single file

ElyKar:
Enable debug logs using a configuration parameter and put all logs in a single file as per #346 .
Linked to the PR on [dugong](https://github.com/matrix-org/dugong/pull/6) so vendors have not yet been updated.

### DIFF
--- a/dendrite-config.yaml
+++ b/dendrite-config.yaml
@@ -117,6 +117,13 @@ listen:
 tracing:
     # Config for the jaeger opentracing reporter.
     # See https://godoc.org/github.com/uber/jaeger-client-go/config#Configuration
-    # for documtation.
+    # for documentation.
     jaeger:
         disabled: true
+
+# The configuration for logs of dendrite
+logging:
+    # The logging level, must be one of debug, info, warn, error, fatal, panic.
+    level: "info"
+    # The file on which save logs. If commented, logs won't be saved by dendrite and only printed to stdout
+    # path: /var/log/dendrite.log
--- a/dendrite-config.yaml
+++ b/dendrite-config.yaml
@@ -121,9 +121,19 @@ tracing:
     jaeger:
         disabled: true
 
-# The configuration for logs of dendrite
+# The configuration for dendrite logs
 logging:
-    # The logging level, must be one of debug, info, warn, error, fatal, panic.
-    level: "info"
-    # The file on which save logs. If commented, logs won't be saved by dendrite and only printed to stdout
-    # path: /var/log/dendrite.log
+    # The logging type, only "file" is supported at the moment
+    - type: "file"
+      # The logging level, must be one of debug, info, warn, error, fatal, panic.
+      level: "info"
+      # Parameters for this type of log
+      params:
+        # File logging must be given a path to a directory. Each component will write to a different file. Logs are rotated each day and gzipped
+        path: "/var/log/dendrite"
+    # It is possible to have multiple logging hooks at the same time.
+    # To save only errors in a different directory, uncomment the following.
+    # - type: "file"
+    #   level: "error"
+    #   params:
+    #     path: "/var/log/dendrite/errors"
--- a/src/github.com/matrix-org/dendrite/common/config/config.go
+++ b/src/github.com/matrix-org/dendrite/common/config/config.go
@@ -206,14 +206,8 @@ type Dendrite struct {
 		Jaeger jaegerconfig.Configuration `yaml:"jaeger"`
 	}
 
-	// The config for logging informations
-	Logging struct {
-		// The path to the log file
-		FPath Path `yaml:"path"`
-
-		// The logging level
-		Level string `yaml:"level"`
-	} `yaml:"logging"`
+	// The config for logging informations. Each hook will be added to logrus.
+	Logging []Hook `yaml:"logging"`
 
 	// Any information derived from the configuration options for later use.
 	Derived struct {
@@ -228,9 +222,6 @@ type Dendrite struct {
 			// registration in order to complete registration stages.
 			Params map[string]interface{} `json:"params"`
 		}
-
-		// The logrus level used for logging configuration
-		LogLevel logrus.Level
 	}
 }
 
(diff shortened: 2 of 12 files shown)


src/github.com/matrix-org/dendrite/common/log.go
```diff
@@ -42,12 +42,15 @@ func SetupLogging(logDir string) {
...
 	})
-	if logDir != "" {
-		_ = os.Mkdir(logDir, os.ModePerm)
+}
+
+// SetupFileLogging configures the logging format to a file.
+func SetupFileLogging(logFile string, logLevel logrus.Level) {
+	logrus.SetLevel(logLevel)
+	if logFile != "" {
+		_ = os.MkdirAll(path.Dir(logFile), os.ModePerm)
```
erikjohnston: Do we not want to fail here if we can't create the directory?

dendrite-config.yaml
```diff
@@ -117,6 +117,13 @@ listen:
...
+    # for documentation.
     jaeger:
         disabled: true
+
+# The configuration for logs of dendrite
+logging:
+    # The logging level, must be one of debug, info, warn, error, fatal, panic.
+    level: "info"
+    # The file on which save logs. If commented, logs won't be saved by dendrite and only printed to stdout
+    # path: /var/log/dendrite.log
```
erikjohnston: If the expectation is for us to support multiple types of loggers, I'd be tempted to have an extensible format, e.g.:

```yaml
logging:
- type: file
  level: info
  params:
    path: /var/log/dendrite.log
```

This would be useful even in just the file case, where you might want to define a separate file for error logs (so that you could e.g. have different retention periods for the error logs)

src/github.com/matrix-org/dendrite/common/log.go
```diff
@@ -44,13 +67,46 @@ func SetupStdLogging() {
...
-func SetupFileLogging(logFile string, logLevel logrus.Level) {
-	logrus.SetLevel(logLevel)
-	if logFile != "" {
-		_ = os.MkdirAll(path.Dir(logFile), os.ModePerm)
-		logrus.AddHook(dugong.NewFSHook(
-			logFile,
+// SetupHookLogging configures the logging hooks defined in the configuration.
+// If something fails here it means that the logging was improperly configured,
+// so we just exit with the error
+func SetupHookLogging(hooks []config.Hook, componentName string) {
```
richvdh: This is an odd name, but I'm not sure I have any better suggestions :/

src/github.com/matrix-org/dendrite/common/log.go
```diff
@@ -44,13 +67,46 @@ func SetupStdLogging() {
 	})
 }
 
-// SetupFileLogging configures the logging format to a file.
-func SetupFileLogging(logFile string, logLevel logrus.Level) {
-	logrus.SetLevel(logLevel)
```
richvdh: I have concerns over the performance implications of getting rid of this. For example, if we use `SetLevel` to set the log level to "Info", then any `logrus.Debug` calls are skipped quickly. 

We should probably use `SetLevel` to set the global log threshold as high as we can.

richvdh:
@erikjohnston any idea why travis isn't running this?

src/github.com/matrix-org/dendrite/common/config/config.go
```diff
@@ -261,6 +252,52 @@ type ThumbnailSize struct {
 	ResizeMethod string `yaml:"method,omitempty"`
 }
 
+// Hook represents a single logrus hook. At this point, only parsing and
+// verification of the proper values for type and level are done.
+// Validity/integrity checks on the parameters are done when configuring logrus.
+type Hook struct {
```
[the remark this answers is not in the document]
ElyKar: LogrusHook maybe ?

src/github.com/matrix-org/dendrite/common/config/config.go
```diff
@@ -261,6 +252,52 @@ type ThumbnailSize struct {
...
+		Type   string                 `yaml:"type"`
+		Level  string                 `yaml:"level"`
+		Params map[string]interface{} `yaml:"params"`
+	}
+
+	if err := unmarshaler(&tmp); err != nil {
+		return err
+	}
+
+	if tmp.Type != "file" {
```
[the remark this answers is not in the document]
ElyKar: I wasn't really either. I am unsure about whether we sould check this in the config.go directly or in the log.go.
To get rid of the UnmarshalYAML, is that fine to just duplicate all logging elements under config.Dendrite.Derived ? (Because the log level needs processing to be retrieved)

src/github.com/matrix-org/dendrite/common/config/config.go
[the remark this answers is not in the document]
richvdh: sure

src/github.com/matrix-org/dendrite/common/config/config.go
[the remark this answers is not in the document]
richvdh: why not keep it as a string?

src/github.com/matrix-org/dendrite/common/config/config.go
```diff
@@ -444,6 +412,7 @@ func (e Error) Error() string {
 
 // check returns an error type containing all errors found within the config
 // file.
+// nolint: gocyclo
```
richvdh: please don't do this. Factor out a separate function to keep the cyclomatic complexity count down.

src/github.com/matrix-org/dendrite/common/log.go
```diff
@@ -70,14 +70,27 @@ func SetupStdLogging() {
 // SetupHookLogging configures the logging hooks defined in the configuration.
 // If something fails here it means that the logging was improperly configured,
 // so we just exit with the error
-func SetupHookLogging(hooks []config.Hook, componentName string) {
+func SetupHookLogging(hooks []config.LogrusHook, componentName string) {
 	for _, hook := range hooks {
+
+		// Ensures we received a proper logging level
```
richvdh: unless you're going to write a long comment block ("The following code ensures we received a proper logging level by doing the X and then the Y"), comments should be written in the imperative: s/Ensures/Check/ please.

src/github.com/matrix-org/dendrite/common/log.go
```diff
@@ -70,14 +70,27 @@ func SetupStdLogging() {
...
+		level, err := logrus.ParseLevel(hook.Level)
+		if err != nil {
+			logrus.Fatalf("Unrecognized logging level %s: %q", hook.Level, err)
+		}
+
+		// Perform a first filter on the logs according to the lowest level of all
+		// (Ex: If we have hook for info and above, prevent logrus from processing debug logs)
```
richvdh: s/ex/eg/

src/github.com/matrix-org/dendrite/common/config/config.go
ElyKar: This function will likely keep growing with the configuration.
Rather than just putting the verification of the logging config in an external function, what about having one function per first-order element in the config ?
Something like `checkMatrix` which checks parameters `matrix.*`, `checkMedia` which checks parameters `media.*`, and so on.
Essentially, refactoring the entire `check` function rather than just moving the last loop.

PR #382, opened 2017-12-19
Make travis builds a bit faster

richvdh:
* cache the kafka download
* only clone the last git commit

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,7 @@
 .*
 
 # Downloads
-kafka.tgz
+/.downloads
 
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -4,7 +4,7 @@
 .*
 
 # Downloads
-kafka.tgz
+/.downloads
```
erikjohnston: Does the forward slash prefix mean only for the top directory of the git checkout?

richvdh: yes.

PR #384, opened 2017-12-19
Factor out component setup from commands

erikjohnston:
Currently the (rather complicated) setup of components is done in the various commands, with everything being duplicated in the monolith command. This makes it hard to a) write new components, b) change how a component works or a change dependency, c) understand the various interdependencies of bits of components, etc

This PR fixes things by:
- Having a base component that reads the config and sets up everything that is common to all components, e.g. kafka producer/consumers, instances of the roomserver APIs, etc.
- Move component specific setup into the package of that component
- Make the commands defer to base component and component specific setup functions.
- Interdependencies are explicit by having the setup component functions take them as parameters, e.g. syncapi components requires instances of the Device and Account databases.


erikjohnston:
Thanks!

PR #387, opened 2017-12-24
Set up a development environment with docker

mujx:
Since dendrite has a lot of moving parts I thought it would be faster and easier to create and maintain a working dev environment with docker and docker-compose. Let me know if you're in favor of such a setup.

#### Starting a monolith server

Listening on `localhost:8008` & `localhost:8448`

```bash
docker-compose up postgres
```

and in a separate terminal

```bash
docker-compose up monolith
```

#### Starting a multiprocess server (defaults)

```bash
docker-compose up postgres kafka zookeeper
```
and then in separate terminals the rest of the components.

```
docker-compose up client_api_proxy
```

```
docker-compose up client_api
```

```
docker-compose up room_server
```

The setup was tested with this [config](https://pastebin.com/5SmLVx1e) (toggling naffka as needed) and Riot as the client.


ara4n:
So we did already have https://github.com/matrix-org/dendrite/pull/140, which unhelpfully has been closed despite presumably having useful stuff in it... and thus making it almost impossible to find.  Sorry about that.  It might be worth comparing the two and looking at the feedback there in case any of it applies here.

I am hugely in favour of having first-class Docker support for Dendrite and would be well up for merging this and ensuring it doesn't get lost like the last attempt.  Many many thanks for contributing it @mujx!

mujx:
I wasn't aware of that PR, thanks for pointing me to it. From a quick look they seem pretty similar (although a few things could be simplified in my PR), so I will wait for your feedback.

Good to hear that you're in favor! Production containers for dendrite would be a big plus with the new architecture.

ara4n:
i should also warn that nobody on the core team is currently particularly enthusiastic about docker, and we don’t use it in our prod or dev envs, so it might be worth getting review on this from someone who’s a docker expert rather than from the core team; we’ll merge it if review is positive :)

mujx:
Thanks for the heads up. Just for the record the PR is meant to setup a simple *dev* environment just to make development and on-boarding (for newbies to the project like me) a little easier. This is *not* a production-ready setup.

erikjohnston:
Sorry that there wasn't an issue for Docker pointing to the old PR, but leaving PRs open for the rest of eternity that may or may not work doesn't seem that helpful either.

I don't really know much about Docker, so can't really review those parts, but I wonder if it's possible to move the Dockerfile to a directory? Not sure if that's possible/sensible/easy, but would potentially make it clearer that this was for development rather than production.

docker/README.md
```diff
@@ -0,0 +1,105 @@
...
+    container_name: dendrite_room_server
+    hostname: new_component
+    # Start up script.
+    entrypoint: ["bash", "./docker/services/new-component.sh"]
+    # Use the common Dockerfile for all the dendrite components.
+    build: ./
+    volumes:
+      - type: bind
+        source: ../
+        target: /build
```
jcgruenhage: Any reason you don't just do ` - ..:/build `? That should work just fine

docker/README.md
```diff
@@ -0,0 +1,105 @@
...
+    depends_on:
+      - another_component
+    networks:
+      - internal
+    # Define which ports to expose to the internal network.
+    ports:
+      - 1234
```
jcgruenhage: afaik the ports section defines which ports are exposed on the host, and all ports are exposed on the internal network by default, so only 8008 and 8448 should need to be exposed ever.

docker/services/client-api-proxy.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+bash ./docker/build.sh
```
jcgruenhage: having this line in every components script means that dendrite is built one time per component for each time we want to (re)start it. do we really want that?

docker/README.md
```diff
@@ -0,0 +1,105 @@
...
+For the monolith server you would need a postgres instance 
+
+```
+docker-compose up postgres
+```
+
+and the dendrite component from `bin/dendrite-monolith-server`
+
+```
+docker-compose up monolith
```
jcgruenhage: Wouldn't it make more sense to split monolith and polylith into two files here? The advantage would be that we could just do `docker-compose -f docker/polylith.yml` instead of having all these commands you put below.

docker/services/client-api-proxy.sh
mujx: We want that during development to compile the changes. If there isn't something to build, it's practically a no-op.

docker/README.md
mujx: Not really, I will use the short syntax.

docker/services/client-api-proxy.sh
jcgruenhage: Oh, okay, if gb caches all that then that's fine.

docker/README.md
mujx: Thanks! I will remove the extra entries.

docker/README.md
mujx: The issue I encountered with running `docker-compose up` is that sometimes the ports won't be available immediately after the container is up (i.e postgres) so the other dependent components will fail.

For this to work we would need to add a script that performs this kind of check (e.g https://docs.docker.com/compose/startup-order/)

erikjohnston: We could probably add an option to the binaries to retry database connections on startup, which would solve this I think.

jcgruenhage: retrying this is probably better anyway, because it will save a lot of trouble in automated deployment situations.

erikjohnston:
Thanks!

vclang:
@mujx @erikjohnston 
I had a try on docker.  No problem on my mac, but get the error below on the centos environment after ".. up monolith", seems to be a path issue?
/* env info
[root@i-jbwdb8jn dendrite]# gb env
GB_PROJECT_DIR="/root/go/src/github.com/matrix-org/dendrite"
GB_SRC_PATH="/root/go/src/github.com/matrix-org/dendrite/src:/root/go/src/github.com/matrix-org/dendrite/vendor/src"
GB_PKG_DIR="/root/go/src/github.com/matrix-org/dendrite/pkg/linux-amd64"
GB_BIN_SUFFIX="-linux-amd64"
GB_GOROOT="/usr/local/go"
[root@i-jbwdb8jn dendrite]# docker version
Client:
 Version:       17.12.0-ce
 API version:   1.35
 Go version:    go1.9.2
 Git commit:    c97c6d6
 Built: Wed Dec 27 20:10:14 2017
 OS/Arch:       linux/amd64

Server:
 Engine:
  Version:      17.12.0-ce
  API version:  1.35 (minimum version 1.12)
  Go version:   go1.9.2
  Git commit:   c97c6d6
  Built:        Wed Dec 27 20:12:46 2017
  OS/Arch:      linux/amd64
  Experimental: false
[root@i-jbwdb8jn dendrite]# docker-compose version
docker-compose version 1.18.0, build 8dd22a9
docker-py version: 2.6.1
CPython version: 2.7.13
OpenSSL version: OpenSSL 1.0.1t  3 May 2016
[root@i-jbwdb8jn dendrite]# 
*/

ERROR while "# docker-compose -f docker/docker-compose.yml up monolith"
......
(12/13) Installing libssl1.0 (1.0.2n-r0)
(13/13) Installing openssl (1.0.2n-r0)
Executing busybox-1.26.2-r9.trigger
Executing ca-certificates-20161130-r2.trigger
OK: 37 MiB in 25 packages
Removing intermediate container 254beffe2947
 ---> f0ce28816118
Step 5/5 : CMD ["bash", "docker/build.sh"]
 ---> Running in c5790fa6f96c
Removing intermediate container c5790fa6f96c
 ---> 060726b63772

Successfully built 060726b63772
Successfully tagged docker_monolith:latest
[root@i-jbwdb8jn dendrite]# docker-compose -f docker/docker-compose.yml up monolith
Creating dendrite_monolith ... done
Creating dendrite_monolith ... 
Attaching to dendrite_monolith
dendrite_monolith       | Generating keys ...
dendrite_monolith       | Generating a 4096 bit RSA private key
dendrite_monolith       | ...............................................++
dendrite_monolith       | ......................++
dendrite_monolith       | writing new private key to 'server.key'
dendrite_monolith       | -----
## **dendrite_monolith       | ./docker/build.sh: line 17: /build/bin/generate-keys: No such file or directory
dendrite_monolith       | ./docker/services/monolith.sh: line 5: ./bin/dendrite-monolith-server: No such file or directory**
dendrite_monolith exited with code 127
[root@i-jbwdb8jn dendrite]#

mujx:
@vclang Did you run any commands before `docker-compose up monolith`? I don't see the output of gb, building dendrite.

vclang:
@mujx    I checked the monolith image, seems that all exec files under /build/bin can not be execute, always report "no such file".     I tried to rebuild everything and cross compile dendrite by "env GOOS=linux GOARCH=arm gb build"   again, this time the docker image works fine.  on mac I use the same compile to generate execute files for docker images.  

on centos, seems the gb build the dendrite execute files as linux amd64, and it can not be executed inside that image.

Anyway, this problem can be resolve by rebuild dendrite into arm ARCH

mujx:
@vclang The point of docker is to build through it and not on the host with cross-compilation. You can remove the build artifacts and try running the container again which will build dendrite.

vclang:
@mujx got your point, yes, I have my dendrite project built on host.  so it means I need to remove all these file before switching to docker mode.  I have a suggestion as below. Since docker build share the folder "bin",  and output execute files into the same folder.  Let's update the build.sh to force fresh gb build "gb build -f", and always override execute files under "bin" folder with right output from docker-compose.  It will prevent this issue, i tried, it works fine.  Another issue is if I want to launch the dendrite from host, I need to build again, but this time, at least the OS will tell me the root cause:

[root@i-c5s3azlu dendrite]# more docker/build.sh
#!/bin/bash

gb build -f
......

mujx:
I agree on using `gb build -f`. @erikjohnston had the same issue.

PR #388, opened 2018-01-05
Filter /sync implementation

CromFr:
Implement filtering for /sync requests

## SQL storage
```sql
-- Quick script to upgrade database (note: columns will be bad ordered, and no index is created => bad perf)
ALTER TABLE syncapi_output_room_events ADD COLUMN IF NOT EXISTS type text;
UPDATE syncapi_output_room_events SET type=(event_json::json->>'type');
ALTER TABLE syncapi_output_room_events ALTER COLUMN type SET NOT NULL;
ALTER TABLE syncapi_output_room_events  ADD COLUMN IF NOT EXISTS sender text;
UPDATE syncapi_output_room_events SET sender=(event_json::json->>'sender');
ALTER TABLE syncapi_output_room_events ALTER COLUMN sender SET NOT NULL;
ALTER TABLE syncapi_output_room_events  ADD COLUMN IF NOT EXISTS contains_url bool;
UPDATE syncapi_output_room_events SET contains_url=(event_json::json->>'content')::json->>'url' IS NOT NULL;
ALTER TABLE syncapi_output_room_events ALTER COLUMN contains_url SET NOT NULL;

ALTER TABLE syncapi_current_room_state  ADD COLUMN IF NOT EXISTS sender text;
UPDATE syncapi_current_room_state SET sender=(event_json::json->>'sender');
ALTER TABLE syncapi_current_room_state ALTER COLUMN sender SET NOT NULL;
ALTER TABLE syncapi_current_room_state  ADD COLUMN IF NOT EXISTS contains_url bool;
UPDATE syncapi_current_room_state SET contains_url=(event_json::json->>'content')::json->>'url' IS NOT NULL;
ALTER TABLE syncapi_current_room_state ALTER COLUMN contains_url SET NOT NULL;

ALTER TABLE syncapi_invite_events ADD COLUMN IF NOT EXISTS type text;
UPDATE syncapi_invite_events SET type=(event_json::json->>'type');
ALTER TABLE syncapi_invite_events ALTER COLUMN type SET NOT NULL;
ALTER TABLE syncapi_invite_events  ADD COLUMN IF NOT EXISTS sender text;
UPDATE syncapi_invite_events SET sender=(event_json::json->>'sender');
ALTER TABLE syncapi_invite_events ALTER COLUMN sender SET NOT NULL;
ALTER TABLE syncapi_invite_events  ADD COLUMN IF NOT EXISTS contains_url bool;
UPDATE syncapi_invite_events SET contains_url=(event_json::json->>'content')::json->>'url' IS NOT NULL;
ALTER TABLE syncapi_invite_events ALTER COLUMN contains_url SET NOT NULL;

ALTER TABLE syncapi_account_data_type ADD COLUMN IF NOT EXISTS sender text;
UPDATE syncapi_account_data_type SET sender=user_id; -- Not accurate, probably best guess
ALTER TABLE syncapi_account_data_type ALTER COLUMN sender SET NOT NULL;
```

## Work left
- filter.event_format not handled (nor implemented on Synapse)
- filter.event_fields not handled (I couldn't find a satisfying implementation with my current golang serialization knowledge)
- __QUESTION__: Should we limit abusive FilterPart.Limit values (ie number of returned events)?

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/filter_table.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/filter_table.go
@@ -17,8 +17,9 @@ package accounts
 import (
 	"context"
 	"database/sql"
+	"encoding/json"
 
-	"github.com/matrix-org/gomatrixserverlib"
+	"github.com/matrix-org/gomatrix"
 )
 
 const filterSchema = `
@@ -71,20 +72,29 @@ func (s *filterStatements) prepare(db *sql.DB) (err error) {
 
 func (s *filterStatements) selectFilter(
 	ctx context.Context, localpart string, filterID string,
-) (filter []byte, err error) {
-	err = s.selectFilterStmt.QueryRowContext(ctx, localpart, filterID).Scan(&filter)
-	return
+) (*gomatrix.Filter, error) {
+	var filterData []byte
+	err := s.selectFilterStmt.QueryRowContext(ctx, localpart, filterID).Scan(&filterData)
+	if err != nil {
+		return nil, err
+	}
+
+	var filter gomatrix.Filter
+	if err = json.Unmarshal(filterData, &filter); err != nil {
+		return nil, err
+	}
+	return &filter, err
 }
 
 func (s *filterStatements) insertFilter(
-	ctx context.Context, filter []byte, localpart string,
+	ctx context.Context, filter *gomatrix.Filter, localpart string,
 ) (filterID string, err error) {
 	var existingFilterID string
 
 	// This can result in a race condition when two clients try to insert the
 	// same filter and localpart at the same time, however this is not a
 	// problem as both calls will result in the same filterID
-	filterJSON, err := gomatrixserverlib.CanonicalJSON(filter)
+	filterJSON, err := json.Marshal(filter)
 	if err != nil {
 		return "", err
 	}
--- a/src/github.com/matrix-org/dendrite/syncapi/storage/syncserver.go
+++ b/src/github.com/matrix-org/dendrite/syncapi/storage/syncserver.go
@@ -178,10 +178,11 @@ func (d *SyncServerDatabase) GetStateEvent(
 // Returns an empty slice if no state events could be found for this room.
 // Returns an error if there was an issue with the retrieval.
 func (d *SyncServerDatabase) GetStateEventsForRoom(
-	ctx context.Context, roomID string,
+	ctx context.Context, roomID string, stateFilter *gomatrix.FilterPart,
 ) (stateEvents []gomatrixserverlib.Event, err error) {
+
 	err = common.WithTransaction(d.db, func(txn *sql.Tx) error {
-		stateEvents, err = d.roomstate.selectCurrentState(ctx, txn, roomID)
+		stateEvents, err = d.roomstate.selectCurrentState(ctx, txn, roomID, stateFilter)
 		return err
 	})
 	return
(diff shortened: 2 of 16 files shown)


src/github.com/matrix-org/dendrite/syncapi/storage/syncserver.go
```diff
@@ -287,39 +292,58 @@ func (d *SyncServerDatabase) CompleteSync(
...
-		recentStreamEvents, err = d.events.selectRecentEvents(
-			ctx, txn, roomID, types.StreamPosition(0), pos, numRecentEventsPerRoom,
-		)
-		if err != nil {
-			return nil, err
+
+		jr := types.NewJoinResponse()
+
+		//Join response should contain events only if room isn't filtered
+		if !isRoomFiltered(roomID, filter, nil) {
```
krombel: AFAIK a wished structure for this kind of checks would be in this case
```
if isRoomFiltered(roomID, filter, nil) {
    continue
}
```
instead of having a closing backet dozens of lines later

vendor/src/github.com/matrix-org/gomatrix/filter.go
```diff
@@ -14,30 +14,71 @@
...
+func DefaultFilter() Filter {
+	return Filter{
+		AccountData: DefaultFilterPart(),
+		EventFields: nil,
+		EventFormat: "client",
+		Presence:    DefaultFilterPart(),
+		Room: FilterRoom{
+			AccountData:  DefaultFilterPart(),
+			Ephemeral:    DefaultFilterPart(),
+			IncludeLeave: false, //TODO check default value on synapse
```
krombel: synapse defaults to false as well. See [here](https://github.com/matrix-org/synapse/blob/2a7e9fa/synapse/api/filtering.py#L244)

src/github.com/matrix-org/dendrite/syncapi/storage/syncserver.go
CromFr: Make sense, however each joined room must appear in the `rooms.join` list, even if the room has been filtered.

In fact, the room filtering only hides the events, but keep everything else as is.

This is an example of a join response for a filtered room on matrix.org:
```json
"!TUYoaUpGiPzdpBzfVM:matrix.org": {
	"unread_notifications": {
		"highlight_count": 0,
		"notification_count": 0
	},
	"timeline": {
		"limited": false,
		"prev_batch": "s419663949_378214671_17641_89713136_35031270_231422_3761477_4179932_4347",
		"events": []
	},
	"state": {
		"events": []
	},
	"ephemeral": {
		"events": []
	},
	"account_data": {
		"events": [
			{
				"content": {
					"primary_color": "#e55e5e",
					"secondary_color": "#f5eaea"
				},
				"type": "org.matrix.room.color_scheme"
			},
			{
				"content": {
					"event_id": "$15162674503862115fsbXn:matrix.org"
				},
				"type": "m.fully_read"
			}
		]
	}
},
```

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/filter_table.go
```diff
@@ -71,20 +72,29 @@ func (s *filterStatements) prepare(db *sql.DB) (err error) {
...
-	ctx context.Context, filter []byte, localpart string,
+	ctx context.Context, filter *gomatrix.Filter, localpart string,
 ) (filterID string, err error) {
 	var existingFilterID string
 
 	// This can result in a race condition when two clients try to insert the
 	// same filter and localpart at the same time, however this is not a
 	// problem as both calls will result in the same filterID
-	filterJSON, err := gomatrixserverlib.CanonicalJSON(filter)
+	filterJSON, err := json.Marshal(filter)
```
richvdh: I suspect there was a reason it was encoding as CanonicalJSON.

CromFr: gomatrixserverlib.CanonicalJSON removes whitespaces & sorts elements, including array content (which is a bit weird IMO since it will cause issues if the array elements need to be in a specific order, but it's not the case for filters) without building a JSON object.

json.Marshall already produces whitespace-less json, but the ordering (I suppose) depends on the hash value of the key, rather than alphabetical order.

Thanks for pointing this, I just found out that in the insert query, there was a check to prevent inserting multiple times the same filter, by comparing contents. I will use `gomatrixserverlib.CanonicalJSON(json.Marshal(filter))` when splitting this PR

richvdh:
ok, I'm going to close this as I believe @CromFr is going to turn it into smaller PRs.

PR #390, opened 2018-01-20
Application Service Registration

anoadragon453:
We now verify and validate loaded application services as well as have the registration flow required for them to register password-less users.

Had to change the database schema to allow for an `appservice_id` for each user. Not sure what this will mean in terms of adapting old databases.

Ready for review, I'm sure there's lots to be said about things :)

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/routing/register.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/routing/register.go
@@ -312,7 +312,7 @@ func validateApplicationService(
 	if matchedApplicationService != nil {
 		return "", &util.JSONResponse{
 			Code: 401,
-			JSON: jsonerror.BadJSON("Supplied access_token does not match any known application service"),
+			JSON: jsonerror.UnknownToken("Supplied access_token does not match any known application service"),
 		}
 	}
 
@@ -321,8 +321,8 @@ func validateApplicationService(
 		// If we didn't find any matches, return M_EXCLUSIVE
 		return "", &util.JSONResponse{
 			Code: 401,
-			JSON: jsonerror.Exclusive("Supplied username " + username +
-				" did not match any namespaces for application service ID: " + matchedApplicationService.ID),
+			JSON: jsonerror.ASExclusive(fmt.Sprintf(
+				"Supplied username %s did not match any namespaces for application service ID: %s", username, matchedApplicationService.ID)),
 		}
 	}
 
@@ -330,8 +330,8 @@ func validateApplicationService(
 	if UsernameMatchesMultipleExclusiveNamespaces(cfg, username) {
 		return "", &util.JSONResponse{
 			Code: 401,
-			JSON: jsonerror.Exclusive("Supplied username " + username +
-				" matches multiple exclusive application service namespaces. Only 1 match allowed"),
+			JSON: jsonerror.ASExclusive(fmt.Sprintf(
+				"Supplied username %s matches multiple exclusive application service namespaces. Only 1 match allowed", username)),
 		}
 	}
 
@@ -386,7 +386,7 @@ func Register(
 		cfg.Derived.ExclusiveApplicationServicesUsernameRegexp.MatchString(r.Username) {
 		return util.JSONResponse{
 			Code: 400,
-			JSON: jsonerror.Exclusive("This username is registered by an application service."),
+			JSON: jsonerror.ASExclusive("This username is reserved by an application service."),
 		}
 	}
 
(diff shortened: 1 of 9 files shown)


t3chguy:
> Not sure what this will mean in terms of adapting old databases.

Nothing until #357 lands

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -233,7 +237,110 @@ func validateRecaptcha(
...
+		}
+	}
+
+	// Ensure the desired username is within at least one of the application service's namespaces.
+	if !UsernameIsWithinApplicationServiceNamespace(cfg, username, matchedApplicationService) {
+		// If we didn't find any matches, return M_EXCLUSIVE
+		return "", &util.JSONResponse{
+			Code: 401,
+			JSON: jsonerror.Exclusive("Supplied username " + username +
+				" did not match any namespaces for application service ID: " + matchedApplicationService.ID),
```
erikjohnston: Generally one should always use string formatting rather than concatenation

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -233,7 +237,110 @@ func validateRecaptcha(
...
+		}
+	}
+
+	// Check this user does not fit multiple application service namespaces
+	if UsernameMatchesMultipleExclusiveNamespaces(cfg, username) {
+		return "", &util.JSONResponse{
+			Code: 401,
+			JSON: jsonerror.Exclusive("Supplied username " + username +
+				" matches multiple exclusive application service namespaces. Only 1 match allowed"),
```
erikjohnston: Ditto

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -233,7 +237,110 @@ func validateRecaptcha(
...
+	for _, appservice := range cfg.Derived.ApplicationServices {
+		if appservice.ASToken == accessToken {
+			matchedApplicationService = &appservice
+			break
+		}
+	}
+	if matchedApplicationService != nil {
+		return "", &util.JSONResponse{
+			Code: 401,
+			JSON: jsonerror.BadJSON("Supplied access_token does not match any known application service"),
```
erikjohnston: `BadJson`?

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -273,6 +380,16 @@ func Register(
 		return *resErr
 	}
 
+	// Make sure normal user isn't registering under an exclusive application
+	// service namespace
+	if r.Auth.Type != "m.login.application_service" &&
+		cfg.Derived.ExclusiveApplicationServicesUsernameRegexp.MatchString(r.Username) {
+		return util.JSONResponse{
+			Code: 400,
+			JSON: jsonerror.Exclusive("This username is registered by an application service."),
```
erikjohnston: The username might not be registered, so we should probably have some different wording

erikjohnston: `jsonerror.Exclusive` is defined as:

> +// Exclusive is an error returned when an application service tries to register
+// an username that is outside of its registered namespace

which doesn't really fit

anoadragon453: Ah yes, the spec return `M_EXCLUSIVE` [for multiple reasons](https://matrix.org/docs/spec/application_service/unstable.html#server-admin-style-permissions). I'll update the comments to reflect that :)

anoadragon453:
Fixed, thanks!

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -233,7 +237,110 @@ func validateRecaptcha(
...
+	// registered in the config.
+	accessToken := req.URL.Query().Get("access_token")
+	var matchedApplicationService *config.ApplicationService
+	for _, appservice := range cfg.Derived.ApplicationServices {
+		if appservice.ASToken == accessToken {
+			matchedApplicationService = &appservice
+			break
+		}
+	}
+	if matchedApplicationService != nil {
```
APwhitehat: @anoadragon453 I feel this should be equality. I'm lost.

anoadragon453: ...yep, looks like it should be. Whoops, not sure how that made it through. Will do some testing to check this is indeed broken and submit a PR if so. For now I'd change it in your PRs if you rely on this.

Issue #391, opened 2018-02-10
If no Application Services are registered, server complains all users are registering in exclusive namespace


anoadragon453:
Need to check if no application services are registered, skip the exclusive check.

PR #392, opened 2018-02-10
 Prevent AS user check if no AS registered

anoadragon453:
When a user registered on a homeserver with no application services
 registered, registration would check if the meta-regexp object matched
 the proposed user's new username.

 Apparently "" is a regex that matches everything, so every user was
 then barred from registering as they were supposedly registering inside
 an AS' exclusive namespace.

 This change prevents that check from happening if no AS's are
 registered.

Fixes: #391


erikjohnston:
TBH, I wonder if it would be better to have ExclusiveApplicationServicesUsernameRegexp default to `^$` or something instead?

anoadragon453:
That is cleaner and prevents any surprises in the future. Good suggestion!

I'll refactor this PR towards that end instead.

anoadragon453:
`^$` by default for the exclusive regex has been implemented. Though I've also kept the check on the total AS number, bypassing the exclusivity check if it's non-necessary.

PR #396, opened 2018-02-26
Add MacOS kafka setup instructions

Trion129:
Setting up Kafka using the provided script wasn't working for me, daemon would start and close instead of staying alive.
Using `brew` for installation is cleaner and easier.
So adding it to the INSTALL.md doc. Would help Mac users in future! :)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -60,13 +60,6 @@ brew services start zookeeper
 brew services start kafka
 ```
 
-You can stop Kafka service by running
-
-```bash
-brew services stop zookeeper
-brew services stop kafka
-```
-
 ## Configuration
 
 ### Postgres database setup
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -52,6 +52,21 @@ kafka/bin/zookeeper-server-start.sh -daemon kafka/config/zookeeper.properties
...
 
+On MacOS, you can use [homebrew](https://brew.sh/) for easier setup of kafka
+
+```bash
+brew install kafka
+brew services start zookeeper
+brew services start kafka
+```
+
+You can stop Kafka service by running
```
richvdh: this won't work on non-MacOS!

richvdh: (isn't it kinda obvious anyway?)

Trion129: ah yes, I felt that while writing too 😋  But added it just in case someone doesn't know.
Will remove it!

richvdh:
thanks!

PR #397, opened 2018-03-01
Update install instructions for mac

nikitavoloboev:
I tried to run the setup as described in INSTALL.md file in [configuration](https://github.com/matrix-org/dendrite/blob/master/INSTALL.md#configuration). But the instructions for setting up Postgres failed on macOS. You don't need to use `sudo -u postgres` there.

I updated the instructions to include the equivalent mac commands. I checked and the commands do indeed create a user and the databases.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -62,10 +62,23 @@ Dendrite requires a postgres database engine, version 9.5 or later.
   ```bash
   sudo -u postgres createuser -P dendrite     # prompts for password
   ```
+
+On macOS you need to run:
+```bash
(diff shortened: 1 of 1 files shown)


richvdh:
looks sensible.

I just wonder if it would be clearer to put:

> (On macOS, omit `sudo -u postgres`.)

... rather than duplicating the whole command.

anoadragon453:
And perhaps mention that once for both commands.

nikitavoloboev:
I have changed it to @richvdh suggestion and duplicated it on both commands.

PR #406, opened 2018-03-03
Return empty list instead of null for new UI-auth sessions

### LINKED ISSUE #399: Registration: "completed" is initially returned as null
**Spec URL**: https://matrix.org/docs/spec/client_server/r0.3.0.html#post-matrix-client-r0-register

**Phase:** 1

As part of the response from `/register`, a `completed` array is returned specifying what registration stages have been completed so far by the client. If a client has not completed any stages (which is always the case initially), the server should send back `"completed"` with an empty array (`[]`). Instead, it sends back `"completed": null`.

This is where we add items to the completed array:

https://github.com/matrix-org/dendrite/blob/139cb7a01e1c391d0828c2d78182c39059b3a0fa/src/github.com/matrix-org/dendrite/clientapi/routing/register.go#L427-L481

If no auth type is specified, `completed` never has anything added to it. We need to initialize it to an empty array (or slice in Go terms) somewhere, probably before we add stuff to it.

## Request

Use a REST API Client such as [RESTClient](https://addons.mozilla.org/en-US/firefox/addon/restclient/) or [Postman](https://chrome.google.com/webstore/detail/postman/fhbjgbiflinjbdggehcddcbncdddomop). 

```
# POST _matrix/client/r0/register

Body: {}
```

## Expected Response Body

```
{
  "flows": [{
    "stages": ["m.login.recaptcha"]
  }],
  "completed": [],
  "params": {
    "m.login.recaptcha": {
      "public_key": "6LdL3i8UAAAAAAxwztdHt348YLXFcA6vx0JRkRge"
    }
  },
  "session": "b32ORZ6ubv8YErtX3sV0h6oA"
}
```

## Given Response Body

```
{
  "flows": [{
    "stages": ["m.login.recaptcha"]
  }],
  "completed": null, # Wrong
  "params": {
    "m.login.recaptcha": {
      "public_key": "6LdL3i8UAAAAAAxwztdHt348YLXFcA6vx0JRkRge"
    }
  },
  "session": "b32ORZ6ubv8YErtX3sV0h6oA"
}
```

**Summary:**

- [ ] Ensure that an empty array `[]` is returned instead of `null`.

APwhitehat:
Map was being accessed without checking that the key exists.
So add a Get method instead.
fixes #399

Signed-off-by: Anant Prakash <anantprakashjsr@gmail.com>

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/routing/register.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/routing/register.go
@@ -50,9 +50,26 @@ const (
 	sessionIDLength   = 24
 )
 
+type sessionsDict struct {
+	sessions map[string][]authtypes.LoginType
+}
+
+func (d sessionsDict) Get(key string) []authtypes.LoginType {
+	if v, ok := d.sessions[key]; ok {
+		return v
+	}
+	return make([]authtypes.LoginType, 0)
+}
+
+func (d *sessionsDict) Set(key string, v []authtypes.LoginType) {
+	d.sessions[key] = v
+}
+
 var (
 	// TODO: Remove old sessions. Need to do so on a session-specific timeout.
-	sessions           = make(map[string][]authtypes.LoginType) // Sessions and completed flow stages
+	sessions = sessionsDict{ // Sessions and completed flow stages
+		sessions: make(map[string][]authtypes.LoginType),
+	}
 	validUsernameRegex = regexp.MustCompile(`^[0-9a-z_\-./]+$`)
 )
 
@@ -112,7 +129,7 @@ func newUserInteractiveResponse(
 	params map[string]interface{},
 ) userInteractiveResponse {
 	return userInteractiveResponse{
-		fs, sessions[sessionID], params, sessionID,
+		fs, sessions.Get(sessionID), params, sessionID,
 	}
 }
 
@@ -433,7 +450,7 @@ func handleRegistrationFlow(
 		}
 
 		// Add Recaptcha to the list of completed registration stages
-		sessions[sessionID] = append(sessions[sessionID], authtypes.LoginTypeRecaptcha)
+		sessions.Set(sessionID, append(sessions.Get(sessionID), authtypes.LoginTypeRecaptcha))
 
 	case authtypes.LoginTypeSharedSecret:
 		// Check shared secret against config
@@ -446,7 +463,7 @@ func handleRegistrationFlow(
 		}
 
 		// Add SharedSecret to the list of completed registration stages
-		sessions[sessionID] = append(sessions[sessionID], authtypes.LoginTypeSharedSecret)
+		sessions.Set(sessionID, append(sessions.Get(sessionID), authtypes.LoginTypeSharedSecret))
 
 	case authtypes.LoginTypeApplicationService:
 		// Check Application Service register user request is valid.
@@ -466,7 +483,7 @@ func handleRegistrationFlow(
 	case authtypes.LoginTypeDummy:
 		// there is nothing to do
 		// Add Dummy to the list of completed registration stages
-		sessions[sessionID] = append(sessions[sessionID], authtypes.LoginTypeDummy)
+		sessions.Set(sessionID, append(sessions.Get(sessionID), authtypes.LoginTypeDummy))
 
 	default:
 		return util.JSONResponse{
@@ -478,7 +495,7 @@ func handleRegistrationFlow(
 	// Check if the user's registration flow has been completed successfully
 	// A response with current registration flow and remaining available methods
 	// will be returned if a flow has not been successfully completed yet
-	return checkAndCompleteFlow(sessions[sessionID], req, r, sessionID, cfg, accountDB, deviceDB)
+	return checkAndCompleteFlow(sessions.Get(sessionID), req, r, sessionID, cfg, accountDB, deviceDB)
 }
 
 // checkAndCompleteFlow checks if a given registration flow is completed given
--- a/src/github.com/matrix-org/dendrite/clientapi/routing/register.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/routing/register.go
@@ -50,26 +50,36 @@ const (
 	sessionIDLength   = 24
 )
 
+// sessionsDict represents every sessions' completed flow stages.
 type sessionsDict struct {
 	sessions map[string][]authtypes.LoginType
 }
 
-func (d sessionsDict) Get(key string) []authtypes.LoginType {
-	if v, ok := d.sessions[key]; ok {
-		return v
+// GetCompletedStages returns the completed stages for a session.
+func (d sessionsDict) GetCompletedStages(sessionID string) []authtypes.LoginType {
+	if completedStages, ok := d.sessions[sessionID]; ok {
+		return completedStages
 	}
+	// Ensure that a empty slice is return and not nil. See gh #399.
 	return make([]authtypes.LoginType, 0)
 }
 
-func (d *sessionsDict) AddCompletedStage(key string, v authtypes.LoginType) {
-	d.sessions[key] = append(d.Get(key), v)
+// AddCompletedStage adds a completed stage to the session.
+func (d *sessionsDict) AddCompletedStage(sessionID string, stage authtypes.LoginType) {
+	d.sessions[sessionID] = append(d.GetCompletedStages(sessionID), stage)
+}
+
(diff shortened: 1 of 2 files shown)


src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -50,9 +50,18 @@ const (
 	sessionIDLength   = 24
 )
 
+type sessionsDict map[string][]authtypes.LoginType
+
+func (d sessionsDict) Get(key string) []authtypes.LoginType {
```
anoadragon453: As far as I know there's no way to override an operator like `[]`, so this is the next best option if we're going for overloading. The other option is just to test if the map isn't initialized yet and do so if not, but this feels cleaner overall.

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -112,7 +121,7 @@ func newUserInteractiveResponse(
 	params map[string]interface{},
 ) userInteractiveResponse {
 	return userInteractiveResponse{
-		fs, sessions[sessionID], params, sessionID,
+		fs, sessions.Get(sessionID), params, sessionID,
```
anoadragon453: My only concern is if a developer doesn't know to use `.Get()` and instead just writes an `if session, ok := sessionsDict; ok {...}`, but this would likely be picked up in review anyway.

erikjohnston: Making `sessionsDict` an actual struct to stop people accessing the map directly. It looks like the only other operation that is done is to append a `authtypes.LoginType` to a session, so should be a small interface to implement.

APwhitehat: @erikjohnston Should I go ahead and do that ?

erikjohnston: Sure, if you're happy

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -466,7 +483,7 @@ func handleRegistrationFlow(
 	case authtypes.LoginTypeDummy:
 		// there is nothing to do
 		// Add Dummy to the list of completed registration stages
-		sessions[sessionID] = append(sessions[sessionID], authtypes.LoginTypeDummy)
+		sessions.Set(sessionID, append(sessions.Get(sessionID), authtypes.LoginTypeDummy))
```
erikjohnston: Given this is the only place we use `Set`, I'd be tempted to instead have something like:

```golang
sessions.AddCompletedStage(sessionID, authtypes.LoginTypeDummy)
```

APwhitehat:
@erikjohnston Made necessary changes, you may review now.

anoadragon453:
Tested. Looks good.

```
{
  "flows": [{
    "stages": ["m.login.dummy"]
  }],
  "completed": [],
  "params": {},
  "session": "z09OtPqdqUnkw15vSof5h8cr"
}
```

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -50,9 +50,26 @@ const (
 	sessionIDLength   = 24
 )
 
+type sessionsDict struct {
+	sessions map[string][]authtypes.LoginType
```
richvdh: likewise, what does this map to and from?

APwhitehat:
@richvdh Thanks for the review. 
Currently substantial portion of dendrite lack tests. Some testing framework would be wonderful.
But that should be addressed in another PR.

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -50,26 +50,34 @@ const (
 	sessionIDLength   = 24
 )
 
+// sessionsDict represents a session's completed flow stages
```
richvdh: does it? it looks like it tracks all sessions' data, not just one.

APwhitehat: Oh, my bad. Correcting.

anoadragon453:
@APwhitehat unit tests can be achieved with the current `endpoint_test.go` file implementation. Integration tests are another story, but unit tests can be completed now.

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -50,9 +50,36 @@ const (
 	sessionIDLength   = 24
 )
 
+// sessionsDict represents every sessions' completed flow stages.
```
richvdh: this might be better phrased as "sessionsDict keeps track of completed auth stages for each session"

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -50,9 +50,36 @@ const (
...
+type sessionsDict struct {
+	sessions map[string][]authtypes.LoginType
+}
+
+// GetCompletedStages returns the completed stages for a session.
+func (d sessionsDict) GetCompletedStages(sessionID string) []authtypes.LoginType {
+	if completedStages, ok := d.sessions[sessionID]; ok {
+		return completedStages
+	}
+	// Ensure that a empty slice is return and not nil. See gh #399.
```
richvdh: s/return/returned/

s/gh//. or s/gh/github issue/

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -50,9 +50,36 @@ const (
...
+	return make([]authtypes.LoginType, 0)
+}
+
+// AddCompletedStage adds a completed stage to the session.
```
richvdh: "AddCompletedStage records that a session has completed an auth stage"

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -50,9 +50,36 @@ const (
...
+func (d *sessionsDict) AddCompletedStage(sessionID string, stage authtypes.LoginType) {
+	d.sessions[sessionID] = append(d.GetCompletedStages(sessionID), stage)
+}
+
+// newSessionsDict returns a sessionsDict whose contained map is initialized and empty.
```
richvdh: this comment is a bit redundant tbh

src/github.com/matrix-org/dendrite/clientapi/routing/register_test.go
```diff
@@ -132,3 +132,14 @@ func TestFlowCheckingExtraneousIncorrectInput(t *testing.T) {
...
 	}
 }
+
+// Completed flows stages should always be a valid slice header.
+// TestEmptyCompletedFlows checks that sessionsDict returns a slice & not nil.
+func TestEmptyCompletedFlows(t *testing.T) {
+	fakeEmptySessions := newSessionsDict()
+	fakeSessionID := "aRandomSessionIDWhichDoesNotExist"
+
+	if fakeEmptySessions.GetCompletedStages(fakeSessionID) == nil {
```
richvdh: it'd be better to test for `[]`, rather than accepting everything that's not `nil`

APwhitehat:
@richvdh stealthily amended last commit.

src/github.com/matrix-org/dendrite/clientapi/routing/register_test.go
```diff
@@ -132,3 +132,16 @@ func TestFlowCheckingExtraneousIncorrectInput(t *testing.T) {
...
+
+// Completed flows stages should always be a valid slice header.
+// TestEmptyCompletedFlows checks that sessionsDict returns a slice & not nil.
+func TestEmptyCompletedFlows(t *testing.T) {
+	fakeEmptySessions := newSessionsDict()
+	fakeSessionID := "aRandomSessionIDWhichDoesNotExist"
+	ret := fakeEmptySessions.GetCompletedStages(fakeSessionID)
+
+	// check for []
+	if !(len(ret) == 0 && ret != nil) {
```
richvdh: the double-negative makes my head hurt.

`if ret == nil || len(ret) != 0`

richvdh:
> @richvdh stealthily amended last commit.

Please don't do that in general, it makes it hard to review your changes.

APwhitehat:
> Please don't do that in general, it makes it hard to review your changes.

Noted.

richvdh:
\o/ thanks!

PR #408, opened 2018-03-04
Login: Add Token based authentication

### LINKED ISSUE #403: Login: Implement Token-based authentication
**Spec URL:** https://matrix.org/docs/spec/client_server/r0.3.0.html#post-matrix-client-r0-login https://matrix.org/docs/spec/client_server/r0.5.0#token-based

**Difficulty:** Medium

https://github.com/matrix-org/dendrite/blob/dfcf31f293a319830e2b8a188413db776930a5e5/src/github.com/matrix-org/dendrite/clientapi/routing/login.go#L60-L146

We currently only allow password-based authentication for login. Token-based authentication will be necessary for basic operation with clients. Currently users have tokens generated for them, we just need to check them on login.

**Summary:**

- [ ] Differentiate between `"type":` "m.login.token" and "m.login.password" in request body
- [ ] Authenticate user based on token and device ID. [Relationship between access tokens and devices](https://matrix.org/docs/spec/client_server/r0.3.0.html#relationship-between-access-tokens-and-devices)
- [ ] Create test case(s) for logging in with access token

APwhitehat:
[WIP] To fix #403 . 
I took a try at this, wanted to check if this appears to be a step in the right direction?

Note: I used the current token generation method. Should we implement macaroons ?
As in synapse ref: [1](https://github.com/matrix-org/synapse/blob/master/synapse/handlers/auth.py#L867) [2](https://github.com/matrix-org/synapse/blob/master/synapse/api/auth.py)

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/routing/login.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/routing/login.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 
 	"github.com/matrix-org/dendrite/clientapi/auth"
+	"github.com/matrix-org/dendrite/clientapi/auth/authtypes"
 	"github.com/matrix-org/dendrite/clientapi/auth/storage/accounts"
 	"github.com/matrix-org/dendrite/clientapi/auth/storage/devices"
 	"github.com/matrix-org/dendrite/clientapi/httputil"
@@ -27,14 +28,6 @@ import (
 	"github.com/matrix-org/dendrite/common/config"
 	"github.com/matrix-org/gomatrixserverlib"
 	"github.com/matrix-org/util"
-)
-
-type loginType string
-
-// https://matrix.org/docs/spec/client_server/r0.3.0.html#login
-const (
-	PasswordBased loginType = "m.login.password"
-	TokenBased    loginType = "m.login.token"
 )
 
 type loginFlows struct {
@@ -46,42 +39,23 @@ type flow struct {
 	Stages []string `json:"stages"`
 }
 
-type loginRequest struct {
-	Type               loginType `json:"type"`
-	User               string    `json:"user"`
-	Medium             string    `json:"medium"`
-	Address            string    `json:"address"`
-	Password           string    `json:"password"`
-	Token              string    `json:"token"`
-	DeviceID           string    `json:"device_id"`
-	InitialDisplayName *string   `json:"initial_device_display_name"`
-}
-
-type loginResponse struct {
-	UserID      string                       `json:"user_id"`
-	AccessToken string                       `json:"access_token"`
-	HomeServer  gomatrixserverlib.ServerName `json:"home_server"`
-	DeviceID    string                       `json:"device_id"`
-}
-
 func defaultPasswordLogin() loginFlows {
 	f := loginFlows{}
-	s := flow{string(PasswordBased), []string{string(PasswordBased)}}
+	s := flow{string(authtypes.LoginTypePassword), []string{string(authtypes.LoginTypePassword)}}
 	f.Flows = append(f.Flows, s)
 	return f
 }
 
 func handlePasswordLogin(
-	r loginRequest, accountDB *accounts.Database, deviceDB *devices.Database,
-	req *http.Request, cfg config.Dendrite) *util.JSONResponse {
+	r authtypes.LoginRequest, localpart string, accountDB *accounts.Database,
+	deviceDB *devices.Database, req *http.Request, cfg config.Dendrite) util.JSONResponse {
 
-	localpart := r.User
 	acc, err := accountDB.GetAccountByPassword(req.Context(), localpart, r.Password)
 
 	if err != nil {
 		// Technically we could tell them if the user does not exist by checking if err == sql.ErrNoRows
 		// but that would leak the existence of the user.
-		return &util.JSONResponse{
+		return util.JSONResponse{
 			Code: 403,
 			JSON: jsonerror.Forbidden("username or password was incorrect, or the account does not exist"),
 		}
@@ -97,15 +71,15 @@ func handlePasswordLogin(
 		req.Context(), acc.Localpart, nil, token, r.InitialDisplayName,
 	)
 	if err != nil {
-		return &util.JSONResponse{
+		return util.JSONResponse{
 			Code: 500,
 			JSON: jsonerror.Unknown("failed to create device: " + err.Error()),
 		}
 	}
 
-	return &util.JSONResponse{
+	return util.JSONResponse{
 		Code: 200,
-		JSON: loginResponse{
+		JSON: authtypes.LoginResponse{
 			UserID:      dev.UserID,
 			AccessToken: dev.AccessToken,
 			HomeServer:  cfg.Matrix.ServerName,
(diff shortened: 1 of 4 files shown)


src/github.com/matrix-org/dendrite/clientapi/routing/login.go
```diff
@@ -28,6 +29,14 @@ import (
 	"github.com/matrix-org/util"
 )
 
+type loginType string
+
+// https://matrix.org/docs/spec/client_server/r0.3.0.html#login
+const (
+	PasswordBased loginType = "m.login.password"
```
anoadragon453: Just a heads up that we already have the known logintypes defined [here](https://github.com/matrix-org/dendrite/blob/master/src/github.com/matrix-org/dendrite/clientapi/auth/authtypes/logintypes.go).

src/github.com/matrix-org/dendrite/clientapi/routing/login.go
```diff
@@ -37,10 +46,15 @@ type flow struct {
 	Stages []string `json:"stages"`
 }
 
-type passwordRequest struct {
-	User               string  `json:"user"`
-	Password           string  `json:"password"`
-	InitialDisplayName *string `json:"initial_device_display_name"`
+type loginRequest struct {
```
anoadragon453: I would put this type in the [authtypes](https://github.com/matrix-org/dendrite/tree/master/src/github.com/matrix-org/dendrite/clientapi/auth/authtypes) folder as well, in a new file.

src/github.com/matrix-org/dendrite/clientapi/routing/login.go
```diff
@@ -50,25 +64,112 @@ type loginResponse struct {
 	DeviceID    string                       `json:"device_id"`
 }
 
-func passwordLogin() loginFlows {
+func defaultPasswordLogin() loginFlows {
 	f := loginFlows{}
-	s := flow{"m.login.password", []string{"m.login.password"}}
+	s := flow{string(PasswordBased), []string{string(PasswordBased)}}
```
anoadragon453: Use the types in [logintypes.go](https://github.com/matrix-org/dendrite/blob/master/src/github.com/matrix-org/dendrite/clientapi/auth/authtypes/logintypes.go).

src/github.com/matrix-org/dendrite/clientapi/routing/login.go
```diff
@@ -50,25 +64,112 @@ type loginResponse struct {
...
 	f.Flows = append(f.Flows, s)
 	return f
 }
 
+func handlePasswordLogin(
+	r loginRequest, accountDB *accounts.Database, deviceDB *devices.Database,
+	req *http.Request, cfg config.Dendrite) *util.JSONResponse {
+
+	localpart := r.User
```
anoadragon453: Necessary?

src/github.com/matrix-org/dendrite/clientapi/routing/login.go
```diff
@@ -50,25 +64,112 @@ type loginResponse struct {
...
+	acc, err := accountDB.GetAccountByPassword(req.Context(), localpart, r.Password)
+
+	if err != nil {
+		// Technically we could tell them if the user does not exist by checking if err == sql.ErrNoRows
+		// but that would leak the existence of the user.
```
anoadragon453: One can always find this out via trying to register as that user :)

src/github.com/matrix-org/dendrite/clientapi/routing/login.go
```diff
@@ -50,25 +64,112 @@ type loginResponse struct {
...
+	if err != nil {
+		if err == sql.ErrNoRows {
+			return &util.JSONResponse{
+				Code: 401,
+				JSON: jsonerror.UnknownToken("Unknown token"),
+			}
+		}
+		return &util.JSONResponse{
+			Code: 401,
+			JSON: jsonerror.Unknown("Unexpected Server error occurred"),
```
anoadragon453: Lowercase 'server'.

src/github.com/matrix-org/dendrite/clientapi/routing/login.go
```diff
@@ -50,25 +64,112 @@ type loginResponse struct {
...
+	dev, err := deviceDB.GetDeviceByAccessToken(req.Context(), r.Token)
+	if err != nil {
+		if err == sql.ErrNoRows {
+			return &util.JSONResponse{
+				Code: 401,
+				JSON: jsonerror.UnknownToken("Unknown token"),
+			}
+		}
+		return &util.JSONResponse{
+			Code: 401,
```
anoadragon453: Should return a 500 on server error.

src/github.com/matrix-org/dendrite/clientapi/routing/login.go
```diff
@@ -50,25 +64,112 @@ type loginResponse struct {
...
+			JSON: jsonerror.Unknown("Unexpected Server error occurred"),
+		}
+	}
+
+	if dev.ID != r.DeviceID {
+		// The access token specified in the request was generated for a
+		// different device.
+		return &util.JSONResponse{
+			Code: 403,
+			JSON: jsonerror.Forbidden("The access token was generated for a different device."),
```
anoadragon453: Remove the "." from the end of the error.

Also not sure if we should tell this information or not. It's certainly useful, but could it be used in any malicious way?

src/github.com/matrix-org/dendrite/clientapi/routing/login.go
```diff
@@ -50,25 +64,112 @@ type loginResponse struct {
...
-	if req.Method == "GET" { // TODO: support other forms of login other than password, depending on config options
+	if req.Method == "GET" {
 		return util.JSONResponse{
 			Code: 200,
-			JSON: passwordLogin(),
+			JSON: defaultPasswordLogin(),
 		}
 	} else if req.Method == "POST" {
-		var r passwordRequest
+		var r loginRequest
```
anoadragon453: Use the types in [authtypes.go](https://github.com/matrix-org/dendrite/blob/master/src/github.com/matrix-org/dendrite/clientapi/auth/authtypes/logintypes.go).

src/github.com/matrix-org/dendrite/clientapi/routing/login.go
```diff
@@ -82,12 +183,10 @@ func Login(
 
 		util.GetLogger(req.Context()).WithField("user", r.User).Info("Processing login request")
 
-		// r.User can either be a user ID or just the localpart... or other things maybe.
-		localpart := r.User
```
anoadragon453: This ensures that `localpart` ("username" in "@username:matrix.org") is actually a localpart and not the full userID ("@username:matrix.org"). By just passing `r` to your two functions and then pulling `r.User`, you're not guaranteeing that `r.User` is just a localpart. 

I would keep this `if` statement to pull out the localpart if necessary and then pass that directly to your functions (`handlePasswordLogin`, `handleTokenLogin`).

src/github.com/matrix-org/dendrite/clientapi/routing/login.go
```diff
@@ -103,41 +202,10 @@ func Login(
...
-		return util.JSONResponse{
-			Code: 200,
-			JSON: loginResponse{
-				UserID:      dev.UserID,
-				AccessToken: dev.AccessToken,
-				HomeServer:  cfg.Matrix.ServerName,
-				DeviceID:    dev.ID,
-			},
+		if r.Type == PasswordBased {
+			return *handleTokenLogin(r, deviceDB, req, cfg)
```
anoadragon453: In the project we need to standardize over whether we're returning a `*util.JSONResponse` or `util.JSONResponse`, because in `handleTokenLogin` we're returning a reference, then removing the reference directly afterwards.

It's a bigger issue than just this PR, but for now I would suggest just returning a `util.JSONResponse` in your `handlePasswordLogin` and `handleTokenLogin` functions and remove the `*` on this line and the next.

anoadragon453:
> Note: I used the current token generation method. Should we implement macaroons ?

Might be nice for the future, but I'd recommend waiting until it's spec'd first.

src/github.com/matrix-org/dendrite/clientapi/routing/login.go
APwhitehat: "Access token not valid for this device" would be less revealing ?

src/github.com/matrix-org/dendrite/clientapi/routing/login.go
APwhitehat: Yeah. Not my comment though, I just repeated this from previous code.

APwhitehat:
Addressed the comments. Need to write tests now.

APwhitehat:
Since no tests are currently written, factoring out the writing of tests into a new issues.
P.S. I tested the tokenLogin manually, things seem to be working good.

src/github.com/matrix-org/dendrite/clientapi/routing/login.go
anoadragon453: Ah yeah, no problem then.

anoadragon453:
Works pretty well. One thing is that the spec states the following for `device_id`:

>ID of the client device. If this does not correspond to a known client device, a new device will be created. The server will auto-generate a device_id if this is not specified.

Currently if I send:

```
{
    "type": "m.login.token",
    "user": "@supercooluser:localhost",
    "token": "dCOOg2SPRHeBVC3BRaser4k6fAZYKEIimb9sXobJRJE"
}
```

I get:

```
{
  "errcode": "M_FORBIDDEN",
  "error": "Access token not valid for this device"
}
```

In this case, you'll need to add a check for whether `device_id` is present in the request or not and generate a new deviceID. There's a method here to do just that:

https://github.com/APwhitehat/dendrite/blob/b4ca664960a6bdf19dcb9ace7a4a4e74c46008dd/src/github.com/matrix-org/dendrite/clientapi/auth/auth.go#L86-L94

anoadragon453:
Alright, so just asked the following in Matrix-dev: https://matrix.to/#/!XqBunHwQIXUiqCaoxq:matrix.org/%2415202424201268rgrPH:amorgan.xyz

Essentially the `token` described in the spec is not an `access token`, but an arbitrary token that one might find get as part of a verification email or text message. It's described here: https://matrix.org/docs/spec/client_server/r0.3.0.html#token-based

Thus a new *access* token is created every time this endpoint is successfully completed.

This kind of changes this PR completely, instead asking for `m.login.token` to be completed as part of the users interactive authentication, and then make use of that in `m.login.token ` type logging in.

I'll update the issue to reflect that, sorry for the confusion on my end.

APwhitehat:
closed in favor of #429

PR #411, opened 2018-03-04
Support user_id query param (and implement whoami endpoint)

### LINKED ISSUE #404: Implement whoami endpoint
**Spec URL:** https://matrix.org/docs/spec/client_server/r0.3.0.html#get-matrix-client-r0-account-whoami

**Phase:** 2

**Difficulty:** Easy

We receive an access token, and with that we return the user_id connected to that access token.

**Summary:**

- [ ] Create endpoint for GET `/account/whoami` in [routing.go](https://github.com/matrix-org/dendrite/blob/master/src/github.com/matrix-org/dendrite/clientapi/routing/routing.go)
- [ ] Return the correct user_id for the given access token. If no correct token was found, return `M_UNKNOWN_TOKEN`
- [ ] Create test case(s) for `/account/whoami` with correct and incorrect access tokens.

APwhitehat:
* Implement `VerifyUserFromRequest` which verifies an user is authenticated.
Handles AS users as well.
* Move `GenerateDeviceID` from `auth.go` to `auth/storage/devices/storage.go` to avoid cyclic import of package auth. This move includes `deviceIDByteLength` .
* Factor out `GetLocalpartFromUsername` & `GetDomainFromUserID` from `routing.login` into `userutil` package since they are useful pretty much everywhere.
* Add `/account/whoami` endpoint. fixes #404

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
@@ -25,11 +25,9 @@ import (
 	"strings"
 
 	"github.com/matrix-org/dendrite/clientapi/auth/authtypes"
-	"github.com/matrix-org/dendrite/clientapi/auth/storage/accounts"
-	"github.com/matrix-org/dendrite/clientapi/auth/storage/devices"
 	"github.com/matrix-org/dendrite/clientapi/httputil"
 	"github.com/matrix-org/dendrite/clientapi/jsonerror"
-	"github.com/matrix-org/dendrite/common"
+	"github.com/matrix-org/dendrite/clientapi/userutil"
 	"github.com/matrix-org/dendrite/common/config"
 	"github.com/matrix-org/util"
 )
@@ -38,13 +36,16 @@ import (
 // 32 bytes => 256 bits
 var tokenByteLength = 32
 
-// The length of generated device IDs
-var deviceIDByteLength = 6
-
 // DeviceDatabase represents a device database.
 type DeviceDatabase interface {
 	// Look up the device matching the given access token.
 	GetDeviceByAccessToken(ctx context.Context, token string) (*authtypes.Device, error)
+}
+
+// AccountDatabase represents a account database.
+type AccountDatabase interface {
+	// Look up the account matching the given localpart.
+	GetAccountByLocalpart(ctx context.Context, localpart string) (*authtypes.Account, error)
 }
 
 // VerifyUserFromRequest authenticates the HTTP request,
@@ -52,8 +53,8 @@ type DeviceDatabase interface {
 // Finds local user or an application service user.
 // On failure returns an JSON error response which can be sent to the client.
 func VerifyUserFromRequest(
-	req *http.Request, accountDB *accounts.Database, deviceDB *devices.Database,
-	cfg *config.Dendrite,
+	req *http.Request, accountDB AccountDatabase, deviceDB DeviceDatabase,
+	applicationServices []config.ApplicationService,
 ) (string, *util.JSONResponse) {
 	// Try to find local user from device database
 	dev, devErr := VerifyAccessToken(req, deviceDB)
@@ -74,7 +75,7 @@ func VerifyUserFromRequest(
 
 	// Search for app service with given access_token
 	var appService *config.ApplicationService
-	for _, as := range cfg.Derived.ApplicationServices {
+	for _, as := range applicationServices {
 		if as.ASToken == token {
 			appService = &as
 			break
@@ -83,7 +84,7 @@ func VerifyUserFromRequest(
 
 	if appService != nil {
 		userID := req.URL.Query().Get("user_id")
-		localpart, err := common.GetLocalpartFromUsername(userID)
+		localpart, err := userutil.GetLocalpartFromUsername(userID)
 
 		if err != nil {
 			return "", &util.JSONResponse{
--- a/src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
@@ -63,7 +63,7 @@ func VerifyUserFromRequest(
 		return dev.UserID, nil
 	}
 
-	// Try to find Application Service user
+	// Try to find the Application Service user
 	token, err := extractAccessToken(req)
 
 	if err != nil {
@@ -96,14 +96,8 @@ func VerifyUserFromRequest(
 		// Verify that the user is registered
 		account, accountErr := accountDB.GetAccountByLocalpart(req.Context(), localpart)
 
-		if accountErr != nil {
-			return "", &util.JSONResponse{
-				Code: http.StatusForbidden,
-				JSON: jsonerror.Forbidden("Application service has not registered this user"),
-			}
-		}
-
-		if account.AppServiceID == appService.ID {
+		// Verify that account exists & appServiceID matches
+		if accountErr == nil && account.AppServiceID == appService.ID {
 			return userID, nil
 		}
 
--- a/src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
@@ -25,8 +25,12 @@ import (
 	"strings"
 
 	"github.com/matrix-org/dendrite/clientapi/auth/authtypes"
+	"github.com/matrix-org/dendrite/clientapi/auth/storage/accounts"
+	"github.com/matrix-org/dendrite/clientapi/auth/storage/devices"
 	"github.com/matrix-org/dendrite/clientapi/httputil"
 	"github.com/matrix-org/dendrite/clientapi/jsonerror"
+	"github.com/matrix-org/dendrite/common"
+	"github.com/matrix-org/dendrite/common/config"
 	"github.com/matrix-org/util"
 )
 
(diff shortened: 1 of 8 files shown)


src/github.com/matrix-org/dendrite/clientapi/routing/whoami.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// Whoami implements /account/whoami endpoint method GET
+// https://matrix.org/docs/spec/client_server/r0.3.0.html#get-matrix-client-r0-account-whoami
+func Whoami(
+	req *http.Request, deviceDB *devices.Database,
+) util.JSONResponse {
+
+	if req.Method == http.MethodGet {
+		token := req.URL.Query().Get("access_token")
```
anoadragon453: Instead of retrieving the `access_token` through the query, it will be sent through the `Authorization` header. See [this document](https://github.com/anoadragon453/dendrite/wiki/How-to-Test-with-Authentication-to-the-Client-Server-API) for clarification. We should pull it from there instead of the URL.

This code within Auth extracts the access token from the Authorization header: https://github.com/APwhitehat/dendrite/blob/b4ca664960a6bdf19dcb9ace7a4a4e74c46008dd/src/github.com/matrix-org/dendrite/clientapi/auth/auth.go#L98

Right now it's not public, but I'd suggest making it a public method and then using it here to get the token instead.

anoadragon453:
> A much wider effort is needed to increase test coverage as a significant part of the repo doesn't have tests. What are your thoughts?

Agreed. I'll try to focus on getting a good base for integration tests in place.

src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
```diff
@@ -43,6 +43,27 @@ type DeviceDatabase interface {
...
 }
 
+// GetUserFromRequest extracts UserID from the given HTTP request,
+// if access token is provided (and correct) or if user_id is given as a parameter of the HTTP request
+// else returns resErr (an error response which can be sent to the client)
+func GetUserFromRequest(req *http.Request, deviceDB DeviceDatabase) (user string, resErr *util.JSONResponse) {
+	dev, err := VerifyAccessToken(req, deviceDB)
+
+	// If access_token is empty try to get UserID from request parameters
+	if err != nil {
```
turt2live: The access token will still be provided by appservices, and should be validated to be from an appservice before considering the value of `user_id`.

src/github.com/matrix-org/dendrite/clientapi/routing/routing.go
```diff
@@ -389,4 +389,10 @@ func Setup(
 			}}
 		}),
 	).Methods("GET", "OPTIONS")
+
+	r0mux.Handle("/account/whoami",
+		common.MakeExternalAPI("whoami", func(req *http.Request) util.JSONResponse {
+			return Whoami(req, deviceDB)
+		}),
+	).Methods("GET")
```
turt2live: I'm not sure what Dendrite's policy is, but considering a bunch of the GET routes above have OPTIONS, might as well do it here too?

src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
APwhitehat: So appsevices request will be same as https://matrix.org/docs/spec/application_service/unstable.html#identity-assertion ?
Also if appsevices token is invalid, we should reject the whoami ?

src/github.com/matrix-org/dendrite/clientapi/routing/routing.go
APwhitehat: Yes

src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
turt2live: yes and yes.

For every client/server API endpoint (including whoami) the appservice must specify it's access token included in the registration file. It may then optionally include the user_id to masquerade as that user (assuming it's also in it's namespace). Otherwise it's supposed to use the `sender_localpart` in the registration file as a default.

This endpoint also needs to make sure that the user, whether it be an appservice user or otherwise, is actually registered on the server. For regular users it's not that hard, but for appservice users it can be a bit annoying.

anoadragon453: @APwhitehat An application service is stored as the following struct: https://github.com/matrix-org/dendrite/blob/6b5597218330b3702cf2b86b2163d5fe769b76a2/src/github.com/matrix-org/dendrite/common/config/appservice.go#L40-L53

Application Services are stored in config.Dendrite.Derived.ApplicationServices: https://github.com/matrix-org/dendrite/blob/6b5597218330b3702cf2b86b2163d5fe769b76a2/src/github.com/matrix-org/dendrite/common/config/config.go#L231-L233

So first iterate through those and check that the token matches a known AppService, then cross reference that with the given user ID in the database and check that it matches the given AppService ID. Users, including AppService users, are stored with the following SQL:

https://github.com/matrix-org/dendrite/blob/6b5597218330b3702cf2b86b2163d5fe769b76a2/src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/accounts_table.go#L43-L44

Reject if something doesn't match up.

AppService implementation is still incomplete, but it should be complete enough for this task to work.

src/github.com/matrix-org/dendrite/clientapi/routing/routing.go
anoadragon453: Yeah not entirely sure what we're using that for at the moment, but best to keep it consistent.

src/github.com/matrix-org/dendrite/clientapi/routing/login.go
```diff
@@ -57,6 +57,21 @@ func passwordLogin() loginFlows {
 	return f
 }
 
+// GetLocalpartDomainFromUserid extracts localpart & domain of server from userID
+// Returns JSON error in case of invalid username.
+func GetLocalpartDomainFromUserid(userID string) (
```
richvdh: firstly, it's not obvious to me that a function like this belongs in `routing.login`: the fact that we're calling it from somewhere that isn't about handling a /login request shows that.

But generally I don't think this is a nice interface. It's confusing that it returns an error-like thing that turns out not to actually be an `error`, and you haven't really saved that much code at the call site.

If you wanted to you *could* factor out something like:

```
func BuildInvalidUsernameResponse() {
    return util.JSONResponse{
        Code: 400,
        JSON: jsonerror.InvalidUsername("Invalid username"),
    }
}
```

... but it's really not obvious to me that is worthwhile.

In short I'd just keep this inline.

src/github.com/matrix-org/dendrite/clientapi/routing/whoami.go
```diff
@@ -0,0 +1,132 @@
...
+)
+
+type whoamiResponse struct {
+	UserID string `json:"user_id"`
+}
+
+// GetUserFromRequest extracts UserID from the given HTTP request,
+// Finds local user or an application service user,
+// On failure returns an error response which can be sent to the client.
+func GetUserFromRequest(
```
richvdh: it feels to me like a mistake to be doing this as a special-case for `whoami`. It's relevant to the majority of the C-S APIs.

Could we not roll it into `common.MakeAuthAPI` ?

APwhitehat: Thanks, 
this is indeed more practical. The correct place for this function should have been `clientapi/auth` but while doing so, I realised that `clientapi/auth/storage/devices/storage.go` imports `clientapi/auth`. This is a recipe for import cycles.
I am going to refactor a little bit and integrate this into `common.MakeAuthAPI`.
I hope it's alright.

src/github.com/matrix-org/dendrite/clientapi/routing/whoami.go
```diff
@@ -0,0 +1,60 @@
...
+
+// Whoami implements `/account/whoami` which enables client to query owner of the HTTP request.
+// https://matrix.org/docs/spec/client_server/r0.3.0.html#get-matrix-client-r0-account-whoami
+func Whoami(
+	req *http.Request, accountDB *accounts.Database, deviceDB *devices.Database,
+	cfg *config.Dendrite,
+) util.JSONResponse {
+
+	if req.Method == http.MethodGet {
+		user, err := auth.VerifyUserFromRequest(req, accountDB, deviceDB, cfg.Derived.ApplicationServices)
```
richvdh: again, why is this being done as a special-case for Whoami rather than being added to MakeAuthAPI?

src/github.com/matrix-org/dendrite/clientapi/routing/routing.go
```diff
@@ -389,4 +389,10 @@ func Setup(
 			}}
 		}),
 	).Methods(http.MethodGet, http.MethodOptions)
+
+	r0mux.Handle("/account/whoami",
```
richvdh: this is not (as the comment at line 374 claims) a stub implementation.

richvdh:
once again I'm finding it regrettable that this PR conflates adding `whoami` and adding support for `?user_id=`. I'd really encourage you to split it up.

richvdh:
(sorry for sitting on this for ages btw)

APwhitehat:
@richvdh Thanks for the review.
According to your suggestion I have split this PR into 4 smaller PRs. And made required changes.
This is recommended order for review #432 -> #433 -> #427 -> #434 .

PR #412, opened 2018-03-05
refactor User Interactive Flow Handling Functions

MohitKS5:
Signed-off-by: MohitKS5 <mohitkumarsingh907@gmail.com>
I have refactored the `RegistrationFlowHandler` which accepted `registrationRequest` along with other config params and returned jsonResponse by calling a complete registration function.
### Some features of UIAAhandler: 
  * The handler accepts generic type `userInteractiveFlowRequest` a struct with min requirements.
  * It accepts an anonymous function which will be called if flow completes without error.
  * It also accepts a `userInteractiveResponse` struct which contains allowed flows and params.  
### Calling the function:
  * If extra params other than those defined for anonymous functions are required then one can user closers as used in my [register.go](https://github.com/MohitKS5/dendrite/commit/5fd2d18acdb119b986b5c9c22a061577396eec2f#diff-789859953fb6c45e704d547fce33b9eeR220).
  * If the request contains extra params say as per spec sheet then we can use `UserInteractiveResponse` as member as in [here.](https://github.com/MohitKS5/dendrite/commit/5fd2d18acdb119b986b5c9c22a061577396eec2f#diff-789859953fb6c45e704d547fce33b9eeR58)

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/routing/register.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/routing/register.go
@@ -202,7 +202,7 @@ func Register(
 
 	// TODO: Handle loading of previous session parameters from database.
 	// TODO: Handle mapping registrationRequest parameters into session parameters
-	
+
 	return HandleUserInteractiveFlow(req, r.userInteractiveFlowRequest, sessionID, cfg,
 		userInteractiveResponse{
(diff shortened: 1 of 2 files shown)


MohitKS5:
will fix lint errors squash into 1 commit and then pull request

PR #420, opened 2018-03-10
Explain gb test output

anoadragon453:
`gb test`s output is confusing as it is similar to that of `gb build`.
Rather than building, it's actually running all the tests. If no output
past this shows, then all tests have succeeded.

Updates CONTRIBUTING.md to include a note about this.


richvdh:
thanks!

Issue #422, opened 2018-03-10
[Discussion] use of interfaces for database operations


MohitKS5:
The functions in clientapi currently accept pointers to databases: `*devices.Database` and `*accounts.Database` as function parameters. I suggest use of defined interfaces in place of these. Suppose we require to use `createDeviceFunction() and GetDevice()` then we can define something like 
``` golang
type deviceDatabase interface {
     createDeviceFunction( params ) returnType
     getDevice ( params ) returnType
}
```
and similar for accounts DB 
#### this helps in various ways : 
- [X] keeping track of database operations being carried out in the file.
- [X] scalability of the project like suppose we want to shift the database to some otherplace like cloudapi or like we have for roomserver.
- [X] Function calls become trivial during development by simulating `external server responses` and database reads and writes.
- [X] testing purposes. Go has built in test features ! 
The testing of functions and that of database operations must be segregated. The the sole purpose of generic drivers is that. In fact an entire mock database server can be generated if we use interface because we now just need **a function that implements those methods.**

It also gives a sort of control over the passed params and of course speed in tests. Fixed Database pointers makes it practically impossible to play with functions in go due to hard typechecking and its policy to avoid overloading and nesting of functions.

#### We have already used this method but in places only where its necessary like in [`state.go`](https://github.com/matrix-org/dendrite/blob/master/src/github.com/matrix-org/dendrite/roomserver/state/state.go#L32) roomserver code . 
I would suggest to use it as a general practice.

Kegsay:
Since the introduction of SQLite support, we now have a storage interface for all components.

PR #423, opened 2018-03-11
use of interfaces for function calls in place of dynamic DB pointers

MohitKS5:
Signed-off-by: MohitKS5 <mohitkumarsingh907@gmail.com>
referencing #422
everything under clientapi has been changed to interface type.

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/threepid/invites.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/threepid/invites.go
@@ -25,13 +25,16 @@ import (
 	"time"
 
 	"github.com/matrix-org/dendrite/clientapi/auth/authtypes"
-	"github.com/matrix-org/dendrite/clientapi/auth/storage/accounts"
 	"github.com/matrix-org/dendrite/clientapi/producers"
 	"github.com/matrix-org/dendrite/common"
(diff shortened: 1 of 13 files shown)


src/github.com/matrix-org/dendrite/clientapi/threepid/invites.go
```diff
@@ -25,14 +25,17 @@ import (
...
 	"github.com/matrix-org/dendrite/clientapi/auth/authtypes"
-	"github.com/matrix-org/dendrite/clientapi/auth/storage/accounts"
 	"github.com/matrix-org/dendrite/clientapi/producers"
 	"github.com/matrix-org/dendrite/common"
 	"github.com/matrix-org/dendrite/common/config"
 	"github.com/matrix-org/dendrite/roomserver/api"
 	"github.com/matrix-org/gomatrixserverlib"
 )
 
+type invitesAccountsData interface {
```
anoadragon453: I would change each instance of `*AccountsData` and `*DevicesData` to `*AccountData` and `*DeviceData`, just so the variable names make more grammatical sense.

Kegsay:
We already use interfaces now because we support both Postgres and SQLite. Thanks for your work on this.

Issue #409, opened 2018-03-04
[Discussion] Registration: Disallow Usernames With Only Numbers


anoadragon453:
Although there's no mention of the correct behavior in the spec, Synapse disallows users to register usernames with only numbers, as this is reserved for guest users and usernames that are automatically generated when registering without a username.

## Synapse behavior

Request:

```
{
    "username": "18",
    "password": "blablabla123",
    "auth": {
        "session": "OTcQLwXyDbakSEfrahByrlqE",
        "type": "m.login.dummy"
    }
}
```

Response:

```
{
  "errcode": "M_UNKNOWN",
  "error": "Numeric user IDs are reserved for guest users."
}
```

## Dendrite behavior

Request:

```
{
    "username": "18",
    "password": "blablabla123",
    "auth": {
        "session": "do4HTxdgchzp9A1iR6m2FbxC",
        "type": "m.login.dummy"
    }
}
```

Response:

```
{
  "user_id": "@18:localhost",
  "access_token": "G-omcHLkHTXD-OePe1Dt1EDLc9sEW31BcApS2VcdHrs",
  "home_server": "localhost",
  "device_id": "z0_EwvF_"
}
```

anoadragon453:
@APwhitehat disallowing numeric IDs has been approved by @erikjohnston for faster username generation.

PR #424, opened 2018-03-11
[WIP] Auto-generate username if none provided during registration

### LINKED ISSUE #400: Registration: Allow registration without a username
**Spec URL:** https://matrix.org/docs/spec/client_server/r0.3.0.html#post-matrix-client-r0-register

> If omitted, the homeserver MUST generate a Matrix ID local part.

**Phase:** 1

We currently do not generate a username if `username` is not set as part of a `POST` request to `/register`.

## Sent Request

```
{
    "password": "asd",
    "auth": {
        "type": "m.login.dummy",
        "session": "Mxa1dMHgfP2VTIFNNgSkDL5n"
    }
}
```

## Received Response

**Synapse:**

```
{
  "access_token": "...",
  "home_server": "amorgan.xyz",
  "user_id": "@15:amorgan.xyz",
  "device_id": "IGKRGSXVFW"
}
```

**Dendrite:**

```
{
  "errcode": "M_INVALID_USERNAME",
  "error": "User ID can only contain characters a-z, 0-9, or '_-./'"
}
```

I believe Synapse will just set the new username to the next available user ID. So for instance, if my homeserver has registered 3 users, the newly generated username would be `@4:homeserver.xyz`. We can either use this method or some other method that generates ID in an increasing manner.

**Summary:**

- [ ] Don't error out if a username is not provided as part of registration
- [ ] Generate and return a new username if one was not provided
- [ ] Create test case(s) that checks for a successful registration without a username, and that the returned generated username is valid

APwhitehat:
fixes #400 .

This is linear over number of registered users for each request. More efficient solution is possible but would require more concurrent (re)design of clientapi, or would seem out of place.
Thoughts are welcome. :)

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/routing/register.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/routing/register.go
@@ -27,6 +27,7 @@ import (
 	"net/url"
 	"regexp"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
 
@@ -237,6 +238,38 @@ func validateRecaptcha(
 	return nil
 }
 
+// generates username for registration if none provided.
+// Sequentially checks for available username and returns the lowest available one.
+func generateUsername(
+	ctx context.Context, accountDB *accounts.Database, cfg *config.Dendrite,
+) (string, error) {
+	numericLps := make(map[string]bool)
+
+	// re matches localparts with only digits
+	re := regexp.MustCompile(`\A\d+\z`)
+	localparts, err := accountDB.GetAllLocalparts(ctx)
+
+	if err != nil {
+		return "", err
+	}
+
+	for _, lp := range localparts {
+		if re.MatchString(lp) {
+			// Length == 2 implies there is a matched localpart
+			numericLps[lp] = true
+		}
+	}
+
+	for i := 1; ; i++ {
+		lp := strconv.Itoa(i)
+		if _, ok := numericLps[lp]; !ok && !UsernameMatchesExclusiveNamespace(
+			cfg, lp,
+		) {
+			return lp, nil
+		}
+	}
+}
+
 // UsernameIsWithinApplicationServiceNamespace checks to see if a username falls
 // within any of the namespaces of a given Application Service. If no
 // Application Service is given, it will check to see if it matches any
@@ -269,6 +302,17 @@ func UsernameIsWithinApplicationServiceNamespace(
 	return false
 }
 
+// UsernameMatchesExclusiveNamespace will check if a given username matches
+// an exclusive namespace. localpart should not match Application Service namespace.
+func UsernameMatchesExclusiveNamespace(
+	cfg *config.Dendrite,
+	username string,
+) bool {
+	// Check namespaces and see if there's a match
+	matchCount := countMatchingNamespaces(cfg, username)
+	return matchCount > 0
+}
+
 // UsernameMatchesMultipleExclusiveNamespaces will check if a given username matches
 // more than one exclusive namespace. More than one is not allowed
 func UsernameMatchesMultipleExclusiveNamespaces(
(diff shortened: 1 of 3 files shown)


src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -237,6 +238,38 @@ func validateRecaptcha(
 	return nil
 }
 
+// generates username for registration if none provided.
+// Sequentially checks for available username and returns the lowest available one.
+func generateUsername(
```
anoadragon453: So I believe this is analogous to how Synapse does username generation, but it seems inherently slow.

I'd need someone from synapse dev to weigh in on whether its for some reason a bit idea to do so, but my suggestion for username generation would instead be:

* Check total number of users in DB on startup. Save in a variable.
* When registering a user, increment this variable.
* When generating a username, set the localpart to this number, then increment.

This would be much more efficient, and I'm interested in hearing anyone's comments on why this wouldn't work.

MohitKS5: @anoadragon453 
Lets say database is empty. I am the first user and registers with a username as 2 by explicitly specifying. Then we get a second a user who registers without username now by your logic he should be given a localpart of 2 since he is 2nd user but it fails. 2 users cant have same localpart.

We can give something really strange like a timestamp for username(after a bit of regExp filtering) and then let register function validate it. If user is found generate a new timestamp(it wd be different since time passed while checking). The probability that anyone would give such username explicitly decreases exponentially. At max we have to do it 2 times. 99% cases one will do.

APwhitehat:
@MohitKS5 generateUsername is run only when user_id is not provided. It
reads the database at that moment.
In case of a race, database may get either request first. In that case we
be assured that statistically if the client tries again the request would
be successful.
Since (as you mentioned) the probability of such an request is very low.

Also we'd like to minimise the number we autogenerate as user_id.

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
APwhitehat: @anoadragon453 Something like that can be very efficiently done by goroutines . But that may require passing another parameter from the setup of clientapi till register function.
This is the reason why I feel it would seem out of place. Thoughts ?

MohitKS5:
@APwhitehat yeah I was referring to Andrew. Your code is fine. Synapse uses the same logic .... 

Its just a bit slow as andrew pointed out. I was just trying to solve a major bug in andrew's logic with the example I gave. Use of goroutines with timestamps wont require any extra field in database too :)

anoadragon453:
@MohitKS5 @APwhitehat Synapse already solves this by disallowing explicit username registertration consisting solely of numbers.

anoadragon453:
But that effectively solves your issue?

> Lets say database is empty. I am the first user and registers with a username as 2 by explicitly specifying. Then we get a second a user who registers without username now by your logic he should be given a localpart of 2 since he is 2nd user but it fails. 2 users cant have same localpart.

anoadragon453:
Right, so @APwhitehat if you could switch to this method:

* Check total number of users in DB on startup. Save in a variable.
* When registering a user, increment this variable.
* When generating a username, set the localpart to this number, then increment.

And prevent users from registering a username that consists of solely numbers, returning the following if so:

```
{
  "errcode": "M_UNKNOWN",
  "error": "Numeric user IDs are reserved for guest users."
}
```

aka

```
return util.JSONResponse{
 	Code: 400,
 	JSON: jsonerror.Unknown("Numeric user IDs are reserved for guest users."),
}
```

That'd be great :D

APwhitehat:
@anoadragon453 I would like wait till #409 gets approval.

anoadragon453:
Synapse's method seems good.

Personally I'd change the `"Numeric user IDs are reserved for guest users."` error message though. Not sure if we still even use them for guest users @erikjohnston?

erikjohnston:
We might want to use numeric IDs for guests, but the error message should probably ignore that for now and give a generic `"Numeric user IDs are reserved"` or something

APwhitehat:
Closed in favor of #470.

PR #427, opened 2018-03-15
Add AS support to common.MakeAuthAPI

APwhitehat:
* Integrate #433 into `common.MakeAuthAPI`
* Update `clientapi.routing` to utilize the AS support.
* Make other Components work ok with the new `common.MakeAuthAPI`, Add "TODO" to help their migration.

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
@@ -49,25 +49,26 @@ type AccountDatabase interface {
 }
 
 // VerifyUserFromRequest authenticates the HTTP request,
-// on success returns UserID of the requester.
+// on success returns UserID, Device of the requester.
 // Finds local user or an application service user.
+// Note: For an AS user, Device is not present.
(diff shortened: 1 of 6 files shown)


erikjohnston:
Couple of comments:

1. We probably want to refactor `MakeAuthAPI`, its got too many parameters as it is and its confusing
2. I wonder if we should make an `authtypes.Device` for AS users?

APwhitehat:
1. I agree.
2. I thought AS users do not have a device. Did you mean a dummy device with AS service's AccessToken & userID ?

APwhitehat:
Once matrix-org/gomatrixserverlib#92 gets merged, we can remove the "AS_Device" with a const.

anoadragon453:
API endpoints such as this one:
https://github.com/matrix-org/dendrite/blob/78440083dfc8f522e55f11f0fce0f3ea4a103f0b/src/github.com/matrix-org/dendrite/clientapi/routing/createroom.go#L160

still check for information from the accountDB, which fails due to an account not actually existing.

What do you think of the solution of just creating a `sender_localpart` account on startup for each appservice if it doesn't already exist?

PR #431, opened 2018-03-16
Fix http responses in validateRecaptcha

### LINKED ISSUE #421: Bad error types in validateRecaptcha
Shouldn't return BadJSON for everything. https://github.com/matrix-org/dendrite/blob/b72ed3e38c33d9a31eb9dd3441a599e13ee75c4b/src/github.com/matrix-org/dendrite/clientapi/routing/register.go#L175-L238

APwhitehat:
fixes #421


MohitKS5:
Merging this will introduce a huge conflict in refactor! though its a minor change. I would rather change this in `user_interactive_auth_flow.go` than resolving. Or just let it wait for a while.

APwhitehat:
@MohitKS5 I think it would be okay if you could include these changes in your PR at the right place.

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -179,8 +179,8 @@ func validateRecaptcha(
 ) *util.JSONResponse {
 	if !cfg.Matrix.RecaptchaEnabled {
 		return &util.JSONResponse{
-			Code: http.StatusBadRequest,
-			JSON: jsonerror.BadJSON("Captcha registration is disabled"),
+			Code: http.StatusServiceUnavailable,
```
richvdh: I think StatusBadRequest is correct here?

APwhitehat: StatusBadRequest would imply there's something wrong with the received JSON which isn't the case here. Service unavailable seems more appropriate.

richvdh: nah, StatusServiceUnavailable is 503. 5xx errors mean "The server is faulty, please try again."

This needs to be a 4xx error.

APwhitehat: ah yes, server isn't faulty.
I'm guessing 403 or 404 ?

anoadragon453: I think `StatusBadRequest` might actually be correct here, seeing as it is a client error if it asked for Recaptcha registration. It shouldn't have done so as recaptcha shouldn't have been listed as one of the approved registration flows.

anoadragon453:
I have rebased this PR on latest master

PR #433, opened 2018-04-10
Add app service authentication functions

APwhitehat:
Add support for AS ?user= parameter in auth.
This is to be integrated with MakeAuthAPI in #427.

This includes changes made in #432.
Note: This was earlier part of #411.

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
@@ -42,7 +42,7 @@ type DeviceDatabase interface {
 	GetDeviceByAccessToken(ctx context.Context, token string) (*authtypes.Device, error)
 }
 
-// AccountDatabase represents a account database.
+// AccountDatabase represents an account database.
 type AccountDatabase interface {
 	// Look up the account matching the given localpart.
 	GetAccountByLocalpart(ctx context.Context, localpart string) (*authtypes.Account, error)
--- a/src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
@@ -27,6 +27,8 @@ import (
 	"github.com/matrix-org/dendrite/clientapi/auth/authtypes"
 	"github.com/matrix-org/dendrite/clientapi/httputil"
 	"github.com/matrix-org/dendrite/clientapi/jsonerror"
+	"github.com/matrix-org/dendrite/clientapi/userutil"
+	"github.com/matrix-org/dendrite/common/config"
 	"github.com/matrix-org/util"
 )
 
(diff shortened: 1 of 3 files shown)


APwhitehat:
Build failure not related to this PR. One may trigger another build.

src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
```diff
@@ -27,22 +27,92 @@ import (
...
+type AccountDatabase interface {
+	// Look up the account matching the given localpart.
+	GetAccountByLocalpart(ctx context.Context, localpart string) (*authtypes.Account, error)
+}
+
+// VerifyUserFromRequest authenticates the HTTP request,
+// on success returns UserID of the requester.
+// Finds local user or an application service user.
+// On failure returns an JSON error response which can be sent to the client.
+func VerifyUserFromRequest(
```
erikjohnston: It feels odd for there to be both this and `VerifyAccessToken`? Can we replace all instances of `VerifyAccessToken` and make it private? Or does that not make sense?

APwhitehat: I cannot make `VerifyAccessToken` private in this PR, since it's being used in `dendrite/common/httpapi.go`.
I do this change in #427. I hope it's alright.

APwhitehat: Made VerifyUserFromRequest private in #427. You may skip review of this PR but I must warn, it's big.

src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
```diff
@@ -27,22 +27,92 @@ import (
...
-// The length of generated device IDs
-var deviceIDByteLength = 6
-
 // DeviceDatabase represents a device database.
 type DeviceDatabase interface {
 	// Look up the device matching the given access token.
 	GetDeviceByAccessToken(ctx context.Context, token string) (*authtypes.Device, error)
 }
 
+// AccountDatabase represents a account database.
```
anoadragon453: "a account" -> "an account"

src/github.com/matrix-org/dendrite/clientapi/auth/storage/devices/storage.go
```diff
@@ -111,6 +115,18 @@ func (d *Database) CreateDevice(
 	return
 }
 
+// generateDeviceID creates a new device id. Returns an error if failed to generate
+// random bytes.
+func generateDeviceID() (string, error) {
```
anoadragon453: Why move this function from auth to here?

APwhitehat: Import cycles. 😖 

Should never import parent directory/package.

PR #435, opened 2018-04-12
Added notes for contributing to dependencies

CromFr:
Please double check my english ;)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,7 +43,7 @@ in the dendrite repository to update the dependency.
 
 You can then create a commit containing only the modified vendor files (along
 with the `vendor/manifest` file), name it with the command you just ran (ie
-_gb vendor update github.com/matrix-org/gomatrix_), and open a PR on Dendrite.
+`gb vendor update github.com/matrix-org/gomatrix`), and open a PR on Dendrite.
 
 ## Getting Help
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,6 +33,12 @@ nonetheless fairly well-contained.
 We ask people who are familiar with Dendrite to leave the [good first issue](https://github.com/matrix-org/dendrite/labels/good%20first%20issue)
 issues so that there is always a way for new people to come and get involved.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -33,6 +33,12 @@ nonetheless fairly well-contained.
 We ask people who are familiar with Dendrite to leave the [good first issue](https://github.com/matrix-org/dendrite/labels/good%20first%20issue)
 issues so that there is always a way for new people to come and get involved.
 
+## Contributing to dependencies
+
+Dependencies are located in `vendor/src` and are managed by `gb`. If you need make some changes in those directories, you first need to open a PR in the dependency repository. Once your PR is merged, you need to run `gb vendor update $repo_url` (example: `gb vendor update github.com/matrix-org/gomatrix`) in dendrite repository to update the dependency.
```
richvdh: please could you wrap at 80 chars or so?

richvdh: "If you need *to* make"

richvdh: in *the* dendrite repository.

CONTRIBUTING.md
```diff
@@ -33,6 +33,12 @@ nonetheless fairly well-contained.
...
+
+You can then commit the modified files and name the commit with the command you just ran.
```
richvdh: ... and open a PR in dendrite.

might be worth noting that, where possible, it's preferable to avoid mixing changes to the vendored dependencies with other changes in the same PR.

CromFr:
I shouldn't edit things on Github, I always get something wrong with it ^^

Btw, why are the dependencies versioned inside this repository? wouldn't it be simpler to have only the vendor/manifest file versioned, and let the `vendor/{pkg,src}` be installed/upgraded by gb ?

richvdh:
> Btw, why are the dependencies versioned inside this repository? wouldn't it be simpler to have only the vendor/manifest file versioned, and let the vendor/{pkg,src} be installed/upgraded by gb ?

Suggest you take that one to #dendrite-dev. (I can't really remember, tbh)

PR #419, opened 2018-03-09
Implement transactions cache to ensure idempotency of sendEvents

APwhitehat:
Fixes #327 .
This implements transactions cache, which is used to ensure idempotency of clientapi:`/rooms/{roomID}/send/{eventType}/{txnID}` .
This can be extended as required to other events with txnID.

@erikjohnston: How to proceed further ?

PR #440, opened 2018-05-10
Add module for transactions cache

APwhitehat:
Add a transactions module in dendrite/common. This is needed for idempotent APIs.

based on #419.

### DIFF
--- a/src/github.com/matrix-org/dendrite/common/transactions/transactions.go
+++ b/src/github.com/matrix-org/dendrite/common/transactions/transactions.go
@@ -1,0 +1,87 @@
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package transactions
+
+import (
+	"errors"
+	"sync"
+	"time"
+
+	"github.com/matrix-org/util"
+)
+
+// CleanupPeriod represents time in nanoseconds after which cacheCleanService runs.
+const CleanupPeriod time.Duration = 30 * time.Minute
+
+type entry struct {
+	value     *util.JSONResponse
+	entryTime time.Time
+}
+
+// Cache represents a temporary store for entries.
+// Entries are evicted after a certain period, defined by CleanupPeriod.
+type Cache struct {
+	sync.RWMutex
+	txns map[string]entry
+}
+
+// New creates a new Cache object, starts cacheCleanService.
+// Returns a referance to newly created Cache.
+func New() *Cache {
+	t := Cache{txns: make(map[string]entry)}
+
+	// Start cleanup service as the Cache is created
+	go cacheCleanService(&t)
+	return &t
+}
+
+// FetchTransaction looks up an entry for txnID in Cache.
+// Returns a JSON response if txnID is found, else returns error.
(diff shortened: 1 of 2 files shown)


src/github.com/matrix-org/dendrite/common/transactions/transactions_test.go
```diff
@@ -0,0 +1,75 @@
...
+// TestConcurentAccess provides some guarantee against corruption.
+func TestConcurentAccess(t *testing.T) {
+	fakeTxnCache := New()
+	// Signal that this test should run in parallel
+	t.Parallel()
+	// Add entries concurrently to test concurrent access.
+	for i := 1; i <= 1000; i++ {
+		go check(t, fakeTxnCache, i)
+	}
+}
```
erikjohnston: Honestly, I'm not sure this test is really testing anything much; if there is a concurrency bug then this is unlikely to pick it up. The concurrency model here is to simply use `RWLock`, so I think its safe to simply not have a unit test and to manually make sure we lock in the appropriate methods.

(If the cache was more complicated, then we'd probably want to add hooks into the cache to let us write deterministic tests.)

src/github.com/matrix-org/dendrite/common/transactions/transactions.go
```diff
@@ -0,0 +1,87 @@
...
+import (
+	"errors"
+	"sync"
+	"time"
+
+	"github.com/matrix-org/util"
+)
+
+// CleanupPeriod represents time in nanoseconds after which cacheCleanService runs.
+const CleanupPeriod time.Duration = 30 * time.Minute
```
erikjohnston: Personally I would make this an option for `Cache`

erikjohnston:
Thinking about it, I wonder if we should be a little concerned about the clean up method. We'll need to lock for during iteration of the map, which could take a little while if the map is large.

It's probably fine for now, but an alternative trick that you can use to make the clean up efficient is to:
1. Have two maps instead of one
2. During lookup check both maps
3. Add new transactions to the first map
3. Every N minutes: set the second map point to the first, and set the first map to a new empty map.

This works by effectively rotating the maps. A transaction will get added to the first map, it will at some point get rotated to the second map, and finally (after at least N minutes) the map with the transaction gets dropped.

src/github.com/matrix-org/dendrite/common/transactions/transactions.go
```diff
@@ -0,0 +1,86 @@
...
+type Cache struct {
+	sync.RWMutex
+	txnsMaps      [2]txnsMap
+	cleanupPeriod time.Duration
+}
+
+// New creates a new Cache object, starts cacheCleanService.
+// Takes cleanupPeriod (in minutes) as argument, in case of 0 DefaultCleanupPeriod is used instead.
+// Returns a reference to newly created Cache.
+func New(cleanupPeriodInMins int) *Cache {
```
erikjohnston: Given we have a `time.Duration` type, I'd use that. If you want to make it optional then either take a pointer and check for null or have `New()` and `NewWithPeriod(..)`

src/github.com/matrix-org/dendrite/common/transactions/transactions.go
```diff
@@ -0,0 +1,86 @@
...
+// DefaultCleanupPeriod represents the default time duration after which cacheCleanService runs.
+const DefaultCleanupPeriod time.Duration = 30 * time.Minute
+
+type txnsMap map[string]*util.JSONResponse
+
+// Cache represents a temporary store for response entries.
+// Entries are evicted after a certain period, defined by cleanupPeriod.
+type Cache struct {
+	sync.RWMutex
+	txnsMaps      [2]txnsMap
```
erikjohnston: I'd probably give a quick comment here to say why there are two maps, and how this all works internally

src/github.com/matrix-org/dendrite/common/transactions/transactions.go
APwhitehat: In most scenario one would let the Default be as it is, passing nil for that case is cumbersome.
So I will avoid the nil-able pointer parameter.

PR #445, opened 2018-05-17
Fix typo

l2dy:
```
// Returns a list of output events to write to the kafka log to inform the
// consumers about the invites added or retired by the change in current state.
```
So we probably shouldn't return `nil` on success.


anoadragon453:
I believe this is the intended behavior. Thanks!

PR #446, opened 2018-05-21
[roomserver]Store transaction ID and prevent reprocessing of events

APwhitehat:
Reconfiguration of postgresql is required to test this.

` Signed-off-by: Anant Prakash <anantprakashjsr@gmail.com>`

### DIFF
--- a/src/github.com/matrix-org/dendrite/roomserver/storage/transactions_table.go
+++ b/src/github.com/matrix-org/dendrite/roomserver/storage/transactions_table.go
@@ -32,14 +32,12 @@ CREATE TABLE IF NOT EXISTS roomserver_transactions (
 	event_id TEXT NOT NULL,
 	-- A transaction ID is unique for a user and device
 	-- This automatically creates an index.
-	CONSTRAINT roomserver_transaction_unique PRIMARY KEY (transaction_id, device_id, user_id)
+	PRIMARY KEY (transaction_id, device_id, user_id)
 );
 `
 const insertTransactionSQL = "" +
 	"INSERT INTO roomserver_transactions (transaction_id, device_id, user_id, event_id)" +
-	" VALUES ($1, $2, $3, $4)" +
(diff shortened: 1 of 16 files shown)


erikjohnston:
Unfortunately, I forgot two things that makes things a bit more annoying:
1. Transactions IDs are scoped to a particular user's device. This means that the same transaction ID may be used across different devices, so all queries for transaction IDs need to take that into account.
2. We need to make sure the client API returns the right event ID, i.e. if we have already processed the transaction we need to return the existing event's ID, rather than the ID for the event we are trying to persist. This means that the API the clientapi uses to send events needs to return an event ID.

The second point should be relatively easy to do. However the first point is more annoying, since the events table currently stores neither the sender user_id nor device ID. Given that, I think I've changed my mind and we should probably have it as a a separate table (like you originally suggested).

erikjohnston:
Oh, and also: whenever you add new queries (especially when using new columns), its a good idea to stop and consider whether it needs a new index to be added.

For example, currently in this PR we're adding a query to select based on `transaction_id`, which is not indexed. This means that the database would need to scan the entire table to find the row, which isn't ideal, so we should in that case add an index on `transaction_id`.

src/github.com/matrix-org/dendrite/roomserver/storage/transactions_table.go
```diff
@@ -0,0 +1,88 @@
...
+	-- A transaction ID is unique for a user and device
+	-- This automatically creates an index.
+	CONSTRAINT roomserver_transaction_unique PRIMARY KEY (transaction_id, device_id, user_id)
+);
+`
+const insertTransactionSQL = "" +
+	"INSERT INTO roomserver_transactions (transaction_id, device_id, user_id, event_id)" +
+	" VALUES ($1, $2, $3, $4)" +
+	" ON CONFLICT ON CONSTRAINT roomserver_transaction_unique" +
+	" DO NOTHING"
```
erikjohnston: Why do we want to noop if the row already exists?

PR #429, opened 2018-03-16
[WIP] Add Macaroon generator & handler, implement Token-based login

APwhitehat:
* Adds macaroon generator & handler, uses vendor package `github.com/go-macaroon/macaroon`
* Include helper functions to read , validate, etc tokens (Macaroons).
* `login.go` is littered, #432 will help immensely in reducing the same.
* move LoginResponse & LoginRequest from `login.go` to `auth.authtypes`

PR #447, opened 2018-05-22
Add macaroon token handlers

APwhitehat:
Add token validation and userID query.

was earlier part of #429

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/auth/tokens/tokens_handlers.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/auth/tokens/tokens_handlers.go
@@ -1,0 +1,115 @@
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tokens
+
+import (
+	"errors"
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/matrix-org/dendrite/clientapi/jsonerror"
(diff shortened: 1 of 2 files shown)


src/github.com/matrix-org/dendrite/clientapi/auth/tokens/tokens_handlers.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	return nil
+}
+
+// verifyCaveats verifies caveats associated with a login token macaroon.
+// which are "gen = 1", "user_id = ...", "time < ..."
+// Returns nil on successful verification, else returns an error.
+func verifyCaveats(caveats []string, userID string) error {
+	// variable verified represents a bitmap
+	// last 4 bit is Uvvv where,
```
anoadragon453: "last 4 bit is Uvvv" -> "last 4 bits are Uvvv"

src/github.com/matrix-org/dendrite/clientapi/auth/tokens/tokens_handlers.go
```diff
@@ -0,0 +1,115 @@
...
+		}
+	}
+	return nil
+}
+
+// verifyCaveats verifies caveats associated with a login token macaroon.
+// which are "gen = 1", "user_id = ...", "time < ..."
+// Returns nil on successful verification, else returns an error.
+func verifyCaveats(caveats []string, userID string) error {
+	// variable verified represents a bitmap
```
anoadragon453: This is certainly one way to do it :) Since we only really need to differentiate between good versus bad caveats, I'd say this approach works fine.

src/github.com/matrix-org/dendrite/clientapi/auth/tokens/tokens_handlers_test.go
```diff
@@ -0,0 +1,56 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tokens
+
+import "testing"
+
+// If any of these options are missing, validation should fail
+var invalidMissings = []string{"ServerPrivateKey", "UserID"}
+
```
anoadragon453: Would just add a couple more test cases for an expired token and/or invalid userID.

APwhitehat:
Closed in favour of matrix-org/gomatrixserverlib#89

Issue #454, opened 2018-05-22
Set up Prometheus handler for each component


anoadragon453:
The plan is to have a prometheus HTTP handler running per component. This should be set up in the `cmd/dendrite-*` file for each component.

```
// The Handler function provides a default handler to expose metrics
// via an HTTP server. "/metrics" is the usual endpoint for that.
http.Handle("/metrics", promhttp.Handler())
log.Fatal(http.ListenAndServe(":8080", nil))
```

Port should be that of each component of course.

anoadragon453:
Done in #459

PR #455, opened 2018-05-22
Add prometheus metrics tracking

anoadragon453:
This adds Prometheus as a dependency and initializes the http server for the monolith setup (per-component http server tracked at #454) at http://127.0.0.1:8080/metrics

It also implements a very basic counter for the total amount of registered users (`dendrite_clientapi_reg_users_total`). Metric name is currently `dendrite-componentName-name`. Any feedback on that structure is welcome.


anoadragon453:
Closed in favor of #459

PR #459, opened 2018-05-23
Update Prometheus metrics tracking

anoadragon453:
So turns out we already had a bit of Prometheus metrics tracking, but the methods and libraries it used were outdated/deprecated.

This updates it to the latest stable commit, and serves metrics on /metrics for each component separately, as well as for the monolith.

Note: I haven't tested without naffka yet, but it should work in theory.

This still adds a basic metrics of total amount of registered users.


anoadragon453:
@erikjohnston So only allowing localhost to access `/metrics` you mean?

PR #462, opened 2018-05-23
Add Application Service component

### LINKED ISSUE #449: AppServices: Create New Component
We need to make a new component for Application Services. It will take care of the following tasks:

* Subscribe to the Roomserver kafka stream and pull out any events that need to be forwarded to registered Application Services.
* Make requests to the various [Application Service HTTP APIs](https://matrix.org/docs/spec/application_service/unstable.html#http-apis).
* Handling `/_matrix/app/*` HTTP requests from Application Services.

It will **not** handle:

* Registering Application Services on startup.
* Handling any `/_matrix/client/*` HTTP requests from Application Services.

anoadragon453:
The component subscribes to the Roomserver kafka stream, filtering events to be eventually sent off to relevant app services, as well as handling incoming HTTP traffic from app services on the `/matrix/app/r0/*` route.

Linting will fail on some unused variables in `appservice/routing/routing.go` as the endpoints currently don't have any functionality.

This is only the beginning. Fixes #449

### DIFF
--- a/dendrite-config.yaml
+++ b/dendrite-config.yaml
@@ -112,6 +112,7 @@ listen:
     media_api: "localhost:7774"
     public_rooms_api: "localhost:7775"
     federation_sender: "localhost:7776"
+    appservice: "localhost:7777"
 
 # The configuration for tracing the dendrite components.
 tracing:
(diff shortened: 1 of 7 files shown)


erikjohnston:
This looks good. Is there a way for you to make the linting pass? I'm not keen on having the linting fail on master

anoadragon453:
Sure, I'll stick a nolint in there for the time being.

PR #467, opened 2018-05-25
Filter Roomserver Events Based on Application Service

anoadragon453:
When Application Services are registered, they specify a number of namespaces (regexes based on userID, RoomID and Room Alias) for events that they want to know about. We pull these namespaces in at startup and compile their regex into Regexp objects, which we later use to quickly check if several properties of an incoming event match what a given application service wants.

Checking sender names and room IDs were quite simple as they are part of the event, however to get all the aliases of a room, a new aliasAPI was necessary. `GetAliasesForRoomID` was added for this, right alongside `GetRoomIDForAlias` which was renamed from its earlier quite vague name of `GetAliasRoomID`.

I believe this new alias API is the correct way to get aliases for a given RoomID, though a database query for Aliases from RoomID was already available, so let me know if this was already possible somehow and the new call isn't necessary.

At the moment it just prints out whether an app service was interested in an event or not, but the next step is to queue up that event and have a module batch and send them off to the appservice.

### DIFF
--- a/src/github.com/matrix-org/dendrite/appservice/consumers/roomserver.go
+++ b/src/github.com/matrix-org/dendrite/appservice/consumers/roomserver.go
@@ -17,6 +17,7 @@ package consumers
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 
 	"github.com/matrix-org/dendrite/clientapi/auth/storage/accounts"
 	"github.com/matrix-org/dendrite/common"
@@ -85,6 +86,7 @@ func (s *OutputRoomEventConsumer) onMessage(msg *sarama.ConsumerMessage) error {
 	}
 
 	ev := output.NewRoomEvent.Event
+	fmt.Println("got event", ev)
 	log.WithFields(log.Fields{
 		"event_id": ev.EventID(),
 		"room_id":  ev.RoomID(),
--- a/src/github.com/matrix-org/dendrite/appservice/consumers/roomserver.go
+++ b/src/github.com/matrix-org/dendrite/appservice/consumers/roomserver.go
@@ -29,11 +29,16 @@ import (
 	sarama "gopkg.in/Shopify/sarama.v1"
 )
 
+var (
+	appServices []config.ApplicationService
+)
+
 // OutputRoomEventConsumer consumes events that originated in the room server.
 type OutputRoomEventConsumer struct {
 	roomServerConsumer *common.ContinualConsumer
 	db                 *accounts.Database
 	query              api.RoomserverQueryAPI
+	alias              api.RoomserverAliasAPI
 	serverName         string
 }
 
@@ -43,7 +48,9 @@ func NewOutputRoomEventConsumer(
 	kafkaConsumer sarama.Consumer,
 	store *accounts.Database,
 	queryAPI api.RoomserverQueryAPI,
+	aliasAPI api.RoomserverAliasAPI,
 ) *OutputRoomEventConsumer {
+	appServices = cfg.Derived.ApplicationServices
 
 	consumer := common.ContinualConsumer{
 		Topic:          string(cfg.Kafka.Topics.OutputRoomEvent),
@@ -54,6 +61,7 @@ func NewOutputRoomEventConsumer(
 		roomServerConsumer: &consumer,
 		db:                 store,
 		query:              queryAPI,
+		alias:              aliasAPI,
 		serverName:         string(cfg.Matrix.ServerName),
 	}
 	consumer.ProcessMessage = s.onMessage
@@ -86,7 +94,6 @@ func (s *OutputRoomEventConsumer) onMessage(msg *sarama.ConsumerMessage) error {
 	}
 
 	ev := output.NewRoomEvent.Event
-	fmt.Println("got event", ev)
 	log.WithFields(log.Fields{
 		"event_id": ev.EventID(),
 		"room_id":  ev.RoomID(),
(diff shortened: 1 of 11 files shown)


anoadragon453:
Ok, should be filtering all known events now.

src/github.com/matrix-org/dendrite/appservice/consumers/roomserver.go
```diff
@@ -142,3 +156,57 @@ func (s *OutputRoomEventConsumer) lookupStateEvents(
...
+	for _, roomNamespace := range appservice.NamespaceMap["rooms"] {
+		if roomNamespace.RegexpObject.MatchString(event.RoomID()) {
+			return true
+		}
+	}
+
+	// Check all known room aliases of the room the event came from
+	queryReq := api.GetAliasesForRoomIDRequest{RoomID: event.RoomID()}
+	var queryRes api.GetAliasesForRoomIDResponse
+	if err := s.alias.GetAliasesForRoomID(context.TODO(), &queryReq, &queryRes); err == nil {
```
erikjohnston: You should be threading the context through. `context.TODO()` should only be used when you otherwise don't know quite where to get one yet, but we already have made a context for `UpdateMemberships`, so lets use that one.

src/github.com/matrix-org/dendrite/appservice/consumers/roomserver.go
```diff
@@ -142,3 +156,57 @@ func (s *OutputRoomEventConsumer) lookupStateEvents(
...
+		for _, alias := range queryRes.Aliases {
+			for _, aliasNamespace := range appservice.NamespaceMap["aliases"] {
+				if aliasNamespace.RegexpObject.MatchString(alias) {
+					return true
+				}
+			}
+		}
+	} else {
+		log.WithError(err).Errorf("Unable to get aliases for Room with ID: %s", event.RoomID())
```
erikjohnston: Probably better to use the structured logging, so something like:

```golang
log.WithFields(log.Fields{
    "room_id": event.RoomID(),
}).WithError(err).Error("Unable to get aliases for room")
```

src/github.com/matrix-org/dendrite/common/config/appservice.go
```diff
@@ -182,9 +199,9 @@ func checkErrors(config *Dendrite) error {
 			}
 		}
 	}
-	setupRegexps(config)
 
-	return nil
+	err = setupRegexps(config)
+	return err
```
erikjohnston: Why are we not just `return setupRegexps(config)`, and why do we need to name the return value?

src/github.com/matrix-org/dendrite/common/config/config.go
```diff
@@ -235,12 +235,19 @@ type Dendrite struct {
...
+		// exclusive Application Service namespaces
 		ExclusiveApplicationServicesUsernameRegexp *regexp.Regexp
-
-		// TODO: Exclusive alias, room regexp's
+		// When a user creates a room alias, we check that it isn't already
+		// reserved by an application service
+		ExclusiveApplicationServicesAliasRegexp *regexp.Regexp
+		// TODO: Huh? When a room ID is created, we block the client from
+		// creating it if it falls under an application service's exclusive
+		// roomID regex? But why tho
```
erikjohnston: We should probably find out the answer to this question before writing the code and landing it :)

src/github.com/matrix-org/dendrite/roomserver/alias/alias.go
```diff
@@ -98,14 +98,30 @@ func (r *RoomserverAliasAPI) GetAliasRoomID(
...
+) error {
+	// Look up the aliases in the database for the given RoomID
+	aliases, err := r.DB.GetAliasesForRoomID(ctx, request.RoomID)
+	if err != nil {
+		return err
+	}
+
+	response.Aliases = aliases
+	return nil
+}
```
erikjohnston: Where was the implementation for `GetAliasesFromRoomID`? I'm surprised I'm not seeing that removed in the diff.

src/github.com/matrix-org/dendrite/common/config/appservice.go
anoadragon453: Derp, yes that really should be `return setupRegexps(config)`.

src/github.com/matrix-org/dendrite/appservice/consumers/roomserver.go
anoadragon453: Gotcha. Still figuring out how context works :)

src/github.com/matrix-org/dendrite/roomserver/alias/alias.go
anoadragon453: There wasn't one. I've had to add it :)

anoadragon453:
* Prevented user from creating more than one regex per namespace type

  - Got caught out by realizing I had an extra '-' in the config file. This
prevents anyone from making the same mistake :)

* There was a pointer reference to a string slice. Slices are already
pointers, so it wasn't necessary. Testing shows this hasn't broken
anything.

* Removed exclusive RoomID namespace regex, as we won't need to check
upon room creation if the ID is reserved exclusively by an AS (as this
is silly and horribly inefficient).

* Fixed all else mentioned

@erikjohnston Back to you

src/github.com/matrix-org/dendrite/roomserver/alias/alias.go
erikjohnston: Then how was `RoomserverAliasAPIDatabase` implemented?

anoadragon453: The database queries and supporting functions were already there, a method to expose it to the RoomserverAliasAPI was all that was needed.

PR #470, opened 2018-05-26
Auto-generate username if none provided during registration

### LINKED ISSUE #400: Registration: Allow registration without a username

APwhitehat:
Auto-generate username if none provided during registration.
Reserve numeric user IDs.

fixes #400.


src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -399,10 +400,27 @@ func Register(
 	// Retrieve or generate the sessionID
 	sessionID := r.Auth.Session
 	if sessionID == "" {
-		// Generate a new, random session ID
+		// Generate a new, random session ID`
 		sessionID = util.RandomString(sessionIDLength)
 	}
 
+	// Don't allow numeric usernames less than MAX_INT64.
```
anoadragon453: greater than? :P

APwhitehat: No ?
We are using numbers less than int_max 64, so only disallow numbers less than the same.

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -403,6 +404,23 @@ func Register(
...
+		return util.JSONResponse{
+			Code: http.StatusBadRequest,
+			JSON: jsonerror.InvalidUsername("Numeric user IDs are reserved"),
+		}
+	}
+	// Auto generate a numeric username if r.Username is empty
+	if r.Username == "" {
+		id, err := accountDB.GetNewNumericLocalpart(req.Context())
+		if err != nil {
+			return jsonerror.InternalServerError()
```
erikjohnston: We use `return httputil.LogThenError(req, err)` so that it includes an appropriate error message in the response.

src/github.com/matrix-org/dendrite/clientapi/routing/register.go
anoadragon453: Ah, for users trying to register a username. Thought this was referencing generated usernames.

anoadragon453:
@APwhitehat Could you rebase on master please?

PR #474, opened 2018-05-30
Scan AppServiceID from accounts table and fix typo

APwhitehat:
We should scan `appservice_id` as well.
Also a typo somehow crept in here.


anoadragon453:
Heh, that's what happens when you write `if err != nil` about 20 times a day :P

PR #475, opened 2018-05-30
Move makeUserID into userutil and prevent code duplication

APwhitehat:
It's raining in my city but I love DRY.


anoadragon453:
To be fair, this is probably something else that should go in gomatrixserverlib, but so should the rest of userutil IMO.

Will merge this for now for code cleanup. Then we should at some point take a good hard look at the utility stuff in Dendrite and see what should be extracted into the general server lib.

APwhitehat:
I feel the same way. 

Many stuff could be moved to gomatrixserverlib.
Should definitely try to do that in near future.

APwhitehat:
ah, ungraceful merge. I'll fix this.

anoadragon453:
Looks good. Thanks!

Issue #450, opened 2018-05-22
AppServices: PUT /transactions/{txnId}


anoadragon453:
**Spec:** https://matrix.org/docs/spec/application_service/unstable.html#put-transactions-txnid

As events come in, the AppServices component will read and filter those that are relevant to application services by checking several things:

* The `user_id` of the event is within the App Service's `users` namespace.
* The `room_id` of the event is within the App Service's `rooms` namespace.
* One of the aliases of the room the event was sent in is within the App Service's `aliases` namespace.

The app service is considered to be "interested" in such events. And before you ask, yes, app services are interested in their own events. These events need to be batched up in an array and sent off to the app service via a `PUT` to `/transactions/{txnId}`. `txnId` should just be an incrementing integer, only requirement is the `txnId` be different between each batch of requests.

We also need to employ exponential backoff if the AS is unable to return a `200` after a quick timeout. We should keep retrying to send the same batch over and over, waiting longer and longer each time.

PR #477, opened 2018-06-01
Send Application Service Events

anoadragon453:
Adds the database backend for the application service component, stores filtered events for goroutines for each app service to grab them and send them off to the transactions API.

The database consists of two tables, one for events, along with the app service they're destined for and some metadata. The second table is a simple relation between an app service and the last transaction ID that was used to send data to it.

The full bullet-point list of changes is below:

* Modify INSTALL.md and dendrite-config.yaml for the new appservice database.
* Correct all instances of casing on 'application service' to align with
spec.
* Store incoming events that an app service is interested in in the
database to be later read by transaction workers.
* Retrieve these events from transaction workers, one per AS.
* Minimal transaction ID data is stored as well to recover after
server failure.
* Send events to AS and exponentially backoff on failure.

Fixes: #450


src/github.com/matrix-org/dendrite/appservice/workers/transaction_scheduler.go
```diff
@@ -0,0 +1,210 @@
...
+	for _, appservice := range cfg.Derived.ApplicationServices {
+		// Don't create a worker if this AS doesn't want to receive events
+		if appservice.URL != "" {
+			go worker(appserviceDB, appservice, eventCounterMap)
+		}
+	}
+	return nil
+}
+
+// worker is a goroutine that
```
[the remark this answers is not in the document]
anoadragon453: Guess I was still deciding exactly what it should do :P Lemme fix that up.

src/github.com/matrix-org/dendrite/appservice/consumers/roomserver.go
```diff
@@ -165,13 +173,20 @@ func (s *OutputRoomEventConsumer) lookupStateEvents(
...
 		for _, appservice := range appServices {
 			// Check if this event is interesting to this application service
 			if s.appserviceIsInterestedInEvent(ctx, event, appservice) {
-				// TODO: Queue this event to be sent off to the application service
-				fmt.Println(appservice.ID, "was interested in", event.Sender(), event.Type(), event.RoomID())
+				// Queue this event to be sent off to the application service
+				if err := s.asDB.StoreEvent(ctx, appservice.ID, event); err != nil {
+					log.WithError(err).Warn("failed to insert incoming event into appservices database")
+				} else {
+					ecm[appservice.ID]++
```
APwhitehat: concurrent map writes, yikes!

src/github.com/matrix-org/dendrite/appservice/consumers/roomserver.go
```diff
@@ -31,26 +31,32 @@ import (
 
 var (
 	appServices []config.ApplicationService
+	ecm         map[string]int
```
APwhitehat: Please don't use globals 😢

src/github.com/matrix-org/dendrite/appservice/storage/appservice_events_table.go
```diff
@@ -0,0 +1,198 @@
...
+) (
+	eventIDs []string,
+	events []gomatrixserverlib.ApplicationServiceEvent,
+	err error,
+) {
+	eventRows, err := s.selectEventsByApplicationServiceIDStmt.QueryContext(ctx, applicationServiceID, limit)
+	if err != nil {
+		return nil, nil, err
+	}
+	defer eventRows.Close() // nolint: errcheck
```
APwhitehat: make errcheck happy. :)
```go
	defer func() {
		err := eventRows.Close()
		if err != nil {
			logrus.WithError(err).Fatal("This is very serious error, better exit")
		}
	}()
```

src/github.com/matrix-org/dendrite/appservice/storage/appservice_events_table.go
```diff
@@ -0,0 +1,198 @@
...
+	for eventRows.Next() {
+		var eventID, originServerTimestamp, roomID, eventType, sender, eventContent *string
+		if err = eventRows.Scan(
+			&eventID,
+			&originServerTimestamp,
+			&roomID,
+			&eventType,
+			&sender,
+			&eventContent,
+		); err != nil || eventID == nil || roomID == nil || eventType == nil || sender == nil || eventContent == nil {
```
APwhitehat: As Erik said, better to make fields NOT NULL than checking after Scan.

src/github.com/matrix-org/dendrite/appservice/storage/appservice_events_table.go
```diff
@@ -0,0 +1,198 @@
...
+	return
+}
+
+// countEventsByApplicationServiceID inserts an event mapped to its corresponding application service
+// IDs into the db.
+func (s *eventsStatements) countEventsByApplicationServiceID(
+	ctx context.Context,
+	appServiceID string,
+) (int, error) {
+	var count *int
```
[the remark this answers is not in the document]
APwhitehat: This actually works :P (although not idiomatic).

anoadragon453:
Yeah, I agree. That is a nasty line.

anoadragon453:
Good point. Luckily Erik's paper scribbles describe a method that removes the need for it :)

src/github.com/matrix-org/dendrite/appservice/storage/appservice_events_table.go
```diff
@@ -0,0 +1,198 @@
...
+	-- The JSON representation of the event. Text to avoid db JSON parsing
+	event_json TEXT,
+	-- The ID of the transaction that this event is a part of
+	txn_id INTEGER
+);
+`
+
+const selectEventsByApplicationServiceIDSQL = "" +
+	"SELECT event_id, origin_server_ts, room_id, type, sender, event_json FROM appservice_events " +
+	"WHERE as_id = $1 ORDER BY as_id LIMIT $2"
```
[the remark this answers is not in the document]
anoadragon453: Huh, TIL `ASC` isn't guaranteed unless explicitly specified.

src/github.com/matrix-org/dendrite/appservice/storage/appservice_events_table.go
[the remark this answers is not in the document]
anoadragon453: Hm, would there ever really be a case where the count was NULL, though?

src/github.com/matrix-org/dendrite/appservice/consumers/roomserver.go
anoadragon453: This'll thankfully no longer be necessary with the new event-based method of sending events Erik proposed.

src/github.com/matrix-org/dendrite/appservice/workers/transaction_scheduler.go
```diff
@@ -0,0 +1,211 @@
...
+// worker is a goroutine that sends any queued events to the application service
+// it is given.
+func worker(db *storage.Database, as config.ApplicationService, ecm map[string]int) {
+	// Initialize transaction ID counter
+	var err error
+	currentTransactionID, err = db.GetTxnIDWithAppServiceID(context.TODO(), as.ID)
+	if err != nil {
+		logrus.WithError(err).Warnf("appservice worker for %s unable to get latest transaction ID from DB",
+			as.ID)
+	}
```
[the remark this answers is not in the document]
anoadragon453: Would require a restart of Dendrite to re-enable the appservice then, no?

src/github.com/matrix-org/dendrite/appservice/workers/transaction_scheduler.go
```diff
@@ -0,0 +1,211 @@
...
+				logrus.WithError(err).Error("appservice worker unable to marshal events")
+
+				// Wait a little bit before trying again
+				time.Sleep(transactionBreakTime)
+				continue
+			}
+
+			// Send the events off to the application service
+			eventsSent, err := send(client, as, transactionJSON, len(events))
+			if err != nil {
```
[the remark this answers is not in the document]
anoadragon453: I believe that's the case already. If there's an error, we backoff, continue, and the next time round try to send the same transaction again.

src/github.com/matrix-org/dendrite/appservice/workers/transaction_scheduler.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	// Initialize counter to amount of events currently in the database
+	eventCount, err := db.CountEventsWithAppServiceID(context.TODO(), as.ID)
+	if err != nil {
+		logrus.WithError(err).Warn("appservice worker unable to count queued events from DB")
+	}
+	ecm[as.ID] = eventCount
+
+	// Initialize backoff exponent (2^x secs). Max 9, aka 512s.
```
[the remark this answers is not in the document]
anoadragon453: This number was taken from Synapse, but I agree that 512s is a little excessive. I'll cap it to 64s.

src/github.com/matrix-org/dendrite/appservice/workers/transaction_scheduler.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+	return nil
+}
+
+// worker is a goroutine that sends any queued events to the application service
+// it is given.
+func worker(db *storage.Database, as config.ApplicationService, ecm map[string]int) {
+	// Initialize transaction ID counter
+	var err error
+	currentTransactionID, err = db.GetTxnIDWithAppServiceID(context.TODO(), as.ID)
```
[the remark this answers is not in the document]
anoadragon453: Would this context exist for the whole lifetime of the worker, or would we make once every time we pull new messages from the database?

src/github.com/matrix-org/dendrite/appservice/workers/transaction_scheduler.go
```diff
@@ -0,0 +1,226 @@
...
+		transactionJSON, err := createTransaction(events)
+		if err != nil {
+			log.WithError(err).Fatalf("appservice %s worker unable to marshal events",
+				ws.AppService.ID)
+
+			return
+		}
+
+		// Send the events off to the application service
+		err = send(client, ws.AppService, transactionJSON)
```
krombel: As `transactionJSON` is recomputed every time it is possible that a later request might contain more events then the previous one - even with having the same transaction ID.
This might lead to loosing events as the app-service might deduplicate on the first event so will never handle the events which came in between the first and later tries

src/github.com/matrix-org/dendrite/appservice/workers/transaction_scheduler.go
```diff
@@ -0,0 +1,226 @@
...
+		}
+
+		// Update transactionID
+		currentTransactionID++
+		if err = db.UpsertTxnIDWithAppServiceID(ctx, ws.AppService.ID, currentTransactionID); err != nil {
+			log.WithError(err).Fatalf("unable to update transaction ID for %s",
+				ws.AppService.ID)
+			return
+		}
+		waitForEvents(&ws)
```
krombel: What happens if we hit `transactionBatchSize` in the transaction in [104](https://github.com/matrix-org/dendrite/pull/477/files#diff-d84306218ea8217c1024f21eb34d9fcbR104)
We should not wait for the next Broadcast until we process further queries when we have pending events in the database

src/github.com/matrix-org/dendrite/appservice/workers/transaction_scheduler.go
```diff
@@ -0,0 +1,226 @@
...
+
+	"github.com/matrix-org/dendrite/appservice/storage"
+	"github.com/matrix-org/dendrite/appservice/types"
+	"github.com/matrix-org/dendrite/common/config"
+	"github.com/matrix-org/gomatrixserverlib"
+	log "github.com/sirupsen/logrus"
+)
+
+var (
+	// TODO: Expose these in the config?
```
anoadragon453: Feedback on this is appreciated.

src/github.com/matrix-org/dendrite/appservice/workers/transaction_scheduler.go
anoadragon453: Good point, the earlier implementation dealt with this, but as we no longer use an event counter, this did not.

I've added to the query a count for the total events, and we won't sleep until that is 0. Adding this count is slightly more expensive that before, but everything is still done in one query, so it should still be fast.

src/github.com/matrix-org/dendrite/appservice/workers/transaction_scheduler.go
```diff
@@ -0,0 +1,242 @@
...
+		eventsCount, maxEventID, transactionID, transactionJSON, err := createTransaction(ctx, db, ws.AppService.ID)
+		if err != nil {
+			log.WithError(err).Fatalf("appservice %s worker unable to create transaction",
+				ws.AppService.ID)
+
+			return
+		}
+
+		// Send the events off to the application service
+		err = send(client, ws.AppService, transactionID, transactionJSON)
```
krombel: why not introducing another loop here which can reuse `transactionJSON`? Something like the following:
```
for {
    err = send (...)
    if err != nil {
        backoff(err, &ws)
        continue
    }
    break
}
```

src/github.com/matrix-org/dendrite/appservice/storage/appservice_events_table.go
```diff
@@ -0,0 +1,247 @@
...
+	-- The ID of the transaction that this event is a part of
+	txn_id BIGINT NOT NULL
+);
+
+CREATE INDEX IF NOT EXISTS appservice_events_as_id ON appservice_events(as_id);
+`
+
+const selectEventsByApplicationServiceIDSQL = "" +
+	"SELECT id, event_id, origin_server_ts, room_id, type, sender, event_content, txn_id " +
+	"FROM appservice_events WHERE as_id = $1 ORDER BY txn_id DESC LIMIT $2"
```
krombel: Is it reasonable to change this to `ORDER BY txn_id DESC, id ASC`?

anoadragon453: Yes, that is supported and will prevent events from going out of order :)

anoadragon453:
Just a quick update that this PR is currently blocked on https://github.com/matrix-org/gomatrixserverlib/pull/94. That PR should be hopefully good now @erikjohnston @APwhitehat :+1:

src/github.com/matrix-org/dendrite/appservice/consumers/roomserver.go
```diff
@@ -74,9 +76,10 @@ func (s *OutputRoomEventConsumer) Start() error {
 	return s.roomServerConsumer.Start()
 }
 
-// onMessage is called when the sync server receives a new event from the room server output log.
-// It is not safe for this function to be called from multiple goroutines, or else the
-// sync stream position may race and be incorrectly calculated.
+// onMessage is called when the sync server receives a new event from the room
+// server output log. It is not safe for this function to be called from
+// multiple goroutines, or else the sync stream position may race and be
+// incorrectly calculated.
```
erikjohnston: We're not in the sync server.

src/github.com/matrix-org/dendrite/appservice/consumers/roomserver.go
```diff
@@ -113,6 +116,10 @@ func (s *OutputRoomEventConsumer) onMessage(msg *sarama.ConsumerMessage) error {
 	}
 
 	// Check if any events need to passed on to external application services
+	if len(events) > 0 {
+		// Check if this was a membership event
+		return s.filterRoomserverEvents(ctx, events)
```
erikjohnston: What is this doing? Why are we returning? The fact that we then call this again with `append(events, ev)` after feels weird.

erikjohnston: I'm also not sure we want to be resending state events to the AS. Does synapse do this?

src/github.com/matrix-org/dendrite/appservice/storage/appservice_events_table.go
```diff
@@ -0,0 +1,256 @@
...
+			&event.EventID,
+			&event.OriginServerTimestamp,
+			&event.RoomID,
+			&event.Type,
+			&event.UserID,
+			&eventContent,
+			&txnID,
+		)
+		if err != nil {
+			fmt.Println("Failed:", err.Error())
```
erikjohnston: No printlns.

src/github.com/matrix-org/dendrite/appservice/storage/txn_id_counter_table.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+const txnIDSchema = `
+-- Keeps a count of the current transaction ID per application service 
+CREATE TABLE IF NOT EXISTS txn_id_counter (
+	-- The ID of the application service the this count belongs to
+	as_id TEXT NOT NULL PRIMARY KEY,
+	-- The last-used transaction ID
+	txn_id INTEGER NOT NULL
+);
```
erikjohnston: You can use https://www.postgresql.org/docs/9.6/static/sql-createsequence.html, and have a transaction ID be globally unique rather than scoped to an AS.

src/github.com/matrix-org/dendrite/appservice/workers/transaction_scheduler.go
```diff
@@ -0,0 +1,239 @@
...
+func worker(db *storage.Database, ws types.ApplicationServiceWorkerState) {
+	log.Infof("Starting Application Service %s", ws.AppService.ID)
+	ctx := context.Background()
+
+	// Initialize transaction ID counter
+	var err error
+	currentTransactionID, err = db.GetTxnIDWithAppServiceID(ctx, ws.AppService.ID)
+	if err != nil && err != sql.ErrNoRows {
+		log.WithError(err).Fatalf("appservice %s worker unable to get latest transaction ID from DB",
+			ws.AppService.ID)
```
erikjohnston: Please you fields for things like this so that structured loggers can e.g. search for all logs for a given AS.

src/github.com/matrix-org/dendrite/appservice/workers/transaction_scheduler.go
```diff
@@ -0,0 +1,239 @@
...
+		return
+	}
+	ws.Cond.L.Lock()
+	*ws.EventsReady = eventCount
+	ws.Cond.L.Unlock()
+
+	// Loop forever and keep waiting for more events to send
+	for {
+		// Wait for more events if we've sent all the events in the database
+		if *ws.EventsReady <= 0 {
```
erikjohnston: This check should happen in the lock. I think things can spuriously wake up (they can in most other langs), so may people tend to do:

```golang
ws.Cond.L.Lock()
for *ws.EventsReady <= 0 {
    ws.Cond.Wait()
}
ws.Cond.L.Unlock()
```

src/github.com/matrix-org/dendrite/appservice/consumers/roomserver.go
```diff
@@ -165,13 +172,25 @@ func (s *OutputRoomEventConsumer) lookupStateEvents(
...
+				// Queue this event to be sent off to the application service
+				if err := s.asDB.StoreEvent(ctx, ws.AppService.ID, &event); err != nil {
+					log.WithError(err).Warn("failed to insert incoming event into appservices database")
+				} else {
+					// Tell our worker to send out new messages by updating remaining message
+					// count and waking them up with a broadcast
+					ws.Cond.L.Lock()
+					*ws.EventsReady++
+					ws.Cond.Broadcast()
+					ws.Cond.L.Unlock()
```
erikjohnston: I would pull this into `ApplicationServiceWorkerState` so that random parts of the code don't need to care about the details of the lock, e.g. implement a `NotifyNewEvent` and `WaitForNewEvents` methods

src/github.com/matrix-org/dendrite/appservice/storage/appservice_events_table.go
```diff
@@ -0,0 +1,256 @@
...
+func (s *eventsStatements) insertEvent(
+	ctx context.Context,
+	appServiceID string,
+	event *gomatrixserverlib.Event,
+) (err error) {
+	// If event has no content, strip the json
+	content := event.Content()
+	if string(content) == "{\"disable\":true}" {
+		content = []byte("{}")
+	}
```
erikjohnston: What on earth is this?

src/github.com/matrix-org/dendrite/appservice/storage/appservice_events_table.go
```diff
@@ -0,0 +1,256 @@
...
+		)
+		if err != nil {
+			fmt.Println("Failed:", err.Error())
+			return nil, 0, 0, err
+		}
+
+		// If txnID has changed on this event from the previous event, then we've
+		// reached the end of a transaction's events. Return only those events.
+		if lastTxnID > -2 && lastTxnID != txnID {
+			return events, maxID, lastTxnID, nil
```
erikjohnston: Instead of having this complicated logic for dealing with events that have already been assigned a transaction ID, I'd probably do it as two queries: the first would pull out all events with an assigned txn ID (without a limit), and then if there aren't any events then we can pull out a limited set of events that don't have a txn ID.

This is slightly less efficient but a single DB RTT is probably not going to cause an issue, and if it turns out it does there are a couple of nice ways of skipping the first step in most cases.

src/github.com/matrix-org/dendrite/appservice/consumers/roomserver.go
anoadragon453: It apparently does. And yeah, I agree the formatting here is a bit weird, will fix up.

src/github.com/matrix-org/dendrite/appservice/storage/appservice_events_table.go
anoadragon453: I have spent quite a few hours trying to reduce this into a single-query ordeal, including talking it over with @krombel. I agree a dual-query setup would be easier to understand, but do you think it is too crazy that we need to switch it back?

erikjohnston: My main concern is the fact that includes the shotgun of: https://github.com/matrix-org/dendrite/pull/477/files#diff-d84306218ea8217c1024f21eb34d9fcbR39, which feels a little scary

anoadragon453: Well, we could actually get rid of that if we remove the `LIMIT` in the query, instead only enforcing it in the `for` loop and opting not to enforce it once we've completely sent the last transaction.

src/github.com/matrix-org/dendrite/appservice/storage/txn_id_counter_table.go
anoadragon453: Is there any reason not to have it scoped to an AS though? I did so on purpose so that each AS would get an sequentially incrementing number for transaction ID even if other AS's are working.

src/github.com/matrix-org/dendrite/appservice/storage/appservice_events_table.go
anoadragon453: I have done the about in 801b58c9277511b1b5c26a908dd101f8026db555. Let me know what you think.

anoadragon453:
@erikjohnston @APwhitehat Do you know if any of this is relevant to/necessary for application services? https://github.com/matrix-org/dendrite/blob/b0f209b91fad7a565161a9ab1573472f1a85316d/src/github.com/matrix-org/dendrite/appservice/consumers/roomserver.go#L104-L114

src/github.com/matrix-org/dendrite/appservice/storage/appservice_events_table.go
```diff
@@ -109,59 +100,60 @@ func (s *eventsStatements) selectEventsByApplicationServiceID(
...
-func retrieveEvents(eventRows *sql.Rows, limit int) (events []gomatrixserverlib.ApplicationServiceEvent, maxID, txnID int, err error) {
+func retrieveEvents(eventRows *sql.Rows, limit int) (events []gomatrixserverlib.Event, maxID, txnID int, err error) {
 	// Get current time for use in calculating event age
 	nowMilli := time.Now().UnixNano() / int64(time.Millisecond)
 
 	// Iterate through each row and store event contents
 	// If txn_id changes dramatically, we've switched from collecting old events to
 	// new ones. Send back those events first.
-	lastTxnID := -2 // Invalid transaction ID
+	invalidTxnID := -2
```
krombel: this as well can be a const at the top of the file as it does not change and don't have to be recreated on every call.

anoadragon453: Good point, moved to a global.

src/github.com/matrix-org/dendrite/appservice/types/types.go
```diff
@@ -12,7 +12,46 @@
...
+
+// ApplicationServiceWorkerState is a type that couples an application service,
+// a lockable condition as well as some other state variables, allowing the
+// roomserver to notify appservice workers when there are events ready to send
+// externally to application services.
+type ApplicationServiceWorkerState struct {
+	AppService config.ApplicationService
+	Cond       *sync.Cond
+	// Events ready to be sent
+	EventsReady *int
```
erikjohnston: I would really just have this as a `bool` that gets set when there are more events, and unset when `WaitForEvents` returns, as if there is an off by one error (or whatever) anywhere things will get weird quickly. If you use a boolean then it will self correct.

erikjohnston:
Otherwise looks good!

Issue #471, opened 2018-05-29
AppServices: Implement Synapse-supported registration config options


anoadragon453:
Once matrix-org/matrix-doc#1259 is closed and these options are properly spec'd, they will need to be added in Dendrite.

Essentially, these are:

- [x] Ability to have a null `url` field
- [ ] Namespace `group_id` field (need to figure out how synapse handles this)
- [ ] Rate limited option (to be implemented once general rate limit support is implemented)
- [ ] Supported protocols option (used in 3PE lookup)

Edit: Options have been added to the config in #478, but their functionality is still not implemented.

kegsay:
Groups were subsequently removed from the spec. We generally do not like issues which are blanket "implement feature X" - we rely on Are We Synapse Yet? to expose areas which are unimplemented in an automated fashion.

PR #478, opened 2018-06-04
Add group_id, rate_limit and protocol AS config options

anoadragon453:
We currently just record and error check these options. Things like rate limiting are not yet implemented.

Working towards completion of #471.

### DIFF
--- a/src/github.com/matrix-org/dendrite/common/config/appservice.go
+++ b/src/github.com/matrix-org/dendrite/common/config/appservice.go
@@ -180,9 +180,10 @@ func checkErrors(config *Dendrite) (err error) {
 
 	// Check each application service for any config errors
 	for _, appservice := range config.Derived.ApplicationServices {
-		// Check that namespace(s) are valid regex
+		// Namespace-related checks
 		for key, namespaceSlice := range appservice.NamespaceMap {
 			for _, namespace := range namespaceSlice {
+				// Check that namespace(s) are valid regex
 				if !IsValidRegex(namespace.Regex) {
 					return configErrors([]string{fmt.Sprintf(
 						"Invalid regex string for Application Service %s", appservice.ID,
@@ -236,7 +237,7 @@ func checkErrors(config *Dendrite) (err error) {
 			}
 		}
 
-		// TODO: Remove once group_id is implemented
+		// TODO: Remove once rate_limited is implemented
 		if appservice.RateLimited {
 			log.Warn("WARNING: Application service option rate_limited is currently unimplemented")
 		}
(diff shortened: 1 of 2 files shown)


src/github.com/matrix-org/dendrite/common/config/appservice.go
```diff
@@ -51,6 +59,10 @@ type ApplicationService struct {
 	// Information about an application service's namespaces. Key is either
 	// "users", "aliases" or "rooms"
 	NamespaceMap map[string][]ApplicationServiceNamespace `yaml:"namespaces"`
+	// Whether rate limiting is applied to each application service user
+	RateLimited bool `yaml:"rate_limited"`
```
erikjohnston: I think this should default to true?

anoadragon453: In true go fashion this defaults to false. Not sure if we can check whether it's false because it hasn't been set, or because the user set it to false.

anoadragon453:
Warns loudly (enough?) now. Still not quite sure how to do the default rate limiting to true..

src/github.com/matrix-org/dendrite/common/config/appservice.go
```diff
@@ -157,12 +170,41 @@ func appendExclusiveNamespaceRegexs(
...
+		for key, namespaceSlice := range appservice.NamespaceMap {
+			for _, namespace := range namespaceSlice {
+				if !IsValidRegex(namespace.Regex) {
+					return configErrors([]string{fmt.Sprintf(
+						"Invalid regex string for Application Service %s", appservice.ID,
+					)})
+				}
+
+				// Check if GroupID for the users namespace is in the correct format
+				if key == "users" && namespace.GroupID != "" {
```
APwhitehat: would looping over `appservice.NamespaceMap["users"]` again and removing the if statement reduce cyclomatic complexity ?

I know gocyclo is a pain. Still cyclo of over 12 should be avoided.

anoadragon453: Seems to make it even worse, 15 > 14.

erikjohnston: Usually the thing to do is to move logic out into a separate function, e.g. pull out the inner loop contents into a `ValidatAppserviceConfig` func

erikjohnston:
> Still not quite sure how to do the default rate limiting to true..

So a way of doing this sort of thing is to note that when you call `Unmarshal` and give it a pointer to the output struct, you can prefill that struct with any defaults you want.

src/github.com/matrix-org/dendrite/common/config/appservice.go
anoadragon453: The `checkErrors` function essentially was `ValidateAppserviceConfig`. I've moved the namespace checks into a separate function and that's made gocyclo happy.

anoadragon453:
Both points fixed up.

PR #479, opened 2018-06-04
Prevent sql scanning into nil value in accounts_table

anoadragon453:
acc was a null pointer. Dereferencing this causes problems, but not any that the compiler picked up.

This fixes login again. Also adds in some logging so that this doesn't just return an error to the client and nothing in the log.

Oh SyTest, where are you :)

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/accounts_table.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/accounts_table.go
@@ -124,11 +124,11 @@ func (s *accountsStatements) selectPasswordHash(
 func (s *accountsStatements) selectAccountByLocalpart(
 	ctx context.Context, localpart string,
 ) (*authtypes.Account, error) {
-	var localpartPtr, appserviceIDPtr sql.NullString
+	var appserviceIDPtr sql.NullString
 	var acc authtypes.Account
 
 	stmt := s.selectAccountByLocalpartStmt
-	err := stmt.QueryRowContext(ctx, localpart).Scan(&localpartPtr, &appserviceIDPtr)
+	err := stmt.QueryRowContext(ctx, localpart).Scan(&acc.Localpart, &appserviceIDPtr)
 	if err != nil {
 		switch err {
 		case sql.ErrNoRows:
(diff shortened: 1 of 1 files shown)


src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/accounts_table.go
```diff
@@ -121,14 +123,27 @@ func (s *accountsStatements) selectPasswordHash(
...
+
 	stmt := s.selectAccountByLocalpartStmt
-	err = stmt.QueryRowContext(ctx, localpart).Scan(&acc.Localpart, &acc.AppServiceID)
-	if err == nil {
-		acc.UserID = userutil.MakeUserID(localpart, s.serverName)
-		acc.ServerName = s.serverName
+	err := stmt.QueryRowContext(ctx, localpart).Scan(&localpartPtr, &appserviceIDPtr)
+	if err != nil {
+		log.WithError(err).Error("Unable to retrieve user from the db")
+		return nil, err
```
erikjohnston: We probably don't want to log if the error is `sql.ErrNoRows`

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/accounts_table.go
```diff
@@ -121,14 +123,27 @@ func (s *accountsStatements) selectPasswordHash(
...
 	}
-	return
+	if appserviceIDPtr.Valid {
+		acc.AppServiceID = appserviceIDPtr.String
+	}
+	if localpartPtr.Valid {
+		acc.Localpart = localpartPtr.String
```
erikjohnston: `localpart` is a `NOT NULL` comlumn, so we don't need to do this check

anoadragon453:
Fixed

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/accounts_table.go
```diff
@@ -121,14 +123,28 @@ func (s *accountsStatements) selectPasswordHash(
...
-	if err == nil {
-		acc.UserID = userutil.MakeUserID(localpart, s.serverName)
-		acc.ServerName = s.serverName
+	err := stmt.QueryRowContext(ctx, localpart).Scan(&acc.Localpart, &appserviceIDPtr)
+	if err != nil {
+		switch err {
+		case sql.ErrNoRows:
+		default:
+			log.WithError(err).Error("Unable to retrieve user from the db")
+		}
```
APwhitehat: is this simpler ?

```go
if err != sql.ErrNoRows {
    log.WithError(err).Error("Unable to retrieve user from the db")
}
```

anoadragon453: Indeed, thanks!

anoadragon453:
This issue still prevents login on Dendrite. Is this PR good to go @erikjohnston ?

anoadragon453:
Ta!

PR #480, opened 2018-06-04
 Prevent alias registration in AS exclusive namespace 

### LINKED ISSUE #451: AppServices: Ensure users cannot create a room alias that an Application Service manages
https://github.com/matrix-org/dendrite/blob/6fed3a1be2bfce46bd446198e762ef3a0072818f/src/github.com/matrix-org/dendrite/clientapi/routing/createroom.go#L116-L125

It would be bad if a user could reserve an alias that would otherwise be used for, say, a very popular freenode room. Thus when creating a room the homeserver should first check if that room is already managed by an app service's alias namespace (exclusive or not).

The only problem I'm not sure about is if someone creates a room, and *afterwards* the app service is registered. At that point I think you might just be out of luck :|

anoadragon453:
Fixes #451, includes fix from #479.

Users should not be allowed to create a room alias in an application service's exclusive namespace. We now return an `M_EXCLUSIVE` if they attempt to do so.

Also, according to the spec, `M_EXCLUSIVE` should return a status `400`, not a `403`.

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/routing/directory.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/routing/directory.go
@@ -113,6 +113,23 @@ func SetLocalAlias(
 		}
 	}
 
+	// Check that the alias does not fall within an exclusive namespace of an
+	// application service
+	for _, appservice := range cfg.Derived.ApplicationServices {
+		for key, namespaceMaps := range appservice.NamespaceMap {
+			if key == "aliases" {
+				for _, namespace := range namespaceMaps {
+					if namespace.Exclusive && namespace.RegexpObject.MatchString(alias) {
+						return util.JSONResponse{
+							Code: http.StatusBadRequest,
+							JSON: jsonerror.ASExclusive("Alias is reserved by an application service"),
+						}
+					}
+				}
+			}
+		}
+	}
+
 	var r struct {
 		RoomID string `json:"room_id"`
 	}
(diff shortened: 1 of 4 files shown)


erikjohnston:
> Also, according to the spec, M_EXCLUSIVE should return a status 400, not a 403.

Is this what synapse does?

anoadragon453:
image

Indeed it is.

src/github.com/matrix-org/dendrite/clientapi/routing/directory.go
```diff
@@ -113,6 +113,23 @@ func SetLocalAlias(
 		}
 	}
 
+	// Check that the alias does not fall within an exclusive namespace of an
+	// application service
+	for _, appservice := range cfg.Derived.ApplicationServices {
+		for key, namespaceMaps := range appservice.NamespaceMap {
+			if key == "aliases" {
```
APwhitehat: why are you looping over a map to find a key ?
```go
namespaces, ok := appservice.NamespaceMap["aliases"]
```
Then check `if ok`

anoadragon453: Change, thanks. I think I'll be doing a separate PR to be removing the whole NamespaceMap->[]Namespace relationship, as app services aren't ever, for instance, going to have more than one "users" namespace.

anoadragon453:
Merging based on prior approval.

PR #486, opened 2018-06-06
[federation] Implement state APIs

### LINKED ISSUE #485: [federation] Implement State API
- [ ] GET /_matrix/federation/v1/state_ids/{roomID}/{eventID}
- [ ] GET /_matrix/federation/v1/state/{roomID}/{eventID}

response formats: https://github.com/matrix-org/dendrite/blob/master/vendor/src/github.com/matrix-org/gomatrixserverlib/federationtypes.go#L24

APwhitehat:
fixes #485.

Other TODOs (will create gh issues to track):  Figure out how & when to use keyring to verify JSON requests.
`event_auth` can be implemented via the state apis, but is it too much of extra work ?

### DIFF
--- a/src/github.com/matrix-org/dendrite/federationapi/routing/events.go
+++ b/src/github.com/matrix-org/dendrite/federationapi/routing/events.go
@@ -43,8 +43,8 @@ func GetEvent(
 	return util.JSONResponse{Code: http.StatusOK, JSON: event}
 }
 
-// getEvent returns the requested event.
-// Or returns an error response which can be sent to the client.
+// getEvent returns the requested event,
+// otherwise it returns an error response which can be sent to the client.
 func getEvent(
 	ctx context.Context,
 	request *gomatrixserverlib.FederationRequest,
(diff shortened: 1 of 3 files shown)


src/github.com/matrix-org/dendrite/federationapi/routing/events.go
```diff
@@ -35,6 +35,22 @@ func GetEvent(
...
+	event, err := getEvent(ctx, request, query, eventID)
+	if err != nil {
+		return *err
+	}
+
+	return util.JSONResponse{Code: http.StatusOK, JSON: event}
+}
+
+// getEvent returns the requested event.
+// Or returns an error response which can be sent to the client.
```
anoadragon453: "returns the requested event, otherwise it returns"

src/github.com/matrix-org/dendrite/federationapi/routing/events.go
```diff
@@ -45,11 +61,13 @@ func GetEvent(
 		&authResponse,
 	)
 	if err != nil {
-		return util.ErrorResponse(err)
+		resErr := util.ErrorResponse(err)
```
anoadragon453: These could also just be

```
return nil, &util.ErrorResponse(err)
```

APwhitehat: I couldn't believe it either. That is not allowed in go.

erikjohnston:
At some point we should make it so that `/state_ids/` don't pull the actual events from the room server, but that is an optimisation that can happen later.

PR #494, opened 2018-06-11
Implement GET /rooms/{roomAlias}

### LINKED ISSUE #452: AppServices: GET /rooms/{roomAlias}
**Spec:** https://matrix.org/docs/spec/application_service/unstable.html#get-rooms-roomalias

When we receive a user's request to join a room that is on our homeserver (i.e. `#coolroom:myserver.org`), but we don't have a record of it in the database, it likely means that it belongs to an application service. In this case, the flow is then:

* Figure out which app service handles this room alias (what to do if more than one here? [Link](https://github.com/matrix-org/matrix-doc/issues/1253)).
* Send a `GET` request to `/rooms/{roomAlias}`.
* Wait for AS to create corresponding room and returns `200 OK`.
* Join the user to the new room.

Some error handling of `401`, `403`, and `404` errors are necessary as per the spec link.

With the current architecture (having a separate client-server and appservice component), we'll need to create a new internal http api on the appservice component so that client-server can command it to ask app services whether the room exists or not.

anoadragon453:
Closes #452 

This PR implements a method for querying /rooms/{roomAlias} on an appservice, as defined in [the spec](https://matrix.org/docs/spec/application_service/unstable.html#get-rooms-roomalias).

This PR depends on #541

### DIFF
--- a/src/github.com/matrix-org/dendrite/appservice/api/query.go
+++ b/src/github.com/matrix-org/dendrite/appservice/api/query.go
@@ -18,15 +18,40 @@
 package api
 
 import (
+	"context"
 	"net/http"
+
+	commonHTTP "github.com/matrix-org/dendrite/common/http"
+	opentracing "github.com/opentracing/opentracing-go"
 )
+
+// RoomAliasExistsRequest is a request to an application service
+// about whether a room alias exists
+type RoomAliasExistsRequest struct {
+	// Alias we want to lookup
+	Alias string `json:"alias"`
+}
+
+// RoomAliasExistsResponse is a response from an application service
+// about whether a room alias exists
+type RoomAliasExistsResponse struct {
+	AliasExists bool `json:"exists"`
+}
 
 // AppServiceQueryAPI is used to query user and room alias data from application
 // services
 type AppServiceQueryAPI interface {
-	// TODO: Check whether a room alias exists within any application service namespaces
+	// Check whether a room alias exists within any application service namespaces
+	RoomAliasExists(
+		ctx context.Context,
+		req *RoomAliasExistsRequest,
+		response *RoomAliasExistsResponse,
+	) error
 	// TODO: QueryUserIDExists
 }
+
+// AppServiceRoomAliasExistsPath is the HTTP path for the RoomAliasExists API
+const AppServiceRoomAliasExistsPath = "/api/appservice/RoomAliasExists"
 
 // httpAppServiceQueryAPI contains the URL to an appservice query API and a
 // reference to a httpClient used to reach it
@@ -46,4 +71,17 @@ func NewAppServiceQueryAPIHTTP(
 		httpClient = http.DefaultClient
 	}
 	return &httpAppServiceQueryAPI{appserviceURL, httpClient}
+}
+
+// RoomAliasExists implements AppServiceQueryAPI
+func (h *httpAppServiceQueryAPI) RoomAliasExists(
+	ctx context.Context,
+	request *RoomAliasExistsRequest,
+	response *RoomAliasExistsResponse,
+) error {
+	span, ctx := opentracing.StartSpanFromContext(ctx, "appserviceRoomAliasExists")
+	defer span.Finish()
+
+	apiURL := h.appserviceURL + AppServiceRoomAliasExistsPath
+	return commonHTTP.PostJSON(ctx, span, h.httpClient, apiURL, request, response)
 }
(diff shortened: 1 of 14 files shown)


anoadragon453:
This will need to be rebased on master once #477 is merged.

anoadragon453:
We also need to add AS query call to the /join event, as some clients (nheko) do not call `/directory/room` first.

anoadragon453:
This PR has now been partially separated into #541.

src/github.com/matrix-org/dendrite/appservice/api/query.go
```diff
@@ -0,0 +1,95 @@
...
+// RoomAliasExistsRequest is a request to an application service
+// about whether a room alias exists
+type RoomAliasExistsRequest struct {
+	// Alias we want to lookup
+	Alias string `json:"alias"`
+}
+
+// RoomAliasExistsRequestAccessToken is a request to an application service
+// about whether a room alias exists. Includes an access token
+type RoomAliasExistsRequestAccessToken struct {
```
erikjohnston: Is this used anywhere?

src/github.com/matrix-org/dendrite/appservice/query/query.go
```diff
@@ -0,0 +1,136 @@
...
+			}
+			resp, err := a.HTTPClient.Do(req.WithContext(ctx))
+			if resp != nil {
+				defer func() {
+					err = resp.Body.Close()
+					if err != nil {
+						log.WithFields(log.Fields{
+							"appservice_id": appservice.ID,
+							"status_code":   resp.StatusCode,
+						}).Error("Unable to close application service response body")
```
erikjohnston: You should add a `.WithError(err)`

src/github.com/matrix-org/dendrite/appservice/query/query.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// makeHTTPClient creates an HTTP client with certain options that will be used for all query requests to application services
+func makeHTTPClient() *http.Client {
+	return &http.Client{
+		Timeout: time.Second * 30,
+		// TODO: Verify certificates
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: true, // nolint: gas
```
erikjohnston: I'm not sure we should turn this off. If the appservice doesn't have valid certs they should be using plain http anyway.

src/github.com/matrix-org/dendrite/appservice/query/query.go
```diff
@@ -0,0 +1,136 @@
...
+			URL.Path += request.Alias
+			apiURL := URL.String() + "?access_token=" + appservice.HSToken
+
+			// Send a request to each application service. If one responds that it has
+			// created the room, immediately return.
+			req, err := http.NewRequest(http.MethodGet, apiURL, nil)
+			if err != nil {
+				return err
+			}
+			resp, err := a.HTTPClient.Do(req.WithContext(ctx))
```
erikjohnston: I'd separate out the `req.WithContext(ctx)` into a separate line. Combining them is a bit confusing

src/github.com/matrix-org/dendrite/appservice/query/query.go
```diff
@@ -0,0 +1,136 @@
...
+				// StatusOK received from appservice. Room exists
+				response.AliasExists = true
+				return nil
+			}
+
+			// Log non OK
+			log.WithFields(log.Fields{
+				"appservice_id": appservice.ID,
+				"status_code":   resp.StatusCode,
+			}).Warn("Application service responded with non-OK status code")
```
erikjohnston: This feels odd. Isn't there a way for the remote AS to respond with no to the alias request without erroring?

src/github.com/matrix-org/dendrite/appservice/query/query.go
anoadragon453: Good point, yeah.

src/github.com/matrix-org/dendrite/appservice/query/query.go
anoadragon453: They respond with `404`. I could create an error case just for this?

erikjohnston: Yeah, I'd check for a 404 and otherwise error, to save WARNing unnecessarily for expected failures

anoadragon453:
@erikjohnston Fixed the warning

PR #496, opened 2018-06-12
Don't require preset to create room

anoadragon453:
**Spec:** https://matrix.org/docs/spec/client_server/r0.3.0.html#post-matrix-client-r0-createroom

The `preset` field is not required for `/createRoom`. It is only a convenience.

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/routing/createroom.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/routing/createroom.go
@@ -169,7 +169,7 @@ func createRoom(req *http.Request, device *authtypes.Device,
 
 	var joinRules, historyVisibility string
 	switch r.Preset {
-	case presetPrivateChat, "":
+	case presetPrivateChat:
 		joinRules = joinRuleInvite
 		historyVisibility = historyVisibilityShared
 	case presetTrustedPrivateChat:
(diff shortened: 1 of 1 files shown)


APwhitehat:
Since none of the request parameter are marked *Required*, it may be case that all are required.
W.r.t. this PR, I think you'll need to add a Default for `switch r.Preset` on line 174.

anoadragon453:
The bridges I've come across all create rooms without a preset, so I'm *pretty sure* it's not required, especially since the spec refers to it as a convenience parameter just to set other settings.

> W.r.t. this PR, I think you'll need to add a Default for switch r.Preset on line 174.

Yes, good point.

src/github.com/matrix-org/dendrite/clientapi/routing/createroom.go
```diff
@@ -87,13 +87,15 @@ func (r createRoomRequest) Validate() *util.JSONResponse {
 			}
 		}
 	}
-	switch r.Preset {
-	case presetPrivateChat, presetTrustedPrivateChat, presetPublicChat:
```
krombel: Why not just add `case "": break`?
Would make the diff shorter :smirk:

anoadragon453: Oh, duh. And go doesn't even need `break`. Super short!

anoadragon453:
Fixed up, thanks!

src/github.com/matrix-org/dendrite/clientapi/routing/createroom.go
```diff
@@ -170,7 +169,7 @@ func createRoom(req *http.Request, device *authtypes.Device,
 
 	var joinRules, historyVisibility string
 	switch r.Preset {
-	case presetPrivateChat:
+	case presetPrivateChat, "":
```
erikjohnston: I would probably have a `default` case to handle none specified, with a comment that we've already applied checks to ensure that we've checked valid values previously. As it is its non-obvious that `joinRules` and `historyVisibility` are always set to valid values

PR #498, opened 2018-06-12
[federation] Add user_devices query

APwhitehat:
Skipping `stream_id` since we are not using it in Dendrite.

### DIFF
--- a/src/github.com/matrix-org/dendrite/federationapi/routing/devices.go
+++ b/src/github.com/matrix-org/dendrite/federationapi/routing/devices.go
@@ -1,0 +1,53 @@
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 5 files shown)


src/github.com/matrix-org/dendrite/federationapi/routing/devices.go
```diff
@@ -0,0 +1,53 @@
...
+func GetUserDevices(
+	req *http.Request,
+	deviceDB *devices.Database,
+	userID string,
+) util.JSONResponse {
+	localpart, err := userutil.ParseUsernameParam(userID, nil)
+	if err != nil {
+		return util.JSONResponse{
+			Code: http.StatusBadRequest,
+			JSON: jsonerror.InvalidArgumentValue("Invalid userID"),
```
anoadragon453: I think we usually write "user ID" in external responses.

APwhitehat: Force of habit. Thanks for pointing out.

PR #502, opened 2018-06-13
Check for masqueraded user before real user on login

anoadragon453:
We need to first check if the requested user is an application service fake user, before trying to pull the userID from the database. Trying to check from the database requires a device for the user to exist, which it of course doesn't yet.

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
@@ -65,14 +65,33 @@ type Data struct {
 func VerifyUserFromRequest(
 	req *http.Request, data Data,
 ) (*authtypes.Device, *util.JSONResponse) {
-	// Try to find local user from device database
-	dev, devErr := verifyAccessToken(req, data.DeviceDB)
-	if devErr == nil {
+	// Check to see if this is an application service user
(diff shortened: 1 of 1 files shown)


src/github.com/matrix-org/dendrite/clientapi/auth/auth.go
```diff
@@ -172,20 +192,26 @@ func GenerateAccessToken() (string, error) {
 	return base64.RawURLEncoding.EncodeToString(b), nil
 }
 
-// extractAccessToken from a request, or return an error detailing what went wrong. The
-// error message MUST be human-readable and comprehensible to the client.
-func extractAccessToken(req *http.Request) (string, error) {
-	// cf https://github.com/matrix-org/synapse/blob/v0.19.2/synapse/api/auth.py#L631
+// extractApplicationServiceAccessToken retreives an access token from a request
```
anoadragon453: I separated out the `extractAccessToken` function into two so that we wouldn't need to do unnecessary checks depending on whether a request was originating from an application service or not. We also maintain the "mixing query and Header" error message by only checking for that edge case in `extractApplicationServiceAccessToken` which is always called first.

babolivier:
I just renamed the PR so I don't forget what it's about when I'll have a look at it.

Kegsay:

PR #508, opened 2018-06-15
Switch to golangci-lint

anoadragon453:
As gometalinter has been discontinued, I thought it best to switch to the next popular linter, which boasts speed gains of up to 5-7x. It also includes a few more linters if we ever want to enable them.

This PR switches travis to the new linter, adds a new `./scripts/find-lint-ci.sh` script specifically for CI, whereas the user can continue to use `./scripts/find-lint.sh`. A few linting issues were also fixed to ensure the build doesn't fail.


krombel:
You need to remove this folder(s) in vendor as well

anoadragon453:
@krombel Good point, thanks!

erikjohnston:
> As gometalinter has been discontinued

Where has that been announced? There is still activity on the GH rep

anoadragon453:
Oh, so there is. I heard it had been discontinued a while ago on hackernews, however looking now it seems activity has picked up again in the last month.

This linter is still much faster though, so I'd still recommend it for that reason.

APwhitehat:
I notice that safesql is absent.
Are linter.json & linter-fast.json still being used ?

anoadragon453:
No they aren't, good point. For fast minting you can just add `--fast` to golangci-lint. I should probably add that to CONTRIBUTING and any docs that already mention linting.

Cnly:
@anoadragon453 Is this already handled in #713?

anoadragon453:
Indeed :) Forgot about this PR completely. Thanks for the reminder!

PR #510, opened 2018-06-16
Use userutil to parse localpart

APwhitehat:
Replace SplitID with userutil wherever it's more apt.

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/routing/account_data.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/routing/account_data.go
@@ -23,7 +23,7 @@ import (
 	"github.com/matrix-org/dendrite/clientapi/httputil"
 	"github.com/matrix-org/dendrite/clientapi/jsonerror"
 	"github.com/matrix-org/dendrite/clientapi/producers"
-	"github.com/matrix-org/gomatrixserverlib"
+	"github.com/matrix-org/dendrite/clientapi/userutil"
 
 	"github.com/matrix-org/util"
 )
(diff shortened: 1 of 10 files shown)


anoadragon453:
So ParseUsernameParam just uses SplitID internally, so this seems to just add more overhead. Surely we should just remove ParseUsernameParam, unless it has better error reporting?

APwhitehat:
ParseUsernameParam handles cases where argument is just the localpart.
Also it looks kind of clearer to me :P.
Overhead should be small enough to not be noticeable.

anoadragon453:
> ParseUsernameParam handles cases where argument is just the localpart.

Wait but then why even run the method, you already have the localpart? :P 

> Overhead should be small enough to not be noticeable.

Yeah, I do think we should try to use gomatrixserverlib functions wherever possible. Maybe we should move `ParseUsernameParam` to it ¯\\\_(ツ)\_/¯

APwhitehat:
> Wait but then why even run the method, you already have the localpart? :P

Sometimes we don't know if client has sent localpart or the full userID, hence it becomes problematic.
eg.
https://github.com/matrix-org/dendrite/blob/a76e31f1b3ff7035b53e755c07058ff8c5f91177/src/github.com/matrix-org/dendrite/clientapi/routing/login.go#L85

> Yeah, I do think we should try to use gomatrixserverlib functions wherever possible. Maybe we should move ParseUsernameParam to it ¯\_(ツ)_/¯

definitely

anoadragon453:
Under login on the C-S spec:

> The fully qualified user ID or just local part of the user ID, to log in.

So yes you're right, we should account for both. Still +1 for adding pretty much all userutil methods to gomatrixserverlib.

erikjohnston:
I'd be a bit cautious about not checking that something is a valid userID in all these cases, some of the APIs the spec requires fully qualified IDs for. While it may be safe to make this change, I'm not sure if we want dendrite to be so lax compared with the spec.

APwhitehat:
It's only safe to assume user id can be just the localpart for /login.
We may be better off having stricter checks albeit a bit ugly.

Issue #488, opened 2018-06-08
[federation] Verify JSON using keyring


APwhitehat:
Figure out how & when to use keyring to verify JSON requests.

Since this looks to be just left in a hurry [ref](https://github.com/matrix-org/dendrite/blob/master/src/github.com/matrix-org/dendrite/federationapi/routing/events.go#L29)

APwhitehat:
turns out we already do this.

PR #511, opened 2018-06-16
Remove redundant federation request verification

### LINKED ISSUE #509: [cleanup] remove unwanted signature verification
We use `common.MakeFedAPI` which verifies the request JSON.

examples:

https://github.com/matrix-org/dendrite/blob/78440083dfc8f522e55f11f0fce0f3ea4a103f0b/src/github.com/matrix-org/dendrite/federationapi/routing/invite.go#L74-L82

https://github.com/matrix-org/dendrite/blob/78440083dfc8f522e55f11f0fce0f3ea4a103f0b/src/github.com/matrix-org/dendrite/federationapi/routing/join.go#L141-L149

APwhitehat:
Fixes #509 
Since makeFedAPI checks signatures, we don't need to do that again.
No need to pass the keyring.

### DIFF
--- a/src/github.com/matrix-org/dendrite/federationapi/routing/events.go
+++ b/src/github.com/matrix-org/dendrite/federationapi/routing/events.go
@@ -17,9 +17,7 @@ package routing
 import (
 	"context"
 	"net/http"
-	"time"
 
-	"github.com/matrix-org/dendrite/common/config"
 	"github.com/matrix-org/dendrite/roomserver/api"
 	"github.com/matrix-org/gomatrixserverlib"
 	"github.com/matrix-org/util"
@@ -29,10 +27,7 @@ import (
 func GetEvent(
 	ctx context.Context,
 	request *gomatrixserverlib.FederationRequest,
-	_ config.Dendrite,
 	query api.RoomserverQueryAPI,
-	_ time.Time,
-	_ gomatrixserverlib.KeyRing,
 	eventID string,
 ) util.JSONResponse {
 	var authResponse api.QueryServerAllowedToSeeEventResponse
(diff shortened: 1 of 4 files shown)


src/github.com/matrix-org/dendrite/federationapi/routing/events.go
```diff
@@ -29,10 +27,7 @@ import (
 func GetEvent(
 	ctx context.Context,
 	request *gomatrixserverlib.FederationRequest,
-	_ config.Dendrite,
```
anoadragon453: Can anyone tell me why it was useful to have a function take in parameters that were never used?

APwhitehat: These were expected to be used to verify http request (around August 17).
But then in October or so everything was ported to common.MakeFedAPI which did the verification.
As a result some these parameters became redundant.

erikjohnston:
So the reason for checking the signatures of the events are not so much to ensure that they come from the right origin (you're correct that that is protected by the request level signing), but to ensure that the signatures in the event are actually valid. It would be quite bad for dendrite to serve up events where the signatures are wrong.

APwhitehat:
What I notice is that VerifyHTTPRequest does the exact same thing.
https://github.com/matrix-org/dendrite/blob/241b1b5ace677d81207299e52e2f6ac5f3f547b1/vendor/src/github.com/matrix-org/gomatrixserverlib/request.go#L218-L222

> .. but to ensure that the signatures in the event are actually valid

What more is required ?

erikjohnston:
The difference is:
- VerifyHTTPRequest is a signature over the entire request and isn't stored
- the event signatures only protect the events and are stored in the DB

The reason to do the second check is to ensure that other servers accept the event if we send it to them (e.g. when they ask for the current state of the room from us). If the second check fails but we accept the event anyway then we risk causing confusion about whether the event is valid or not between different servers.

APwhitehat:
Oh I see. Thanks.
I think #488 is required w.r.t. `get_event`

PR #438, opened 2018-05-09
State events filtering database api

CromFr:
This PR adds a `gomatrixserverlib.Filter` parameter to functions handling the `syncapi_current_room_state` table. It does not implement any filtering logic inside the syncapi IncrementalSync/CompleteSync functions, just the APIs for future use.

Default filters are provided as placeholders in IncrementalSync/CompleteSync, so behaviour should be unchanged (except the default 20 event limit)

SQL table will be changed. You can upgrade an existing database using:
```sql
ALTER TABLE syncapi_current_room_state  ADD COLUMN IF NOT EXISTS sender text;
UPDATE syncapi_current_room_state SET sender=(event_json::json->>'sender');
ALTER TABLE syncapi_current_room_state ALTER COLUMN sender SET NOT NULL;
ALTER TABLE syncapi_current_room_state  ADD COLUMN IF NOT EXISTS contains_url bool;
UPDATE syncapi_current_room_state SET contains_url=(event_json::json->>'content')::json->>'url' IS NOT NULL;
ALTER TABLE syncapi_current_room_state ALTER COLUMN contains_url SET NOT NULL;
```

Note: This depends on #436 (and includes all its commits). I'm not sure if Github will remove the duplicated commits once #436 is merged.

PR #436, opened 2018-04-12
Store & retrieve filters as structs rather than []byte

CromFr:
Manipulate filters as gomatrix.Filter structures, instead of their []byte JSON representation.
This lays ground work for using filters in dendrite for /sync requests.

Note: this PR is taken from #388 , with some additional comment & improvements (fixed the canonical JSON storage)

PR #513, opened 2018-06-18
Account data events filtering database API

CromFr:
Very similar to #438, adds a gomatrixserverlib.Filter parameter to functions handling the syncapi_account_data_type table. It does not implement any filtering logic inside the syncapi IncrementalSync/CompleteSync functions, just the APIs for future use.

Default filters are provided as placeholders in IncrementalSync/CompleteSync, so behaviour should be unchanged (except the default 20 event limit)

SQL table will be changed. You can upgrade an existing database using:
```sql
ALTER TABLE syncapi_account_data_type ADD COLUMN IF NOT EXISTS sender text;
UPDATE syncapi_account_data_type SET sender=user_id; -- Not accurate, probably best guess
ALTER TABLE syncapi_account_data_type ALTER COLUMN sender SET NOT NULL;
```

Note: This depends on #436


Cnly:
Given #436, #438 have both been merged, this PR doesn't seem a big one. I'll also try to merge master onto it and work on merging it.

PR #518, opened 2018-06-26
[federation] Fix state API endpoints

APwhitehat:
event_id should be given as parameter.

sytest is still failing. I reckon it's a URI enconding issues with federation Client of gomatrixserverlib.

### DIFF
--- a/src/github.com/matrix-org/dendrite/federationapi/routing/state.go
+++ b/src/github.com/matrix-org/dendrite/federationapi/routing/state.go
@@ -35,7 +35,7 @@ func GetState(
 	_ gomatrixserverlib.KeyRing,
 	roomID string,
 ) util.JSONResponse {
-	eventID, err := parseParams(request)
+	eventID, err := parseEventIDParam(request)
 	if err != nil {
 		return *err
 	}
@@ -58,7 +58,7 @@ func GetStateIDs(
 	_ gomatrixserverlib.KeyRing,
 	roomID string,
 ) util.JSONResponse {
-	eventID, err := parseParams(request)
+	eventID, err := parseEventIDParam(request)
 	if err != nil {
 		return *err
 	}
@@ -78,7 +78,7 @@ func GetStateIDs(
 	}
 }
 
-func parseParams(
+func parseEventIDParam(
 	request *gomatrixserverlib.FederationRequest,
 ) (eventID string, resErr *util.JSONResponse) {
 	URL, err := url.Parse(request.RequestURI())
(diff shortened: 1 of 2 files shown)


anoadragon453:
Can you not print out what is getting sent and see where it differs from the proper URL encoding?

src/github.com/matrix-org/dendrite/federationapi/routing/state.go
```diff
@@ -68,6 +78,26 @@ func GetStateIDs(
 	}
 }
 
+func parseParams(
```
anoadragon453: I might rename this to `parseEventIDParam` or something more specific if it's just designed to pull out `"event_id"`.

APwhitehat: Good point.

anoadragon453:
What's the current state of this? Do you want to merge even though sytest is still failing or continue iterating?

APwhitehat:
This is ready for review. We'll track the failing tests in #519.

src/github.com/matrix-org/dendrite/federationapi/routing/state.go
```diff
@@ -68,6 +78,26 @@ func GetStateIDs(
 	}
 }
 
+func parseEventIDParam(
+	request *gomatrixserverlib.FederationRequest,
+) (eventID string, resErr *util.JSONResponse) {
+	URL, err := url.Parse(request.RequestURI())
```
anoadragon453: Was there a reason you parsed the URL first here instead of simply doing:

```
eventID = req.URL.Query().Get("event_id")
```

Like we do elsewhere? https://github.com/matrix-org/dendrite/blob/b86b55360a6b46b5edc0f6d049fd997f46a9b074/src/github.com/matrix-org/dendrite/clientapi/auth/auth.go#L159

anoadragon453:
Looks good otherwise.

src/github.com/matrix-org/dendrite/federationapi/routing/state.go
APwhitehat: Intention of `request *gomatrixserverlib.FederationRequest` is to be a signature verified version of raw federation `http.Request`.
https://github.com/matrix-org/gomatrixserverlib/blob/4c45af876608399b13cf891e2d9f89f01c67137a/request.go#L183-L194

Although it may be argued that after verification, we are allowed to use `http.Request`'s parameters, it's safer to just use `request *gomatrixserverlib.FederationRequest`.

anoadragon453: Fair enough.

APwhitehat:
Deleted this while doing local cleanup :P

PR #522, opened 2018-06-27
Make use of /users/{userID} in relavent APIs

anoadragon453:
Have C-S and S-S endpoints query application services for users within their namespace.

Depends on and is the "hookup" portion of #521.


erikjohnston:
This feels slightly wrong, in that we're asking the AS component for profile information?! But OTOH its probably going more faff than its worth right now to make it cleaner somehow

PR #524, opened 2018-06-27
Return 404 when profile is not found

### LINKED ISSUE #520: [clientapi] 404 on profile not found
spec: https://matrix.org/docs/spec/client_server/r0.3.0.html#get-matrix-client-r0-profile-userid

We are encountering sql.NoRows which returns internal server error

APwhitehat:
Fixes #520

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/routing/profile.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/routing/profile.go
@@ -41,15 +41,12 @@ func GetProfile(
 			JSON: jsonerror.NotFound("Bad method"),
 		}
 	}
-	localpart, _, err := gomatrixserverlib.SplitID('@', userID)
+
+	profile, err := getProfileByUserID(req, accountDB, userID)
 	if err != nil {
-		return httputil.LogThenError(req, err)
+		return *err
 	}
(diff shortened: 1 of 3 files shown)


src/github.com/matrix-org/dendrite/clientapi/routing/profile.go
```diff
@@ -60,19 +57,40 @@ func GetProfile(
...
+			Code: http.StatusNotFound,
+			JSON: jsonerror.NotFound("no profile information for this user or this user does not exist"),
+		}
+	}
+
+	return profile, nil
+}
+
+// GetAvatarURL implements GET /profile/{userID}/avatar_url
+func GetAvatarURL(
```
anoadragon453: Should this method (and others) not also check for a `nil` profile?

APwhitehat: getProfileByUserID does that, part of the reason why I made this common.

anoadragon453: Ah I see. Misread the diff, thought you were still doing the `nil` check in `GetAvatarURL`.

Issue #519, opened 2018-06-27
Make sytest's federation tests pass 


APwhitehat:
- [x] 00prepare.pl
- [x] 01keys.pl (notary server not implemented)
- [x] 10query-profile.pl (remote profile query not implemented)
- [x] 11query-directory.pl
- [x] 30room-join.pl
- [x] 31room-send.pl (we don't support /events, deprecated)
- [ ] 32room-getevent.pl
- [ ] 33room-get-missing-events.pl
- [ ] 34room-backfill.pl
- [ ] 35room-invite.pl
- [ ] 36-state.pl
- [ ] 37public-rooms.pl
- [ ] 40devicelists.pl
- [ ] 41power-levels.pl

At least the ones which are supposed to be implemented by dendrite

Kegsay:
I don't mind having a "are-we-synapse-ready-yet?" issue to track parity with Synapse, but I do object to having lots of issues for this. Please keep issues Specific, Measurable, Achievable/Actionable, Realistic, and Time-Bounded.

PR #525, opened 2018-06-28
[WIP] Improve /query/directory endpoint

anoadragon453:
WIP for fixing #519 

Attempts to make `11query-directory.pl` pass.

All that's left is returning the other servers that are aware of the given room alias, as per this comment:

https://github.com/matrix-org/dendrite/blob/7f2db95a06f31a96bae8f9f1a072861fe8d5545f/src/github.com/matrix-org/dendrite/federationapi/routing/query.go#L62-L66

### DIFF
--- a/src/github.com/matrix-org/dendrite/federationapi/routing/routing.go
+++ b/src/github.com/matrix-org/dendrite/federationapi/routing/routing.go
@@ -130,7 +130,7 @@ func Setup(
 		},
 	)).Methods(http.MethodGet)
 
-	v1fedmux.Handle("/query/directory/", common.MakeFedAPI(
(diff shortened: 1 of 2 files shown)


src/github.com/matrix-org/dendrite/federationapi/routing/routing.go
```diff
@@ -130,7 +130,7 @@ func Setup(
 		},
 	)).Methods(http.MethodGet)
 
-	v1fedmux.Handle("/query/directory/", common.MakeFedAPI(
```
APwhitehat: me wonders if this was intentional

anoadragon453: Yes, it would 404 otherwise :P

PR #527, opened 2018-06-29
Run TravisCI with Go 1.10

anoadragon453:
Go 1.10 is out. We should test with it.

Only downside is this makes CI jobs a couple minutes longer.


APwhitehat:
we should consider removing older version, i.e. 1.8

anoadragon453:
@APwhitehat We still support Go 1.8+ though, unless we want to bump the oldest supported version.

PR #528, opened 2018-06-29
Allow multiple regexes per as namespace

anoadragon453:
Turns out that it is allowed to have multiple regexes per application service namespace, for instance an AS that bridges across multiple IRC networks. You'd want one users regex for each network.


APwhitehat:
Looks good code wise.

PR #537, opened 2018-07-04
Remove unnecessary req.Close()

APwhitehat:
This is not necessary according to [doc](https://golang.org/pkg/net/http/#Request).
Server requests are closed automatically.
Also avoid `nolint` directives


anoadragon453:
```
        // Body is the request's body.
        //
        // For client requests a nil body means the request has no
        // body, such as a GET request. The HTTP Client's Transport
        // is responsible for calling the Close method.
        //
        // For server requests the Request Body is always non-nil
        // but will return EOF immediately when no body is present.
        // The Server will close the request body. The ServeHTTP
        // Handler does not need to.
        Body io.ReadCloser
```

Yup, from what I can tell, this will have no ill effects.

PR #538, opened 2018-07-04
Remove unused params

APwhitehat:
These parameters won't be required since don't need to verify the JSON in a GET request :P.


anoadragon453:
Got a verbal OK from @erikjohnston.

PR #539, opened 2018-07-06
Consolidate AS interest checking

anoadragon453:
Combined multiple instances of the same loop into methods that can be called from anywhere.

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/routing/register.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/routing/register.go
@@ -337,17 +337,15 @@ func UsernameMatchesMultipleExclusiveNamespaces(
 ) bool {
 	// Check namespaces and see if more than one match
 	matchCount := 0
+	userID := userutil.MakeUserID(username, cfg.Matrix.ServerName)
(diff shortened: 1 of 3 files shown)


src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -337,17 +337,15 @@ func UsernameMatchesMultipleExclusiveNamespaces(
...
+	userID := userutil.MakeUserID(username, cfg.Matrix.ServerName)
 	for _, appservice := range cfg.Derived.ApplicationServices {
-		for _, namespaceSlice := range appservice.NamespaceMap {
-			for _, namespace := range namespaceSlice {
-				// Check if we have a match on this username
-				if namespace.RegexpObject.MatchString(username) {
-					matchCount++
-				}
+		if appservice.IsInterestedInUserID(userID) {
+			if matchCount++; matchCount > 1 {
```
erikjohnston: Ermh, won't this always be True?

krombel: It will be true on the second iteration.
The first match sets `matchCount` to 1 and since the second match it is `> 1`

Issue #463, opened 2018-05-23
ClientAPI: POST /_matrix/client/r0/delete_devices


anoadragon453:
**Spec:** https://matrix.org/docs/spec/client_server/unstable.html#post-matrix-client-r0-delete-devices

**Phase**: 2

Need to add an endpoint in the client-api router:

https://github.com/matrix-org/dendrite/blob/c2380485991f0aba55381da5f6bc8b7ba40296dc/src/github.com/matrix-org/dendrite/clientapi/routing/routing.go#L358-L376

And then a method for deleting devices in `device.go`: https://github.com/matrix-org/dendrite/blob/master/src/github.com/matrix-org/dendrite/clientapi/routing/device.go

APwhitehat:
Breakdown of what's required:
- Take a look at [dendrite/clientapi/auth/storage/devices/storage.go](https://github.com/matrix-org/dendrite/blob/master/src/github.com/matrix-org/dendrite/clientapi/auth/storage/devices/storage.go#L144), we'll use the function `RemoveDevice `.
- Then create a (http request) handler function [here](https://github.com/matrix-org/dendrite/blob/master/src/github.com/matrix-org/dendrite/clientapi/routing/device.go),  which takes a request of type [spec](https://matrix.org/docs/spec/client_server/unstable.html#post-matrix-client-r0-delete-devices).
- Then add the handler to [routing](https://github.com/matrix-org/dendrite/blob/master/src/github.com/matrix-org/dendrite/clientapi/routing/routing.go#L377) at path `/delete_devices`.

Note: Feel free to bug us at [#dendrite-dev:matrix.org](https://riot.im/app/#/room/#dendrite-dev:matrix.org).

t3chguy:
I'd argue that there should be another storage method for deleting all devices, it should be parallelised at the database instead of done sequentially/parallelised at the app level

APwhitehat:
@t3chguy I'm sorry, I don't understand.
pq supports parallel queries ? is there a speedup if we parallelise at the app level ?

(I looked and found that they do in fact support parallel read, but that's in version 10. Interesting though)

t3chguy:
I'm saying it'd be best if we ran a `DELETE... WHERE X IN (...) ` rather than N queries

APwhitehat:
True, we should be using a bulk query for this.

t3chguy:
[...]

PR #540, opened 2018-07-08
[#489] Trim room alias when receiving a directory request

### LINKED ISSUE #489: Trim room alias when receiving a directory request
Some clients **cough** Riot **cough** won't always trim strings when they send them to you. So if you request to join a room "#happy:matrix.org ", the request will fail because there's a space in there.

Easy fix, just need to trim vars["roomAlias"] here:

https://github.com/matrix-org/dendrite/blob/78440083dfc8f522e55f11f0fce0f3ea4a103f0b/src/github.com/matrix-org/dendrite/clientapi/routing/routing.go#L140-L145

Empty2k12:
> Some clients cough Riot cough won't always trim strings when they send them to you. So if you request to join a room "#happy:matrix.org ", the request will fail because there's a space in there.

Fixes #489

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/routing/routing.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/routing/routing.go
@@ -140,21 +140,21 @@ func Setup(
 	r0mux.Handle("/directory/room/{roomAlias}",
 		common.MakeAuthAPI("directory_room", deviceDB, func(req *http.Request, device *authtypes.Device) util.JSONResponse {
 			vars := mux.Vars(req)
-			return DirectoryRoom(req, vars["roomAlias"], federation, &cfg, aliasAPI)
+			return DirectoryRoom(req, strings.TrimSpace(vars["roomAlias"]), federation, &cfg, aliasAPI)
 		}),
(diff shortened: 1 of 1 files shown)


Empty2k12:
Signed-off-by: Gero Gerke <matrix.contributions@gerogerke.de>

anoadragon453:
Just out of curiosity, is there any reason not to trim every endpoint variable?

Empty2k12:
User input should be trimmed, everything else should be considered a bug by the sending software.

msrd0:
Just wondering why dendrite needs to trim anything at all? I guess that riot doesn't trim user inputs is a problem with riot and needs to be fixed client-side

anoadragon453:
The consensus here is that we're going to keep trimming a client-side task, especially as trimming on our side would require a lot more work on the server side as it is accepting much more requests than any single client is sending out.

If you've run into a string not being properly trimmed on a client, I suggest making an issue/PR on the client repo itself.

Thanks for the effort though, its helped us define what we do and do not want to cover on our endpoints!

APwhitehat:
@Empty2k12 Good job. Sorry this could not be merged.
If you're up for it, you could take a look at #463. It requires little more work but definitely doable.

PR #542, opened 2018-07-10
AppServices: Implement Timestamp Massaging

### LINKED ISSUE #456: AppServices: Add ability to perform Timestamp Massaging
**Spec:** https://matrix.org/docs/spec/application_service/unstable.html#timestamp-massaging

Application Services have the ability to determine the timestamp of the events that they send. In order to do this, a `?ts=` parameter can be added to the end of any Client-Server API endpoint.

Effectively all that's required of the homeserver here is to change the `origin_server_ts` value of the event to that given by the `?ts=` parameter if it exists in the query.

anoadragon453:
Timestamp massaging allows for an AS to send events with a custom `origin_server_ts` parameter.

Spec: https://matrix.org/docs/spec/application_service/unstable.html#timestamp-massaging

Closes #456. Depends on #427

### DIFF
--- a/src/github.com/matrix-org/dendrite/common/events.go
+++ b/src/github.com/matrix-org/dendrite/common/events.go
@@ -18,6 +18,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/http"
+	"strconv"
 	"time"
 
 	"github.com/matrix-org/dendrite/common/config"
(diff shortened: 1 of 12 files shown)


src/github.com/matrix-org/dendrite/common/events.go
```diff
@@ -38,25 +40,44 @@ var ErrRoomNoExists = errors.New("Room does not exist")
 // the room doesn't exist
 // Returns an error if something else went wrong
 func BuildEvent(
-	ctx context.Context,
+	req *http.Request,
```
APwhitehat: I am sorry I could not review this before.
I feel it's not common's job to parse the http request. Since this is used from all components, it will be very messy to send in the http.Request whenever we want a event to be built.
e.g. If someone wants to build an event by reading from kafka logs.

It'd be a lot cleaner if we could just pass the timestamp as an argument to BuildEvent.

src/github.com/matrix-org/dendrite/federationapi/routing/threepid.go
```diff
@@ -274,8 +273,8 @@ func buildMembershipEvent(
 	builder.AuthEvents = refs
 
 	eventID := fmt.Sprintf("$%s:%s", util.RandomString(16), cfg.Matrix.ServerName)
-	now := time.Now()
-	event, err := builder.Build(eventID, now, cfg.Matrix.ServerName, cfg.Matrix.KeyID, cfg.Matrix.PrivateKey)
+	eventTime := common.ParseTSParam(req)
```
APwhitehat: Shouldn't accept the ts parameter in federation APIs, right ?

PR #546, opened 2018-07-12
Remove trailing slashes from appservice url

anoadragon453:
User had their appservice broken from an extra trailing slash in the `url` field of their application service registration file.

While this is a user problem, it's nice to handle automatically.

### DIFF
--- a/src/github.com/matrix-org/dendrite/common/config/appservice.go
+++ b/src/github.com/matrix-org/dendrite/common/config/appservice.go
@@ -188,6 +188,11 @@ func checkErrors(config *Dendrite) (err error) {
 			}
 		}
 
+		// Check if the url has trailing /'s. If so, remove them
+		for strings.HasSuffix(appservice.URL, "/") {
+			appservice.URL = strings.TrimSuffix(appservice.URL, "/")
+		}
+
 		// Check if we've already seen this ID. No two application services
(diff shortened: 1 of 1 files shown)


src/github.com/matrix-org/dendrite/common/config/appservice.go
```diff
@@ -188,6 +188,11 @@ func checkErrors(config *Dendrite) (err error) {
 			}
 		}
 
+		// Check if the url has trailing /'s. If so, remove them
+		for strings.HasSuffix(appservice.URL, "/") {
+			appservice.URL = strings.TrimSuffix(appservice.URL, "/")
+		}
```
erikjohnston: You don't actually need to have the guard there

anoadragon453: This covers the use case of multiple trailing `/`s though?

erikjohnston: Oh, it's a `for` loop. I need to learn to read. Anyway, you can just use `TrimRight`, which will strip off all trailing slashes.

anoadragon453: Oh, that's handy.

PR #550, opened 2018-07-17
Support PUTing typing status in clientapi

APwhitehat:
Commits can be easily reviewed incrementally, :)

- Support PUT /rooms/{roomID}/typing/{userID}, add handler that passes it to `clientapi/producers` typingServerProducer
- the producer talks with typing server (currently mock implementation) via internal API
- Setup the typing server (mock)

### DIFF
--- a/src/github.com/matrix-org/dendrite/common/config/config_test.go
+++ b/src/github.com/matrix-org/dendrite/common/config/config_test.go
@@ -59,6 +59,7 @@ listen:
   federation_api: "localhost:7772"
   sync_api: "localhost:7773"
   media_api: "localhost:7774"
+  typing_server: "localhost:7778"
 logging:
   - type: "file"
     level: "info"
--- a/src/github.com/matrix-org/dendrite/typingserver/api/input.go
+++ b/src/github.com/matrix-org/dendrite/typingserver/api/input.go
@@ -1,5 +1,3 @@
-// Copyright 2017 Vector Creations Ltd
-//
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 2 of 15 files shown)


src/github.com/matrix-org/dendrite/typingserver/api/input.go
```diff
@@ -0,0 +1,85 @@
...
+	Typing bool `json:"typing"`
+	// Timeout is the interval for which the user should be marked as typing.
+	Timeout int64 `json:"timeout"`
+	// OriginServerTS when the server received the update.
+	OriginServerTS gomatrixserverlib.Timestamp `json:"origin_server_ts"`
+}
+
+// InputTypingEventsRequest is a request to TypingServerInputAPI
+type InputTypingEventsRequest struct {
+	InputTypingEvents []InputTypingEvent `json:"input_typing_events"`
```
krombel: Why do you want to input an array?
If one - but not all - requests are failing e.g. because of missing permissions for one request: How do you find out?
I think the easier way is to have one request per request. Where should the consolidating and separation happen?

APwhitehat: so, 1st thing is that these EDUs are already permitted (have gone through auth).
And from an perspective of drawing out an API, it seemed short-sighted to support single event per API call.

P.S. I remembered now that we'll need to put multiple events at a time for federation :P

erikjohnston: Given that we only currently send one typing event at a time I'd not use a slice for now. Its an internal API so we can change it down the road if we need it to handle multiple typing events :)

src/github.com/matrix-org/dendrite/common/config/config_test.go
```diff
@@ -59,6 +59,7 @@ listen:
   federation_api: "localhost:7772"
   sync_api: "localhost:7773"
   media_api: "localhost:7774"
```
anoadragon453: /me wonders why the other components haven't been added here.

APwhitehat: tracking this via #558

PR #554, opened 2018-07-21
Ensure appservices have their devices checked

turt2live:
The regular device check will return the device for the appservice's bot user instead of going through the user_id branch. The check has been moved to below the user_id check to ensure the right virtual user's device is chosen.


APwhitehat:
@turt2live are we now adding AS to the device DB ?

turt2live:
@APwhitehat I have no idea what that means. All I know is that by moving the block down further, `user_id` works again.

anoadragon453:
> are we now adding AS to the device DB ?

We already were, no? The only AS user that's generated automatically is the `sender_localpart` one. The rest have to be created by the AS through the C-S API, so they should exist in the DB.

PR #556, opened 2018-07-23
OpenTracing support across multiple components

anoadragon453:
Adds Jaeger OpenTracing support per-component. The same tracing behavior is observed externally for both Monolith and Multi-Process mode. A tracer is created per component, even in Monolith mode.

Currently Internal APIs do not seem to be traced, at least in my own testing. Will need to figure out why that is, as the code for it is in place.

OpenTracing support was also added to the Appservices component.


src/github.com/matrix-org/dendrite/common/httpapi.go
```diff
@@ -24,20 +38,32 @@ func MakeAuthAPI(
...
 	withSpan := func(w http.ResponseWriter, req *http.Request) {
-		span := opentracing.StartSpan(metricsName)
+		span := tracer.StartSpan(metricsName)
 		defer span.Finish()
+
+		ext.HTTPUrl.Set(span, req.URL.String())
+		ext.HTTPMethod.Set(span, req.Method)
+
 		req = req.WithContext(opentracing.ContextWithSpan(req.Context(), span))
+		rw := newStatusCodeResponseWriter(w)
```
anoadragon453: I'm a little confused on the few lines above, hence no comments. I get that this is essentially adding metadata to the trace such as request URI and status code. Do we need to do this for internalAPIs as well.

src/github.com/matrix-org/dendrite/common/tracers.go
```diff
@@ -0,0 +1,92 @@
...
+func (t *Tracers) Close() error {
+	for _, c := range t.closers {
+		c.Close() // nolint: errcheck
+	}
+
+	return nil
+}
+
+// logrusLogger is a small wrapper that implements jaeger.Logger using logrus.
+type logrusLogger struct {
```
anoadragon453: Do we want any more logging methods here than `Error` and `Infof`?

src/github.com/matrix-org/dendrite/mediaapi/routing/routing.go
```diff
@@ -48,9 +50,11 @@ func Setup(
 	}
 	authData := auth.Data{nil, deviceDB, nil}
 
+	// TODO: How to use tracing with these endpoints?
```
anoadragon453: I think this was essentially asking whether or not we need to use tracing with the DownloadAPI.

src/github.com/matrix-org/dendrite/roomserver/query/query.go
```diff
@@ -709,3 +710,130 @@ func (r *RoomserverQueryAPI) SetupHTTP(servMux *http.ServeMux) {
...
 }
+
+// InProcessRoomServerQueryAPI is a RoomserverQueryAPI database linked with a
+// OpenTracing tracer, providing metrics
+type InProcessRoomServerQueryAPI struct {
+	db     RoomserverQueryAPI
+	tracer opentracing.Tracer
+}
+
+// NewInProcessRoomServerQueryAPI creates a new InProcessRoomServerQueryAPI
```
anoadragon453: Hm, not sure if this is entirely relevant to this PR, but was a part of the other branches. Thoughts @erikjohnston?

src/github.com/matrix-org/dendrite/common/basecomponent/base.go
```diff
@@ -78,7 +76,7 @@ func NewBaseDendrite(cfg *config.Dendrite, componentName string) *BaseDendrite {
 
 // Close implements io.Closer
 func (b *BaseDendrite) Close() error {
-	return b.tracerCloser.Close()
+	return b.tracers.Close()
```
krombel: You call tracers.Close() twice. This is just one example
https://github.com/matrix-org/dendrite/blob/9831fcb855a793f4c3571999b32b16b251035374/src/github.com/matrix-org/dendrite/cmd/dendrite-sync-api-server/main.go#L26-L30

anoadragon453: Oh yes, good catch!

src/github.com/matrix-org/dendrite/appservice/storage/storage.go
```diff
@@ -20,6 +20,7 @@ import (
 
 	// Import postgres database driver
 	_ "github.com/lib/pq"
```
krombel: You are not using this here anymore

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/storage.go
```diff
@@ -24,7 +24,6 @@ import (
 	"github.com/matrix-org/gomatrixserverlib"
 	"golang.org/x/crypto/bcrypt"
 	// Import the postgres database driver.
```
krombel: This comment can be removed now as well (not just the import)

src/github.com/matrix-org/dendrite/common/tracers.go
krombel: This is just copied from [here](https://github.com/matrix-org/dendrite/pull/556/files#diff-37e96baa4412da7918486fbc488561f9L662).
So I would not change more here as well. Only add once we start using it

src/github.com/matrix-org/dendrite/appservice/storage/storage.go
krombel: There are some other appearances where you just removed the import but not the comment for it.
And some some more of this

src/github.com/matrix-org/dendrite/common/httpapi.go
erikjohnston: Yeah, this is just annotating the span with some info. We probably want to annotate internalAPI spans with if there was an error or not.

src/github.com/matrix-org/dendrite/common/httpapi.go
```diff
@@ -62,6 +87,12 @@ func MakeInternalAPI(metricsName string, f func(*http.Request) util.JSONResponse
 			span = tracer.StartSpan(metricsName, ext.RPCServerOption(clientContext))
 		}
 		defer span.Finish()
+
+		ext.HTTPUrl.Set(span, req.URL.String())
+		ext.HTTPMethod.Set(span, req.Method)
+
+		// TODO: Do we need to do the newStatusCodeResponseWriter stuff?
```
erikjohnston: Yes, seeing if an internal api succeeded or failed would be useful i think

anoadragon453: Ah, that's what that is for. Will do!

src/github.com/matrix-org/dendrite/mediaapi/routing/routing.go
erikjohnston: Yeah, looks like it.

src/github.com/matrix-org/dendrite/roomserver/query/query.go
erikjohnston: Yes this is. It's so that we can trace calls between different components in the same process, like we do for the InternalAPI stuff

Kegsay:

PR #559, opened 2018-07-24
Add TypingCache to maintain a list of users typing

APwhitehat:
ready for review.

### DIFF
--- a/src/github.com/matrix-org/dendrite/typingserver/cache/cache.go
+++ b/src/github.com/matrix-org/dendrite/typingserver/cache/cache.go
@@ -53,7 +53,7 @@ func (t *TypingCache) GetTypingUsers(roomID string) (users []string) {
 // if expire is nil, defaultTypingTimeout is assumed.
 func (t *TypingCache) AddTypingUser(userID, roomID string, expire *time.Time) {
 	expireTime := getExpireTime(expire)
-	if until := time.Until(expireTime); until > 0 {
+	if time.Until(expireTime) > 0 {
 		t.addUser(userID, roomID, expireTime)
 		t.removeUserAfterTime(userID, roomID, expireTime)
 	}
--- a/src/github.com/matrix-org/dendrite/typingserver/cache/cache.go
+++ b/src/github.com/matrix-org/dendrite/typingserver/cache/cache.go
@@ -19,8 +19,8 @@ import (
 
 var defaultTypingTimeout = 10 * time.Second
 
-// userSet is a map of user IDs to their time of expiry.
-type userSet map[string]time.Time
+// userSet is a map of user IDs to a timer, timer fires at expiry.
+type userSet map[string]*time.Timer
 
 // TypingCache maintains a list of users typing in each room.
 type TypingCache struct {
@@ -53,14 +53,14 @@ func (t *TypingCache) GetTypingUsers(roomID string) (users []string) {
 // if expire is nil, defaultTypingTimeout is assumed.
 func (t *TypingCache) AddTypingUser(userID, roomID string, expire *time.Time) {
 	expireTime := getExpireTime(expire)
-	if time.Until(expireTime) > 0 {
-		t.addUser(userID, roomID, expireTime)
-		t.removeUserAfterTime(userID, roomID, expireTime)
+	if until := time.Until(expireTime); until > 0 {
+		timer := time.AfterFunc(until, t.timeoutCallback(userID, roomID))
+		t.addUser(userID, roomID, timer)
 	}
 }
 
-// addUser with mutex lock.
-func (t *TypingCache) addUser(userID, roomID string, expireTime time.Time) {
+// addUser with mutex lock & replace the previous timer.
+func (t *TypingCache) addUser(userID, roomID string, expiryTimer *time.Timer) {
 	t.Lock()
 	defer t.Unlock()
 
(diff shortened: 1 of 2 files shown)


src/github.com/matrix-org/dendrite/typingserver/cache/cache_test.go
```diff
@@ -0,0 +1,97 @@
...
+		gotUsers := tCache.GetTypingUsers(tt.roomID)
+		sort.Strings(gotUsers)
+		sort.Strings(tt.wantUsers)
+		if !reflect.DeepEqual(gotUsers, tt.wantUsers) {
+			t.Errorf("TypingCache.GetTypingUsers(%s) = %v, want %v", tt.roomID, gotUsers, tt.wantUsers)
+		}
+	}
+}
+
+func testGetTypingUsersAfterTimeout(t *testing.T, tCache *TypingCache) {
```
APwhitehat: This test fails since the `removeUserAfterDuration` function will remove user1 after the first timeout occurs regardless of the fact that user1 is again added with a longer expiry.

https://github.com/APwhitehat/dendrite/blob/3e697873a65ccb25ea984b3ac66db39a5e2513bb/src/github.com/matrix-org/dendrite/typingserver/cache/cache.go#L75-L81

src/github.com/matrix-org/dendrite/typingserver/cache/cache_test.go
```diff
@@ -0,0 +1,97 @@
...
+	time.Sleep(defaultInterval)
```
erikjohnston: Please don't use `Sleep` in tests, they should complete quickly. Instead have the cache accept a "clock", this can then be mocked in the unit tests

src/github.com/matrix-org/dendrite/typingserver/cache/cache.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func (t *TypingCache) addUser(userID, roomID string) {
+	t.Lock()
+	if t.data[roomID] == nil {
+		t.data[roomID] = make(userSet)
+	}
+
+	t.data[roomID][userID] = userExists
+	t.Unlock()
```
erikjohnston: I would heavily recommend using `defer` where possible.

```golang
t.Lock()
defer t.Unlock()
...
```

APwhitehat: I got swept away by the argument that `defer` introduce an overhead. :P
But good programming practice.

src/github.com/matrix-org/dendrite/typingserver/cache/cache_test.go
```diff
@@ -0,0 +1,97 @@
...
+	}{
+		{"room1", []string{"user1", "user2", "user3", "user4"}},
+		{"room2", []string{"user1"}},
+	}
+
+	for _, tt := range tests {
+		gotUsers := tCache.GetTypingUsers(tt.roomID)
+		sort.Strings(gotUsers)
+		sort.Strings(tt.wantUsers)
+		if !reflect.DeepEqual(gotUsers, tt.wantUsers) {
```
krombel: DeepEqual fails when the order of the entries does not match.
This makes this test flaky

src/github.com/matrix-org/dendrite/typingserver/cache/cache_test.go
```diff
@@ -0,0 +1,97 @@
...
+		roomID    string
+		userID    string
+		wantUsers []string
+	}{
+		{"room2", "user1", []string{"user1"}},
+	}
+
+	for _, tt := range tests {
+		tCache.removeUserIfExpired(tt.userID, tt.roomID)
+		if gotUsers := tCache.GetTypingUsers(tt.roomID); !reflect.DeepEqual(gotUsers, tt.wantUsers) {
```
krombel: DeepEqual fails when the order of the entries does not match.
This makes this test flaky.

krombel: BTW there is this. You might reuse that somehow:
https://github.com/matrix-org/dendrite/blob/9e352e7311e41853dc151786dceb783f5114c538/src/github.com/matrix-org/dendrite/roomserver/query/query_test.go#L94-L109

APwhitehat: Thanks for the suggestion. :)

APwhitehat:
New test from @krombel. Yay!

APwhitehat:
For some reason travis build is not visible. One can it here https://travis-ci.org/matrix-org/dendrite/builds/409862091

src/github.com/matrix-org/dendrite/typingserver/cache/cache_test.go
```diff
@@ -0,0 +1,100 @@
...
+		{"room3", []string{"user1"}},
+		{"room4", []string{"user1", "user2", "user3"}},
+	}
+	for _, tt := range tests {
+		for _, userID := range tt.userIDs {
+			tCache.AddTypingUser(userID, tt.roomID, nil)
+		}
+		i := rand.Intn(len(tt.userIDs))
+		tCache.removeUser(tt.userIDs[i], tt.roomID)
+		expLeftUsers := append(tt.userIDs[:i], tt.userIDs[i+1:]...)
```
erikjohnston: Please don't use random in tests, it can make them flakey and annoying to debug. I'd just pick one to remove, e.g. the last, as there's nothing that should depend on the ordering here.

APwhitehat: sure.

PR #561, opened 2018-07-25
Third Party Lookup Support

anoadragon453:
This PR implements the Third Party Lookup API as defined by https://matrix.org/docs/spec/client_server/unstable.html#id127

Third party lookups allow clients to provide rich bridging interfaces by using the homeserver as a proxy to retrieve third party network metadata from application services acting as bridges.

These endpoints are mostly designed to simply be proxied by the homeserver to the application service, and thus the majority of them use the same handler function. New internal API endpoints for appservices were created to allow communication between the C-S and AS components.

This PR depends on #522. Needs a rebase onto master.

### DIFF
--- a/src/github.com/matrix-org/dendrite/appservice/query/query.go
+++ b/src/github.com/matrix-org/dendrite/appservice/query/query.go
@@ -90,8 +90,8 @@ func (a *AppServiceQueryAPI) RoomAliasExists(
 					err = resp.Body.Close()
 					if err != nil {
 						log.WithFields(log.Fields{
-							"appservice":  appservice.ID,
-							"status_code": resp.StatusCode,
+							"appservice_id": appservice.ID,
+							"status_code":   resp.StatusCode,
 						}).WithError(err).Error("Unable to close application service response body")
 					}
 				}()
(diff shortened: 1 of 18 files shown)


src/github.com/matrix-org/dendrite/appservice/query/query.go
```diff
@@ -17,18 +17,351 @@
...
+		// TODO: Verify certificates
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: true, // nolint: gas
+			},
+		},
+	}
 }
 
 // SetupHTTP adds the AppServiceQueryPAI handlers to the http.ServeMux. This
```
rederekt: Minor typo 😄

anoadragon453:
@rederekt Reviewing this PR is taking a bit longer due to parallel work happening. Which exact functions do you need? I'll see if I can separate them out and make them more generalized then the ones presented in this PR.

For ThirdPartyProxy, I'm thinking of making a generic function that contacts all application services and merges results, and having ThirdPartyProxy then use that.

rederekt:
@anoadragon453 I think for a number of the App Service APIs, we'll need a way to contact application services. For mine specifically, the [ThirdPartyProxy funtion](https://github.com/matrix-org/dendrite/pull/561/files#diff-f3464f1c9dd37b2cdcc0d85ac980d65fR223), and everything involved to make it work, would be most useful. Then, implementing URI to UID and URI to room alias would be simple.

Kegsay:

PR #562, opened 2018-07-26
Refactor string slice helper to common/test

APwhitehat:
quick refactor before using this in #559

### DIFF
--- a/src/github.com/matrix-org/dendrite/common/test/slice.go
+++ b/src/github.com/matrix-org/dendrite/common/test/slice.go
@@ -14,9 +14,9 @@ package test
 
 import "sort"
 
-// StringSliceEqual returns true if the slices have same length & elements.
+// UnsortedStringSliceEqual returns true if the slices have same length & elements.
 // Does not modify the given slice.
-func StringSliceEqual(first, second []string) bool {
+func UnsortedStringSliceEqual(first, second []string) bool {
 	if len(first) != len(second) {
 		return false
 	}
--- a/src/github.com/matrix-org/dendrite/common/test/slice.go
+++ b/src/github.com/matrix-org/dendrite/common/test/slice.go
@@ -15,11 +15,13 @@ package test
 import "sort"
 
 // StringSliceEqual returns true if the slices have same length & elements.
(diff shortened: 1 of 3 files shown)


src/github.com/matrix-org/dendrite/common/test/strings.go
```diff
@@ -0,0 +1,32 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package test
+
+import "sort"
+
+// StringSliceEqual returns true if the slices have same length & elements.
```
t3chguy: Probably worth noting that it does mutate the inputs

APwhitehat: yes, good catch.

APwhitehat: I should probably sort a copy of the slices.
since in tests we usually have small slices, this is okay.

anoadragon453:
This would likely fit better in a new file in https://github.com/matrix-org/util

src/github.com/matrix-org/dendrite/common/test/slice.go
```diff
@@ -0,0 +1,34 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package test
+
+import "sort"
+
+// StringSliceEqual returns true if the slices have same length & elements.
+// Does not modify the given slice.
+func StringSliceEqual(first, second []string) bool {
```
erikjohnston: I'd keep some mention of "unsorted" in the name to avoid confusion

APwhitehat:
@anoadragon453 I seem to have not noticed this repo :P 
util seems to be good place. And we're using this in tests exclusively so common/test makes sense too.

erikjohnston:
I don't think its worth moving things to util just for the sake of it, it just makes dev'ing more complicated if you now need to track changes across multi repos

PR #563, opened 2018-07-26
Differentiate between users and AS upon registration

anoadragon453:
#548 unfortunately broke registration as users also register without an auth type initially, while AS's also do so when they are acting as their SenderLocalpart user.

We now check beforehand if this is occurring before we switch over auth types.

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/routing/register.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/routing/register.go
@@ -497,7 +497,8 @@ func Register(
 	// Application services can register users with no auth type, but require
 	// access token. Differentiate from users who are initially hitting register
 	// without an auth type
-	if r.Auth.Type == "" && req.URL.Query().Get("access_token") == "" {
+	if r.Auth.Type == "" && req.URL.Query().Get("access_token") == "" &&
+		req.Header.Get("Authorization") == "" {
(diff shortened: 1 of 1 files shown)


src/github.com/matrix-org/dendrite/clientapi/routing/register.go
```diff
@@ -485,6 +494,17 @@ func Register(
 		}
 	}
 
+	// Application services can register users with no auth type, but require
+	// access token. Differentiate from users who are initially hitting register
+	// without an auth type
+	if r.Auth.Type == "" && req.URL.Query().Get("access_token") == "" {
```
krombel: Application services can use Auth headers for this endpoint as well I assume.
So this check is "incomplete"

anoadragon453: Good point, let me add that in.

krombel:
@anoadragon453 There is already this. I would just reuse it here:
https://github.com/matrix-org/dendrite/blob/68131ca7a37bded8a00f745cb71fa538d26746e9/src/github.com/matrix-org/dendrite/clientapi/auth/auth.go#L189

anoadragon453:
Ah yes, that should work and stay under the cyclo limit.

Kegsay:

PR #566, opened 2018-07-28
removed outdated "flag" comments on clientapi membership storage

rederekt:
Signed-off-by: Derek Meer <derekmeer@protonmail.com>

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/storage.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/storage.go
@@ -143,8 +143,7 @@ func (d *Database) CreateAccount(
 }
 
 // SaveMembership saves the user matching a given localpart as a member of a given
-// room. It also stores the ID of the membership event and a flag on whether the user
-// is still in the room.
+// room. It also stores the ID of the membership event.
 // If a membership already exists between the user and the room, or of the
 // insert fails, returns the SQL error
 func (d *Database) saveMembership(
@@ -205,11 +204,9 @@ func (d *Database) GetMembershipsByLocalpart(
 	return d.memberships.selectMembershipsByLocalpart(ctx, localpart)
 }
 
-// newMembership will save a new membership in the database, with a flag on whether
-// the user is still in the room. This flag is set to true if the given state
-// event is a "join" membership event and false if the event is a "leave" or "ban"
-// membership. If the event isn't a m.room.member event with one of these three
-// values, does nothing.
+// newMembership will save a new membership in the database.
+// If the event isn't a m.room.member event or there's no state key
+// (which stores the server name and localpart), does nothing.
 // If the event isn't a "join" membership event, does nothing
 // If an error occurred, returns it
 func (d *Database) newMembership(
--- a/src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/storage.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/storage.go
@@ -144,7 +144,7 @@ func (d *Database) CreateAccount(
 
 // SaveMembership saves the user matching a given localpart as a member of a given
 // room. It also stores the ID of the membership event.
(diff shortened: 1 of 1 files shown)


src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/storage.go
```diff
@@ -143,8 +143,7 @@ func (d *Database) CreateAccount(
 }
 
 // SaveMembership saves the user matching a given localpart as a member of a given
-// room. It also stores the ID of the membership event and a flag on whether the user
-// is still in the room.
+// room. It also stores the ID of the membership event.
 // If a membership already exists between the user and the room, or of the
```
APwhitehat: correction: ~of~ if

APwhitehat:
https://github.com/matrix-org/dendrite/pull/566/files?utf8=%E2%9C%93&diff=unified#diff-b411bc269702e1949607e9e5aab0772cR155
could you change this to to something like the following ?
... removes the memberships corresponding to the `join` membership events IDs.
If the removal fails, or if there is no membership to remove, returns an error.

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/storage.go
```diff
@@ -205,11 +204,9 @@ func (d *Database) GetMembershipsByLocalpart(
...
-// newMembership will save a new membership in the database, with a flag on whether
-// the user is still in the room. This flag is set to true if the given state
-// event is a "join" membership event and false if the event is a "leave" or "ban"
-// membership. If the event isn't a m.room.member event with one of these three
-// values, does nothing.
+// newMembership will save a new membership in the database.
+// If the event isn't a m.room.member event or there's no state key
+// (which stores the server name and localpart), does nothing.
 // If the event isn't a "join" membership event, does nothing
 // If an error occurred, returns it
```
APwhitehat: correction: ~returns it~ returns the SQL error

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/storage.go
```diff
@@ -205,11 +204,9 @@ func (d *Database) GetMembershipsByLocalpart(
 	return d.memberships.selectMembershipsByLocalpart(ctx, localpart)
 }
 
-// newMembership will save a new membership in the database, with a flag on whether
-// the user is still in the room. This flag is set to true if the given state
-// event is a "join" membership event and false if the event is a "leave" or "ban"
-// membership. If the event isn't a m.room.member event with one of these three
-// values, does nothing.
+// newMembership will save a new membership in the database.
+// If the event isn't a m.room.member event or there's no state key
```
APwhitehat: ftr state key is the user id corresponding to the m.room.member event, i.e. the user who joins or leaves.
You're correct that it contains server name and the localpart. I would rather not have this description since it's documented elsewhere or spec. (if not file an issue :P)

APwhitehat: since a valid m.room.member event would always have the state key. perhaps this would be better (& shorter):

If the event isn't a valid m.room.member event with type `join`, does nothing.

src/github.com/matrix-org/dendrite/clientapi/auth/storage/accounts/storage.go
```diff
@@ -205,11 +204,9 @@ func (d *Database) GetMembershipsByLocalpart(
 	return d.memberships.selectMembershipsByLocalpart(ctx, localpart)
 }
 
-// newMembership will save a new membership in the database, with a flag on whether
-// the user is still in the room. This flag is set to true if the given state
-// event is a "join" membership event and false if the event is a "leave" or "ban"
-// membership. If the event isn't a m.room.member event with one of these three
-// values, does nothing.
+// newMembership will save a new membership in the database.
```
APwhitehat: correction: ~will save~ saves

PR #568, opened 2018-08-02
URI to UID API

### LINKED ISSUE #457: AppServices: GET /_matrix/app/r0/user?uri={url_encoded_uri}
**Spec:** https://matrix.org/docs/spec/application_service/unstable.html#user-ids

This endpoint takes an encoded URI (such as `tel://123.1234`) that represents an appservice user, and returns the complete Matrix ID on the homeserver (`@_tel_1231234:matrix.org`) which can then be used to contact them directly.

dcrck:
I'm looking for guidance on how to properly implement the URI-to-User ID mapping since the [spec](https://matrix.org/docs/spec/application_service/unstable.html#user-ids) leaves the implementation pretty open-ended. I tried looking to synapse and matrix-docs for guidance, but I couldn't find anything substantial or definitive. Am I missing anything? If I am, I'd appreciate a link to the agreed-upon mapping.  If not, what are your thoughts on mapping URIs to User IDs? Right now, I just replace all the illegal characters with underscores. I'm guessing this isn't good enough, but it's a starting point for a discussion (if one is necessary). Thank you!

(Fixes #457 )

### DIFF
--- a/src/github.com/matrix-org/dendrite/appservice/routing/user.go
+++ b/src/github.com/matrix-org/dendrite/appservice/routing/user.go
@@ -16,8 +16,9 @@ type URIToUIDResponse struct {
 // URI and turning it into a Matrix ID on the homeserver.
 // https://matrix.org/docs/spec/application_service/unstable.html#user-ids
 
-func URIToUID(req *http.Request) util.JSONResponse {
+func URIToUID(req *http.Request, cfg config.Dendrite) util.JSONResponse {
 	// TODO: Implement
+	homeserver := cfg.Matrix.ServerName
 	return util.JSONResponse{
 		Code: http.StatusOK,
 		JSON: nil,
--- a/src/github.com/matrix-org/dendrite/appservice/routing/user.go
+++ b/src/github.com/matrix-org/dendrite/appservice/routing/user.go
@@ -1,0 +1,25 @@
+package routing
+
+import (
+	"github.com/matrix-org/util"
+	"net/http"
+)
+
+// represents response to an AppService URI to User Id
+// (/_matrix/app/r0/user?uri={url_encoded_uri}) request
+type URIToUIDResponse struct {
+	UserID string `json:"user_id"`
+}
+
+// URIToUID implements `/_matrix/app/r0/user?uri={url_encoded_uri}`, which
+// enables users to contact appservice users directly by taking an encoded
+// URI and turning it into a Matrix ID on the homeserver.
+// https://matrix.org/docs/spec/application_service/unstable.html#user-ids
+
+func URIToUID(req *http.Request) util.JSONResponse {
+	// TODO: Implement
+	return util.JSONResponse{
+		Code: http.StatusOK,
+		JSON: nil,
+	}
+}
--- a/src/github.com/matrix-org/dendrite/appservice/routing/user.go
+++ b/src/github.com/matrix-org/dendrite/appservice/routing/user.go
@@ -27,14 +27,14 @@ func URIToUID(req *http.Request, cfg config.Dendrite) util.JSONResponse {
 			JSON: nil,
 		}
 	}
-	baseReqURL := "http://" + string(cfg.Matrix.ServerName) + "/_matrix/app/unstable/thirdparty/user/"
 	for _, appservice := range cfg.Derived.ApplicationServices {
 		// Check all the fields associated with each application service
 		if !appservice.IsInterestedInUserID(uri) {
 			continue
 		}
 		// call the application service
-		reqURL := baseReqURL + appservice.ID + "?access_token=" + appservice.HSToken +
+		reqURL := "http://" + appservice.URL + "/_matrix/app/unstable/thirdparty/user/" +
+			appservice.ID + "?access_token=" + appservice.HSToken +
 			"&fields=" + uri
 		resp, err := http.Get(reqURL)
 		// take the first successful match and send that back to the user
--- a/src/github.com/matrix-org/dendrite/appservice/routing/user.go
+++ b/src/github.com/matrix-org/dendrite/appservice/routing/user.go
@@ -14,7 +14,7 @@ type URIToUIDResponse struct {
 }
 
 // URIToUID implements `/_matrix/app/r0/user?uri={url_encoded_uri}`, which
-// enables users to contact appservice users directly by taking an encoded
+// enables users to contact App Service users directly by taking an encoded
 // URI and turning it into a Matrix ID on the homeserver.
 // https://matrix.org/docs/spec/application_service/unstable.html#user-ids
 // tel://123.1234 -> @tel_//123.1234:matrix.org
--- a/src/github.com/matrix-org/dendrite/appservice/routing/user.go
+++ b/src/github.com/matrix-org/dendrite/appservice/routing/user.go
@@ -17,7 +17,7 @@ type URIToUIDResponse struct {
 	UserID string `json:"user_id"`
 }
 
-const PathPrefixUnstableThirdPartyUser = "/_matrix/app/unstable/thirdparty/user/"
+const pathPrefixUnstableThirdPartyUser = "/_matrix/app/unstable/thirdparty/user/"
 
 // URIToUID implements `/_matrix/app/r0/user?uri={url_encoded_uri}`, which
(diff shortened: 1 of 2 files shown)


anoadragon453:
OK so what I think you'd do is use [this endpoint](https://matrix.org/docs/spec/application_service/unstable.html#get-matrix-app-unstable-thirdparty-user-protocol) with one of the fields filled with the URI which will return a list of matching Matrix UserIDs. Check all the fields associated with each application service first, if one or more matches, send a request to that endpoint on each matched application service. Since this endpoint only returns one User ID, take the first successful match and send that back to the user.

rederekt:
I think it's fixed now...let me know if there's a better way to call the App Service API since mine is kind of messy. Thanks!

src/github.com/matrix-org/dendrite/appservice/routing/user.go
```diff
@@ -0,0 +1,60 @@
...
+			reqURL := baseReqURL + appservice.ID + "?access_token=" + appservice.HSToken +
+				"&fields=" + uri
+			resp, err := http.Get(reqURL)
+			// take the first successful match and send that back to the user
+			if err == nil {
+				body, _ := ioutil.ReadAll(resp.Body)
+				respMap := map[string]interface{}{}
+				err := json.Unmarshal(body, &respMap)
+				if err != nil {
+					panic(err)
```
krombel: In general you don't want to panic here but return that the content is not acceptable. I don't know how far this goes but in worst case it can kill whole dendrite.

In this situation it might be best to use [UnmarshalJSONRequest](https://github.com/matrix-org/dendrite/blob/c87fb4d/src/github.com/matrix-org/dendrite/clientapi/httputil/httputil.go#L27) where you can return what that method returns if it is not `nil`.

src/github.com/matrix-org/dendrite/appservice/routing/user.go
```diff
@@ -0,0 +1,60 @@
...
+func URIToUID(req *http.Request, cfg config.Dendrite) util.JSONResponse {
+	uri := req.URL.Query().Get("uri")
+	if uri == "" {
+		return util.JSONResponse{
+			Code: http.StatusOK,
+			JSON: nil,
+		}
+	}
+	baseReqURL := "http://" + string(cfg.Matrix.ServerName) + "/_matrix/app/unstable/thirdparty/user/"
+	//appServices := cfg.Derived.ApplicationServices
```
krombel: you should remove this

src/github.com/matrix-org/dendrite/appservice/routing/user.go
```diff
@@ -0,0 +1,60 @@
...
+	for _, appservice := range cfg.Derived.ApplicationServices {
+		// Check all the fields associated with each application service
+		if appservice.IsInterestedInUserID(uri) {
```
krombel: I am not sure anymore if that is a rule to go here, but I would prefer the following
```
if !appservice.IsInterestedInUserID(uri) {
    continue
}
```

src/github.com/matrix-org/dendrite/appservice/routing/user.go
```diff
@@ -0,0 +1,60 @@
...
+			// call the application service
+			reqURL := baseReqURL + appservice.ID + "?access_token=" + appservice.HSToken +
+				"&fields=" + uri
+			resp, err := http.Get(reqURL)
+			// take the first successful match and send that back to the user
+			if err == nil {
```
krombel: nearly same as above: AFAIK it is preferred to first do the error-check and abort/return and then the other.
That way you have not that many indentations

rederekt:
@krombel Thanks for your feedback! I've made changes based on that.

src/github.com/matrix-org/dendrite/appservice/routing/user.go
```diff
@@ -0,0 +1,64 @@
...
+// https://matrix.org/docs/spec/application_service/unstable.html#user-ids
+func URIToUID(req *http.Request, cfg config.Dendrite) util.JSONResponse {
+	uri := req.URL.Query().Get("uri")
+	if uri == "" {
+		return util.JSONResponse{
+			Code: http.StatusOK,
+			JSON: nil,
+		}
+	}
+	baseReqURL := "http://" + string(cfg.Matrix.ServerName) + "/_matrix/app/unstable/thirdparty/user/"
```
krombel: Why do you try to trigger that URL? Isn't it right that this will hit dendrite again (which does not implement this URL currently btw)?
Or which instance should answer that request?

rederekt: @anoadragon453 made some relevant comments above. Basically, I think I'm using [this endpoint](https://matrix.org/docs/spec/application_service/unstable.html#get-matrix-app-unstable-thirdparty-user-protocol) to access the third party network and get the right data back. I think I messed up, though, since I used the homeserver URL here instead of the app service URL. I'll go ahead and change that.

src/github.com/matrix-org/dendrite/appservice/routing/user.go
```diff
@@ -27,14 +27,14 @@ func URIToUID(req *http.Request, cfg config.Dendrite) util.JSONResponse {
...
 	}
-	baseReqURL := "http://" + string(cfg.Matrix.ServerName) + "/_matrix/app/unstable/thirdparty/user/"
 	for _, appservice := range cfg.Derived.ApplicationServices {
 		// Check all the fields associated with each application service
 		if !appservice.IsInterestedInUserID(uri) {
 			continue
 		}
 		// call the application service
-		reqURL := baseReqURL + appservice.ID + "?access_token=" + appservice.HSToken +
+		reqURL := "http://" + appservice.URL + "/_matrix/app/unstable/thirdparty/user/" +
```
anoadragon453: The appservice.URL should include "http://"

src/github.com/matrix-org/dendrite/appservice/routing/user.go
```diff
@@ -0,0 +1,64 @@
...
+			JSON: nil,
+		}
+	}
+	for _, appservice := range cfg.Derived.ApplicationServices {
+		// Check all the fields associated with each application service
+		if !appservice.IsInterestedInUserID(uri) {
+			continue
+		}
+		// call the application service
+		reqURL := "http://" + appservice.URL + "/_matrix/app/unstable/thirdparty/user/" +
```
anoadragon453: It's also recommended to change `"/_matrix/app/unstable/thirdparty/user/"` to a const that's set at the top of the file.

src/github.com/matrix-org/dendrite/appservice/routing/user.go
```diff
@@ -0,0 +1,64 @@
...
+// https://matrix.org/docs/spec/application_service/unstable.html#user-ids
+func URIToUID(req *http.Request, cfg config.Dendrite) util.JSONResponse {
+	uri := req.URL.Query().Get("uri")
+	if uri == "" {
+		return util.JSONResponse{
+			Code: http.StatusOK,
+			JSON: nil,
+		}
+	}
+	for _, appservice := range cfg.Derived.ApplicationServices {
```
anoadragon453: It would be ideal to contact all application services simultaneously in parallel and then collect and present the results once all have answered or timed out.

You can see an example of doing so here: https://github.com/matrix-org/dendrite/pull/561/files#diff-f3464f1c9dd37b2cdcc0d85ac980d65fR195

rederekt: That PR has a lot of helpful stuff in it! I may have to refactor mine after that one gets merged...

rederekt:
I added some parallelization, although taking a look at #561 , it may be more appropriate to wait for that PR to merge before moving forward with this one, since it contains several helpful functions for App Service APIs. I can put this on hold until that point if you agree.

anoadragon453:
Sounds good. I'll fix it up so it's mergable.

anoadragon453:
Yes, but that brings up a good point in that that method should probably be abstracted further into a asynchronous method for communicating with application services that anything can use, including the ThirdPartyProxy method. I'll get on that tomorrow so you aren't blocked further.

rederekt:
Sounds good, just let me know how to use the method once you're done with it!

Kegsay:

PR #570, opened 2018-08-03
Fix up timestamp messaging

APwhitehat:
fixes #542 

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/httputil/parse.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/httputil/parse.go
@@ -1,0 +1,38 @@
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package httputil
+
+import (
+	"net/http"
+	"strconv"
+	"time"
+)
+
+// ParseTSParam takes a req (typically from an application service) and parses a Time object
+// from the req if it exists in the query parameters. If it doesn't exist, the
+// current time is returned.
+func ParseTSParam(req *http.Request) time.Time {
+	// Use the ts parameter's value for event time if present
+	tsStr := req.URL.Query().Get("ts")
+	if tsStr == "" {
+		return time.Now()
+	}
+
+	// The parameter exists, parse into a Time object
+	ts, err := strconv.ParseInt(tsStr, 10, 64)
+	if err != nil {
+		return time.Unix(ts/1000, 0)
+	}
+
+	return time.Unix(ts/1000, 0)
+}
--- a/src/github.com/matrix-org/dendrite/clientapi/routing/createroom.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/routing/createroom.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/matrix-org/dendrite/roomserver/api"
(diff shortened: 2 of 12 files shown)


src/github.com/matrix-org/dendrite/federationapi/routing/routing.go
```diff
@@ -165,7 +165,7 @@ func Setup(
 			roomID := vars["roomID"]
 			userID := vars["userID"]
 			return MakeJoin(
-				httpReq, request, cfg, query, roomID, userID,
+				httpReq.Context(), httpReq, request, cfg, query, roomID, userID,
```
anoadragon453: I feel like we should simplify this to just take `httpReq`...

src/github.com/matrix-org/dendrite/federationapi/routing/threepid.go
```diff
@@ -273,8 +274,8 @@ func buildMembershipEvent(
 	builder.AuthEvents = refs
 
 	eventID := fmt.Sprintf("$%s:%s", util.RandomString(16), cfg.Matrix.ServerName)
-	eventTime := common.ParseTSParam(req)
-	event, err := builder.Build(eventID, eventTime, cfg.Matrix.ServerName, cfg.Matrix.KeyID, cfg.Matrix.PrivateKey)
+	now := time.Now()
+	event, err := builder.Build(eventID, now, cfg.Matrix.ServerName, cfg.Matrix.KeyID, cfg.Matrix.PrivateKey)
```
anoadragon453: Above you put `time.Now()` directly in the function call. We should keep things consistent.

src/github.com/matrix-org/dendrite/federationapi/routing/routing.go
APwhitehat: true. Rooted out some more deadcode #571

src/github.com/matrix-org/dendrite/clientapi/httputil/parse.go
```diff
@@ -0,0 +1,38 @@
...
+	// Use the ts parameter's value for event time if present
+	tsStr := req.URL.Query().Get("ts")
+	if tsStr == "" {
+		return time.Now()
+	}
+
+	// The parameter exists, parse into a Time object
+	ts, err := strconv.ParseInt(tsStr, 10, 64)
+	if err != nil {
+		return time.Unix(ts/1000, 0)
```
krombel: In this case it might be possible that `ts` is not correctly parseable to an int. In that case a [0 is returned](https://golang.org/pkg/strconv/#ParseInt) which will make this command cause at least an exception.

src/github.com/matrix-org/dendrite/clientapi/routing/createroom.go
```diff
@@ -309,12 +310,10 @@ func createRoom(
 
 // buildEvent fills out auth_events for the builder then builds the event
 func buildEvent(
-	req *http.Request,
 	builder *gomatrixserverlib.EventBuilder,
 	provider gomatrixserverlib.AuthEventProvider,
-	cfg config.Dendrite,
+	cfg config.Dendrite, evTime time.Time,
```
krombel: Why don't you use a dedicated line for the new content?

src/github.com/matrix-org/dendrite/clientapi/routing/profile.go
```diff
@@ -258,10 +264,10 @@ func SetDisplayName(
 }
 
 func buildMembershipEvents(
-	req *http.Request,
+	ctx context.Context,
 	memberships []authtypes.Membership,
 	newProfile authtypes.Profile, userID string, cfg *config.Dendrite,
-	queryAPI api.RoomserverQueryAPI,
+	evTime time.Time, queryAPI api.RoomserverQueryAPI,
```
krombel: as above

src/github.com/matrix-org/dendrite/clientapi/httputil/parse.go
APwhitehat: I see.
@anoadragon453: Shouldn't we return current time when ts is not parsable ?

APwhitehat:
> I think we should error for invalid timestamps, but otherwise looks good.

Will track this in a new issue.

Issue #569, opened 2018-08-02
TypingEvent should be replaced by gomatrixserverlib.Event ?


APwhitehat:
https://github.com/matrix-org/dendrite/blob/1165b49da78f4e4c92e9cbd5f5a633528a304068/src/github.com/matrix-org/dendrite/typingserver/api/output.go#L21-L26

We should probably have an gomatrixserverlib.Event instead. Thoughts ?

anoadragon453:
Yep, I went through the same thing with ApplicationServiceEvent. Saves having to keep two things in sync.

PR #572, opened 2018-08-04
[Federation] Send typing events

APwhitehat:
* Consume typing events from typing server.
 * Send the events to all joined hosts in a room

P.S. dummy api is used to emulate the type of api.OutputTypingEvent.Event as gomatrixserverlib.Event, will be removed when #569 is fixed.


src/github.com/matrix-org/dendrite/typingserver/dummy/api/api.go
```diff
@@ -0,0 +1,23 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package api
+
+import "github.com/matrix-org/gomatrixserverlib"
+
+// TODO: Remove this package after, typingserver/api is updated to contain a gomatrixserverlib.Event
```
erikjohnston: We should probably move this TODO out of dummy package now that the plan is no longer to change to using Event

src/github.com/matrix-org/dendrite/typingserver/api/output.go
```diff
@@ -16,16 +16,14 @@ package api
 type OutputTypingEvent struct {
 	// The Event for the typing edu event.
 	Event TypingEvent `json:"event"`
+	// Users typing in the room at the event.
+	TypingUsers []string `json:"typing_users"`
```
erikjohnston: Can you expand the `OutputTypingEvent` doc comment to explain why we're sending both a TypingEvent and list of currently typing users?

APwhitehat:
Nuances with the current implementation (forgot to mention earlier):
 - federation EDU doesn't have the destination field. [link to travis's PR](https://287-24998719-gh.circle-artifacts.com/0/root/project/scripts/gen/server_server/unstable.html#edus).
It feels rather unused, since the (federation) transaction would have the destination. 
- processing edus does not increment the `destinationQueue.sendCounter`. I am not sure if this is correct.

@erikjohnston: Hope you see this before accepting this PR.

erikjohnston:
> federation EDU doesn't have the destination field. link to travis's PR.
It feels rather unused, since the (federation) transaction would have the destination.

I've just got travis to remove the `destination` and `origin` fields from his PR ftr

APwhitehat:
> processing edus does not increment the destinationQueue.sendCounter. I am not sure if this is correct.

Took a look at the origin of this, I believe this is used for logging only.
Since EDUs are short lived and can be a lot, we could ignore them in sendCounter.
Pretty easy to add if required.
/cc @erikjohnston

PR #576, opened 2018-08-06
Propagate error with wrong ?ts= param back to client

### LINKED ISSUE #575: Propagate error back to client
All the client APIs can have a `ts` parameter in the request which is used by AS (Application Services) to set the timestamp of the event.
If this parameter isn't a valid int, there is a parse error.
Currently the ParseTSParam ignores the error in parsing. here:
https://github.com/matrix-org/dendrite/blob/b11591fdd1606f29ee6da7f8cde01a49b12214a1/src/github.com/matrix-org/dendrite/clientapi/httputil/parse.go#L31-L35

Instead we need to return a `*util.JSONResponse` instead which acts an an error.
This can be sent back to client.
Then update the calls to `common. ParseTSParam` to check for returned error.

krombel:
Fixes #575

This as well introduces `ParseIntParam()` which might be used for other values as well.

Signed-Off-by: Matthias Kesler <krombel@krombel.de>

### DIFF
--- a/src/github.com/matrix-org/dendrite/clientapi/httputil/parse.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/httputil/parse.go
@@ -13,26 +13,49 @@
 package httputil
 
 import (
+	"fmt"
 	"net/http"
 	"strconv"
 	"time"
+
+	"github.com/matrix-org/dendrite/clientapi/jsonerror"
+	"github.com/matrix-org/util"
 )
+
+// ParseIntParam takes a request and parses the query param to an int
+// returns 0 when not present and -1 when parsing failed as int
+func ParseIntParam(req *http.Request, param string) (int64, *util.JSONResponse) {
+	paramStr := req.URL.Query().Get(param)
+	if paramStr == "" {
+		return 0, &util.JSONResponse{
+			Code: http.StatusBadRequest,
+			JSON: jsonerror.MissingArgument(fmt.Sprintf("Param %s not found", param)),
+		}
+	}
+	paramInt, err := strconv.ParseInt(paramStr, 0, 64)
+	if err != nil {
+		return -1, &util.JSONResponse{
+			Code: http.StatusBadRequest,
+			JSON: jsonerror.InvalidArgumentValue(
+				fmt.Sprintf("Param %s is no valid int (%s)", param, err.Error()),
+			),
+		}
+	}
+	return paramInt, nil
+}
 
 // ParseTSParam takes a req (typically from an application service) and parses a Time object
 // from the req if it exists in the query parameters. If it doesn't exist, the
 // current time is returned.
-func ParseTSParam(req *http.Request) time.Time {
-	// Use the ts parameter's value for event time if present
-	tsStr := req.URL.Query().Get("ts")
-	if tsStr == "" {
-		return time.Now()
-	}
-
-	// The parameter exists, parse into a Time object
-	ts, err := strconv.ParseInt(tsStr, 10, 64)
+func ParseTSParam(req *http.Request) (time.Time, *util.JSONResponse) {
+	ts, err := ParseIntParam(req, "ts")
 	if err != nil {
-		return time.Unix(ts/1000, 0)
+		if ts == 0 {
+			// param was not available
+			return time.Now(), nil
+		}
+		return time.Time{}, err
 	}
 
-	return time.Unix(ts/1000, 0)
+	return time.Unix(ts/1000, 0), nil
 }
--- a/src/github.com/matrix-org/dendrite/clientapi/httputil/parse.go
+++ b/src/github.com/matrix-org/dendrite/clientapi/httputil/parse.go
@@ -17,15 +17,12 @@ import (
 	"net/http"
 	"strconv"
 	"time"
-
-	"github.com/matrix-org/dendrite/clientapi/jsonerror"
-	"github.com/matrix-org/util"
 )
 
 // ParseTSParam takes a req (typically from an application service) and parses a Time object
 // from the req if it exists in the query parameters. If it doesn't exist, the
 // current time is returned.
-func ParseTSParam(req *http.Request) (time.Time, *util.JSONResponse) {
+func ParseTSParam(req *http.Request) (time.Time, error) {
 	// Use the ts parameter's value for event time if present
 	tsStr := req.URL.Query().Get("ts")
 	if tsStr == "" {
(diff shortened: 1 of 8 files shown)


src/github.com/matrix-org/dendrite/clientapi/httputil/parse.go
```diff
@@ -13,26 +13,49 @@
...
 	"strconv"
 	"time"
+
+	"github.com/matrix-org/dendrite/clientapi/jsonerror"
+	"github.com/matrix-org/util"
 )
 
+// ParseIntParam takes a request and parses the query param to an int
+// returns 0 when not present and -1 when parsing failed as int
+func ParseIntParam(req *http.Request, param string) (int64, *util.JSONResponse) {
```
APwhitehat: Will this be used somewhere ?

I'd rather not have unnecessary Public functions.

src/github.com/matrix-org/dendrite/clientapi/httputil/parse.go
```diff
@@ -13,26 +13,49 @@
...
+	paramStr := req.URL.Query().Get(param)
+	if paramStr == "" {
+		return 0, &util.JSONResponse{
```
APwhitehat: This case is not an error.

I'm guessing you wanted to make `ParseIntParam` do strict check for ts parameter ?

src/github.com/matrix-org/dendrite/clientapi/routing/joinroom.go
```diff
@@ -272,6 +276,12 @@ func (r joinRoomReq) joinRoomUsingServers(
 // server was invalid this returns an error.
 // Otherwise this returns a JSONResponse.
 func (r joinRoomReq) joinRoomUsingServer(roomID string, server gomatrixserverlib.ServerName) (*util.JSONResponse, error) {
+	// parse all data (and return an error) before doing some work
+	evTime, resErr := httputil.ParseTSParam(r.req)
+	if resErr != nil {
+		// this looks weird here but does what we want.
+		return resErr, resErr.JSON.(jsonerror.MatrixError)
```
APwhitehat: The error here is not due to communication with the other server, hence
`resErr, nil` makes more sense.

src/github.com/matrix-org/dendrite/clientapi/routing/profile.go
```diff
@@ -153,7 +158,7 @@ func SetAvatarURL(
 	}
 
 	events, err := buildMembershipEvents(
-		req.Context(), memberships, newProfile, userID, cfg, httputil.ParseTSParam(req), queryAPI,
```
APwhitehat: This older code is quite funny, thanks for correcting it.

src/github.com/matrix-org/dendrite/clientapi/routing/sendevent.go
```diff
@@ -35,35 +35,25 @@ type sendEventResponse struct {
 	EventID string `json:"event_id"`
 }
 
-// SendEvent implements:
-//   /rooms/{roomID}/send/{eventType}
-//   /rooms/{roomID}/send/{eventType}/{txnID}
-//   /rooms/{roomID}/state/{eventType}/{stateKey}
-func SendEvent(
+func generateSendEvent(
```
APwhitehat: Could you move this (helper) function after SendEvent ?
This pattern is usually followed in Dendrite (and hopefully other places ..)
Pattern being: placing the public function first then all it's helpers.

p.s. Inner me wants this function be named just `sendEvent`, but anything is fine. :P

src/github.com/matrix-org/dendrite/clientapi/httputil/parse.go
krombel: If I want to parse an int of nothing that is some kind of error. At least sth. I would want to return to the caller. I thought it would be a good approach to return 0 in that case.
So the calling function can distinguish from a "cannot parse" error.

src/github.com/matrix-org/dendrite/clientapi/routing/joinroom.go
krombel: I expect calling functions to check for the response of the `error` param to identify if there is an issue. This structure returns the prase which the creator of `resErr` intended to have there

src/github.com/matrix-org/dendrite/clientapi/routing/sendevent.go
krombel: Of course I can move it. I left it on top for a better readable diff.
But I would leave it as `generateSendEvent` to avoid name clashing. I would assume `generateEvent` might arise in other contexts as well.

src/github.com/matrix-org/dendrite/clientapi/httputil/parse.go
krombel: removed it now. To reintroduce it it is possible to revert https://github.com/matrix-org/dendrite/pull/576/commits/767ad6e36f9b2a5585006510d2ba18e3ac745078

src/github.com/matrix-org/dendrite/clientapi/routing/joinroom.go
```diff
@@ -207,13 +207,17 @@ func (r joinRoomReq) writeToBuilder(eb *gomatrixserverlib.EventBuilder, roomID s
 func (r joinRoomReq) joinRoomUsingServers(
 	roomID string, servers []gomatrixserverlib.ServerName,
 ) util.JSONResponse {
+	evTime, resErr := httputil.ParseTSParam(r.req)
+	if resErr != nil {
+		return *resErr
```
APwhitehat: Re https://github.com/matrix-org/dendrite/pull/576/files#discussion_r208269878

Here you have an error, yet you return it as the response.
This is okay, since the error has been handled and the caller doesn't need to.

With similar reasoning you could argue the you can just return `resErr, nil`

krombel: the ts-param parsing is now done before this method gets called. This is more clear and avoids having this confusion with MatrixError (from the linked discussion)

src/github.com/matrix-org/dendrite/clientapi/routing/joinroom.go
```diff
@@ -50,6 +51,10 @@ func JoinRoomByIDOrAlias(
 	if resErr := httputil.UnmarshalJSONRequest(req, &content); resErr != nil {
 		return *resErr
 	}
+	evTime, resErr := httputil.ParseTSParam(req)
```
APwhitehat: you may add an empty line above to maintain consistency.

src/github.com/matrix-org/dendrite/clientapi/routing/joinroom.go
```diff
@@ -65,7 +70,18 @@ func JoinRoomByIDOrAlias(
 	content["displayname"] = profile.DisplayName
 	content["avatar_url"] = profile.AvatarURL
 
-	r := joinRoomReq{req, content, device.UserID, cfg, federation, producer, queryAPI, aliasAPI, keyRing}
+	r := joinRoomReq{
```
APwhitehat: I would've preferred the single/double line initialisation, since
1) this is a private struct.
2) Field names are more or less exactly the same as variable names.
But I'll leave it upto you.

krombel: done

APwhitehat: Thanks :)

krombel:
It seems there is something flaky...
Here is a travis-ci run for the same commit-id (d32640b) which suceeds:
https://travis-ci.org/krombel/dendrite/builds/414168122

APwhitehat:
Hey, re-triggered the build. Travis tends to misbehave once in a while.

PR #583, opened 2018-11-06
Fix interactive registration failing

babolivier:
Registration using the User-Interactive auth was failing because it was being confused with AS registration, thus expecting an access token. This PR moves the check on whether a token is present earlier in the workflow and details the case in which we should consider the registration request as coming from an AS.


erikjohnston:
Except the test fail

babolivier:
I'm working on making the tests pass and the linter not complain ftr.

Issue #495, opened 2018-06-11
[federation] Implement backfill


APwhitehat:
GET _matrix/federation/v1/backfill/{roomID}

params: `v []event, limit int`

Response: 
```js
{
     pdus: []Event
}
```

babolivier:
Fixed in #585 (forgot the "fixes" mention there, sorry about that)

PR #585, opened 2018-11-06
Implement the /backfill federation endpoint

babolivier:
The comments I made might be a bit messy, I'll have another look at them tomorrow.


src/github.com/matrix-org/dendrite/roomserver/query/query.go
```diff
@@ -487,6 +487,80 @@ BFSLoop:
 	return err
 }
 
+// QueryPreviousEvents implements api.RoomServerQueryAPI
+func (r *RoomserverQueryAPI) QueryPreviousEvents(
```
erikjohnston: `QueryPreviousEvents` feels a little overly generic. Can we call it like `QueryForBackfill` or something? I'm expecting eventually we'll want to optimise the query to be more efficient (BFS is fine for now, but isn't the best)

APwhitehat:
(self) note: This fixes #495.

PR #588, opened 2018-11-07
Fix the ordering of events in a response to /sync

babolivier:
A request to `/sync` returns the list of the oldest `n` events that happened in rooms from a specific position in the stream while it should be the opposite (return the `n` latest events that happened later than that position).

### DIFF
--- a/src/github.com/matrix-org/dendrite/syncapi/storage/output_room_events_table.go
+++ b/src/github.com/matrix-org/dendrite/syncapi/storage/output_room_events_table.go
@@ -17,6 +17,7 @@ package storage
 import (
 	"context"
 	"database/sql"
+	"sort"
 
 	"github.com/matrix-org/dendrite/roomserver/api"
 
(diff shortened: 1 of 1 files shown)


src/github.com/matrix-org/dendrite/syncapi/storage/output_room_events_table.go
```diff
@@ -234,6 +235,12 @@ func (s *outputRoomEventsStatements) selectRecentEvents(
 	if err != nil {
 		return nil, err
 	}
+	// The events need to be returned from oldest to latest, which isn't
+	// necessary the way the SQL query returns them, so a sort is necessary to
+	// ensure the events are in the right order in the slice.
```
erikjohnston: Is this actually true, given we've go the `ORDER BY id DESC` in the SQL? Or is `rowsToStreamEvents` messing with the order?

babolivier: If we consider an example stream with positions `1`, `2` and `3` (and an event on each of these positions), switching to a `DESC` ordering will make the slice be `[eventAtPos(3), eventAtPos(2), eventAtPos(1)]`, but the expected output is `[eventAtPos(1), eventAtPos(2), eventAtPos(3)]`, so we need to sort the slice *because* we switched to a `DESC` ordering.

PR #589, opened 2018-11-08
Fill the prev_batch property in responses from /sync

babolivier:
For each room, fill the `prev_batch` property with the position in the stream of the earliest event in the response -1. If that equals to 0 (which wouldn't match anything since stream positions start at 1), uses the position of the earliest event in the response (which is, as far as I've observed, what synapse does).

### DIFF
--- a/src/github.com/matrix-org/dendrite/syncapi/storage/syncserver.go
+++ b/src/github.com/matrix-org/dendrite/syncapi/storage/syncserver.go
@@ -310,6 +310,11 @@ func (d *SyncServerDatabase) CompleteSync(
 
 		stateEvents = removeDuplicates(stateEvents, recentEvents)
 		jr := types.NewJoinResponse()
+		if prevBatch := recentStreamEvents[0].streamPosition - 1; prevBatch > 0 {
+			jr.Timeline.PrevBatch = types.StreamPosition(prevBatch).String()
+		} else {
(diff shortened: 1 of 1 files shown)


src/github.com/matrix-org/dendrite/syncapi/storage/syncserver.go
```diff
@@ -310,6 +310,11 @@ func (d *SyncServerDatabase) CompleteSync(
 
 		stateEvents = removeDuplicates(stateEvents, recentEvents)
 		jr := types.NewJoinResponse()
+		if prevBatch := recentStreamEvents[0].streamPosition - 1; prevBatch > 0 {
+			jr.Timeline.PrevBatch = types.StreamPosition(prevBatch).String()
+		} else {
+			jr.Timeline.PrevBatch = recentStreamEvents[0].streamPosition.String()
```
erikjohnston: It's probably clearer to explicitly set this to `types.StreamPosition(0).String()` or whatever, otherwise it feels a bit confusing.

babolivier: Fixed in https://github.com/matrix-org/dendrite/pull/589/commits/ac0a928702853a17b36c793f44c778b6d883ebda. I used `types.StreamPosition(0).String()` as the stream position sequence starts at 1.

PR #593, opened 2018-11-12
use go module for vendoring

rbns:
this is unfortunately a giant-commit due to the removal of the vendor folder.
in a nutshell:

- `go mod init` uses the contents of vendor/manifest to initialize go.mod
- `git mv src/github/matrix-org/dendrite/* .` to allow building with default go tools
- `git rm -r vendor` remove vendors
- `for f in $(ls); do cd $f; go build ; cd ..; echo $f; done` in folder cmd to build every command to fetch the dependency sources, this also creates go.sum (checksums of deps).

i don't expect this being merged, just my 2 cents of how to convert to go modules


anoadragon453:
/me wonders about implications of `import "github.com/matrix-org/dendrite/github.com/alecthomas/gometalinter"` lines

rbns:
replaced by #594

PR #595, opened 2018-12-05
Fix the sync api returning an empty sync response when reaching timeout, regardless of the since token

babolivier:
Also got rid of the loop which didn't make much sense to me.

### DIFF
--- a/src/github.com/matrix-org/dendrite/syncapi/sync/requestpool.go
+++ b/src/github.com/matrix-org/dendrite/syncapi/sync/requestpool.go
@@ -84,38 +84,33 @@ func (rp *RequestPool) OnIncomingSyncRequest(req *http.Request, device *authtype
 	userStreamListener := rp.notifier.GetListener(*syncReq)
 	defer userStreamListener.Close()
 
-	for {
-		select {
-		// Wait for notifier to wake us up
-		case <-userStreamListener.GetNotifyChannel(currPos):
-			currPos = userStreamListener.GetStreamPosition()
-		// Or for timeout to expire
-		case <-timer.C:
-			return util.JSONResponse{
-				Code: http.StatusOK,
-				JSON: types.NewResponse(currPos),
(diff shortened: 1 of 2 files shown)


babolivier:
FTR this fixes all of sytest's tests that have to do with room creation (and possibly others).

src/github.com/matrix-org/dendrite/syncapi/sync/requestpool.go
```diff
@@ -84,38 +86,33 @@ func (rp *RequestPool) OnIncomingSyncRequest(req *http.Request, device *authtype
 	userStreamListener := rp.notifier.GetListener(*syncReq)
 	defer userStreamListener.Close()
 
-	for {
```
erikjohnston: We probably don't want to remove the loop here, as there will be times when `userStreamListener` wakes up even if there isn't anything to send down. I think we should just add the loop back, and add a check like `if !syncData.IsEmpty() || hasTimedOut { return ... }` or whatever

babolivier: Hah. The reason I removed it was because I didn't know what it was doing there, but that makes sense, I'll add it back.

erikjohnston: Ah yeah, it should absolutely have had a comment there! (I'm somewhat reminded of this quote: https://www.chesterton.org/taking-a-fence-down/ )

babolivier: (that's a great quote)

babolivier: Fixed in https://github.com/matrix-org/dendrite/pull/595/commits/deb181ea500267ada816745472584892d71cd426

babolivier:
:tada:

Issue #607, opened 2019-01-08
Response to /rooms/{roomId}/join should contain the joined room's ID


babolivier:
If the user successfully joined the room, we currently respond with an empty object. This object should have a `room_id` attribute containing the room's ID, [as described in the specs](https://matrix.org/docs/spec/client_server/r0.4.0.html#post-matrix-client-r0-rooms-roomid-join).

Cnly:
Closed by #734.

PR #683, opened 2019-01-17
Fix response to /rooms/{roomId}/join

BEHOUBA:
when user successfully joined the room, he receive an object having a room_id attribute containing the room's ID, [ described in the specs](https://matrix.org/docs/spec/client_server/r0.4.0.html#post-matrix-client-r0-rooms-roomid-join).
Also missing letter 'n' added in comment inside joinRoomByID method of joinRoomReq struct.

This PR is for #607 
Signed-off-by: Kouame Behouba Manasse behouba@gmail.com


BEHOUBA:
> Amazing, thanks! Could you please make it so that the CI is happy? (I see some linting errors there)
Ok, i will work on that. thanks for reviewing :+1:

BEHOUBA:
@babolivier since the CI is complaining about the complexity of SendMembership function, i wonder if it is a good idea to make the check of the status and the membership inside the handler before returning the JSON response.

PR #684, opened 2019-01-18
Fix response to /rooms/{roomId}/join

### LINKED ISSUE #607: Response to /rooms/{roomId}/join should contain the joined room's ID

behouba:
@babolivier since the CI is complaining about the complexity of SendMembership function, i wonder if it is a good idea to make the check of the status code and the membership inside the handler before returning the JSON response as i did here.
This should fix #607

### DIFF
--- a/clientapi/routing/filter.go
+++ b/clientapi/routing/filter.go
@@ -62,7 +62,7 @@ func GetFilter(
 	filter := gomatrix.Filter{}
 	err = json.Unmarshal(res, &filter)
 	if err != nil {
-		httputil.LogThenError(req, err)
+		return httputil.LogThenError(req, err)
 	}
 
 	return util.JSONResponse{
--- a/clientapi/routing/login.go
+++ b/clientapi/routing/login.go
@@ -107,7 +107,7 @@ func Login(
 
 		token, err := auth.GenerateAccessToken()
 		if err != nil {
-			httputil.LogThenError(req, err)
+			return httputil.LogThenError(req, err)
 		}
 
 		dev, err := getDevice(req.Context(), r, deviceDB, acc, localpart, token)


babolivier:
I don't really like this idea, and would like to avoid sending responses directly in `routing.go` as much as possible.

Another solution would be to move [this code](https://github.com/matrix-org/dendrite/blob/9329b571bbecd76ccdfa5f991c8d66f2ab408af7/src/github.com/matrix-org/dendrite/clientapi/routing/membership.go#L61-L82) out of `SendMembership` into a small, separate function. The function would then return `inviteStored`, and a `*util.JSONResponse` which would be `nil` if there's no need to respond from there. This way, 

```go
inviteStored, err := threepid.CheckAndProcessInvite(
	req.Context(), device, &body, cfg, queryAPI, accountDB, producer,
	membership, roomID, evTime,
)
// Error handling
```

would then look like:

```go
inviteStored, errRes := checkAndProcessThreepid(
	req, device, &body, cfg, queryAPI, accountDB, producer,
	membership, roomID, evTime,
)
if errRes != nil {
	return *errRes
}
```

This would reduce the complexity of `SendMembership` enough to make the linter no longer complain about it, and might also make it more readable :slightly_smiling_face: 

Btw, my example is, well, a quickly-drafted, improvised example, feel free to address this the way you prefer (though I'd prefer it not to bleed out on `routing.go` as much as possible).

Also btw, feel free to join the [#dendrite-dev:matrix.org](https://matrix.to/#/#dendrite-dev:matrix.org) room, as it would be the best place to voice concerns or questions about what your PR should look like, instead of recreating it everytime :wink:

behouba:
copy that! thanks @babolivier !

behouba:
You were right @babolivier. My first commit was not very clean so I did it as you suggested, have a look.

Cnly:
@behouba I think this generally looks good, but can you rebase it onto master so we don't have conflicts? Thanks!

anoadragon453:
This has been superseded by https://github.com/matrix-org/dendrite/pull/734.

PR #685, opened 2019-01-23
Content-Disposition HTTP header in mediaapi's responses added

### LINKED ISSUE #628: Use Content-Disposition HTTP header in mediaapi's responses
The media API needs to return a `Content-Disposition` header in responses for both [GET /_matrix/media/r0/download/{serverName}/{mediaId}](https://matrix.org/docs/spec/client_server/r0.4.0.html#get-matrix-media-r0-download-servername-mediaid) and [GET /_matrix/media/r0/download/{serverName}/{mediaId}/{fileName}](https://matrix.org/docs/spec/client_server/r0.4.0.html#get-matrix-media-r0-download-servername-mediaid-filename).

This header must contain the file name and its encoding such as `inline; filename*=utf-8"foo.png"`.

It must be omitted if the file has been uploaded with no name.

behouba:
This should fix #628 
I also Noticed that [GET /_matrix/media/r0/download/{serverName}/{mediaId}/{fileName}](https://matrix.org/docs/spec/client_server/r0.4.0.html#get-matrix-media-r0-download-servername-mediaid-filename) is not yet implemented, but it should work for both.

Signed-off-by: Kouamé Behouba Manassé <behouba@gmail.com>


Cnly:
I think this looks good and just did a simple merge...

PR #688, opened 2019-01-27
Remove riot.im from the list of trusted ID servers in the default config

aaronraimist:
### Pull Request Checklist

* [x] I have made sure any new dependencies have been checked into the `vendor/` directory
* [x] Pull request includes a [sign off](/CONTRIBUTING.rst#sign-off)

Synapse removed it here https://github.com/matrix-org/synapse/commit/78ba0e7ab8f1d57eee6aa56e9d496d838a24c6f3#diff-e23363fca80aa05eb0f8068551166e94 because it doesn't actually exist and was causing issues with deactivating accounts. Maybe Dendrite's way of doing things allows it to be here even though it doesn't exist in which case feel free to close this.

Also the comment above it seems to lie. "Defaults to no trusted servers.", isn't this the list of default trusted ID servers?

### DIFF
--- a/dendrite-config.yaml
+++ b/dendrite-config.yaml
@@ -17,7 +17,6 @@ matrix:
     trusted_third_party_id_servers:
       - vector.im
       - matrix.org
-      - riot.im
 
 # The media repository config
 media:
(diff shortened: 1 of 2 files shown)


babolivier:
> Also the comment above it seems to lie. "Defaults to no trusted servers.", isn't this the list of default trusted ID servers?

This configuration file is a template users should copy and adapt to their needs (though it needs to be more obvious). The "Defaults to XXX" defines what happens if this key is omitted from the configuration.

PR #693, opened 2019-03-10
Implement client single event retrieval - #671

### LINKED ISSUE #671: Implement single event retrieval
Full specs: https://matrix.org/docs/spec/client_server/r0.4.0.html#get-matrix-client-r0-rooms-roomid-event-eventid

Cnly:
This should fix #671.
Implemented spec: https://matrix.org/docs/spec/client_server/r0.4.0.html#get-matrix-client-r0-rooms-roomid-event-eventid

When the event requested is not found locally, a federation query is done.


src/github.com/matrix-org/dendrite/clientapi/routing/getevent.go
```diff
@@ -0,0 +1,151 @@
+// Copyright 2017 Vector Creations Ltd
```
anoadragon453: It's 2019! Also please put yourself here instead of Vector Creations Ltd, as you've written the code :)

Cnly: Ah, they've been changed. Thanks!

src/github.com/matrix-org/dendrite/clientapi/routing/getevent.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/matrix-org/dendrite/clientapi/jsonerror"
+	"github.com/matrix-org/dendrite/common/config"
+	"github.com/matrix-org/dendrite/roomserver/api"
+	"github.com/matrix-org/gomatrixserverlib"
+	"github.com/matrix-org/util"
+	"net/http"
+)
+
+var (
+	unknownEventResponse = util.JSONResponse{
```
APwhitehat: This style (global variables) is not followed in this project. Repetition is okay considering that we would like to give specific error msg which may be different sometimes.
However you may as well change this to a function and that would be fine as well.

src/github.com/matrix-org/dendrite/clientapi/routing/getevent.go
```diff
@@ -0,0 +1,151 @@
...
+		// Missing some events locally so stateResp.StateEvents will be unavailable.
+		// Do a federation query in hope of getting the state events needed.
+		return r.proceedWithMissingState()
+	}
+
+	return r.proceedWithStateEvents(stateResp.StateEvents)
+}
+
+func (r *getEventRequest) proceedWithMissingState() util.JSONResponse {
+	// It's not guaranteed that the server we query has these events.
```
APwhitehat: This comment refers to this entire function.
So I would prefer this to be replaced by direct comment on top of the function.

src/github.com/matrix-org/dendrite/clientapi/routing/getevent.go
```diff
@@ -0,0 +1,151 @@
...
+	if err != nil {
+		return httputil.LogThenError(req, err)
+	}
+
+	if len(eventsResp.Events) == 0 {
+		// TODO: Event not found locally. May need a federation query here.
+		return unknownEventResponse
+	}
+
+	r := getEventRequest{
```
APwhitehat: Is this really necessary, since this struct is only ever used by GetEvent?
It seems that all this could easily be done by passing relevant arguments to functions.

Cnly: For this struct, I followed the style in [`federationapi/routing/send.go`](https://github.com/matrix-org/dendrite/blob/master/src/github.com/matrix-org/dendrite/federationapi/routing/send.go#L73). I originally put the things tighter in one function, but the code was reported by the linter as having too high cyclomatic complexity.

APwhitehat: I proposed that the struct methods rather be funtions to which you pass relevant arguments.
Then you could do away with the struct. :)

Cnly: I got what you mean. I was just guessing why things in `send.go` are divided into different functions. I'm unsure why the team didn't do it the way you mentioned, though.
Then I'll mark this as resolved. :)

Cnly:
This PR is now complete and ready for review.

clientapi/routing/routing.go
```diff
@@ -116,6 +116,12 @@ func Setup(
 				nil, cfg, queryAPI, producer, transactionsCache)
 		}),
 	).Methods(http.MethodPut, http.MethodOptions)
+	r0mux.Handle("/rooms/{roomID}/event/{eventID}",
+		common.MakeAuthAPI("room_get_event", authData, func(req *http.Request, device *authtypes.Device) util.JSONResponse {
+			vars := mux.Vars(req)
```
anoadragon453: We do this silly thing now for URL encoding purposes:

https://github.com/matrix-org/dendrite/blob/c72517687878267a021641a689fdf73fe6d1c127/clientapi/routing/routing.go#L97

clientapi/routing/routing.go
```diff
@@ -116,6 +116,12 @@ func Setup(
 				nil, cfg, queryAPI, producer, transactionsCache)
 		}),
 	).Methods(http.MethodPut, http.MethodOptions)
+	r0mux.Handle("/rooms/{roomID}/event/{eventID}",
+		common.MakeAuthAPI("room_get_event", authData, func(req *http.Request, device *authtypes.Device) util.JSONResponse {
```
anoadragon453:
```suggestion
		common.MakeAuthAPI("rooms_get_event", authData, func(req *http.Request, device *authtypes.Device) util.JSONResponse {
```

clientapi/routing/getevent.go
```diff
@@ -0,0 +1,178 @@
...
+	if err != nil {
+		return httputil.LogThenError(req, err)
+	}
+
+	var requestedEvent gomatrixserverlib.Event
+	if len(eventsResp.Events) == 0 {
+		// Event not found locally. Do a federation query in hope of getting
+		// the event from another server.
+		// TODO: May need a better way to determine which server to query
+		_, domain, err := gomatrixserverlib.SplitID('$', eventID)
```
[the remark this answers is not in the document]
Cnly: Hmmm, then I think I'll need to change this back to a `TODO` for now.

clientapi/routing/getevent.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	var stateResp api.QueryStateAfterEventsResponse
+	if err := queryAPI.QueryStateAfterEvents(req.Context(), &stateReq, &stateResp); err != nil {
+		return httputil.LogThenError(req, err)
+	}
+
+	if !stateResp.RoomExists {
+		return util.JSONResponse{
+			Code: http.StatusNotFound,
+			JSON: jsonerror.NotFound("The event was not found or you do not have permission to read this event."),
```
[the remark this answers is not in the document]
Cnly: Indeed, it should be found now. According to the previous conversation we'll remove federation event retrieval for the endpoint, then `!stateResp.RoomExists` would mean Dendrite doesn't know about the room the *local* event is in, which seems quite unlikely and problematic. Maybe a `LogThenError` will be better here?

clientapi/routing/getevent.go
```diff
@@ -0,0 +1,178 @@
...
+	return r.proceedWithStateEvents(state.StateEvents)
+}
+
+func (r *getEventRequest) proceedWithStateEvents(stateEvents []gomatrixserverlib.Event) util.JSONResponse {
+	allowed := false
+	for _, stateEvent := range stateEvents {
+		if stateEvent.StateKeyEquals(r.device.UserID) {
+			membership, err := stateEvent.Membership()
+			if err == nil && membership == "join" {
+				allowed = true
```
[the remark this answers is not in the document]
Cnly: We can't use `QueryMembershipForUser` before we want the state at a specific event, not the current state.

But I think there's a new problem after I read the implementation of `QueryStateAfterEvents` again: If the user wants to get a `leave` event for themselves, then the state after that event will be the user is not in the room, and the request will thus be denied. So we probably need to get the state *at* the event instead of *after* the event.

clientapi/routing/getevent.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	return r.proceedWithStateEvents(stateResp.StateEvents)
+}
+
+// proceedWithMissingState tries to proceed by fetching the missing states with
+// federation.
+// Note: It's not guaranteed that the server(s) we query have the state events.
+func (r *getEventRequest) proceedWithMissingState() util.JSONResponse {
+	_, domain, err := gomatrixserverlib.SplitID('$', r.eventID)
```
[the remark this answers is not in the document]
Cnly: > Again, this won't work with new event IDs.

I'll use room ID and that should solve the problem.

> Synapse takes quite a simple approach to all this

The [function](https://github.com/matrix-org/synapse/blob/463b072b1290a8cb75bc1d7b6688fa76bbfdb14f/synapse/visibility.py#L45) called in these lines doesn't look simple though... But Synapse isn't trying to get missing state over federation at all in the endpoint. Should we remove that here too?

clientapi/routing/getevent.go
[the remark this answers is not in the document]
anoadragon453: Yes, `LogThenError` sounds sensible.

clientapi/routing/getevent.go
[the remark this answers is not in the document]
anoadragon453: I think so. There's also no guarantee that the server that created the room (and thus that would be in the roomID) even exists anymore. RoomIDs are supposed to be opaque identifiers.

There are other ways to find out what servers are in a room and to request state from them. I don't think in this method is where it should happen though.

clientapi/routing/getevent.go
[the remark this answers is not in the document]
anoadragon453: Interestingly I just tested the above edge case with Synapse and it will 403 you from retrieving the leave event since you aren't in the room...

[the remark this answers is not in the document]
Cnly: The spec doesn't say anything specific about this, so let's keep it simple for now and go Synapse's way.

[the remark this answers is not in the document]
Cnly: Oh, it seems `QueryMembershipForUser` still isn't the best because we need to ensure the event isn't before the user's join event.

Cnly:
The linting [issue](https://buildkite.com/matrix-dot-org/dendrite/builds/116#63d4de53-fe35-4475-adf5-c01627daced3) for this PR is fixed in #752, so we can see if that can be merged first.

Cnly:
> This is getting there, but some of the keys returned are reserved for federation usage, and shouldn't be returned to the client.

[`gomatrixserverlib.ToClientEvent`](https://github.com/matrix-org/gomatrixserverlib/blob/master/clientevent.go#L52) should help.

Cnly:
Re-requesting review from @anoadragon453. Doing this manually because GitHub keeps giving me 404 when I click the re-request button...

clientapi/routing/getevent.go
```diff
@@ -0,0 +1,128 @@
...
+	err := queryAPI.QueryEventsByID(req.Context(), &eventsReq, &eventsResp)
+	if err != nil {
+		return httputil.LogThenError(req, err)
+	}
+
+	if len(eventsResp.Events) == 0 {
+		// Event not found locally
+		return util.JSONResponse{
+			Code: http.StatusNotFound,
+			JSON: jsonerror.NotFound("The event was not found or you do not have permission to read this event."),
```
anoadragon453:
```suggestion
			JSON: jsonerror.NotFound("The event was not found or you do not have permission to read this event"),
```

clientapi/routing/getevent.go
```diff
@@ -0,0 +1,128 @@
...
+	if !stateResp.RoomExists {
+		util.GetLogger(req.Context()).Errorf("Expected to find room for event %s but failed", r.requestedEvent.EventID())
+		return jsonerror.InternalServerError()
+	}
+
+	if !stateResp.PrevEventsExist {
+		// Missing some events locally; stateResp.StateEvents unavailable.
+		return util.JSONResponse{
+			Code: http.StatusNotFound,
+			JSON: jsonerror.NotFound("The event was not found or you do not have permission to read this event."),
```
anoadragon453:
```suggestion
			JSON: jsonerror.NotFound("The event was not found or you do not have permission to read this event"),
```

clientapi/routing/getevent.go
```diff
@@ -0,0 +1,128 @@
...
+		}
+	}
+
+	allowed := false
+	for _, stateEvent := range stateResp.StateEvents {
+		if stateEvent.StateKeyEquals(r.device.UserID) {
+			membership, err := stateEvent.Membership()
+			if err == nil && membership == "join" {
+				allowed = true
```
anoadragon453: I think you can just move lines 118-121 into this `if` statement instead of using `allowed`?

clientapi/routing/getevent.go
```diff
@@ -0,0 +1,128 @@
...
+	if allowed {
+		return util.JSONResponse{
+			Code: http.StatusOK,
+			JSON: gomatrixserverlib.ToClientEvent(r.requestedEvent, gomatrixserverlib.FormatAll),
+		}
+	}
+
+	return util.JSONResponse{
+		Code: http.StatusNotFound,
+		JSON: jsonerror.NotFound("The event was not found or you do not have permission to read this event."),
```
anoadragon453: When I test this out, I find that I am able to retrieve events that I've sent, but not events that have been sent by others in the same room. This is where the 404 is returned from.

clientapi/routing/getevent.go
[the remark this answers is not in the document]
Cnly: Previously I talked about the edge case if we use `QueryStateAfterEvents`. That problem should be nonexistent because `QueryStateAfterEvents` queries the state after _prevEvents_, which means the state at the _requested_ event.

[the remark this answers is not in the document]
Cnly: But Synapse should probably fix its 403. From the [spec](https://matrix.org/docs/spec/client_server/r0.5.0#post-matrix-client-r0-rooms-roomid-leave) for `/leave`:

> The user will still be allowed to retrieve history from the room which they were previously allowed to see.

Cnly:
Umm, now we have yet another `string `join` has 3 occurrences, make it a constant`.

clientapi/routing/getevent.go
Cnly: This should have been fixed now.

Cnly:
Linting for this PR won't pass perhaps until #765.
Three tests are blocked from passing by https://github.com/matrix-org/sytest/issues/653 and https://github.com/matrix-org/sytest/issues/652.

clientapi/routing/getevent.go
```diff
@@ -0,0 +1,127 @@
...
+		cfg:            cfg,
+		federation:     federation,
+		keyRing:        keyRing,
+		requestedEvent: requestedEvent,
+	}
+
+	stateReq := api.QueryStateAfterEventsRequest{
+		RoomID:       r.requestedEvent.RoomID(),
+		PrevEventIDs: r.requestedEvent.PrevEventIDs(),
+		StateToFetch: []gomatrixserverlib.StateKeyTuple{{
```
anoadragon453: Is this supposed to be `{{`?

clientapi/routing/getevent.go
```diff
@@ -0,0 +1,127 @@
...
+		device:         device,
+		roomID:         roomID,
+		eventID:        eventID,
+		cfg:            cfg,
+		federation:     federation,
+		keyRing:        keyRing,
+		requestedEvent: requestedEvent,
+	}
+
+	stateReq := api.QueryStateAfterEventsRequest{
```
anoadragon453: All of this event retrieval from the DB along with auth checking seems like it could be useful in other parts of Dendrite and go in its own function. What do you think?

clientapi/routing/getevent.go
```diff
@@ -0,0 +1,127 @@
...
+		}
+	}
+
+	for _, stateEvent := range stateResp.StateEvents {
+		if stateEvent.StateKeyEquals(r.device.UserID) {
+			membership, err := stateEvent.Membership()
+			if err != nil {
+				return httputil.LogThenError(req, err)
+			}
+			if membership == "join" {
```
anoadragon453: This can be updated to gomatrixserverlib's constant now.

clientapi/routing/getevent.go
```diff
@@ -0,0 +1,127 @@
...
+					Code: http.StatusOK,
+					JSON: gomatrixserverlib.ToClientEvent(r.requestedEvent, gomatrixserverlib.FormatAll),
+				}
+			}
+		}
+	}
+
+	return util.JSONResponse{
+		Code: http.StatusNotFound,
+		JSON: jsonerror.NotFound("The event was not found or you do not have permission to read this event."),
```
anoadragon453:
```suggestion
		JSON: jsonerror.NotFound("The event was not found or you do not have permission to read this event"),
```

It might be nice to consolidate this error text as well, lest we want to change it.

clientapi/routing/getevent.go
Cnly: We only want the `m.room.member` event for `device.UserID` to check the user's membership, so there's indeed only one event to be fetched, if that's what you're asking. Or talking about coding style, I think `{{` here is better than putting the second `{` on the next line as the array only has one thing.

anoadragon453: Ahh, I wasn't familiar with this syntax. I see now that you're creating a `gomatrixserverlib.StateKeyTuple` inside a slice of them. I first read it as just creating a `gomatrixserverlib.StateKeyTuple` and was confused why you were putting an object inside of one.

clientapi/routing/getevent.go
Cnly: But it seems there're several bits that need to be customized when the function is called, e.g. the actions on `!stateResp.RoomExists`, `!stateResp.PrevEventsExist`. How should these fit in?

anoadragon453: Hmm, yeah, you could do it with returning custom error types but then at that point I question if doing all the error checking would be worse than just calling the functions directly.

I also looked through Synapse expecting it to have such a function, but while it has [one for retrieving state events in a room](https://github.com/matrix-org/synapse/blob/cf89266b980b62a6d8547f8e1ae9394359a05fc8/synapse/handlers/message.py#L93), it's only used in one place.

So yeah, until we actually start having duplicated code lets leave this as it is for now.

PR #702, opened 2019-05-21
rommserver/alias: Do not call appserviceAPI in GetRoomIDForAlias if local alias found #631

### LINKED ISSUE #631: Fix alias lookup in roomserver
Currently, the roomserver:

* looks up the alias in the database
* query the appservices to try to find one in charge of creating the alias

This isn't really useful as even if an appservice creates the alias we don't look it up afterwards. On top of that, we should query appservices only if no alias was found.

My take on that would be to change [GetRoomIDForAlias](https://github.com/matrix-org/dendrite/blob/master/src/github.com/matrix-org/dendrite/roomserver/alias/alias.go#L88) this way:

```go
// GetRoomIDForAlias implements alias.RoomserverAliasAPI
func (r *RoomserverAliasAPI) GetRoomIDForAlias(
	ctx context.Context,
	request *roomserverAPI.GetRoomIDForAliasRequest,
	response *roomserverAPI.GetRoomIDForAliasResponse,
) error {
	// Look up the room ID in the database
	roomID, err := r.DB.GetRoomIDForAlias(ctx, request.Alias)
	if err != nil {
		return err
	}

	if len(roomID) == 0 {
		// No rooms found locally, try our application services by making a call to
		// the appservice component
		aliasReq := appserviceAPI.RoomAliasExistsRequest{Alias: request.Alias}
		var aliasResp appserviceAPI.RoomAliasExistsResponse
		if err := r.AppserviceAPI.RoomAliasExists(ctx, &aliasReq, &aliasResp); err != nil {
			return err
		}

		// Look up the room ID in the database
		roomID, err = r.DB.GetRoomIDForAlias(ctx, request.Alias)
		if err != nil {
			return err
		}
	}

	response.RoomID = roomID
	return nil
}
```

serra-allgood:
### Pull Request Checklist

* [x] I have made sure any new dependencies have been checked into the `vendor/` directory

### Changes
A conditional is added to wrap the call to appserviceAPI if a local alias is not found in the database.

Fixes #631


roomserver/alias/alias.go
```diff
@@ -96,12 +96,14 @@ func (r *RoomserverAliasAPI) GetRoomIDForAlias(
 		return err
 	}
 
-	// No rooms found locally, try our application services by making a call to
-	// the appservice component
-	aliasReq := appserviceAPI.RoomAliasExistsRequest{Alias: request.Alias}
-	var aliasResp appserviceAPI.RoomAliasExistsResponse
-	if err = r.AppserviceAPI.RoomAliasExists(ctx, &aliasReq, &aliasResp); err != nil {
-		return err
+	if len(roomID) == 0 {
```
anoadragon453: I might make this `if roomID == ""` just to be clear that roomID is a string versus a Slice that could also have zero length.

roomserver/alias/alias.go
```diff
@@ -96,12 +96,14 @@ func (r *RoomserverAliasAPI) GetRoomIDForAlias(
...
+		// No rooms found locally, try our application services by making a call to
+		// the appservice component
+		aliasReq := appserviceAPI.RoomAliasExistsRequest{Alias: request.Alias}
+		var aliasResp appserviceAPI.RoomAliasExistsResponse
+		if err = r.AppserviceAPI.RoomAliasExists(ctx, &aliasReq, &aliasResp); err != nil {
+			return err
+		}
 	}
 
 	response.RoomID = roomID
```
anoadragon453: If `aliasResp.RoomExists` is `true` we'll want to check again for whether the roomID exists in the db. That's because when we ask the application service whether the room exists, it can create one in the background before answering the request "yes, it exists! (now)".

So we'll need to check again after it does that. More info: https://matrix.org/docs/spec/application_service/unstable.html#get-matrix-app-v1-rooms-roomalias

roomserver/alias/alias_test.go
```diff
@@ -0,0 +1,90 @@
+// Copyright 2017 Vector Creations Ltd
```
anoadragon453: Can you update this to the current year and your name please? :)

roomserver/alias/alias_test.go
```diff
@@ -0,0 +1,90 @@
...
+) error {
+	return nil
+}
+
+func (q MockAppServiceQueryAPI) RoomAliasExists(
+	ctx context.Context,
+	req *appserviceAPI.RoomAliasExistsRequest,
+	resp *appserviceAPI.RoomAliasExistsResponse,
+) error {
+	return fmt.Errorf("Should not have called this")
```
anoadragon453: Would be nice to also have a test case for the Room ID not existing, for which `GetRoomIDForAlias` will then call  `AppServiceQueryAPI.RoomAliasExists". Also checking how everything works when the appservice responses with `RoomExists: False`. would be useful.

roomserver/alias/alias_test.go
serra-allgood: Whoops, new to contributing like this!

roomserver/alias/alias.go
serra-allgood: ok, cool. If `aliasResp.RoomExists` is `false`, should this method return an error instead of setting `response.RoomID` to an empty string?

serra-allgood: Nvm, just answered my own question by reading the method's callers

serra-allgood:
@anoadragon453 Changes made!

roomserver/alias/alias_test.go
```diff
@@ -0,0 +1,187 @@
...
+	"strings"
+	"testing"
+
+	appserviceAPI "github.com/matrix-org/dendrite/appservice/api"
+	roomserverAPI "github.com/matrix-org/dendrite/roomserver/api"
+)
+
+type MockRoomserverAliasAPIDatabase struct {
+	mode string
+	attempts    int
```
anoadragon453: Could you fix the spacing here please?

roomserver/alias/alias.go
```diff
@@ -84,24 +84,46 @@ func (r *RoomserverAliasAPI) SetRoomAlias(
 	return r.sendUpdatedAliasesEvent(context.TODO(), request.UserID, request.RoomID)
 }
 
+func getRoomIDFromDB(
```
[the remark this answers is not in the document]
serra-allgood: Ah, good point! Knee jerk attempt at DRY

serra-allgood:
Thanks for the compliment on the tests!

anoadragon453:
@serra-allgood Can you rebase this PR upon the latest master so CI changes don't appear in the diff?

anoadragon453:
Nevermind, your PR branch was accessible to maintainers so I handled it :)

PR #594, opened 2018-11-12
use go module for dependencies

rbns:
same steps as #593, only without the deletion of the vendor folder.

Signed-off-by: Ruben Schuller <code@rbn.im>

PR #703, opened 2019-05-22
Update Docker files and remove references to gb

babolivier:
Fix Docker files which had been forgotten in #594 and update scripts and docs to remove references to gb.

I also removed the `jenkins` directory which is outdated. We can revert f8d7e99 if we realise that it was a mistake to do so.


CONTRIBUTING.md
```diff
@@ -12,10 +12,11 @@ See [INSTALL.md](INSTALL.md) for instructions on setting up a running dev
 instance of dendrite, and [CODE_STYLE.md](CODE_STYLE.md) for the code style
 guide.
 
-We use `gb` for managing our dependencies, so `gb build` and `gb test` is how
-to build dendrite and run the unit tests respectively. Be aware that a list of
-all dendrite packages is the expected output for all tests succeeding with `gb
-test`. There are also [scripts](scripts) for [linting](scripts/find-lint.sh)
+As of May 2019, we're not using `gb` anymore, which is the tool we had been
+using for managing our dependencies, and switched to Go modules. To build
```
anoadragon453: First sentence is a bit of a run-on. I'd suggest putting a period after "dependencies" and then "Instead, we switched to Go modules".

CONTRIBUTING.md
```diff
@@ -12,10 +12,11 @@ See [INSTALL.md](INSTALL.md) for instructions on setting up a running dev
...
+Dendrite, run the `build.sh` script at the root of this repository, and to run
+unit tests, run `go test ./...` (which should pick up any unit test and run
```
anoadragon453:
```suggestion
unit tests, run `go test ./...` (which should pick up any unit tests and run
```

mediaapi/README.md
```diff
@@ -18,7 +18,7 @@ See the sample below for image quality with nfnt/resize:
 
 ### bimg (uses libvips C library)
 
-Alternatively one can use `gb build -tags bimg` to use bimg from https://github.com/h2non/bimg (MIT-licensed) which uses libvips from https://github.com/jcupitt/libvips (LGPL v2.1+ -licensed). libvips is a C library and must be installed/built separately. See the github page for details. Also note that libvips in turn has dependencies with a selection of FOSS licenses.
+Alternatively one can use `go build -tags bimg` to use bimg from https://github.com/h2non/bimg (MIT-licensed) which uses libvips from https://github.com/jcupitt/libvips (LGPL v2.1+ -licensed). libvips is a C library and must be installed/built separately. See the github page for details. Also note that libvips in turn has dependencies with a selection of FOSS licenses.
```
anoadragon453: Not sure if this works but I'll take your word for it.

scripts/build-test-lint.sh
```diff
@@ -5,7 +5,7 @@
 set -eu
 
 echo "Checking that it builds"
-gb build
+go build
```
anoadragon453: This will check that everything builds right? `./build.sh` is just for outputting all the different binaries?

CONTRIBUTING.md
```diff
@@ -12,10 +12,11 @@ See [INSTALL.md](INSTALL.md) for instructions on setting up a running dev
 instance of dendrite, and [CODE_STYLE.md](CODE_STYLE.md) for the code style
 guide.
 
-We use `gb` for managing our dependencies, so `gb build` and `gb test` is how
-to build dendrite and run the unit tests respectively. Be aware that a list of
-all dendrite packages is the expected output for all tests succeeding with `gb
-test`. There are also [scripts](scripts) for [linting](scripts/find-lint.sh)
+As of May 2019, we're not using `gb` anymore, which is the tool we had been
+using for managing our dependencies, and switched to Go modules. To build
+Dendrite, run the `build.sh` script at the root of this repository, and to run
```
anoadragon453: Might want to specify what `build.sh` does exactly and why it's different from `go build`.

scripts/build-test-lint.sh
babolivier: Yep. This line is useless though, as I just realised the right thing is done below it.

mediaapi/README.md
babolivier: It looks like it does the right thing.

Issue #654, opened 2019-01-08
Fix alias deletion


babolivier:
- [ ] Keep track of an alias's owner (someone else shouldn't be able to delete an alias I set)
- [ ] `DELETE /directory/room/{roomAlias}` should 404 if the alias doesn't exist

anoadragon453:
> Keep track of an alias's owner (someone else shouldn't be able to delete an alias I set)

Apparently this is an [implementation detail](https://matrix.to/#/!YkZelGRiqijtzXZODa:matrix.org/$15614741096174Svjyb:maunium.net?via=matrix.org&via=matrix.allmende.io&via=amorgan.xyz), so we could either prevent room admins from deleting aliases or allow it.

We could also provide a config option for dendrite to let the homeserver admin choose to allow server admins this ability.

Cnly:
I'll say that we should always allow room and homeserver admins to delete aliases, even if they'll only be able to delete the local ones. This is their control over the room/server. The fact that admins can't remove alias mappings on other servers anyways *is* making their deleting local aliases less meaningful, but I don't think it makes enough sense either that we remove admins' control over local data because of this.

anoadragon453:
[The spec](https://matrix.org/docs/spec/client_server/unstable#room-aliases) says we have free reign here, so let's go for allowing admins to remove local room aliases.

PR #706, opened 2019-05-28
Fix permission and 404 response for alias deletion - #654

Cnly:
This PR should fix https://github.com/matrix-org/dendrite/issues/654. Checklist from the issue:
- [x] Keep track of an alias's owner (someone else shouldn't be able to delete an alias I set)
- [x] DELETE /directory/room/{roomAlias} should 404 if the alias doesn't exist

### Pull Request Checklist


anoadragon453:
I'd like to get an answer to [this](https://github.com/matrix-org/dendrite/issues/654#issuecomment-505483222) before reviewing this.

Cnly:
> There's less bookkeeping we need to do if we're allowing room admins to delete room aliases, so this PR can be cut down quite considerably.

I think that's still a [`TODO`](https://github.com/matrix-org/dendrite/pull/706/files#diff-d058e5d2047e10ee4a65553b6e1923e1R190) here? This PR only concerns about users deleting aliases they set.

roomserver/storage/room_aliases_table.go
```diff
@@ -25,29 +25,35 @@ CREATE TABLE IF NOT EXISTS roomserver_room_aliases (
     -- Alias of the room
     alias TEXT NOT NULL PRIMARY KEY,
     -- Room ID the alias refers to
-    room_id TEXT NOT NULL
+    room_id TEXT NOT NULL,
+    -- ID of the user created this alias
```
anoadragon453:
```suggestion
    -- User ID of the creator of this alias
```

PR #712, opened 2019-05-29
Switch to Buildkite

anoadragon453:
Need to figure out how to actually test if this works...

~~Removed the TravisCI badge from the README but we don't have one for BuildKite just yet so that's currently been left out.~~ Got one.

Synapse buildkite pipeline for reference: https://github.com/matrix-org/synapse/blob/develop/.buildkite/pipeline.yml


.buildkite/pipeline.yaml
```diff
@@ -0,0 +1,30 @@
+steps:
+  - command:
+    - "./build.sh"
```
babolivier: Do we care about having the binaries built? If not, then we should use `go build ./cmd/...`, if yes, then we should expose them as artifacts.

.buildkite/pipeline.yaml
```diff
@@ -0,0 +1,30 @@
...
+    label: "\U0001F528 Build - Go 1.12"
```
richvdh: the emoji have names - it would be nice to use them.

See https://github.com/buildkite/emojis

.buildkite/pipeline.yaml
anoadragon453: I don't think we care at this point.

.buildkite/pipeline.yaml
anoadragon453: We can't as they clutter up the PR "Checks" display I'm afraid.

Issue #697, opened 2019-03-26
Replace gometalinter in CI


babolivier:
According to https://github.com/alecthomas/gometalinter/issues/590, gometalinter is being deprecated and needs to be replaced with [golangci-lint](https://github.com/golangci/golangci-lint).

PR #713, opened 2019-05-29
Fix pipeline, emoji and syntax

anoadragon453:
Fixes https://github.com/matrix-org/dendrite/issues/697

Emoji syntax has `-` in some places and `_` in some places :')

`image` indentation was my bad whoopsie


anoadragon453:
This'll require a review of golangci-lint's config. Also I took out `misspell` as it's included in `golangci-lint` but presumably we have to tell it which files/folders to run on.

.golangci.yml
```diff
@@ -0,0 +1,254 @@
...
+  fast: false
+
+
+issues:
+  # List of regexps of issue texts to exclude, empty list by default.
+  # But independently from this option we use default exclude patterns,
+  # it can be disabled by `exclude-use-default: false`. To list all
+  # excluded by default patterns execute `golangci-lint run --help`
+  exclude:
+    - abcdef
```
anoadragon453: Stuff like this should be commented out for instance.

anoadragon453:
Note to self: consider pipeline order before merging. Should we be waiting on lint?

.golangci.yml
```diff
@@ -0,0 +1,280 @@
...
+  disable:
+    - bodyclose
+    - depguard
+    - dupl
+    - gochecknoglobals
+    - gochecknoinits
+    - gocritic
+    - gofmt
+    - golint
+    - gosec
```
babolivier:
```suggestion
    - gosec # Should turn back on soon
```

PR #717, opened 2019-06-17
Replace deprecated prometheus.InstrumentHandler method

anoadragon453:
`prometheus.InstrumentHandler()` was deprecated in favour of `promhttp.Handler()`. This deprecated function was finally removed in prometheus' client_golang v1.0.0, and was [breaking buils](https://circleci.com/gh/matrix-org/dendrite/47).


anoadragon453:
Will be handled in https://github.com/matrix-org/dendrite/pull/713

PR #722, opened 2019-06-26
Fix test POST /rooms/:room_id/join can join a room

anoadragon453:
There's not much point in reusing SendMembership (which doesn't seem to probably support join) if we already have a method specifically for joining.

However, this may not be the cleanest way to fix this issue, so let me know if there's a nicer method of doing so.


Cnly:
I did some investigation and it looks like [SendMembership](https://github.com/matrix-org/dendrite/blob/74827428bd3e11faab65f12204449c1b9469b0ae/clientapi/routing/membership.go#L42) is lacking a `room_id` in response to `/join` requests, which should be easy to fix.

Maybe one downside of using `joinRoomByIDOrAlias` will be that function also accepts aliases?

Cnly:
It turns out #684 has already fixed what I said.

anoadragon453:
Superseded by #734.

PR #728, opened 2019-07-02
Encode URLs properly

anoadragon453:
We were escaping the URL before performing any pattern matching on it.
This meant that if you sent data that URLdecoded to a "/", it would count as
a "/" in the URL, potentially causing a 404. This was causing some flaky tests
with some randomly-generated query parameters.

Now, we keep URLs encoded while doing the pattern matching, and only afterwards
do we URL decode each query parameter individually before passing them to their
respective handler functions.

`github.com/gorilla/mux` was also updated to v1.7.3 to fix a bug with URL encoding and subrouters.


common/routing.go
```diff
@@ -0,0 +1,35 @@
...
+package common
+
+import (
+	"net/url"
+)
+
+// URLDecodeVarMap is a function that iterates through each of the items in a
+// map, URL decodes the values, and returns a new map with the decoded values
+// under the same key names
+func URLDecodeVarMap(vars map[string]string) (map[string]string, error) {
```
Cnly: Maybe it's better to explain in comments why we need this function?

testfile
```diff
@@ -142,3 +142,5 @@ Trying to get push rules with unknown rule_id fails with 404
 Events come down the correct room
 local user can join room with version 5
 User can invite local user to room with version 5
+Should reject keys claiming to belong to a different user
+Enabling an unknown default rule fails with 404
```
Cnly: These two lines are already in the file.

common/routing.go
anoadragon453: Mm, not sure though, as if another use is found for it the comment will need to be updated.

I can make a comment about it where it's used however.

testfile
anoadragon453: Have removed the lines.

common/routing.go
anoadragon453: Make the function name more generic. Does this help?

Cnly: I think this will do it. Didn't realise this can be a generic function before.

anoadragon453: Yeah :D

Issue #659, opened 2019-01-08
Tell servers and clients which servers know about an alias


babolivier:
Implement the `server` attribute in [`GET /_matrix/client/r0/directory/room/{roomAlias}`](https://matrix.org/docs/spec/client_server/r0.5.0#get-matrix-client-r0-directory-room-roomalias) responses.

This can be done by setting its value to a list of the servers that are in the room at the time of the request.
As a bonus point, if we needed to use federation to perform the lookup request, we can check and make sure the list contains the server we talked to.

--

Edit: added link to spec ~ @anoadragon453

PR #731, opened 2019-07-08
Fix #661 appservice can't set aliases in its own namespace

### LINKED ISSUE #661: Allow setting aliases within an appservice's namespace if it's the appservice that's setting the alias

Cnly:
Fixes #661.

This should allow new tests to pass after #659 is also fixed.

### Pull Request Checklist

### DIFF
--- a/clientapi/routing/directory.go
+++ b/clientapi/routing/directory.go
@@ -117,12 +117,14 @@ func SetLocalAlias(
 	// 1. The new method for checking for things matching an AS's namespace
 	// 2. Using an overall Regex object for all AS's just like we did for usernames
 	for _, appservice := range cfg.Derived.ApplicationServices {
-		if aliasNamespaces, ok := appservice.NamespaceMap["aliases"]; ok {
-			for _, namespace := range aliasNamespaces {
-				if namespace.Exclusive && namespace.RegexpObject.MatchString(alias) {
-					return util.JSONResponse{
-						Code: http.StatusBadRequest,
-						JSON: jsonerror.ASExclusive("Alias is reserved by an application service"),
+		if device.UserID != appservice.SenderLocalpart {
+			if aliasNamespaces, ok := appservice.NamespaceMap["aliases"]; ok {
+				for _, namespace := range aliasNamespaces {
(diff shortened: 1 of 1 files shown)


clientapi/routing/directory.go
```diff
@@ -117,12 +117,14 @@ func SetLocalAlias(
 	// 1. The new method for checking for things matching an AS's namespace
 	// 2. Using an overall Regex object for all AS's just like we did for usernames
 	for _, appservice := range cfg.Derived.ApplicationServices {
-		if aliasNamespaces, ok := appservice.NamespaceMap["aliases"]; ok {
-			for _, namespace := range aliasNamespaces {
-				if namespace.Exclusive && namespace.RegexpObject.MatchString(alias) {
-					return util.JSONResponse{
-						Code: http.StatusBadRequest,
-						JSON: jsonerror.ASExclusive("Alias is reserved by an application service"),
+		if device.UserID != appservice.SenderLocalpart {
```
anoadragon453: May be worth leaving a comment here that the UserID of the appservice's device will be a localpart, instead of a full UserID.

Frankly, I wonder if this will be slightly confusing down the line and instead we should set a full ID with the local serverName here. https://github.com/matrix-org/dendrite/blob/bc382bba4623a1c7ae16535541f2f4b23708999c/clientapi/auth/auth.go#L121

clientapi/routing/directory.go
```diff
@@ -117,12 +117,14 @@ func SetLocalAlias(
 	// 1. The new method for checking for things matching an AS's namespace
```
anoadragon453: Perhaps it's time to put this in?

Cnly: Actually there doesn't seem to be an equivalent... There is [`IsInterestedInRoomAlias`](https://github.com/matrix-org/dendrite/blob/master/common/config/appservice.go#L103), but it doesn't check for `namespace.Exclusive`. Another reason for not doing it here is we should probably refactor this with all other occurrences in one batch.

anoadragon453: I meant that we could just use the method in conjunction with your `if device.UserID ...` check.

But yes, it would be nice to fix them all in one batch.

PR #732, opened 2019-07-09
Add missing servers field in /directory/room/:alias response

### LINKED ISSUE #659: Tell servers and clients which servers know about an alias

Cnly:
Fixes #659.

### Pull Request Checklist


anoadragon453:
So looking at Synapse's code, it:

Checks that the room is owned by itself:

https://github.com/matrix-org/synapse/blob/4806651744616bf48abf408034ab9560e33f60ce/synapse/handlers/directory.py#L235-L280

Then if it is, pulls the list of servers it knows are in a room from the DB and returns that:

https://github.com/matrix-org/synapse/blob/4806651744616bf48abf408034ab9560e33f60ce/synapse/storage/directory.py#L30-L63

I'm not quite sure why, if I'm not the server in the room_id, that I can't answer this query, and presumably this will break if the homeserver that originally created the room leaves it...

Cnly:
Depends on #781.

anoadragon453:
Can we add our own server to the `servers` slice?

Cnly:
> I'm not quite sure why, if I'm not the server in the room_id, that I can't answer this query, and presumably this will break if the homeserver that originally created the room leaves it...

I'm afraid that'll also be Dendrite's behaviour because we don't (?) keep track of remote aliases...

Here Dendrite will first directly check if the alias association is local. If not, it sends a federation query to the domain in the ID. So if the server in the ID is down, this will break.

> Can we add our own server to the `servers` slice?

If the alias is local and we're in that room (which should be impled), then we'll be in the list. If it's resolved via federation, then we'll just use whatever the remote server gives us.

anoadragon453:
So this will fail if the server that the alias is assigned to is down, but that's expected by definition of an alias. That's why you want to create a myriad of aliases for a room across a number of servers to keep it resilient.

So actually I think this is ok! Could you merge this branch forward please? :)

PR #734, opened 2019-07-11
Fix response to /rooms/{roomId}/join v2

anoadragon453:
Continuation of #684 but merged-forward.

Also did a little code cleanup and added a new, passing test to the testfile.


Cnly:
Does this obsolete #722?

anoadragon453:
Indeed it does, thanks :)

clientapi/routing/membership.go
```diff
@@ -215,3 +209,41 @@ func getMembershipStateKey(
 
 	return
 }
+
+func checkAndProcessThreepid(
+	req *http.Request,
+	device *authtypes.Device,
+	body *threepid.MembershipRequest,
+	cfg config.Dendrite,
```
Cnly: It seems a lot of places are passing `cfg` not as a pointer. Maybe they need to be changed at some point?

anoadragon453: Mm, indeed. I don't believe there's any reason not too, and there's RAM savings in doing so.

anoadragon453: https://github.com/matrix-org/dendrite/issues/736

PR #735, opened 2019-07-12
Backup and restore go.mod & go.sum during linting

anoadragon453:
Every time before sending a PR I like to run `./scripts/build-test-lint.sh` to make sure the CI won't complain about anything.

The problem is that this script attempts to install golangci-lint, which causes modifications to go.mod/go.sum. This PR backs up and restores those files before and after linting.

Ideally instead of this hacky backing up/restoring we'd use `go get`s `-mod=readonly` option, but that still modifies `go.sum`. This will be fixed in go 1.13 apparently. https://github.com/golang/go/issues/30667

### DIFF
--- a/scripts/find-lint.sh
+++ b/scripts/find-lint.sh
@@ -22,7 +22,15 @@ then args="--fast"
 fi
 
 echo "Installing golangci-lint..."
+
+# Make a backup of go.{mod,sum} first
+# TODO: Once go 1.13 is out, use go get's -mod=readonly option
+# https://github.com/golang/go/issues/30667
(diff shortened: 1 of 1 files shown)


Cnly:
+1 for this. But is `mktemp` a better choice than using a `.bak` file? It minimizes the chance that we overwrite the user's own file. Not sure about it's cross-platform compatibility though.

anoadragon453:
Yeah, I chose the simple `cp` and `mv` commands because they're pretty standard (even on Windows apparently).

We'll remove this when go 1.13 comes anyways, so a little hacky isn't a problem imo.

PR #718, opened 2019-06-18
Add typing notifications to /sync responses - fixes #635

Cnly:
This PR adds a new consumer for typing notifications in syncapi. It also brings changes to syncserver.go and some related files so EDUs can better fit in /sync responses.

Fixes #635.
Fixes #574.

Signed-off-by: Alex Chen <minecnly@gmail.com>

### Pull Request Checklist

* [x] I have added any new tests that need to pass to `testfile` as specified in [docs/sytest.md](https://github.com/matrix-org/dendrite/blob/master/docs/sytest.md)
* [x] Pull request includes a [sign off](https://github.com/matrix-org/dendrite/blob/master/CONTRIBUTING.md#sign-off)

PR #752, opened 2019-07-19
Fix /sync may contain duplicate EDUs and EDUs for left rooms

Cnly:
In 29841bed (#718), EDUs are added to /sync responses for rooms listed
in joinedRoomIDs returned by addPDUDeltaToResponse. However this list
may contain rooms other than those currently joined.

Some variable renamings are done to make golangci-lint pass.

### Pull Request Checklist


syncapi/storage/syncserver.go
```diff
@@ -253,9 +259,14 @@ func (d *SyncServerDatasource) addPDUDeltaToResponse(
 		return nil, err
 	}
 
+	// len(deltas) should be a good enough initial capacity
 	joinedRoomIDs := make([]string, 0, len(deltas))
+
 	for _, delta := range deltas {
-		joinedRoomIDs = append(joinedRoomIDs, delta.roomID)
+		if delta.membership == membershipJoin {
```
anoadragon453: getStateDeltas states that it may produce duplicate join events. Is it fine to pass multiple stateDeltas for the same room down `/sync`?

https://github.com/matrix-org/dendrite/blob/29841bed6b1a88787211368e6052a87a658c5714/syncapi/storage/syncserver.go#L723-L731

Seems like `getStateDeltas` is currently incomplete and some of this work could be moved into there instead?

Cnly: > getStateDeltas states that it may produce duplicate join events. Is it fine to pass multiple stateDeltas for the same room down /sync?

It's not fine, indeed. There will be duplicate EDUs added to the /sync response.

> Seems like getStateDeltas is currently incomplete and some of this work could be moved into there instead?

It seems it'd be more efficient if we just return `joinedRoomIDs` from `getStateDeltas`, so I think that is worth a change. This will also fix the duplication above. But it isn't really related to the completeness of the function I guess?

anoadragon453: No, but we can fix that function in another PR I suppose.

Cnly: In fact the duplicate join events will not only affect EDUs, but also affect PDUs I believe; there'll be no visual effects though. Duplicate join events will cause the joined room response for a room to be calculated multiple times in [`addRoomDeltaToResponse`](https://github.com/matrix-org/dendrite/blob/0f0941ba25e64c4a3784f9b746b4f2692c242127/syncapi/storage/syncserver.go#L575).

syncapi/storage/syncserver.go
```diff
@@ -716,10 +720,13 @@ func (d *SyncServerDatasource) fetchMissingStateEvents(
 	return events, nil
 }
 
+// getStateDeltas returns the state deltas between fromPos and toPos for the
```
anoadragon453: Inclusive or not?

syncapi/storage/syncserver.go
anoadragon453: I think Riot (and possibly other clients) will display that a user has joined a room multiple times, but it's not the end of the world.

Issue #401, opened 2018-03-02
Registration: Don't create a new device_id during login if one is supplied


anoadragon453:
**Spec URL:** https://matrix.org/docs/spec/client_server/r0.3.0.html#post-matrix-client-r0-login

Under "Response Format":
> device_id | string | ID of the logged-in device. Will be the same as the corresponding parameter in the request, if one was specified.

**Phase:** 1

**Difficulty:** Easy

When we login, we supply a `device` parameter so that we can tell the homeserver what device we're logging in from. If we don't supply this parameter, the homeserver presumes we are logging in from a new device and creates a new device ID for us. If we supply one, that ID should just be returned in the response. The problem is that a new ID is being generated whether we supply one or not.

## Request

```
Note: "ClaaatBG" is a valid device ID
{
    "user": "@supercoolname:localhost",
    "password": "greatpassword123",
    "type": "m.login.password",
    "device_id": "ClaaatBG"
}
```

## Response

```
{
  "user_id": "@supercoolname:localhost",
  "access_token": "HUUXAhG8eU6qrN1MiJ-3-1hCH4kotGE2q2uD28D7VIQ",
  "home_server": "localhost",
  "device_id": "8KRmr3IY" # This should have been "ClaaatBG" again.
}
```

We got back `"device_id": "8KRmr3IY"`, a new device ID even though we supplied a valid ID already. It should have responded with the original device ID.

**Summary:**

- [ ] Prevent a new Device ID from being generated if we supply one on login
- [ ] Make sure that if a Device ID is supplied in the request, it is returned in the response.
- [ ] Create test case(s) to verify a new Device ID is created, and that the same Device ID is sent back if we supply one to `/login`.

MohitKS5:
Although I implemented it, I guess its not a necessary thing to do because we remove device at logout. So unless you want to login twice with same device simultaneously its no use at all.

anoadragon453:
Spec mandates it, so ¯\\\_(ツ)\_/¯

PR #753, opened 2019-07-19
Correctly create new device when device_id is passed to /login

anoadragon453:
Fixes https://github.com/matrix-org/dendrite/issues/401

Currently when passing a `device_id` parameter to `/login`, which is [supposed](https://matrix.org/docs/spec/client_server/unstable#post-matrix-client-r0-login) to return a device with that ID set, it instead just generates a random `device_id` and hands that back to you.

The code was already there to do this correctly, it looks like it had just been broken during some change. Hopefully sytest will prevent this from becoming broken again.

### DIFF
--- a/clientapi/auth/storage/devices/devices_table.go
+++ b/clientapi/auth/storage/devices/devices_table.go
@@ -169,6 +169,8 @@ func (s *devicesStatements) selectDeviceByToken(
 	return &dev, err
 }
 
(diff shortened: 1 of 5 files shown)


anoadragon453:
Hm, the fact that this passed means that sytest may not have a check for this behaviour.

Will add one real quick.

anoadragon453:
Have added in https://github.com/matrix-org/sytest/pull/655

PR #772, opened 2019-08-01
Fix transaction IDs in transaction cache have global scope

### LINKED ISSUE #756: Transaction IDs in transaction cache have global scope
Currently a transaction cache is [implemented](https://github.com/matrix-org/dendrite/blob/master/clientapi/routing/sendevent.go#L53) for request idempotency, however the transactions cached are only associated with their IDs regardless of the senders' access tokens. So if a transaction ID is used by more than one user before the cache expires, only one among these requests will actually have effect.

From the [spec](https://matrix.org/docs/spec/client_server/r0.5.0#api-standards):

> The scope of a transaction identifier is a particular access token.

Cnly:
Fixes #756.

### Pull Request Checklist

### DIFF
--- a/clientapi/routing/sendevent.go
+++ b/clientapi/routing/sendevent.go
@@ -50,7 +50,7 @@ func SendEvent(
 ) util.JSONResponse {
 	if txnID != nil {
 		// Try to fetch response from transactionsCache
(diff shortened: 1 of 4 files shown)


Cnly:
Interestingly this allowed a new test where there's a txnID shared in SyTest to pass.

PR #783, opened 2019-08-13
sort some gitignores

hawkowl:


.gitignore
```diff
@@ -7,18 +7,18 @@
...
-/bin
-/pkg
 /_obj
 /_test
-/vendor/bin
+/bin
 /docker/build
+/kafka
+/pkg
+/vendor/bin
```
Cnly: `/vendor` and `/pkg` are not used anymore.

PR #786, opened 2019-08-14
Error when recaptcha configs are not present

Trion129:
During setup of Client API if recaptcha is enabled but the settings are missing, error is logged and panic is raised.

Testing:
Enabled enable_registration_captcha, causes error.
Providing keys and api url passes the check and server starts.


Cnly:
A better place for these might be:

https://github.com/matrix-org/dendrite/blob/d21a2fb152143b2ce7600213abe18013519ccedf/common/config/config.go#L566

Cnly:
Done! I merged it with a slight change to the title.

Issue #767, opened 2019-07-26
Dendrite shouldn't try to unmarshal the request body for /leave


Cnly:
[`/rooms/:roomID/leave`](https://matrix.org/docs/spec/client_server/r0.5.0#post-matrix-client-r0-rooms-roomid-leave) accepts an empty request body, but Dendrite [tries](https://github.com/matrix-org/dendrite/blob/d21a2fb152143b2ce7600213abe18013519ccedf/clientapi/routing/membership.go#L49) to unmarshal it anyway.

yozamacs:
I would like to work on this....would it be okay to start with adding a basic test for that case? I see there aren't many tests in that directory

Cnly:
@yozamacs Sure, you can take it, and feel free to help add more tests. Currently we use [SyTest](https://github.com/matrix-org/sytest) to test Dendrite as a whole, but finer, language-specific tests are surely good to have! You can also join us in [#dendrite-dev:matrix.org](https://matrix.to/#/#dendrite-dev:matrix.org) if you want to discuss or ask more.

yozamacs:
I did some work on this in https://github.com/matrix-org/dendrite/pull/790. Let me know if that was the direction you wanted to go in

anoadragon453:
@babolivier and I are arguing over what's considered spec compliant here. Whether Unmarshalling even if there's an empty body is spec compliant or not. We'll get back to you.

yozamacs:
great, thanks!

anoadragon453:
So here we are getting back to you :)

Neither the Matrix Spec nor HTTP define whether a `POST` request need contain a body with content. We're going to be opinionated an require this. Therefore the rule should be:

if the content is `nil`, return a `400 M_NOT_JSON`, otherwise attempt to unmarshal.

yozamacs:
@anoadragon453 I'll get on it, thanks!!

Kegsay:
> if the content is nil, return a 400 M_NOT_JSON, otherwise attempt to unmarshal.

What does this even mean? A literal body with the text `null` or `nil`? Because `req.Body` will always be defined even if it's empty.

Kegsay:
https://github.com/matrix-org/dendrite/pull/839#issuecomment-632027384

PR #790, opened 2019-08-19
PR #767 - Checks to see if the request body is nil

yozamacs:
Signed-off-by: Maryum Styles <yozamacs@msn.com>

### Pull Request Checklist

### DIFF
--- a/federationsender/federationsender.go
+++ b/federationsender/federationsender.go
@@ -15,11 +15,15 @@
 package federationsender
 
 import (
+	"net/http"
+
 	"github.com/matrix-org/dendrite/common/basecomponent"
+	"github.com/matrix-org/dendrite/federationsender/api"
 	"github.com/matrix-org/dendrite/federationsender/consumers"
+	"github.com/matrix-org/dendrite/federationsender/query"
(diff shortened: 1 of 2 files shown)


clientapi/routing/membership.go
```diff
@@ -46,6 +46,11 @@ func SendMembership(
 	producer *producers.RoomserverProducer,
 ) util.JSONResponse {
 	var body threepid.MembershipRequest
+
+	if req.Body == nil {
+		body = threepid.MembershipRequest{}
+		req.Body = body
```
anoadragon453: You can't assign to `req.Body` unfortunately as it's read only.

Line 54 is unmarshalling `req.Body`'s contents into `body`, so if you already plan to act on a `nil` body, then there's no need to Unmarshal.

yozamacs: I thought that was the case. I wasn't sure what to do with the request body. Since it's nil, it'll still return in the next block with an error and I think we wanted to avoid that. I also considered doing an if statement and not unmarshaling the request if the body was nil

anoadragon453:
I borked up the update of this PR to latest master, sorry. Recreated here: https://github.com/matrix-org/dendrite/pull/799

PR #795, opened 2019-09-22
patch dendrite microservices with bind config

aditsachde:
This PR adds a block in the dendrite config for the services to bind to. The microservices should bind to the addresses in the bind block, and will be contacted at the address in the listen block.

This fixes an issue with the microservices and kubernetes services. 

### Pull Request Checklist

### DIFF
--- a/cmd/dendrite-appservice-server/main.go
+++ b/cmd/dendrite-appservice-server/main.go
@@ -35,9 +35,6 @@ func main() {
 		base, accountDB, deviceDB, federation, alias, query, cache,
 	)
 
-	if base.Cfg.Bind.AppServiceAPI != "" {
-		base.SetupAndServeHTTP(string(base.Cfg.Bind.AppServiceAPI))
-	} else {
-		base.SetupAndServeHTTP(string(base.Cfg.Listen.AppServiceAPI))
-	}
+	base.SetupAndServeHTTP(string(base.Cfg.Bind.AppServiceAPI), string(base.Cfg.Listen.AppServiceAPI))
+
 }
--- a/cmd/dendrite-appservice-server/main.go
+++ b/cmd/dendrite-appservice-server/main.go
@@ -35,5 +35,5 @@ func main() {
 		base, accountDB, deviceDB, federation, alias, query, cache,
 	)
 
-	base.SetupAndServeHTTP(string(base.Cfg.Listen.FederationSender))
+	base.SetupAndServeHTTP(string(base.Cfg.Bind.FederationSender))
 }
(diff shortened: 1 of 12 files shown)


aditsachde:
I added a check that causes the service to listen on the address specified in the listen block if no specific bind addresses have been specified.

I'm not really sure how to proceed with tests, it looks like a test config is generated in `test/config.go` with the function `assignAddress()`. In a testing environment, the listen address and bind address will likely be the same, so if something like `cfg.Bind.ClientAPI = assignAddress()` is added alongside `cfg.Listen.ClientAPI = assignAddress()` for each service, the bind and listen addresses will be different causing the tests to fail. 

---
Signed-off-by: Adit Sachde <me@aditsachde.com>

cmd/dendrite-appservice-server/main.go
```diff
@@ -35,5 +35,9 @@ func main() {
 		base, accountDB, deviceDB, federation, alias, query, cache,
 	)
 
-	base.SetupAndServeHTTP(string(base.Cfg.Listen.FederationSender))
+	if base.Cfg.Bind.AppServiceAPI != "" {
```
anoadragon453: It might be cleaner just to change `SetupAndServeHTTP` to take both addresses and have the fallback logic inside of it instead?

aditsachde: Pushed a change, can you make sure its alright?

aditsachde:
> Mm, I'm not too familiar with the unit tests myself unfortunately, but my naive thought if you can just modify assignAddress to return the port twice, and assign that to both bind/listen config values. 

In `test/config.go`, `cfg.Listen.ClientAPI` is defined as `cfg.Listen.ClientAPI = assignAddress()`. Could `cfg.Bind.ClientAPI` then be defined like `cfg.Bind.ClientAPI = cfg.Listen.ClientAPI`?

```
	cfg.Bind.ClientAPI = cfg.Listen.ClientAPI
	cfg.Bind.FederationAPI = cfg.Listen.FederationAPI
	cfg.Bind.MediaAPI = cfg.Listen.MediaAPI
	cfg.Bind.RoomServer = cfg.Listen.RoomServer
	cfg.Bind.SyncAPI = cfg.Listen.SyncAPI
	cfg.Bind.PublicRoomsAPI = cfg.Listen.PublicRoomsAPI
	cfg.Bind.TypingServer = cfg.Listen.TypingServer
```

aditsachde:
I believe all the changes should be resolved, just need buildkite to run

PR #798, opened 2019-09-29
selectAccountDataByType return ClientEvent pointer instead of slice of ClientEvent

### LINKED ISSUE #773: selectAccountDataByType shouldn't return a slice
[`selectAccountDataByType`](https://github.com/matrix-org/dendrite/blob/3578d77d259c852a16dc430725d348e2c62c4ff5/clientapi/auth/storage/accounts/account_data_table.go#L121) currently returns a `[]gomatrixserverlib.ClientEvent`. However since there can only be [one](https://github.com/matrix-org/dendrite/blob/3578d77d259c852a16dc430725d348e2c62c4ff5/clientapi/auth/storage/accounts/account_data_table.go#L42) account data row for a given `(localpart, room_id, type)` tuple, we can either change the return type to a pointer or a plain `ClientEvent` with an error to indicate empty data. This should also make the return value less confusing.

behouba:
### Pull Request Checklist

This pull request is an attempt to fix #773. 

Signed-off-by: Kouame Behouba Manassé behouba@gmail.com

### DIFF
--- a/clientapi/auth/storage/accounts/account_data_table.go
+++ b/clientapi/auth/storage/accounts/account_data_table.go
@@ -122,7 +122,6 @@ func (s *accountDataStatements) selectAccountDataByType(
 	ctx context.Context, localpart, roomID, dataType string,
 ) (data *gomatrixserverlib.ClientEvent, err error) {
 	stmt := s.selectAccountDataByTypeStmt
-
 	var content []byte
 
 	if err = stmt.QueryRowContext(ctx, localpart, roomID, dataType).Scan(&content); err != nil {
--- a/clientapi/auth/storage/accounts/account_data_table.go
+++ b/clientapi/auth/storage/accounts/account_data_table.go
@@ -120,28 +120,20 @@ func (s *accountDataStatements) selectAccountData(
 
 func (s *accountDataStatements) selectAccountDataByType(
 	ctx context.Context, localpart, roomID, dataType string,
-) (data []gomatrixserverlib.ClientEvent, err error) {
-	data = []gomatrixserverlib.ClientEvent{}
-
+) (data *gomatrixserverlib.ClientEvent, err error) {
 	stmt := s.selectAccountDataByTypeStmt
-	rows, err := stmt.QueryContext(ctx, localpart, roomID, dataType)
+	row := stmt.QueryRowContext(ctx, localpart, roomID, dataType)
+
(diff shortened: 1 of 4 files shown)


clientapi/auth/storage/accounts/account_data_table.go
```diff
@@ -120,28 +120,20 @@ func (s *accountDataStatements) selectAccountData(
 
 func (s *accountDataStatements) selectAccountDataByType(
 	ctx context.Context, localpart, roomID, dataType string,
-) (data []gomatrixserverlib.ClientEvent, err error) {
-	data = []gomatrixserverlib.ClientEvent{}
-
+) (data *gomatrixserverlib.ClientEvent, err error) {
 	stmt := s.selectAccountDataByTypeStmt
-	rows, err := stmt.QueryContext(ctx, localpart, roomID, dataType)
+	row := stmt.QueryRowContext(ctx, localpart, roomID, dataType)
```
Cnly:
```suggestion
```

This line and the `Scan` statement below can be merged into the `if` block.

clientapi/auth/storage/accounts/account_data_table.go
```diff
@@ -120,28 +120,20 @@ func (s *accountDataStatements) selectAccountData(
...
+
```
Cnly:
```suggestion
```

clientapi/auth/storage/accounts/account_data_table.go
```diff
@@ -120,28 +120,20 @@ func (s *accountDataStatements) selectAccountData(
...
+	var content []byte
+
+	err = row.Scan(&content)
```
Cnly:
```suggestion
```

clientapi/auth/storage/accounts/account_data_table.go
```diff
@@ -120,28 +120,20 @@ func (s *accountDataStatements) selectAccountData(
...
 	if err != nil {
```
Cnly:
```suggestion
	if err = stmt.QueryRowContext(ctx, localpart, roomID, dataType).Scan(&content); err != nil {
```

clientapi/auth/storage/accounts/account_data_table.go
behouba: indeed it would be much better, i will change that.

PR #803, opened 2019-10-04
Add Dockerfiles with proper microservices

aditsachde:
### Pull Request Checklist


aditsachde:
This PR is pretty opinionated as I was building via Github actions, not buildkite. Couple things I would like feedback on:

- Where in the repo should these dockerfiles exist
- Should they be build the way they are in the readme, or should each dockerfile be selfcontained? I think they will need to be selfcontained to allow buildkite to properly build them at once, although it will come at the expense of some compute
- Is the method that I am passing configuration to the dendrite microservices ok? I don't think dendrite can directly read environment variables yet. 
- Other things

I know that this is not really a huge priority and I have the images already at [here](https://hub.docker.com/r/aditsachde/dendrite), but it would encourage people to run dendrite. It also might make Sytest easier, but I'm not too familiar with Sytest's internal workings.

aditsachde:
I'll update naming scheme based on the outcome of #806

Issue #393, opened 2018-02-10
Load Dendrite configuration from multiple hierarchical sources


anoadragon453:
Yan Minari:

[...] the config file should be optional, setting the server address using env vars should be a thing, etc

1. load sane defaults (hardcoded on binary)
2. override with config file (if any)
3. override with --flags
4. override with env vars
5. override with secret file
6. etc.

not exactly everything in this list or in this order but you get the idea

yangm97:
More on that: https://medium.com/@kelseyhightower/12-fractured-apps-1080c73d481c

CromFr:
I'm currently looking at this issue :)

This is what I'm planing to do for configuration loading:

Configuration is build from multiple sources. Each override replace scalars and arrays with the provided value, mappings are only looked recursively (seems to be `gopkg.in/yaml.v2` default behaviour).

Here are the configuration sources:

0. Compiled default config. Should produce obvious errors when users don't override those value (ie panic & print "you must set matrix.server_name" if not present)
0. System configuration file (ie `/etc/dendrite.yaml`). Fallback to an empty file if not found.
0. Configuration file provided as command line argument (`--config`). Two approaches here:
    1. Try to open `dendrite.yaml` and fallback to an empty file if not found (eventually post a warning). If a config path is provided as command-line argument, panic if not found.
    2. Use an empty file if no config file is provided by command-line (no more `dendrite.yaml` used by default). If a config path is provided as command-line argument, panic if not found.
0. Command-line argument, with `--set` option, that can be provided multiple times. The value format should be valid yaml (seems a legit choice since config is in yaml and you can inline anything using json syntax):
    - `--set='matrix.server_name: example.com'`
    - `--set='matrix.trusted_third_party_id_servers: [vector.im, matrix.org, riot.im]'`
0. Override with a single environment var
    - `DENDRITE_CFG='{matrix.server_name: example.com, media.base_path: /dev/null}' dendrite-room-server`
0. Any ideas? (I didn't get what @anoadragon453  meant by "secret file")


I think this cover most use cases, but we could eventually add specific syntax like (not very useful IMO):
- `matrix.trusted_third_party_id_servers+: [ example.im ]` append elements to an array
- `matrix.trusted_third_party_id_servers-: [ riot.im ]` remove elements from an array

---

Some other ideas:
- Allow to provide multiple `--config=file` arguments, so the starting config is build using multiple configuration fragments

ptman:
[...]

PR #807, opened 2019-10-06
Update Docker stuff

feffes:
### Pull Request Checklist

-----

Updates Docker build implementation. Builds everything in a container then makes a minimal container for each service. This is by no means the final version and simply want to add this implementation to the discussion. However, considering that master branch is currently not in a stable state it would be fine to merge, and fix things like CI-implementation and compose examples.

This also adds a (addmitedly pretty hacky) solution to Environment variable parsing where needed. 
This is because from the docker implementation persepective that should be an expected form of interfacing configs, and how that implementation looks in the code can be fixed to look nicer later. #393 

### Things this doesn't implement(yet(possibly(but not probably))):
* [ ] dockerfiles for test tools
* [ ] updated ci ( this one I might actually do, but need more discussion about naming conventions and stuff)
* [ ] docker compose files for example configurations
* [ ] updated documentation (will do if people like the implementation)

### Things I'm unsure about with current implementation:
What requires ca-certificates? As of right now only the monolith server has it.

### Image sizes
```
appservice-server         latest              7be53849f9f4        26 seconds ago      16.1MB
kafka-producer            latest              851bcbc5c608        40 minutes ago      12.2MB
create-room-events-tool   latest              d59c5eccf47e        40 minutes ago      11.7MB
create-account-tool       latest              e73bb902d3f4        40 minutes ago      12.2MB
generate-keys-tool        latest              d7d338dc4563        40 minutes ago      11.4MB
federation-sender         latest              ad6a203c48d6        40 minutes ago      16.1MB
federation-api            latest              0f9be9576e90        40 minutes ago      16.3MB
federation-api-proxy      latest              14c54522107b        41 minutes ago      11.9MB
typing-server             latest              6900920f095b        41 minutes ago      16.9MB
room-server               latest              ae8e0c2d8ef8        41 minutes ago      17.5MB
sync-api                  latest              e517196eaae9        41 minutes ago      16.4MB
public-rooms-api          latest              baef78ac0868        41 minutes ago      16.1MB
media-api                 latest              22111ed53b1c        41 minutes ago      16.9MB
client-api                latest              76c4aa57be44        41 minutes ago      17.5MB
client-api-proxy          latest              272f3ff2c37c        41 minutes ago      11.9MB
monolith                  latest              4dffa823001e        41 minutes ago      19.9MB
dendrite-build            latest              56bd0be732b7        41 minutes ago      780MB
```


frzifus:
what's the current status here?

feffes:
@frzifus 
I haven't really been working on it since the draft was made, but since there seems to be some more sign of life here lately and the CI is actually working I might give it some work in the coming weeks. As it stands now I just rebased it onto master and it still builds the containers fine but they haven't been tested yet, but you could try to build them to test for yourself if you want. If you come back with results that would be super cool as well.

What needs to get done before being ready for review is I need to test all images and make sure they work in a complete environment. I also need to check with the maintainers if it's fine to include the environment variable config implementation in this pr or if it should be separated. I also need to update documentation but perhaps it can be opened for review before then.

Kegsay:
I like this a lot. There's a few conversations to be had around how we want to deploy dendrite in general, but I much prefer this style to what is there currently.

I would be very interested to see if you could incorporate https://docs.docker.com/engine/reference/builder/#onbuild into your build image here, it might come in handy to reduce the code duplication on downstream images, but I haven't given it much thought yet.

I would like to merge this in, but I will need to handshake with other core team members first, and modify our CI pipeline to support this new format.

Great job!

feffes:
> I would be very interested to see if you could incorporate https://docs.docker.com/engine/reference/builder/#onbuild into your build image here, it might come in handy to reduce the code duplication on downstream images, but I haven't given it much thought yet.

I guess you could make a conf image on top of scratch that just copies the config file and so everything that uses the config file can build on it, but I don't think onbuild helps any in this case.

I think there's something to be said about not introducing too much complexity though, so people have an easier time parsing the files in the future. And really there isn't much that can be abstracted up the chain anyway. Only things I can think of are ssl certs and config file setup.

Kegsay:
FYI I'll be bringing this up (the Docker modifications) with folks next week.

feffes:
@Kegsay Nice. I'll hopefully have some time to polish it up and test it more after the 20th.

Kegsay:
Superseded by #1053 - and we're holding off on moving to env vars until we refactor the config in #889 - closing when #1053 is merged.

PR #810, opened 2019-10-08
Move files back if linting fails

anoadragon453:
Prevents `go.{mod,sum}.bak` files from being left around if the linting command fails.

This PR makes it so the rest of the script will run even if linting fails.

### DIFF
--- a/scripts/find-lint.sh
+++ b/scripts/find-lint.sh
@@ -31,12 +31,12 @@ go get github.com/golangci/golangci-lint/cmd/golangci-lint
 
 # Run linting
 echo "Looking for lint..."
-# If we're running in CI, a linting fail should fail the CI step
-if [ -n "${CI}" ]; then
-    golangci-lint run $args 
-else
-    # Otherwise continue the script even if linting fails
+# If we're running locally, continue the script even if linting fails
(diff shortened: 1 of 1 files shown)


scripts/find-lint.sh
```diff
@@ -29,8 +29,15 @@ echo "Installing golangci-lint..."
 cp go.mod go.mod.bak && cp go.sum go.sum.bak
 go get github.com/golangci/golangci-lint/cmd/golangci-lint
 
+# Run linting
 echo "Looking for lint..."
-golangci-lint run $args
+# If we're running locally, continue the script even if linting fails
+if [ -z ${CI+x} ]; then
+    golangci-lint run $args || echo "Linting script failed, removing module backups"
```
Cnly: Perhaps we could make it like `docker/dendrite.sh` in SyTest so the script still exits with 1 when something fails.

anoadragon453: Something like:

```
golangci-lint run $args || exit_code=1
mv ...
exit $exit_code
```

?

Yeah, that's not a bad idea.

anoadragon453:
Merging as CI is known broken currently.

Issue #668, opened 2019-01-08
Implement guest access


babolivier:
- [ ] Handle `kind=guest` query parameter on `/register`
- [ ] Handle `m.room.guest_access` event type

Full spec : https://matrix.org/docs/spec/client_server/r0.4.0.html#guest-access

Kegsay:
Failing sytests:
```
    × Guest users can send messages to guest_access rooms if joined
    × Guest users are kicked from guest_access rooms on revocation of guest_access
    × Guest users are kicked from guest_access rooms on revocation of guest_access over federation
    × Guest user can upgrade to fully featured user
    × Guest users denied access over federation if guest access prohibited
    × m.room.history_visibility == "shared" allows/forbids appropriately for Guest users
    × m.room.history_visibility == "invited" allows/forbids appropriately for Guest users
    × m.room.history_visibility == "joined" allows/forbids appropriately for Guest users
    × m.room.history_visibility == "default" allows/forbids appropriately for Guest users
    × Guest users can sync from shared guest_access rooms if joined
    × Guest users can sync from invited guest_access rooms if joined
    × Guest users can sync from joined guest_access rooms if joined
```

kegsay:
Basic support is definitely there.

PR #816, opened 2019-10-11
Allow registration of Guest users (issue #668)

CromFr:
This PR should allow guest users to be registered on the server, with a unique numeric ID.

I added a prometheus counter `dendrite_clientapi_guest_users_total` for counting guest users. I'm not familiar with prometheus, so I'm not sure if there are other things to configure.

### Pull Request Checklist


clientapi/routing/register.go
```diff
@@ -597,6 +625,27 @@ func handleRegistrationFlow(
...
+
+	// Exit if guest registration is forbidden
+	if cfg.Matrix.GuestAccessDisabled {
+		return util.MessageResponse(http.StatusForbidden, "Guest access has been disabled")
+	}
+
+	// Create guest user
+	return completeRegistration(
+		req.Context(), accountDB, deviceDB, true, r.Username, r.Password, "",
+		r.InhibitLogin, r.InitialDisplayName, r.DeviceID,
```
babolivier: Won't `r.Username` be at its 0-value (`""`) if we're ignoring everything but `initial_device_display_name` in the JSON body? And even if not, we still need to generate a user ID because `@:server_name` is not a valid MXID. What Synapse does here is use the lowest integer that isn't already in use by another guest user as the localpart, but we could also use a random string (using `util.RandomString`) so that we don't have to scan the whole users table to determine the localpart to use.

clientapi/routing/register.go
```diff
@@ -791,7 +841,11 @@ func completeRegistration(
 	}
 
 	// Increment prometheus counter for created users
-	amtRegUsers.Inc()
+	if isGuest {
+		amtRegUsers.Inc()
```
babolivier: You mixed up the counters here (you're increasing `amtRegUsers` if we're registering a guest user and `amtGuestUsers` if we're registering a normal user).

dendrite-config.yaml
```diff
@@ -18,6 +18,9 @@ matrix:
       - vector.im
       - matrix.org
 
+    # Setting this to false will allow guest user to register on the server
+    guest_access_disabled: true
```
babolivier: We usually name config options from the action they manage, in this case it would rather be `disable_guest_access` (and `DisableGuestAccess` in `config.go` and `register.go`).

neilalexander:
Thanks for the contribution here, but I'm going to close for now as we have merged #860. If you are willing, it'd be great to add the prometheus counting for guests into the current `master`.

PR #818, opened 2019-10-23
check for nil body prior to unmarshal

yozamacs:
### Pull Request Checklist

This is a redo of https://github.com/matrix-org/dendrite/pull/790 regarding issue https://github.com/matrix-org/dendrite/issues/767.

I'm not entirely certain that checking that the body is nil in this way will work. According to the go documentation in [request.go](https://golang.org/src/net/http/request.go) it says that 
```
	// For client requests, a nil body means the request has no
	// body, such as a GET request. The HTTP Client's Transport
	// is responsible for calling the Close method.
```
which seems like the use case that the issue is referring to. 

I haven't added any tests yet, I would like some guidance on how tests work for this project.

### DIFF
--- a/clientapi/routing/membership.go
+++ b/clientapi/routing/membership.go
@@ -45,6 +45,12 @@ func SendMembership(
 	queryAPI roomserverAPI.RoomserverQueryAPI, asAPI appserviceAPI.AppServiceQueryAPI,
 	producer *producers.RoomserverProducer,
 ) util.JSONResponse {
+	if req.Body==nil {
+		return util.JSONResponse{
+			Code: http.StatusBadRequest,
+			JSON: jsonerror.NotJSON("Content is nil"),
+		}
+	}
 	var body threepid.MembershipRequest
 	if reqErr := httputil.UnmarshalJSONRequest(req, &body); reqErr != nil {
 		return *reqErr


yozamacs:
I don't understand this error, `clientapi/routing/membership.go:48: File is not `goimports`-ed (goimports)`. I ran `goimports -e -d` manually and didn't get any output. I see the same error when I run `./scripts/build-test-lint.sh` locally but not sure how to address

Cnly:
From my reading I think `Body` will always be non-`nil` since we're on the server side:

```go
	// For server requests, the Request Body is always non-nil
	// but will return EOF immediately when no body is present.
	// The Server will close the request body. The ServeHTTP
	// Handler does not need to.
	Body io.ReadCloser
```

A better way to do this may be to check `ContentLength`.

And "File is not `goimports`-ed" means you haven't formatted the file with `goimports`; you need to run `goimports -w clientapi/routing/membership.go` ("-ed" is the past particle here, not flags to use). You may also run `goimports -h` to see the help text. Basically if you're using an IDE, it may already have integrated the tool and it shouldn't be hard to let your IDE do this automatically.

Kegsay:
Missing sign-off and this could really do with a sytest (POSTing to /leave with an empty body).

Kegsay:
https://github.com/matrix-org/dendrite/pull/839 is a better fix for this.

Issue #649, opened 2019-01-08
Implement user directory


babolivier:
This might need its own component like publicroomsapi does, though its logic is simpler so I'm not that sure about it.

- [ ] Implement [POST /_matrix/client/r0/user_directory/search](https://matrix.org/docs/spec/client_server/r0.4.0.html#post-matrix-client-r0-user-directory-search)

anoadragon453:
Probably wouldn't hurt to put it in its own dedicated component since matrix.org sees a large amount of traffic from user searches, and taking load off roomserver is probably a good idea.

PrivyDG:
maybe GAIA from blockstack project would be suitables for such. https://github.com/blockstack/gaia . Only problem is that GAIA  is not written in  GOLANG but their have have couple go projects and libs

anoadragon453:
That would require a change to the [Matrix spec](https://matrix.org/docs/spec) which Dendrite is currently implementing.

I agree that decentralizing the user directory more is a good idea though. One could even use something like https://zeronet.io, which has the concept of user accounts and search.

Would need some PoW scheme though to dissuade Sybil attacks.

PrivyDG:
well Gaia is very impressive since it does not sacrifice speed, does not depend on a crypto currency, support delegations already which would be very useful for matrix.I like the fact it does allow the users to select where his data would be hosted. 
![gaia](https://user-images.githubusercontent.com/48634491/61579521-08dcfd80-aad4-11e9-8ad4-889864657e7e.png)

anstue:
Anyone implementing this issue? Otherwise I grab it ;)

anoadragon453:
> Anyone implementing this issue? Otherwise I grab it ;)

Not to my knowledge!

PR #822, opened 2019-11-09
Implement user directory for local users #649

anstue:
Atm it queries only local users.

Signed-off-by: Anton Stuetz <opensource@ti-zero.com>

### Pull Request Checklist


clientapi/auth/storage/accounts/profile_table.go
```diff
@@ -92,6 +105,37 @@ func (s *profilesStatements) selectProfileByLocalpart(
...
+) (*[]searchtypes.SearchResult, bool, error) {
+	//increase limit by one to find out if the query was limited
+	rows, err := s.selectSearchTermLikeLocalpartOrDisplayNameStmt.QueryContext(ctx, searchTerm, limit+1)
+	if err != nil {
+		return nil, false, err
+	}
+	searchResults := []searchtypes.SearchResult{}
+	counter := int8(1)
+	limited := false
+	for rows.Next() {
```
Kegsay: missing `defer rows.Close()`

clientapi/auth/storage/accounts/storage.go
```diff
@@ -28,6 +30,69 @@ import (
 	_ "github.com/lib/pq"
 )
 
+type AccountDatabase interface {
+	GetAccountByPassword(
+		ctx context.Context, localpart, plaintextPassword string,
+	) (*authtypes.Account, error)
+	GetProfileByLocalpart(
+		ctx context.Context, localpart string,
+	) (*authtypes.Profile, error)
```
Kegsay: Please document the exact contract you wish to expose here. E.g if there is no profile for a given localpart, do we:
 - return an error?
 - return a nil profile and nil error?

Can the storage impl do string coercion e.g case insensitive matching? I appreciate that there may be no 'right' answer here, in which case please document what your code does under these circumstances so we can check the behaviour.

cmd/dendrite-userdirectory-api-server/main.go
```diff
@@ -0,0 +1,34 @@
+// Copyright 2017 Vector Creations Ltd
```
Kegsay: This is probably wrong, though I have no idea what to put here.

Kegsay:
This PR was created in 2019 and has now been marked as stale. Please can someone resurrect this or else it will be closed in 1 week.

PR #823, opened 2019-11-13
use custom http client instead of the http DefaultClient

### LINKED ISSUE #820: Use http.Client instead of http.DefaultClient
Currently there are several places (eg [query.go](https://github.com/matrix-org/dendrite/blob/5eb63f1d1eafae0111d46def5cf5bf431d2e8169/federationsender/api/query.go#L64)) where http.DefaultClient is used, which doesn't come with a timeout. This could cause unwanted locks.

frzifus:
This commit replaces the default client from the http lib with a custom one.
The previously used default client doesn't come with a timeout. This could cause
unwanted locks.
That solution chosen here creates a http client in the base component dendrite
with a constant timeout of 30 seconds. If it should be necessary to overwrite
this, we could include the timeout in the dendrite configuration.
Here it would be a good idea to extend the type "Address" by a timeout and
create an http client for each service.

Closes #820

Signed-off-by: Benedikt Bongartz <benne@klimlive.de>

### DIFF
--- a/appservice/api/query.go
+++ b/appservice/api/query.go
@@ -20,6 +20,7 @@ package api
 import (
 	"context"
 	"database/sql"
+	"errors"
 	"net/http"
 
 	"github.com/matrix-org/dendrite/clientapi/auth/authtypes"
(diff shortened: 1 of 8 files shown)


frzifus:
ping, any updates?

frzifus:
rebased

anoadragon453:
The test failure is an issue with the testing infrastructure and will be fixed by #859

Kegsay:
It's slightly frustrating to see all the `NewXXXAPI` calls sprout an `error` return, but there's no other nice way of doing it. We definitely should *not* be using the default client either way, for the aforementioned timeout issue.

@frzifus please rebase this and I'll merge this in.

frzifus:
All right, I'll do it today after work or during my lunch break. :+1:

Kegsay:
Alright, one final thing and then we can merge this: please rebase (typingserver => eduserver) and then I'll merge this in. Sorry, I should've done this before that refactor :/

frzifus:
mhpf? :D

Kegsay:
When CI passes I'M GONNA DO IT

Issue #750, opened 2019-07-18
Support the new "identifier" login parameter


anoadragon453:
Spec: https://matrix.org/docs/spec/client_server/unstable#post-matrix-client-r0-login

The login API has a new `identifier` field which deprecates the `user` and `address` fields.

MohitKS5:
Assuming 3PID endpoint remains for now (to be depreciated as per spec), It requires

-  Define identifier struct , preferably in login.go as at this point only login endpoint uses that.
-  In /login handler , go over [interface types](https://matrix.org/docs/spec/client_server/unstable#identifier-types)
    - extract user from identifier in m.id.user
    - send back not-implemented for others

Kegsay:
https://github.com/matrix-org/dendrite/pull/825 added this

PR #825, opened 2019-11-27
#750 Fix login using identifier key

serra-allgood:
Not a full fix, it only really supports logging in with the localpart of an mxid.

### Pull Request Checklist


serra-allgood:
How serious are those CI failures?

PR #829, opened 2019-12-15
add identifier field to login

MohitKS5:
Signed-off-by: mohitks <mohitkumarsingh907@gmail.com>
Addresses #750. Checks for identifier field and its type and extracts user favouring it over `user` field.

I am a bit confused about logging, should I use separate logging formats for login using identifier and login using depreciated `user` field or log both `user` and `identifier` field [here](https://github.com/matrix-org/dendrite/blob/master/clientapi/routing/login.go#L89) only.

### DIFF
--- a/clientapi/routing/login.go
+++ b/clientapi/routing/login.go
@@ -89,7 +89,7 @@ func Login(
 			// go over identifier types
 			switch r.Identifier.Type {
 			case "m.type.user":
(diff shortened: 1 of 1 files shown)


neilalexander:
Thanks for the contribution here, but I think this is already solved by #825 which was merged a couple weeks back.

PR #830, opened 2019-12-17
build.sh: Make cmd directory path absolute

anoadragon453:
The `GOBIN` parameter in `build.sh` was made to work from anywhere, but `./cmd/...` is still relative to the directory the script was run in.

This PR makes that absolute as well, so that `build.sh` can be run outside of the dendrite folder (i.e. by sytest).


anoadragon453:
I wish for this to be the case.

PR #831, opened 2019-12-17
Replace deprecated prometheus.InstrumentHandler and unsafe time.Ticker

anoadragon453:
This should fix the remaining `golangci-lint` staticcheck errors.

Will need a merge with broken sytests, as there's a dependency loop between this PR and #832.

Fixes #717 for the InstrumentHandler.

`golangci-lint` complains about `time.Ticker` as:

> SA1015: using time.Tick leaks the underlying ticker, consider using it only in endless functions, tests and the main package, and use time.NewTicker here.


mediaapi/routing/routing.go
```diff
@@ -83,26 +85,33 @@ func makeDownloadAPI(
 	activeRemoteRequests *types.ActiveRemoteRequests,
 	activeThumbnailGeneration *types.ActiveThumbnailGeneration,
 ) http.HandlerFunc {
-	return prometheus.InstrumentHandler(name, http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
-		req = util.RequestWithLogging(req)
+	return promhttp.InstrumentHandlerCounter(
```
babolivier: This is a bit more difficult to read than it should be, can't you define the `CounterVec` and the `HandlerFunc` separately and then do `promhttp.InstrumentHandlerCounter(counterVec, handlerFunc)` or something like that? Or just defining separately the function given to `http.HandlerFunc` like you do in `common/httpapi.go` is also good enough.

mediaapi/routing/routing.go
```diff
@@ -83,26 +85,33 @@ func makeDownloadAPI(
...
+				types.MediaID(vars["mediaId"]),
+				cfg,
+				db,
+				client,
+				activeRemoteRequests,
+				activeThumbnailGeneration,
+				name == "thumbnail",
+			)
+		},
+		))
```
babolivier: The indentation doesn't look right to me here.

anoadragon453: image

I think it's right? It's due to a function inside of an argument.

anoadragon453: Anyways, refactored now.

mediaapi/routing/routing.go
```diff
@@ -83,15 +85,21 @@ func makeDownloadAPI(
...
+	httpHandler := func(w http.ResponseWriter, req *http.Request) {
 		req = util.RequestWithLogging(req)
 
 		// Set common headers returned regardless of the outcome of the request
 		util.SetCORSHeaders(w)
 		// Content-Type will be overridden in case of returning file data, else we respond with JSON-formatted errors
 		w.Header().Set("Content-Type", "application/json")
-
-		vars, _ := common.URLDecodeMapValues(mux.Vars(req))
+		vars := mux.Vars(req)
```
babolivier: Why drop the call to `common.URLDecodeMapValues`? It sounds like something we'd like to keep here.

anoadragon453: Ah, yes. This is because this was changed after I made the original fix which I just copy-pasted for this PR.

PR #832, opened 2019-12-17
Resync testfile with current sytest pass/fail

anoadragon453:
At some point we should go back over these specific tests and see why they're failing all of a sudden (seems to be due to an internal server error somewhere).

CI won't pass until #831 is merged.


anoadragon453:
Force merging to fix CI.

PR #796, opened 2019-09-23
Response from /send_join now use gomatrixserverlib.RespSendJoin

behouba:
Signed-off-by: Kouame Behouba Manassé behouba@gmail.com

Should fix #614

### Pull Request Checklist

* [ ] I have added any new tests that need to pass to `testfile` as specified in [docs/sytest.md](https://github.com/matrix-org/dendrite/blob/master/docs/sytest.md)
* [x] Pull request includes a [sign off](https://github.com/matrix-org/dendrite/blob/master/CONTRIBUTING.md#sign-off)

PR #834, opened 2019-12-20
Federation bugfixes

neilalexander:
This PR tries to fix some federation bugs:

- The `send_join` endpoint was not returning as a `[200, {"state": ..., ...}]` tuple which causes unmarshaling errors on the remote side.


Cnly:
I think #796 is already a work on this.

neilalexander:
Replaced by #845.

PR #839, opened 2019-12-31
Return 400 M_NOT_JSON for empty request body in /rooms/:roomID/:action

### LINKED ISSUE #767: Dendrite shouldn't try to unmarshal the request body for /leave

dryairship:
Fixes #767 according to the guidelines in [this comment](https://github.com/matrix-org/dendrite/issues/767#issuecomment-536639195).

Signed-off-by: Priydarshi Singh dryairship@gmail.com

### Pull Request Checklist

### DIFF
--- a/clientapi/routing/membership.go
+++ b/clientapi/routing/membership.go
@@ -46,6 +46,14 @@ func SendMembership(
 	producer *producers.RoomserverProducer,
 ) util.JSONResponse {
 	var body threepid.MembershipRequest
+
+	if req.Body == http.NoBody {
+		return util.JSONResponse{
+			Code: http.StatusBadRequest,
+			JSON: jsonerror.NotJSON("No request body"),
+		}
+	}
+
 	if reqErr := httputil.UnmarshalJSONRequest(req, &body); reqErr != nil {
 		return *reqErr
 	}


anoadragon453:
Some recommend:

```
if http.Request().Body == http.NoBody {
  // TODO.
}
```

But do we not already check for this elsewhere in Dendrite?

dryairship:
@babolivier @anoadragon453  I did the changes. Also, I searched for `http.NoBody` in the repository, but did not find it anywhere in the code.

neilalexander:
I think `http.NoBody` is fine as a test, but it feels strange implementing it *here* for this *specific* endpoint and not really solving the problem elsewhere.

It might be better if we define a matcher function in `common` which does a check like this, and then add it to the routing for each endpoint as an additional `MatcherFunc` for requests that are already `http.MethodPost` or `http.MethodPut`.

@anoadragon453, @babolivier - thoughts?

babolivier:
> I think `http.NoBody` is fine as a test, but it feels strange implementing it _here_ for this _specific_ endpoint and not really solving the problem elsewhere.
> 
> It might be better if we define a matcher function in `common` which does a check like this, and then add it to the routing for each endpoint as an additional `MatcherFunc` for requests that are already `http.MethodPost` or `http.MethodPut`.
> 
> The only thing I'm not sure about if we do this is what the return code will be - I assume it would be 404 if a route is not matched because of a failed `MatcherFunc`.
> 
> @anoadragon453, @babolivier - thoughts?

Sounds good to me.

dryairship:
> I think `http.NoBody` is fine as a test, but it feels strange implementing it _here_ for this _specific_ endpoint and not really solving the problem elsewhere.
> 
> It might be better if we define a matcher function in `common` which does a check like this, and then add it to the routing for each endpoint as an additional `MatcherFunc` for requests that are already `http.MethodPost` or `http.MethodPut`.
> 
> The only thing I'm not sure about if we do this is what the return code will be - I assume it would be 404 if a route is not matched because of a failed `MatcherFunc`.
> 
> @anoadragon453, @babolivier - thoughts?

I'd be glad to take up that task. Here is what I have planned :
 1. We could make a middleware function that checks if the method is POST, if yes, then it checks if the body is empty. If the body is empty, it hijacks the ResponseWriter to return the error, otherwise it calls `next.ServeHTTP()` to continue the handler chain. This way, we do not have to add the MatcherFunc to all the target endpoints.
 2. Now since there is no failed MatcherFunc, the error code could be 400, which I believe is better than 404 for the error under consideration.

@neilalexander @babolivier - Please provide your suggestions.

neilalexander:
> We could make a middleware function that checks if the method is POST, if yes, then it checks if the body is empty. If the body is empty, it hijacks the ResponseWriter to return the error, otherwise it calls next.ServeHTTP() to continue the handler chain.

I'm happy with this approach — we just need to be careful that we don't break `POST /_matrix/media/r0/upload` in the process (but I think we can rely on `Content-Type` there).

Kegsay:
It's somewhat of an aberration that the spec allows POSTs with empty bodies for that API. I don't think we should be adding matchers/middlewares to try to handle this as it should be the exception not the norm. I'd rather we clearly handled it.

Kegsay:
I also don't understand how this PR helps here. It's rejecting what should be valid according to the spec.

Kegsay:
I am of the opinion that nothing needs to be done here. The comment in https://github.com/matrix-org/dendrite/issues/767#issuecomment-536639195 states:

> if the content is nil, return a 400 M_NOT_JSON, otherwise attempt to unmarshal.

This is the current behaviour which https://github.com/matrix-org/dendrite/issues/767 was saying is wrong. Closing this and #767.

PR #840, opened 2020-01-02
Amend syncapi SQL queries to return missing columns

neilalexander:
The `transaction_id` and `session_id` column that were added in #367 have resulted in some breakage in the `/sync` endpoint. These columns are required/expected by `rowsToStreamEvents` in `output_room_events_table.go` - not supplying them results in things like:
```
sql: expected 2 destination arguments in Scan, not 4
```

It's not exactly clear to me yet what `transaction_id` and `session_id` do, but these being added means that state events break sync.

This is a temporary fix. We need to come up with a better solution.

### DIFF
--- a/syncapi/storage/current_room_state_table.go
+++ b/syncapi/storage/current_room_state_table.go
@@ -84,10 +84,12 @@ const selectStateEventSQL = "" +
 	"SELECT event_json FROM syncapi_current_room_state WHERE room_id = $1 AND type = $2 AND state_key = $3"
 
 const selectEventsWithEventIDsSQL = "" +
-	// TODO: The session_id and transaction_id blanks are here because otherwise the rowsToStreamEvents
-	//       expects there to be exactly four columns. We need a better solution to this long term.
-	//       - neilalexander, 2 Jan 2020
-	"SELECT added_at, event_json, 0 AS session_id, '' AS transaction_id FROM syncapi_current_room_state WHERE event_id = ANY($1)"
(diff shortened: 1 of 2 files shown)


syncapi/storage/current_room_state_table.go
```diff
@@ -84,7 +84,10 @@ const selectStateEventSQL = "" +
 	"SELECT event_json FROM syncapi_current_room_state WHERE room_id = $1 AND type = $2 AND state_key = $3"
 
 const selectEventsWithEventIDsSQL = "" +
-	"SELECT added_at, event_json FROM syncapi_current_room_state WHERE event_id = ANY($1)"
+	// TODO: The session_id and transaction_id blanks are here because otherwise the rowsToStreamEvents
+	//       expects there to be exactly four columns. We need a better solution to this long term.
```
babolivier: Maybe also mention that we need to figure out whether we actually need to keep track of `session_id` and `transaction_id` in the db

syncapi/storage/current_room_state_table.go
```diff
@@ -84,7 +84,10 @@ const selectStateEventSQL = "" +
...
+	//       - neilalexander, 2 Jan 2020
+	"SELECT added_at, event_json, 0 AS session_id, '' AS transaction_id FROM syncapi_current_room_state WHERE event_id = ANY($1)"
```
babolivier: Can you wrap that line (e.g. before the `FROM`) please?

PR #841, opened 2020-01-03
Implement storage interfaces

neilalexander:
This PR creates a number of interfaces in the various `storage.go` files of each component. This is so that we can more easily swap out the database engine without modifying the rest of Dendrite.

This is the first of what will probably be a string of PRs. Right now adding support for a different database engine will result in a lot of code duplication - further PRs will reduce that where possible.


mediaapi/storage/postgres/storage.go
```diff
@@ -0,0 +1,105 @@
+// Copyright 2017 Vector Creations Ltd
```
babolivier: The current copyright should be `// Copyright 2020 The Matrix.org Foundation C.I.C.`. The way we do it in Synapse is that every new file gets that copyright, and a file that just got moved or changed has it added under the other mentions (e.g. a random file in Synapse I have open right now has:
```
# Copyright 2014-2016 OpenMarket Ltd
# Copyright 2017-2018 New Vector Ltd
# Copyright 2019 The Matrix.org Foundation C.I.C.
```
)

It would be nice to update the copyrights on the files you're changing (I'm only mentioning it on that file but I think it applies across the whole PR).

PR #844, opened 2020-01-10
Always defer *sql.Rows.Close and consult with Err

### LINKED ISSUE #764: rows.Next() should be followed by checks on rows.Err()
From [docs](https://godoc.org/database/sql#Rows.Next):

> Next prepares the next result row for reading with the Scan method. It returns true on success, or false if there is no next result row or an error happened while preparing it. Err should be consulted to distinguish between the two cases.

vladimiroff:
database/sql.Rows.Next() makes sure to call Close only after exhausting
result rows which would NOT happen when returning early from a bad Scan.
Close being idempotent makes it a great candidate to get always deferred
regardless of what happens later on the result set.

This change also makes sure call Err() after exhausting Next() and
propagate non-nil results from it as the documentation advises.

Closes #764.

Signed-off-by: Kiril Vladimiroff <kiril@vladimiroff.org>

### DIFF
--- a/clientapi/auth/storage/accounts/account_data_table.go
+++ b/clientapi/auth/storage/accounts/account_data_table.go
@@ -115,8 +115,7 @@ func (s *accountDataStatements) selectAccountData(
 			global = append(global, ac)
 		}
 	}
-	err = rows.Err()
-	return
+	return global, rooms, rows.Err()
 }
 
 func (s *accountDataStatements) selectAccountDataByType(
--- a/clientapi/auth/storage/accounts/account_data_table.go
+++ b/clientapi/auth/storage/accounts/account_data_table.go
@@ -90,6 +90,7 @@ func (s *accountDataStatements) selectAccountData(
 	if err != nil {
 		return
 	}
+	defer rows.Close() // nolint: errcheck
 
 	global = []gomatrixserverlib.ClientEvent{}
 	rooms = make(map[string][]gomatrixserverlib.ClientEvent)
@@ -114,7 +115,7 @@ func (s *accountDataStatements) selectAccountData(
 			global = append(global, ac)
 		}
 	}
-
+	err = rows.Err()
 	return
 }
 
--- a/roomserver/storage/postgres/events_table.go
+++ b/roomserver/storage/postgres/events_table.go
@@ -209,6 +209,9 @@ func (s *eventStatements) bulkSelectStateEventByID(
 			return nil, err
 		}
 	}
+	if err = rows.Err(); err != nil {
+		return nil, err
+	}
 	if i != len(eventIDs) {
 		// If there are fewer rows returned than IDs then we were asked to lookup event IDs we don't have.
 		// We don't know which ones were missing because we don't return the string IDs in the query.
@@ -219,7 +222,7 @@ func (s *eventStatements) bulkSelectStateEventByID(
 			fmt.Sprintf("storage: state event IDs missing from the database (%d != %d)", i, len(eventIDs)),
 		)
 	}
-	return results, err
+	return results, nil
 }
 
 // bulkSelectStateAtEventByID lookups the state at a list of events by event ID.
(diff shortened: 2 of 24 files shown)


clientapi/auth/storage/accounts/account_data_table.go
```diff
@@ -114,7 +115,7 @@ func (s *accountDataStatements) selectAccountData(
 			global = append(global, ac)
 		}
 	}
-
+	err = rows.Err()
```
APwhitehat: We could have this instead. I think it will look cleaner.
```go
return rows.Err()
```

APwhitehat: same for subsequent, occurrences.

roomserver/storage/postgres/events_table.go
```diff
@@ -321,6 +327,9 @@ func (s *eventStatements) bulkSelectStateAtEventAndReference(
 		result.EventID = eventID
 		result.EventSHA256 = eventSHA256
 	}
+	if err = rows.Err(); err != nil {
```
APwhitehat: rows.Err() must be checked after the call to rows.Next(). So you might need to bump this code up a few lines.

roomserver/storage/postgres/state_block_table.go
```diff
@@ -169,10 +169,13 @@ func (s *stateBlockStatements) bulkSelectStateBlockEntries(
 		}
 		current.StateEntries = append(current.StateEntries, entry)
 	}
+	if err = rows.Err(); err != nil {
```
APwhitehat: Same here

APwhitehat:
Hey @vladimiroff , good work on your 1st contribution. I have left some comments.

P.S. I not a big fan of wide use of `nolint errcheck` in our codebase, though I would defer this to the active maintainers.

roomserver/storage/postgres/events_table.go
vladimiroff: Not sure I understand what you mean. `rows.Err()` should be called ONLY after `rows.Next()` returns `false`. Thus in cases where we do  `for rows.Next() { .. }` the best place to call `rows.Err()` is right after the for loop.

vladimiroff:
> P.S. I not a big fan of wide use of `nolint errcheck` in our codebase, though I would defer this to the active maintainers.

Yeah, me neither. However, there's really no sense to check `rows.Close` as it does nothing more than propagating the result from `rows.Err` by design. And closing should be happening in defer due to early exists from bad row scan.

IMHO a nice way to solve this is to put a config file telling golangci to ignore `*sql.Rows.Close` checks in general.

vladimiroff:
> I think I generally agree with @APwhitehat. Maybe we can write a tiny wrapper function to close `rows` and ignore the error?

We could do that, indeed. Replacing each of these defer invokes with something like this below will work without errcheck noticing:

```go
defer func() { _ = rows.Close() }()
```

This makes it trivial to introduce a wrapper function somewhere inside `github.com/matrix-org/dendrite/common`, for instance:

```go
func SilentClose(c io.Closer) {
        _ = c.Close()
}
```

and then do `defer common.SilentClose(rows)` instead. However, I find this approach backwards. I would avoid obfuscating code just to make a linter happy not only without improving anything, but making code less readable, instead.

This project uses golangci-lint which exposes configuration inside [`.golangci.yml`](https://github.com/matrix-org/dendrite/blob/bc382bb/.golangci.yml#L78). Thus we could introduce `errcheck_excludes.txt` with this inside:

```
(*database/sql.Rows).Close
```

This way we explicitly ignore error checking only for this method of this specific type, knowing that's what being advised by design.

Said that, I'm no maintainer of dendrite and thus I obviously have no say in this decision. Words above are to be considered merely as my five cents. Whatever you choose I'm going to apply it on this PR.

roomserver/storage/postgres/events_table.go
APwhitehat: > right after the for loop

Yes, that is precisely what I thought.

APwhitehat:
> However, I find this approach backwards. I would avoid obfuscating code just to make a linter happy not only without improving anything, but making code less readable, instead.

I resonate with this opinion.

Something along the lines of `defer func() { _ = rows.Close() }()` should work for this PR.

roomserver/storage/postgres/events_table.go
vladimiroff: I still don't follow. This `rows.Err()` invoke is right after the very `for` looping over `rows.Next()`.

APwhitehat: Oh, you are right. It seemed to me that `rows.Scan` was outside the loop, my bad.
Ignore my comment.

vladimiroff:
@APwhitehat thanks for the review.

I'm rebasing this on top of master and keeping an eye for new `rows.Next()` invokes that might arise in order to cover them as well. Thus this should be ready for merging at any point.

PR #845, opened 2020-01-10
Federation fixes

neilalexander:
This PR makes the following fixes to unbreak federation:

- Updates `go.mod` ref to matrix-org/gomatrixserverlib@5f9a44f2fc671b3d4bf9783b43c780eaa874f520 which uses v2 endpoints for `/send_join` and `/send_leave`
- Returns `respSendJoin` and `respSendLeave` to be composites of `RespState`
- Rewrites `getAuthChain` (maybe not needed but seems clearer this way)
- Fixes some filter references to `gomatrix` instead of `gomatrixserverlib` (as apparently these have changed in more recent commits)
- `go mod tidy` for good measure


roomserver/query/query.go
```diff
@@ -617,42 +617,42 @@ func (r *RoomserverQueryAPI) QueryStateAndAuthChain(
 func getAuthChain(
 	ctx context.Context, dB RoomserverQueryAPIEventDB, authEventIDs []string,
 ) ([]gomatrixserverlib.Event, error) {
-	var authEvents []gomatrixserverlib.Event
-
-	// List of event ids to fetch. These will be added to the result and
-	// their auth events will be fetched (if they haven't been previously)
 	eventsToFetch := authEventIDs
```
babolivier: Can you please keep that comment and add a mention of where we fetch events from (i.e. the DB afaiu)? It wasn't straightforward to me where the events were being fetched from (DB? another server?) and more doc never hurts.

neilalexander: I've updated the comments to be a bit clearer as to what's going on :-)

PR #849, opened 2020-01-21
Add a Sytest blacklist file

anoadragon453:
Adds a blacklist for sytests, since we have some flaky sytests,

For context, we have a `show-expected-fail-tests.sh` script which runs at the end of each CI session and tells the user whether they should add a passing sytest to `testfile`. If so, CI will fail until the user does this. Sytest itself will fail if there is a sytest in `testfile` that doesn't pass. Thus a flaky test will sometimes break one way, and sometimes the other way, meaning we can never win.

The solution is to have a blacklist. With this we can tell the script about tests that we don't care about. The script then will ignore those tests while they remain on the blacklist.

This PR adds a blacklist and modifies `show-expected-fail-tests.sh` to optionally read it.


anoadragon453:
Merging as CI will not pass anyways.

PR #855, opened 2020-01-25
No overwriting of global err before return

### LINKED ISSUE #846: Don't overwrite global err before return
https://github.com/matrix-org/dendrite/blob/74827428bd3e11faab65f12204449c1b9469b0ae/mediaapi/fileutils/fileutils.go#L115

I realised that if we change the value of `err` in a defer statement, we are effectively overwriting the result of `tmpFile.Close()` to `err` which could be non-nil. This behaviour is triggered only for named return values.

> For instance, if the deferred function is a function literal and the surrounding function has named result parameters that are in scope within the literal, the deferred function may access and modify the result parameters before they are returned. If the deferred function has any return values, they are discarded when the function completes.

ref: https://golang.org/ref/spec#Defer_statements

Fix for this issue should also look for any other instances of this issue around the codebase.
e.g. 
https://github.com/matrix-org/dendrite/blob/9a56cf8439304540d0cdae18b30df82962c065cb/common/test/config.go#L145-L148

PC25:
Fixes #846 
Signed-off-by: Pushpam pchoudhary1@ee.iitr.ac.in

### DIFF
--- a/common/test/client.go
+++ b/common/test/client.go
@@ -82,8 +82,8 @@ func (r *Request) Do() (err error) {
 	if err != nil {
 		return err
 	}
-	defer (func() { 
-		if err == nil { err = res.Body.Close() }
+	defer (func() {
+		err = res.Body.Close()
 	})()
 
 	if res.StatusCode != r.WantedStatusCode {
--- a/common/test/client.go
+++ b/common/test/client.go
@@ -82,7 +82,9 @@ func (r *Request) Do() (err error) {
 	if err != nil {
 		return err
 	}
-	defer (func() { err = res.Body.Close() })()
+	defer (func() { 
+		if err == nil { err = res.Body.Close() }
+	})()
 
 	if res.StatusCode != r.WantedStatusCode {
 		return fmt.Errorf("incorrect status code. Expected: %d  Got: %d", r.WantedStatusCode, res.StatusCode)
--- a/common/test/config.go
+++ b/common/test/config.go
@@ -143,7 +143,9 @@ func NewMatrixKey(matrixKeyPath string) (err error) {
 	}
 
 	defer (func() {
-		if err == nil {err = keyOut.Close()}
+		if err == nil {
+			err = keyOut.Close()
+		}
 	})()
 
 	err = pem.Encode(keyOut, &pem.Block{
--- a/common/test/config.go
+++ b/common/test/config.go
@@ -143,7 +143,7 @@ func NewMatrixKey(matrixKeyPath string) (err error) {
 	}
 
 	defer (func() {
-		err = keyOut.Close()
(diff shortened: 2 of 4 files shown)


common/test/client.go
```diff
@@ -82,7 +82,9 @@ func (r *Request) Do() (err error) {
 	if err != nil {
 		return err
 	}
-	defer (func() { err = res.Body.Close() })()
+	defer (func() { 
+		if err == nil { err = res.Body.Close() }
```
Cnly: If `err != nil`, the body won't be closed since `res.Body.Close` won't be called.

mediaapi/fileutils/fileutils.go
```diff
@@ -112,7 +112,10 @@ func WriteTempFile(reqReader io.Reader, maxFileSizeBytes config.FileSizeBytes, a
 	if err != nil {
 		return
 	}
-	defer (func() { err = tmpFile.Close() })()
+	defer (func() {
+		if err == nil {err = tmpFile.Close()
```
Cnly: Please format your code with `goimports` before committing. You may use the automation script `./scripts/build-test-lint.sh` for this.

common/test/client.go
```diff
@@ -82,7 +82,9 @@ func (r *Request) Do() (err error) {
 	if err != nil {
 		return err
 	}
-	defer (func() { err = res.Body.Close() })()
+	defer (func() {
+		err = res.Body.Close()
+	})()
```
Cnly: Now this is exactly the same as the original code.

common/test/config.go
```diff
@@ -143,7 +143,9 @@ func NewMatrixKey(matrixKeyPath string) (err error) {
 	}
 
 	defer (func() {
-		err = keyOut.Close()
+		if err == nil {
```
Cnly: Again, `keyOut.Close` won't be called if `err != nil`. To fix that, you need to call `keyOut.Close()` first outside the `if`, then see what to do with the return value.

This also applies to the remaining occurrences.

frzifus: if an error occurs by calling this method i would guess there is no need to close keyout.

mediaapi/fileutils/fileutils.go
```diff
@@ -112,7 +112,11 @@ func WriteTempFile(reqReader io.Reader, maxFileSizeBytes config.FileSizeBytes, a
 	if err != nil {
 		return
 	}
-	defer (func() { err = tmpFile.Close() })()
+	defer (func() {
```
frzifus: same here. Maybe one of the best ways would be to just call `defer tmpFile.Close()`.

Kegsay:
https://github.com/matrix-org/dendrite/pull/858 is a better fix.

PR #856, opened 2020-01-25
Fix sarama import URLs

neilalexander:
This fixes the sarama import URLs throughout, updates the `naffka` dependency in `go.mod`/`go.sum` to a version that does the same and also runs `go mod tidy`.


neilalexander:
Welp, well that really didn't turn out well.

Kegsay:
@neilalexander please can you merged this forward and update. I'll look at it when it's mergable.

PR #857, opened 2020-01-26
Add create sequence in invites_table.go in syncapi

### LINKED ISSUE #723: invites_table.go should probably CREATE SEQUENCE IF NOT EXISTS syncapi_stream_id
It looks like [invites_table.go](https://github.com/matrix-org/dendrite/blob/master/syncapi/storage/invites_table.go#L11) should also have the `CREATE SEQUENCE` line as in other data tables sharing `syncapi_stream_id` e.g. https://github.com/matrix-org/dendrite/blob/master/syncapi/storage/account_data_table.go#L28.

PC25:
Fixes #723 
Signed-off-by: Pushpam pchoudhary1@ee.iitr.ac.in

### DIFF
--- a/syncapi/storage/postgres/invites_table.go
+++ b/syncapi/storage/postgres/invites_table.go
@@ -25,6 +25,8 @@ import (
 )
 
 const inviteEventsSchema = `
+-- This sequence is shared between all the tables generated from kafka logs.
+CREATE SEQUENCE IF NOT EXISTS syncapi_stream_id;
 CREATE TABLE IF NOT EXISTS syncapi_invite_events (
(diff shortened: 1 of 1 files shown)


Kegsay:
Yes, but for documentation purposes it is nice to see what that mysterious sequence is. It doesn't have to be done on a per-table basis, but we don't really have anywhere to put DB-level SQL statements.

I'm going to make an executive decision and say that we should probably have a place for DB-level SQL statements (the file which implements the storage interface), rather than dupe this everywhere and have it be a pain to update later. This makes this issue null and void, and https://github.com/matrix-org/dendrite/issues/963 A Thing.

PR #858, opened 2020-01-26
Alternate fix for #846, global err is not overwritten before return

### LINKED ISSUE #846: Don't overwrite global err before return

CruxBox:
Fixes #846 . This alternate fix concatenates the errors if required, else the only error is returned.
Signed-off-by: Shubham shubhambhagat111@yahoo.com


common/http/http.go
```diff
@@ -39,7 +39,14 @@ func PostJSON(
 
 	res, err := httpClient.Do(req.WithContext(ctx))
 	if res != nil {
-		defer (func() { err = res.Body.Close() })()
+		defer (func() {
+			finalErr := res.Body.Close()
```
frzifus: I'm not sure that the body here needs to be closed at all. Apart from the fact that in case of a mistake "res" does not have to be considered at all.
SEE: https://github.com/golang/go/blob/master/src/net/http/request.go#L175

common/test/client.go
```diff
@@ -82,7 +82,14 @@ func (r *Request) Do() (err error) {
 	if err != nil {
 		return err
 	}
-	defer (func() { err = res.Body.Close() })()
+	defer (func() {
```
frzifus: Is the function to close the http body called at all if in case of error it is returned before?
SEE: https://github.com/matrix-org/dendrite/pull/858/files#diff-8089a15307bfa30044c79197979a773bR82

mediaapi/fileutils/fileutils.go
```diff
@@ -112,8 +112,14 @@ func WriteTempFile(reqReader io.Reader, maxFileSizeBytes config.FileSizeBytes, a
 	if err != nil {
 		return
 	}
-	defer (func() { err = tmpFile.Close() })()
-
+	defer (func() {
```
frzifus: same here

mediaapi/thumbnailer/thumbnailer_nfnt.go
```diff
@@ -125,7 +126,14 @@ func writeFile(img image.Image, dst string) (err error) {
 	if err != nil {
 		return err
 	}
-	defer (func() { err = out.Close() })()
+	defer (func() {
```
frzifus: same here

roomserver/storage/postgres/invite_table.go
```diff
@@ -120,7 +121,14 @@ func (s *inviteStatements) updateInviteRetired(
 	if err != nil {
 		return nil, err
 	}
-	defer (func() { err = rows.Close() })()
+	defer (func() {
```
frzifus: same here

common/http/http.go
CruxBox: Isn't it required, to prevent resource leak?

> From the http.Client documentation:
> 
>  If the returned error is nil, the Response will contain a non-nil Body which the user is expected to close. If the Body is not both read to EOF and closed, the Client's underlying RoundTripper (typically Transport) may not be able to re-use a persistent TCP connection to the server for a subsequent "keep-alive" request.

CruxBox: In that case we can check for err being nil and then have the defer statement. No need to check for res!=nil

common/test/client.go
CruxBox: You're right. We should move the defer to right after the request is made. How about that?

roomserver/storage/postgres/invite_table.go
CruxBox: Fixed

mediaapi/thumbnailer/thumbnailer_nfnt.go
CruxBox: Fixed

mediaapi/fileutils/fileutils.go
CruxBox: Fixed

common/http/http.go
```diff
@@ -38,12 +38,20 @@ func PostJSON(
 	req.Header.Set("Content-Type", "application/json")
 
 	res, err := httpClient.Do(req.WithContext(ctx))
-	if res != nil {
-		defer (func() { err = res.Body.Close() })()
-	}
+
+	defer (func() {
+		finalErr := res.Body.Close()
```
Kegsay, line 43: Won't this panic if `httpClient.Do` returns an `err` and a `nil` `res`?

common/test/client.go
Kegsay: If an error is returned to do **not** need to close `res.Body`. The docs are very clear on this:

> On error, any Response can be ignored. A non-nil Response with a non-nil error only occurs when CheckRedirect fails, and even then the returned Response.Body is already closed. 

By putting it above the `err != nil` check, you risk cases where `err != nil && res == nil` resulting in a panic e.g on network errors.

Kegsay:
Please make a new PR if you wish to continue work on this. Thanks!

PR #864, opened 2020-02-04
Implemented ignoring specific users

### LINKED ISSUE #600: Implement ignoring specific users
- [ ] Implement special case for [m.ignored_user_list](https://matrix.org/docs/spec/client_server/r0.4.0.html#m-ignored-user-list) account data type

prateek2211:
### Pull Request Checklist
fixed #600 


syncapi/sync/requestpool.go
```diff
@@ -132,10 +133,24 @@ func (rp *RequestPool) OnIncomingSyncRequest(req *http.Request, device *authtype
 
 func (rp *RequestPool) currentSyncForUser(req syncRequest, latestPos types.PaginationToken) (res *types.Response, err error) {
 	// TODO: handle ignored users
+	localpart, _, err := gomatrixserverlib.SplitID('@', req.device.UserID)
+	if err != nil {
+		return
+	}
+	var ignoredUserList []string
+	if data, e := rp.accountDB.GetAccountDataByType(req.ctx, localpart, "", "m.ignored_user_list"); e == nil && data != nil {
```
babolivier: We should handle errors correctly here, instead of ignoring it. Also, let's name the error `err` for consistency and clarity. What I'd suggest would be:

```go
data, err := rp.accountDB.GetAccountDataByType(req.ctx, localpart, "", "m.ignored_user_list")
if err != nil {
    return
}
if data != nil {
    // ...
}
```

syncapi/sync/requestpool.go
```diff
@@ -132,10 +133,24 @@ func (rp *RequestPool) OnIncomingSyncRequest(req *http.Request, device *authtype
 
 func (rp *RequestPool) currentSyncForUser(req syncRequest, latestPos types.PaginationToken) (res *types.Response, err error) {
 	// TODO: handle ignored users
```
babolivier: If you're implementing handling of ignored users then we probably want this comment removed.

syncapi/storage/postgres/syncserver.go
```diff
@@ -642,12 +646,32 @@ func (d *SyncServerDatasource) getResponseWithPDUsForCompleteSync(
 	return res, toPos, joinedRoomIDs, err
 }
 
+func removeIgnoredUserEvents(events []types.StreamEvent, ignoredUsers []string) []types.StreamEvent {
```
babolivier: I'd be more inclined to implement this filtering in the SQL request, so that we return the number of events dictated by the filter as often as possible. FYI you can do that with PostgreSQL by giving an array to `QueryContext` and add `AND sender != ALL($X)` to the corresponding SQL query (obviously replace `X` with the order of the argument).

prateek2211:
@babolivier Apparently there's an out of bound runtime error in https://github.com/matrix-org/dendrite/blob/c20109a57357fbb2cd0857485e2cca4a58c37d1b/syncapi/storage/postgres/syncserver.go#L613 which results in build error. But I am not able to understand why did it appear now, but not before this pr.
Probably I need some help with this

syncapi/storage/postgres/output_room_events_table.go
```diff
@@ -81,12 +81,12 @@ const selectEventsSQL = "" +
 
 const selectRecentEventsSQL = "" +
 	"SELECT id, event_json, session_id, exclude_from_sync, transaction_id FROM syncapi_output_room_events" +
-	" WHERE room_id = $1 AND id > $2 AND id <= $3" +
+	" WHERE room_id = $1 AND id > $2 AND id <= $3 AND sender != ALL($5)" +
```
neilalexander: Is `ALL` right here? I would expect it to be `ANY`.

syncapi/storage/postgres/output_room_events_table.go
```diff
@@ -81,12 +81,12 @@ const selectEventsSQL = "" +
...
 	" ORDER BY id DESC LIMIT $4"
 
 const selectRecentEventsForSyncSQL = "" +
 	"SELECT id, event_json, session_id, exclude_from_sync, transaction_id FROM syncapi_output_room_events" +
-	" WHERE room_id = $1 AND id > $2 AND id <= $3 AND exclude_from_sync = FALSE" +
+	" WHERE room_id = $1 AND id > $2 AND id <= $3 AND exclude_from_sync = FALSE AND sender != ALL($5)" +
```
neilalexander: Same as above - I'd expect this to be `ANY`.

syncapi/storage/postgres/output_room_events_table.go
prateek2211: yeah it works
https://stackoverflow.com/a/11730845/10436060
however it would also work with ANY

neilalexander: Let's use `ANY`, it feels more correct.

prateek2211: ALL was suggested by @babolivier 
https://github.com/matrix-org/dendrite/pull/864#discussion_r375212389

neilalexander: I've done some further testing locally and it looks like `ALL` is OK, so I'll accept this.

prateek2211:
@babolivier @neilalexander I have rebased and resolved the merge conflict. See if the pr could be merged now.

Kegsay:
Closing this as this has bitrotted away. As with previous feature PRs (like redaction) we may use this as inspiration when implementing this feature. Thanks for your time and effort on it.

PR #865, opened 2020-02-05
Room version abstractions

neilalexander:
This PR (still a work in progress) adds some support for differentiating between room versions and state resolution algorithms.


clientapi/routing/capabilities.go
```diff
@@ -0,0 +1,51 @@
+// Copyright 2017 Vector Creations Ltd
```
turt2live: *ahem*

neilalexander: Thanks for the catch - I'll fix 💃

neilalexander:
Notes for review:

- This is one of two broken out PRs now. Sytest will possibly fail but the fixes to the whitelist/blacklist are in the next PR
- Most of the changes are in moving `roomserver/state/state.go` into `roomserver/state/v1/state.go` and changing some things to use function receivers so that the `db` reference is internal (GitHub makes it look like a huge change but in reality it's not much)

roomserver/query/query.go
```diff
@@ -98,6 +99,11 @@ func (r *RoomserverQueryAPI) QueryLatestEventsAndState(
 	request *api.QueryLatestEventsAndStateRequest,
 	response *api.QueryLatestEventsAndStateResponse,
 ) error {
+	// TODO: get the correct room version
+	state, err := state.GetStateResolutionAlgorithm(state.StateResolutionAlgorithmV1, r.DB)
```
babolivier: Not a huge fan of the fact that this variable is shadowing the `state` package, but I guess convenience makes it worthwhile here.

PR #868, opened 2020-02-10
Queue and retry failed federation requests

neilalexander:
This PR adds a new table into the `federationsender` database where PDUs which failed to be sent over federation are queued. A task then runs every minute which looks up these events and then tries sending them again.

Still to do:
- [ ] Exponential backoff, rather than just queuing for another minute
- [ ] Determine what the best number of retries is before giving up and dropping the PDU altogether (currently 5 attempts)

### DIFF
--- a/federationsender/federationsender.go
+++ b/federationsender/federationsender.go
@@ -40,7 +40,7 @@ func SetupFederationSenderComponent(
 		logrus.WithError(err).Panic("failed to connect to federation sender db")
 	}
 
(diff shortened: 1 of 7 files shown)


Kegsay:
What's the status of this @neilalexander ?

neilalexander:
I'm not happy with this PR, it's likely I'll burn it and come up with something better once more of the federation logic is concentrated.

PR #869, opened 2020-02-12
Support sqlite in addition to postgres

Kegsay:
We want to support sqlite so we can potentially embed dendrite into mobile devices / web / etc, and postgres doesn't play nicely with that. This PR does... many things, but the headline items are:
 - Split the `storage` package for some servers into `postgres` and `sqlite3` packages. These packages implement the storage interface for their respective servers.
 - Decide which database engine to use based on the connection string: `postgres://` and `file:`.

The servers which have sprouted sqlite support in this PR are:
 - client API
 - federation sender
 - room server
 - sync API


roomserver/storage/sqlite3/storage.go
```diff
@@ -0,0 +1,864 @@
...
+		cs = uri.Opaque
+	} else if uri.Path != "" { // file:///path/to/filename.db
+		cs = uri.Path
+	} else {
+		return nil, errors.New("no filename or path in connect string")
+	}
+	if d.db, err = sql.Open("sqlite3", cs); err != nil {
+		return nil, err
+	}
+	//d.db.Exec("PRAGMA journal_mode=WAL;")
```
diamondburned, line 55: Why are we not using `journal_mode = WAL` for this?

neilalexander: Largely because we were trying to ensure that we were thread-safe for writers at the time (which is significantly easier to do without the WAL), although there may be some sense in enabling it now. We would need to understand the effects it would have on non-standard targets such as WASM.

diamondburned: Perhaps add build tags to enable a `const sqliteWAL bool` when the target isn't wasm?

PR #870, opened 2020-02-14
Add whitelist/blacklist maintenance complainer to buildkite annotations

anoadragon453:
The complainer's output appears like this now:

image

Instead of you having to dig through the long list of logs.

Required associated SyTest PR: https://github.com/matrix-org/sytest/pull/803


anoadragon453:
Oops, it's not. The newlines in the output are borked. Dangit `xargs`.

anoadragon453:
Well, that's better. But we still have a failing test.

PR #871, opened 2020-02-14
More SQLite

neilalexander:
This PR adds SQLite support to `appservice`, `mediaapi` and `publicroomsapi`. 

Note that there are a couple of caveats with the `publicroomsapi` - the default queries return 30 rows (as we apparently have to specify a limit of some kind) and we don't check aliases in the filters yet.

### DIFF
--- a/publicroomsapi/storage/sqlite3/public_rooms_table.go
+++ b/publicroomsapi/storage/sqlite3/public_rooms_table.go
@@ -73,8 +73,8 @@ const selectPublicRoomsWithFilterSQL = "" +
 	" FROM publicroomsapi_public_rooms" +
 	" WHERE visibility = true" +
 	" AND (LOWER(name) LIKE LOWER($1)" +
-	" OR LOWER(topic) LIKE LOWER($1))" +
-	//" OR LOWER(ARRAY_TO_STRING(aliases, ',')) LIKE LOWER($1))" +
+	" OR LOWER(topic) LIKE LOWER($1)" +
+	" OR LOWER(aliases) LIKE LOWER($1))" + // TODO: Is there a better way to search aliases?
 	" ORDER BY joined_members DESC" +
 	" LIMIT 30 OFFSET $2"
 
@@ -83,8 +83,8 @@ const selectPublicRoomsWithLimitAndFilterSQL = "" +
 	" FROM publicroomsapi_public_rooms" +
 	" WHERE visibility = true" +
 	" AND (LOWER(name) LIKE LOWER($1)" +
-	" OR LOWER(topic) LIKE LOWER($1))" +
-	// " OR LOWER(ARRAY_TO_STRING(aliases, ',')) LIKE LOWER($1))" +
+	" OR LOWER(topic) LIKE LOWER($1)" +
+	" OR LOWER(aliases) LIKE LOWER($1))" + // TODO: Is there a better way to search aliases?
 	" ORDER BY joined_members DESC" +
 	" LIMIT $3 OFFSET $2"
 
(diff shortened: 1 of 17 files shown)


publicroomsapi/storage/sqlite3/public_rooms_table.go
```diff
@@ -39,26 +39,15 @@ var editableAttributes = []string{
...
-	-- Topic of the room (empty string if none)
 	topic TEXT NOT NULL DEFAULT '',
-	-- Is the room world readable?
 	world_readable BOOLEAN NOT NULL DEFAULT false,
-	-- Can guest join the room?
 	guest_can_join BOOLEAN NOT NULL DEFAULT false,
-	-- URL of the room avatar (empty string if none)
 	avatar_url TEXT NOT NULL DEFAULT '',
-	-- Visibility of the room: true means the room is publicly visible, false
-	-- means the room is private
```
anoadragon453: Might be nice to keep these comments?

publicroomsapi/storage/sqlite3/public_rooms_table.go
```diff
@@ -71,33 +60,33 @@ const selectPublicRoomsSQL = "" +
...
 
 const selectPublicRoomsWithFilterSQL = "" +
 	"SELECT room_id, joined_members, aliases, canonical_alias, name, topic, world_readable, guest_can_join, avatar_url" +
 	" FROM publicroomsapi_public_rooms" +
 	" WHERE visibility = true" +
 	" AND (LOWER(name) LIKE LOWER($1)" +
-	" OR LOWER(topic) LIKE LOWER($1)" +
-	" OR LOWER(ARRAY_TO_STRING(aliases, ',')) LIKE LOWER($1))" +
+	" OR LOWER(topic) LIKE LOWER($1))" +
+	//" OR LOWER(ARRAY_TO_STRING(aliases, ',')) LIKE LOWER($1))" +
```
anoadragon453: Do we want to keep this?

publicroomsapi/storage/sqlite3/public_rooms_table.go
```diff
@@ -71,33 +60,33 @@ const selectPublicRoomsSQL = "" +
...
 
 const selectPublicRoomsWithLimitAndFilterSQL = "" +
 	"SELECT room_id, joined_members, aliases, canonical_alias, name, topic, world_readable, guest_can_join, avatar_url" +
 	" FROM publicroomsapi_public_rooms" +
 	" WHERE visibility = true" +
 	" AND (LOWER(name) LIKE LOWER($1)" +
-	" OR LOWER(topic) LIKE LOWER($1)" +
-	" OR LOWER(ARRAY_TO_STRING(aliases, ',')) LIKE LOWER($1))" +
+	" OR LOWER(topic) LIKE LOWER($1))" +
+	// " OR LOWER(ARRAY_TO_STRING(aliases, ',')) LIKE LOWER($1))" +
```
anoadragon453: and this

publicroomsapi/storage/sqlite3/public_rooms_table.go
neilalexander: I suppose I should `TODO` this, as the query as it is won't work with SQLite.

publicroomsapi/storage/sqlite3/public_rooms_table.go
neilalexander: Same with this one - I'll `TODO` it.

publicroomsapi/storage/sqlite3/public_rooms_table.go
neilalexander: For now I don't think any of the SQLite queries have comments as it made them easier to read, but we can re-add those later.

Kegsay:
LGTM

PR #874, opened 2020-02-23
Added special case for m.room.member event in appservice component

### LINKED ISSUE #648: We should check if application services are interested in a membership event's target
We should special case the processing of `m.room.member` events in the appservice component to call `appservice.IsInterestedInUserID` with the event's state key on top of the existing call with the event's sender.

prateek2211:
### Pull Request Checklist

fixed #648


appservice/consumers/roomserver.go
```diff
@@ -191,6 +191,12 @@ func (s *OutputRoomEventConsumer) appserviceIsInterestedInEvent(ctx context.Cont
 		return true
 	}
 
+	if event.Type() == "m.room.member" {
```
Kegsay: Please use `gomatrixserverlib.MRoomMember` as the constant

Kegsay: Also, please do a nil check on `StateKey` e.g:
```
if event.Type() == gomatrixserverlib.MRoomMember && event.StateKey() != nil {
```

Kegsay:
Otherwise LGTM

appservice/consumers/roomserver.go
prateek2211: Done!

PR #878, opened 2020-02-28
Room versions and state resolution v2

neilalexander:
This PR plugs in state resolution v2 support by looking for room versions when persisting `m.room.create` events.


neilalexander:
Closing this as it is becoming unworkable.

PR #880, opened 2020-03-05
Add peer-to-peer support into Dendrite via libp2p and fetch

Kegsay:
This PR adds in support for running peer-to-peer Dendrite instances.

### Overview
- Dendrite can now compile to WebAssembly. `$ GOOS=js GOARCH=wasm go build -o dendrite.wasm ./cmd/dendritejs`
- Dendrite will listen for function calls from JS-land for incoming traffic (both local and federation). These function calls are called by accompanying JS-Go bridge code located at https://github.com/matrix-org/go-http-js-libp2p
- Dendrite will use a new SQL driver when running in JS: https://github.com/matrix-org/go-sqlite3-js
- There is now an in-memory LRU cache around the `KeyDB`. This is required for p2p (though all implementations benefit) in order to keep federation traffic below the threshold which would cause us to be blacklisted (denied) on the relay server.
- There's a new document `p2p.md` which contains instructions on how to set up a complete peer (including a Riot instance).

### Breaking changes

Previously, tracing defaulted to on. Now it is off by default. Set `tracing.enabled = true` in the config YAML to re-enable it.

### Notes

- We now conditionally compile the `storage.go` entry point for all microservices. These files only contain a single `NewWhateverDatabase` function, so the code duplication is minor. For WebAssembly, this means running `storage_wasm.go`. This is required in order to remove a hard dependency on `lib/pq` which does not compile under WASM. **Around 700 lines of this diff is doing this refactoring**.
- We now explicitly JSON serialise "arrays" when working under SQLite. Previously, we relied on `pq.StringArray` to do this serialisation, which as the above point makes, we cannot do anymore.


Kegsay:
The failing tests are SQLite only - which is a known issue I'm going to look into today!

PR #881, opened 2020-03-06
Try to recursively find auth events (to a point) if they are missing

neilalexander:
If auth events are missing then we should request them. This used to fail if the auth event also had a missing auth event - now we try to recursively request them to fill the gap.


federationapi/routing/send.go
```diff
@@ -185,7 +186,11 @@ func checkAllowedByState(e gomatrixserverlib.Event, stateEvents []gomatrixserver
 	return gomatrixserverlib.Allowed(e, &authUsingState)
 }
 
-func (t *txnReq) processEventWithMissingState(e gomatrixserverlib.Event) error {
+func (t *txnReq) processEventWithMissingState(e gomatrixserverlib.Event, depth int) error {
```
Kegsay: Please don't use the term `depth` as I immediately think of event depth, which this is not. Hops/count etc would be better.

federationapi/routing/send.go
```diff
@@ -211,8 +216,27 @@ func (t *txnReq) processEventWithMissingState(e gomatrixserverlib.Event) error {
...
 	}
 	// Check that the event is allowed by the state.
+retryAllowedState:
 	if err := checkAllowedByState(e, state.StateEvents); err != nil {
-		return err
+		switch missing := err.(type) {
+		case gomatrixserverlib.MissingAuthEventError:
+			// An auth event was missing so let's look up that event over federation
+			for _, s := range state.StateEvents {
+				if s.EventID() == missing.AuthEventID {
```
Kegsay: Invert and outdent please.
```
if s.EventID() != missing.AuthEventID {
    continue
}
```

neilalexander: Done

federationapi/routing/send.go
neilalexander: I've removed the recursion limit for now as @erikjohnston suggests Synapse doesn't have one and that adding one might be a failure mode for a room.

federationapi/routing/send.go
```diff
@@ -211,8 +216,27 @@ func (t *txnReq) processEventWithMissingState(e gomatrixserverlib.Event) error {
...
+					if serr := t.processEventWithMissingState(s, depth+1); serr == nil {
+						// If there was no error retrieving the event from federation then
+						// we assume that it succeeded, so retry the original state check
```
Kegsay: Should this retry count as an attempt? I feel like it probably should, or else this could potentially infinite loop.

federationapi/routing/send.go
```diff
@@ -211,8 +216,27 @@ func (t *txnReq) processEventWithMissingState(e gomatrixserverlib.Event) error {
...
+						goto retryAllowedState
+					} else {
+						// An error occurred so let's not do anything further
+						return err
```
Kegsay: This should be `serr`. It feels weird to be doing the `== nil` case first then this. Why are we doing it this way?

federationapi/routing/send.go
neilalexander: The outcomes are either:

1. We fail `t.federation.LookupState` at which point we give up because that won't return a `MissingAuthEventError`
1. We fail `state.Check` at which point we give up for the same reason as above
1. We pass both of the above and hit `checkAllowedByState` again, but this will only recurse if another event is actually missing (since if we retrieved the original one, it'll now satisfy or hard-fail the above two conditions)
1. If all of the above succeeds then no more recursions are made anyway

federationapi/routing/send.go
neilalexander: I've refactored it now, should be more straight-forward.

federationapi/routing/send.go
Kegsay: I just don't like infinite loops which lack clear and obvious terminating cases, *especially* when the data is being provided by potentially malicious parties outside our control.

I can't give you a failure case, but I worry some pathological circular event refs could cause this code to run forever.

Kegsay: Or alternatively even a case where the HS returns an empty set (without triggering an error) could cause this code to loop.

neilalexander: I know what you mean, although this is apparently what Synapse does and I'm hasten to do something different in case we end up with obviously broken rooms as a result. 

Since the only event that can have neither prev events nor auth events are `m.room.create` events, anything else with empty sets should fail `state.Check` before reaching this point.

Kegsay: Okay, I'll have a think about how we can best protect ourselves from this and other things that we need to do for Synapse parity but are just yucky.

PR #882, opened 2020-03-06
bugfix: Fix a race condition when creating guest accounts

Kegsay:
It was possible to both select the same next numeric ID and then both
attempt to INSERT this into the table. This would cause a UNIQUE violation
which then presented itself as an error in sqlite because it does not
implement `common.IsUniqueConstraintViolationErr`.

The fix here is NOT to implement `common.IsUniqueConstraintViolationErr`
otherwise the 2 users would get the SAME guest account. Instead, all of
these operations should be done inside a transaction. This is what this
PR does.

This was the cause of sytest 226 failing in 87283e9de785f5153c5cf9b326d2640e202a36b3

### DIFF
--- a/clientapi/auth/storage/accounts/postgres/accounts_table.go
+++ b/clientapi/auth/storage/accounts/postgres/accounts_table.go
@@ -91,10 +91,10 @@ func (s *accountsStatements) prepare(db *sql.DB, server gomatrixserverlib.Server
 // this account will be passwordless. Returns an error if this account already exists. Returns the account
 // on success.
 func (s *accountsStatements) insertAccount(
-	ctx context.Context, localpart, hash, appserviceID string,
+	ctx context.Context, txn *sql.Tx, localpart, hash, appserviceID string,
 ) (*authtypes.Account, error) {
 	createdTimeMS := time.Now().UnixNano() / 1000000
-	stmt := s.insertAccountStmt
+	stmt := txn.Stmt(s.insertAccountStmt)
 
 	var err error
 	if appserviceID == "" {
--- a/clientapi/auth/storage/accounts/postgres/accounts_table.go
+++ b/clientapi/auth/storage/accounts/postgres/accounts_table.go
@@ -146,8 +146,12 @@ func (s *accountsStatements) selectAccountByLocalpart(
 }
 
 func (s *accountsStatements) selectNewNumericLocalpart(
-	ctx context.Context,
+	ctx context.Context, txn *sql.Tx,
 ) (id int64, err error) {
-	err = s.selectNewNumericLocalpartStmt.QueryRowContext(ctx).Scan(&id)
+	stmt := s.selectNewNumericLocalpartStmt
+	if txn != nil {
+		stmt = txn.Stmt(stmt)
+	}
+	err = stmt.QueryRowContext(ctx).Scan(&id)
 	return
 }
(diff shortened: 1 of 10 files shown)


clientapi/auth/storage/accounts/postgres/accounts_table.go
```diff
@@ -146,8 +146,12 @@ func (s *accountsStatements) selectAccountByLocalpart(
 }
 
 func (s *accountsStatements) selectNewNumericLocalpart(
-	ctx context.Context,
+	ctx context.Context, txn *sql.Tx,
 ) (id int64, err error) {
-	err = s.selectNewNumericLocalpartStmt.QueryRowContext(ctx).Scan(&id)
+	stmt := s.selectNewNumericLocalpartStmt
+	if txn != nil {
+		stmt = txn.Stmt(stmt)
```
neilalexander: This is the sort of thing we used to use `common.TxStmt` for just so it was a bit more terse —
```
func TxStmt(transaction *sql.Tx, statement *sql.Stmt) *sql.Stmt {
	if transaction != nil {
		statement = transaction.Stmt(statement)
	}
	return statement
}
```

clientapi/auth/storage/accounts/sqlite3/storage.go
```diff
@@ -134,13 +159,14 @@ func (d *Database) CreateAccount(
...
-	if err := d.profiles.insertProfile(ctx, localpart); err != nil {
+	if err := d.profiles.insertProfile(ctx, txn, localpart); err != nil {
 		if common.IsUniqueConstraintViolationErr(err) {
 			return nil, nil
 		}
 		return nil, err
 	}
-	if err := d.SaveAccountData(ctx, localpart, "", "m.push_rules", `{
+
+	if err := d.accountDatas.insertAccountData(ctx, txn, localpart, "", "m.push.rules", `{
```
neilalexander: Looks like a typo here - this should be `m.push_rules`.

clientapi/auth/storage/accounts/postgres/accounts_table.go
Kegsay: I want to kill `common.TxStmt` at some point because in 99% of cases we should always be using the `txn` and never passing in `nil`. Making it terser makes it less clear what is going on IMO.

clientapi/auth/storage/accounts/sqlite3/storage.go
Kegsay: Done. https://github.com/matrix-org/dendrite/pull/882/commits/42a9000d4dfc1f743384e3d59bd948fa4bb08598

clientapi/auth/storage/accounts/postgres/accounts_table.go
neilalexander: In that case is it better if the reference code hard-fails if `txn` is `nil`?

Kegsay: In this case I am explicitly passing in `nil` when people call the public function `GetNewNumericLocalpart(ctx context.Context) (int64, error)` - I want to kill this function too, but I also want to keep this PR on point which is why I'm doing the `nil` thing.

neilalexander: No problem - let's merge this as-is then and maybe we can add a kind of assertion function later that checks `txn` is not `nil` and `panic()` if it is (that seems like a good excuse to panic as it would mean something is seriously wrong).

Kegsay: SGTM!

PR #885, opened 2020-03-10
p2p: Use a single ed25519 key for event signing/verifying and p2p peer IDs

Kegsay:
This allows us to fetch keys without making network calls as we can extract it from the server name. This PR adds support for this.

The main benefit to this is in the 3-server scenario:
 - A and B are in a room.
 - A disappears (laptop shut, service worker killed, whatever)
 - C wants to join the room (using a new room alias set up by B)

Previously, C could not join as it could not verify events in the room because it lacked A's public key. With this PR, C can now join the room as it can derive A's public key from the server name.


cmd/dendritejs/main.go
```diff
@@ -69,10 +68,10 @@ func createFederationClient(cfg *config.Dendrite, node *go_http_js_libp2p.P2pLoc
 	return fed
 }
 
-func createP2PNode() (serverName string, node *go_http_js_libp2p.P2pLocalNode) {
-	hosted := "/dns4/ws-star.discovery.libp2p.io/tcp/443/wss/p2p-websocket-star"
-	_ = "/ip4/127.0.0.1/tcp/9090/ws/p2p-websocket-star/"
-	node = go_http_js_libp2p.NewP2pLocalNode("org.matrix.p2p.experiment", []string{hosted})
+func createP2PNode(privKey ed25519.PrivateKey) (serverName string, node *go_http_js_libp2p.P2pLocalNode) {
+	_ = "/dns4/ws-star.discovery.libp2p.io/tcp/443/wss/p2p-websocket-star"
```
neilalexander: No-op

cmd/dendritejs/main.go
```diff
@@ -69,10 +68,10 @@ func createFederationClient(cfg *config.Dendrite, node *go_http_js_libp2p.P2pLoc
...
+	hosted := "/ip4/127.0.0.1/tcp/9090/ws/p2p-websocket-star/"
```
neilalexander: Should this be a public rendezvous for now or is the plan to wait for a Matrix rendezvous? If not then might be worth dropping this requirement into the P2P readme.

cmd/dendritejs/main.go
Kegsay: atm I'm using this as a convenient way to swap between IPFS/local - though I won't ever use IPFS again because it barfs on ed25519 peer IDs, so will remove.

cmd/dendritejs/main.go
Kegsay: It already is in the README - https://github.com/matrix-org/dendrite/blob/master/p2p.md#rendezvous - this will be configurable in the future.

cmd/dendritejs/main.go
Kegsay: Done

PR #900, opened 2020-03-11
Check if user has the power level to edit the room visibility

### LINKED ISSUE #642: Check if user has the power level to edit the room visibility in the public rooms directory

prateek2211:
Fixed #642 

### Pull Request Checklist

* [X] I have added any new tests that need to pass to `testfile` as specified in [docs/sytest.md](https://github.com/matrix-org/dendrite/blob/master/docs/sytest.md)


publicroomsapi/directory/directory.go
```diff
@@ -54,11 +57,50 @@ func GetVisibility(
...
+	var queryEventsRes api.QueryLatestEventsAndStateResponse
+	err = queryAPI.QueryLatestEventsAndState(req.Context(), &queryEventsReq, &queryEventsRes)
+	if err != nil {
+		util.GetLogger(req.Context()).WithError(err).Error("could not query events from room")
+		return jsonerror.InternalServerError()
+	}
+	power, _ := gomatrixserverlib.NewPowerLevelContentFromEvent(queryEventsRes.StateEvents[0])
+
+	//Check if the user's power is greater than power required to change m.room.aliases event
+	if power.UserLevel(dev.UserID) < power.EventLevel(gomatrixserverlib.MRoomAliases, true) {
```
Cnly: Maybe it'll be better to elaborate on why we should use this power level to do this check? Also, checking this power level doesn't necessarily allow server admins to set visibility (though admin accounts haven't been implemented in dendrite yet IIRC, so that's still open for discussion).

prateek2211: We require the user to be a "moderator" in the room. We can do this by checking if user would (theoretically) be able to change the m.room.aliases events.
And because admin account isn't implemented in dendrite, I didn't add a check for user being admin

publicroomsapi/directory/directory.go
```diff
@@ -54,11 +57,50 @@ func GetVisibility(
...
+	}
+	var queryEventsRes api.QueryLatestEventsAndStateResponse
+	err = queryAPI.QueryLatestEventsAndState(req.Context(), &queryEventsReq, &queryEventsRes)
+	if err != nil {
+		util.GetLogger(req.Context()).WithError(err).Error("could not query events from room")
+		return jsonerror.InternalServerError()
+	}
+	power, _ := gomatrixserverlib.NewPowerLevelContentFromEvent(queryEventsRes.StateEvents[0])
+
+	//Check if the user's power is greater than power required to change m.room.aliases event
```
Cnly:
```suggestion
	// Check if the user's power is greater than power required to change m.room.aliases event
```

publicroomsapi/directory/directory.go
```diff
@@ -54,11 +57,50 @@ func GetVisibility(
 }
 
 // SetVisibility implements PUT /directory/list/room/{roomID}
-// TODO: Check if user has the power level to edit the room visibility
```
Cnly: Perhaps can be replaced with a new TODO that talks about the to-be-added admin privileges :)

prateek2211: done

publicroomsapi/directory/directory.go
prateek2211: done

publicroomsapi/directory/directory.go
```diff
@@ -54,11 +57,51 @@ func GetVisibility(
...
+			StateKey:  "",
+		}},
+	}
+	var queryEventsRes api.QueryLatestEventsAndStateResponse
+	err = queryAPI.QueryLatestEventsAndState(req.Context(), &queryEventsReq, &queryEventsRes)
+	if err != nil {
+		util.GetLogger(req.Context()).WithError(err).Error("could not query events from room")
+		return jsonerror.InternalServerError()
+	}
+	power, _ := gomatrixserverlib.NewPowerLevelContentFromEvent(queryEventsRes.StateEvents[0].Event)
```
neilalexander, line 95: Looking at this, I am wondering whether it is correct to assert that `StateEvents[0]` is always correct here?

prateek2211: I suppose, since a room contains only one state event of type `m.room.power_levels`,  StateEvents[0] would give the correct event.
Though, it would be better to add `len(queryEventsRes.StateEvents) == 0` check.

neilalexander: Right - might be good just to avoid a possible panic if it's not set.

prateek2211: done

PR #906, opened 2020-03-16
Update room version descriptors, add error handling

neilalexander:
This updates the room version descriptors to use the new `gomatrixserverlib.RoomVersion` type, adds some error handling.

### DIFF
--- a/roomserver/storage/postgres/rooms_table.go
+++ b/roomserver/storage/postgres/rooms_table.go
@@ -93,11 +93,12 @@ func (s *roomStatements) prepare(db *sql.DB) (err error) {
 }
 
 func (s *roomStatements) insertRoomNID(
-	ctx context.Context, txn *sql.Tx, roomID string,
+	ctx context.Context, txn *sql.Tx,
+	roomID string, roomVersion gomatrixserverlib.RoomVersion,
 ) (types.RoomNID, error) {
 	var roomNID int64
 	stmt := common.TxStmt(txn, s.insertRoomNIDStmt)
-	err := stmt.QueryRowContext(ctx, roomID).Scan(&roomNID)
+	err := stmt.QueryRowContext(ctx, roomID, roomVersion).Scan(&roomNID)
 	return types.RoomNID(roomNID), err
 }
 
--- a/roomserver/storage/postgres/rooms_table.go
+++ b/roomserver/storage/postgres/rooms_table.go
@@ -22,6 +22,7 @@ import (
 	"github.com/lib/pq"
 	"github.com/matrix-org/dendrite/common"
 	"github.com/matrix-org/dendrite/roomserver/types"
+	"github.com/matrix-org/gomatrixserverlib"
 )
 
 const roomsSchema = `
@@ -165,8 +166,8 @@ func (s *roomStatements) updateLatestEventNIDs(
 
 func (s *roomStatements) selectRoomVersionForRoomNID(
 	ctx context.Context, txn *sql.Tx, roomNID types.RoomNID,
-) (int64, error) {
-	var roomVersion int64
+) (gomatrixserverlib.RoomVersion, error) {
+	var roomVersion gomatrixserverlib.RoomVersion
 	stmt := common.TxStmt(txn, s.selectRoomVersionForRoomNIDStmt)
 	err := stmt.QueryRowContext(ctx, roomNID).Scan(&roomVersion)
 	return roomVersion, err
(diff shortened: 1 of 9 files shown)


roomserver/version/version.go
```diff
@@ -15,84 +15,58 @@
...
+// server.
+func GetRoomVersions() map[gomatrixserverlib.RoomVersion]RoomVersionDescription {
 	return roomVersions
 }
 
-func GetSupportedRoomVersions() map[RoomVersionID]RoomVersionDescription {
-	versions := make(map[RoomVersionID]RoomVersionDescription)
+// GetSupportedRoomVersions returns a map of descriptions for room
+// versions that are supported by this homeserver.
+func GetSupportedRoomVersions() map[gomatrixserverlib.RoomVersion]RoomVersionDescription {
```
Kegsay: Nit: I'd prefer if we drop the `Get` as it isn't a common Go idiom - https://golang.org/doc/effective_go.html#Getters

roomserver/version/version.go
```diff
@@ -15,84 +15,58 @@
...
+// namely whether it is marked as supported or stable in this server
+// version.
 type RoomVersionDescription struct {
-	Supported                 bool
-	Stable                    bool
-	StateResolution           state.StateResolutionVersion
-	EventFormat               EventFormatID
-	EnforceSigningKeyValidity bool
+	Supported bool
+	Stable    bool
```
Kegsay: Please document what you actually mean by these two flags. If they refer to specific spec terms, please link to the spec.

roomserver/version/version.go
```diff
@@ -15,84 +15,58 @@
...
 	},
-	RoomVersionV5: RoomVersionDescription{
-		Supported:                 false,
-		Stable:                    true,
-		StateResolution:           state.StateResolutionAlgorithmV2,
-		EventFormat:               EventFormatV3,
-		EnforceSigningKeyValidity: true,
+	gomatrixserverlib.RoomVersionV5: RoomVersionDescription{
+		Supported: false,
+		Stable:    false,
```
Kegsay: Why did this swap to `false` when it was `true` before?

roomserver/storage/postgres/rooms_table.go
```diff
@@ -42,13 +43,13 @@ CREATE TABLE IF NOT EXISTS roomserver_rooms (
     state_snapshot_nid BIGINT NOT NULL DEFAULT 0,
     -- The version of the room, which will assist in determining the state resolution
     -- algorithm, event ID format, etc.
-    room_version BIGINT NOT NULL DEFAULT 1
+    room_version TEXT NOT NULL DEFAULT '1'
```
Kegsay: Please don't default here, as it means to update the default room version we need to remember to do it here and `GetDefaultRoomVersion()`.

roomserver/storage/sqlite3/rooms_table.go
```diff
@@ -31,13 +32,13 @@ const roomsSchema = `
...
     state_snapshot_nid INTEGER NOT NULL DEFAULT 0,
-    room_version INTEGER NOT NULL DEFAULT 1
+    room_version TEXT NOT NULL DEFAULT '1'
   );
 `
 
 // Same as insertEventTypeNIDSQL
 const insertRoomNIDSQL = `
-	INSERT INTO roomserver_rooms (room_id) VALUES ($1)
+	INSERT INTO roomserver_rooms (room_id, room_version) VALUES ($1, $2)
```
Kegsay: We've added an extra param here, but where is the matching bit in code?

roomserver/version/version.go
neilalexander: Will do.

roomserver/version/version.go
neilalexander: Will do.

roomserver/version/version.go
neilalexander: For now I am resetting the baseline back to supporting only v1 rooms. I'll re-add others into the hints as the model makes more sense.

roomserver/storage/postgres/rooms_table.go
neilalexander: The main reason for this was because a `m.room.create` event that doesn't contain a `room_version` option is typically a v1 room. Any room of a later version must contain `room_version` therefore we will always have something to overwrite the default with in those cases.

Kegsay: I'd rather that check resided in code rather than having important semantics tucked away in an easily missed create table statement. Especially since this line gives the impression that by default we support v1 (e.g room version is an optional param) whereas what it's really saying is "by default, if and only if the room version is missing, set it to v1".

roomserver/storage/sqlite3/rooms_table.go
neilalexander: It's committed now.

roomserver/storage/postgres/rooms_table.go
neilalexander: I have removed the default for now. We can handle it in code instead.

PR #912, opened 2020-03-16
Implement gomatrixserverlib.HeaderedEvent in roomserver query API

neilalexander:
This PR implements `gomatrixserverlib.HeaderedEvent` in the roomserver query API. This new struct wraps the room version (and potentially other things in the future, if needed) around the struct.

The logic here is that in the `json.Unmarshaller` implementation for `gomatrixserverlib.HeaderedEvent`, we can intercept the headers to get the room version, prepare the event struct accordingly and then complete unmarshalling the underlying `gomatrixserverlib.Event` into it transparently.

After this PR I will do the same for Kafka.

### DIFF
--- a/roomserver/query/query.go
+++ b/roomserver/query/query.go
@@ -89,6 +89,10 @@ type RoomserverQueryAPIDatabase interface {
 	EventStateKeys(
 		context.Context, []types.EventStateKeyNID,
 	) (map[types.EventStateKeyNID]string, error)
+	// Look up the room version for a given room.
+	GetRoomVersionForRoom(
+		ctx context.Context, roomNID types.RoomNID,
+	) (gomatrixserverlib.RoomVersion, error)
 }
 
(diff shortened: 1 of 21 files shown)


roomserver/query/query.go
```diff
@@ -89,6 +89,10 @@ type RoomserverQueryAPIDatabase interface {
 	EventStateKeys(
 		context.Context, []types.EventStateKeyNID,
 	) (map[types.EventStateKeyNID]string, error)
+	// Look up the room version for a given room.
+	GetRoomVersionForRoom(
+		ctx context.Context, roomNID types.RoomNID,
+	) (gomatrixserverlib.RoomVersion, error)
```
Kegsay: Do we still need this if we're chucking room versions around with each event?

Kegsay: oh sorry, this is the database, not the internal API!

PR #913, opened 2020-03-16
bugfix: Fix #908 by setting the correct state after the event

### LINKED ISSUE #908: The first message sent into a room over federation is not received
Common in P2P.

Kegsay:
Previously, this would only happen if the state already existed
previously!

### DIFF
--- a/federationapi/routing/send.go
+++ b/federationapi/routing/send.go
@@ -116,7 +116,7 @@ func (t *txnReq) processTransaction() (*gomatrixserverlib.RespSend, error) {
 			results[e.EventID()] = gomatrixserverlib.PDUResult{
 				Error: err.Error(),
 			}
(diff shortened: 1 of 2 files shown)


Kegsay:
Manually tested and this works well.

federationapi/routing/send.go
```diff
@@ -116,6 +116,7 @@ func (t *txnReq) processTransaction() (*gomatrixserverlib.RespSend, error) {
 			results[e.EventID()] = gomatrixserverlib.PDUResult{
 				Error: err.Error(),
 			}
+			util.GetLogger(t.context).Error("Failed to process incoming federation event, skipping. Event ID:", e.EventID(), " Err:", err)
```
neilalexander: I think this should be something like:
```
util.GetLogger(t.context).WithError(err).Error("Failed to process...")
```

Kegsay: Done, and dropped it to warning since it's not a huge problem but may signify a problem, as in this case!

PR #915, opened 2020-03-17
Add room version to room create request, persist in storage

neilalexander:
This PR does two things:

1. Adds the room version to the room create request in the Client API, and complains if the room version is not supported
1. Extracts the room version from the room create event when persisting the room to storage

### DIFF
--- a/clientapi/jsonerror/jsonerror.go
+++ b/clientapi/jsonerror/jsonerror.go
@@ -124,6 +124,12 @@ func GuestAccessForbidden(msg string) *MatrixError {
 	return &MatrixError{"M_GUEST_ACCESS_FORBIDDEN", msg}
 }
 
+// UnsupportedRoomVersion is an error which is returned when the client
+// requests a room with a version that is unsupported.
+func UnsupportedRoomVersion(msg string) *MatrixError {
+	return &MatrixError{"M_UNSUPPORTED_ROOM_VERSION", msg}
+}
+
 // LimitExceededError is a rate-limiting error.
 type LimitExceededError struct {
 	MatrixError
--- a/roomserver/storage/postgres/storage.go
+++ b/roomserver/storage/postgres/storage.go
@@ -77,18 +77,9 @@ func (d *Database) StoreEvent(
 	// Note that the below logic depends on the m.room.create event being the
 	// first event that is persisted to the database when creating or joining a
 	// room.
-	roomVersion := roomserverVersion.DefaultRoomVersion()
-	// Look for m.room.create events.
-	if event.Type() == gomatrixserverlib.MRoomCreate {
-		var createContent gomatrixserverlib.CreateContent
-		// The m.room.create event contains an optional "room_version" key in
-		// the event content, so we need to unmarshal that first.
-		if err = json.Unmarshal(event.Content(), &createContent); err == nil {
-			if createContent.RoomVersion != nil {
-				// A room version was specified in the event content.
-				roomVersion = *createContent.RoomVersion
-			}
-		}
+	var roomVersion gomatrixserverlib.RoomVersion
+	if roomVersion, err = extractRoomVersionFromCreateEvent(event); err != nil {
+		return 0, types.StateAtEvent{}, err
 	}
 
 	if roomNID, err = d.assignRoomNID(ctx, nil, event.RoomID(), roomVersion); err != nil {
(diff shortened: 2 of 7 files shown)


clientapi/jsonerror/jsonerror.go
```diff
@@ -124,6 +124,12 @@ func GuestAccessForbidden(msg string) *MatrixError {
 	return &MatrixError{"M_GUEST_ACCESS_FORBIDDEN", msg}
 }
 
+// UnsupportedRoomVersion is an error which is returned when the client
+// requests a room with a version that is unsupported.
+func UnsupportedRoomVersion(msg string) *MatrixError {
```
Kegsay: Please link to the spec for this error code.

Kegsay: I appreciate it the other errors don't - they should be. The purpose is to make clear that these are Matrix spec error codes and not made up Dendrite error codes.

neilalexander: There is a link at the top by `type MatrixError struct` which defines these error codes :-)

roomserver/storage/postgres/storage.go
```diff
@@ -121,6 +134,27 @@ func (d *Database) StoreEvent(
 	}, nil
 }
 
+func extractRoomVersionFromCreateEvent(event gomatrixserverlib.Event) (
+	roomVersion gomatrixserverlib.RoomVersion, err error,
```
Kegsay: Please don't use named return args here. I just spent five minutes writing a comment saying how the API shape is weird until I saw that `:144` assigns the return version to the default value. It would be clearer to explicitly return stuff here.

Kegsay: Though the API shape is still a bit weird in that we return a zero-initialised version and *no error* if the event isn't a create event.

roomserver/storage/sqlite3/storage.go
```diff
@@ -90,8 +93,18 @@ func (d *Database) StoreEvent(
...
+		// https://matrix.org/docs/spec/client_server/r0.6.0#post-matrix-client-r0-createroom
+		// Note that the below logic depends on the m.room.create event being the
+		// first event that is persisted to the database when creating or joining a
+		// room.
+		var roomVersion gomatrixserverlib.RoomVersion
+		if roomVersion, err = extractRoomVersionFromCreateEvent(event); err != nil {
+			return err
+		}
+
+		if roomNID, err = d.assignRoomNID(ctx, txn, event.RoomID(), roomVersion); err != nil {
```
Kegsay: StoreEvent now has a reasonably significant change when handling create events (the first event) vs anything else. I would be tempted to factor this logic out, and keep StoreEvent for the common case where the room NID + version is set.

Kegsay: e.g `StoreInitialEvent(Event, RoomVersion)`

clientapi/jsonerror/jsonerror.go
Kegsay: Very well! Ignore me then :)

roomserver/storage/postgres/storage.go
```diff
@@ -68,8 +71,18 @@ func (d *Database) StoreEvent(
...
 	}
 
-	// TODO: Room version here
-	if roomNID, err = d.assignRoomNID(ctx, nil, event.RoomID(), "1"); err != nil {
+	// Get the default room version. If the client doesn't supply a room_version
+	// then we will use our configured default to create the room.
+	// https://matrix.org/docs/spec/client_server/r0.6.0#post-matrix-client-r0-createroom
+	// Note that the below logic depends on the m.room.create event being the
+	// first event that is persisted to the database when creating or joining a
+	// room.
```
Kegsay: Can we even make that guarantee over federation? I don't think we can.

roomserver/storage/sqlite3/storage.go
neilalexander: So this (and the previous point) are interesting because `StoreEvent` works on the assumption that `assignRoomNID` is idempotent - if a NID already exists, it just returns it. We can remodel into two separate functions.

roomserver/storage/postgres/storage.go
neilalexander: I think we would have to - without the `m.room.create` event, we can't auth anything in the newly joined room, so it really should be the first event that is sent to us. If we request those initial events somewhere then we can at least make the order of persistence determinate.

neilalexander: Looking at this, `func (r RespSendJoin) Check()` in `clientapi/routing/joinroom.go` currently persists the events in the order that they appear in the `send_join` response, which is the same as a regular state response and would therefore include the `m.room.create` event.

With a small update to this function, we can look for and persist the `m.room.create` event first in all cases.

Kegsay: Okay, we'll see how badly things explode.

roomserver/storage/sqlite3/storage.go
Kegsay: Fewer db calls ftw? :)

Kegsay: This doesn't have to happen in this PR.

PR #917, opened 2020-03-17
bugfix: Fix a bug which caused failures to join rooms over federation

Kegsay:
The cause of this was the semantics of `/send_join`'s `auth_chain` response.

Previously, we would only send back the auth chain *for the join event* and
not the entire room state. However, we would then try to check that the
room state is valid, and then be missing auth events.

Now, we send back the entire auth chain for all room state in `/send_join`.
The spec needs to be clarified that this is what the chain should be.


neilalexander:
OK after linter.

neilalexander:
Incidentally this is another of many places where we seem to easily hit gocyclo of 13 instead of 12, so I propose that we increase the threshold in #918.

Issue #909, opened 2020-03-16
Invites over federation don't seem to work


Kegsay:
This could be a P2P thing though, I don't know yet.

neilalexander:
As a sidenote here, Dendrite only implements the `v1` endpoint for invites over federation. The spec states:

> Servers should prefer to use the v2 API for invites instead of the v1 API. 

... therefore I suggest we opt for removing it and implementing the `v2` endpoint only, so that we always have the room version available at invite time (since this matters when we assign the room NID for the first time).

neilalexander:
Fixed following #977, #979, #982, #1007.

PR #918, opened 2020-03-17
Get room versions from database

neilalexander:
This PR updates some code paths in the room server to get the room version from the database.

Currently this doesn't deal with invite events, as right now we only implement  the old `/_matrix/federation/v1/invite` endpoint instead of `/_matrix/federation/v2/invite`, so we don't have awareness of room versions there yet. Suggest that we add fixing invites and implementing v2 of this endpoint to a greater piece of work (#909).

This PR also increases the `gocyclo` linting threshold to 13 instead of 12, since we seem to be hitting that specific number a lot.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -102,7 +102,7 @@ linters-settings:
     #local-prefixes: github.com/org/project
   gocyclo:
     # minimal code complexity to report, 30 by default (but we recommend 10-20)
-    min-complexity: 12
+    min-complexity: 13
   maligned:
     # print struct with more effective memory layout or not, false by default
     suggest-new: true
--- a/roomserver/alias/alias.go
+++ b/roomserver/alias/alias.go
@@ -25,7 +25,6 @@ import (
 	"github.com/matrix-org/dendrite/common"
 	"github.com/matrix-org/dendrite/common/config"
 	roomserverAPI "github.com/matrix-org/dendrite/roomserver/api"
-	"github.com/matrix-org/dendrite/roomserver/types"
 	"github.com/matrix-org/gomatrixserverlib"
 	"github.com/matrix-org/util"
 )
@@ -47,11 +46,9 @@ type RoomserverAliasAPIDatabase interface {
 	// Remove a given room alias.
 	// Returns an error if there was a problem talking to the database.
 	RemoveRoomAlias(ctx context.Context, alias string) error
-	// Return the room NID for a room ID.
-	RoomNID(ctx context.Context, roomID string) (types.RoomNID, error)
 	// Look up the room version for a given room.
 	GetRoomVersionForRoom(
-		ctx context.Context, roomNID types.RoomNID,
+		ctx context.Context, roomID string,
 	) (gomatrixserverlib.RoomVersion, error)
 }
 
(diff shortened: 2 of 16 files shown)


.golangci.yml
```diff
@@ -102,7 +102,7 @@ linters-settings:
     #local-prefixes: github.com/org/project
   gocyclo:
     # minimal code complexity to report, 30 by default (but we recommend 10-20)
-    min-complexity: 12
+    min-complexity: 13
```
Kegsay: I'm unhappy about this, but willing to do it because it's just a bump of 1. No more than that please. The refactor I did in https://github.com/matrix-org/dendrite/pull/917/commits/5f7ff41ed96dd460e3d6ef8278d8e8723adab6c3 because of gocyclo I think feels a lot better, and I wouldn't have done it with this at 13.

roomserver/alias/alias.go
```diff
@@ -46,6 +47,12 @@ type RoomserverAliasAPIDatabase interface {
 	// Remove a given room alias.
 	// Returns an error if there was a problem talking to the database.
 	RemoveRoomAlias(ctx context.Context, alias string) error
+	// Return the room NID for a room ID.
+	RoomNID(ctx context.Context, roomID string) (types.RoomNID, error)
+	// Look up the room version for a given room.
+	GetRoomVersionForRoom(
+		ctx context.Context, roomNID types.RoomNID,
```
Kegsay: Nowhere else in the interface do we pass in room NIDs, why do that here? Leave it as an implementation detail.

roomserver/query/query.go
```diff
@@ -234,8 +238,15 @@ func (r *RoomserverQueryAPI) QueryEventsByID(
...
 
 	for _, event := range events {
-		// TODO: Room version here
-		roomVersion := gomatrixserverlib.RoomVersionV1
+		roomNID, nerr := r.DB.GetRoomNIDForEventID(ctx, event.EventID())
+		if nerr != nil {
+			return nerr
+		}
+
+		roomVersion, verr := r.DB.GetRoomVersionForRoom(ctx, roomNID)
```
Kegsay: I would prefer `GetRoomVersionForRoom(ctx, roomID)`

roomserver/query/query.go
```diff
@@ -93,6 +93,10 @@ type RoomserverQueryAPIDatabase interface {
 	GetRoomVersionForRoom(
 		ctx context.Context, roomNID types.RoomNID,
 	) (gomatrixserverlib.RoomVersion, error)
+	// Look up the room NID that an event ID appears in.
+	GetRoomNIDForEventID(
+		ctx context.Context, eventID string,
+	) (types.RoomNID, error)
```
Kegsay: Why are we doing this? In every single place we call this it's doing `event.EventID()`, why not just directly use `event.RoomID()` and use the existing `RoomNID(ctx context.Context, roomID string) (types.RoomNID, error)`? I don't want the API surface to get bigger than it already is!

.golangci.yml
neilalexander: I've increased this by 1 because there are a few places now where adding *anything*, even simple things, takes us over the cyclo limit due to other changes, like in SQLite where the `common.WithTransaction` wraps automatically take a cyclo away from us, or in cases where we now care about room versions and previously didn't. There isn't otherwise a need to refactor those things and to do so would make some code paths look *less* clear.

roomserver/alias/alias.go
neilalexander: I've stripped NIDs from the interface.

roomserver/query/query.go
neilalexander: Done.

roomserver/query/query.go
neilalexander: Have changed it. Some of this logic came from the old PR which admittedly doesn't make as much sense now. Thanks for pointing it out.

PR #920, opened 2020-03-17
Log errors from rows.Close

### LINKED ISSUE #902: Log errors from rows.Close
We currently `nolint:errcheck` them. Make a wrapper function `LogIfError` which just logs the error so we don't drop it to the floor. Otherwise do nothing.

prateek2211:
Fixed #902 
Logs the error from row.Close() ,so we don't drop it to the floor
### Pull Request Checklist

* [X] I have added any new tests that need to pass to `testfile` as specified in [docs/sytest.md](https://github.com/matrix-org/dendrite/blob/master/docs/sytest.md)

### DIFF
--- a/clientapi/auth/storage/accounts/postgres/account_data_table.go
+++ b/clientapi/auth/storage/accounts/postgres/account_data_table.go
@@ -92,7 +92,7 @@ func (s *accountDataStatements) selectAccountData(
 	if err != nil {
 		return
 	}
-	defer common.LogIfError(ctx, rows.Close(), "selectAccountData: rows.close() failed")
+	defer common.CloseAndLogIfError(ctx, rows, "selectAccountData: rows.close() failed")
 
 	global = []gomatrixserverlib.ClientEvent{}
 	rooms = make(map[string][]gomatrixserverlib.ClientEvent)
--- a/clientapi/auth/storage/accounts/postgres/account_data_table.go
+++ b/clientapi/auth/storage/accounts/postgres/account_data_table.go
@@ -92,7 +92,7 @@ func (s *accountDataStatements) selectAccountData(
 	if err != nil {
 		return
 	}
-	defer common.LogIfError(ctx, rows.Close(), "rows.close() failed")
+	defer common.LogIfError(ctx, rows.Close(), "selectAccountData: rows.close() failed")
 
 	global = []gomatrixserverlib.ClientEvent{}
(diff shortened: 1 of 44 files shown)


Kegsay:
Please update your branch to the latest `master`.

prateek2211:
done

clientapi/auth/storage/accounts/postgres/account_data_table.go
```diff
@@ -90,7 +91,7 @@ func (s *accountDataStatements) selectAccountData(
 	if err != nil {
 		return
 	}
-	defer rows.Close() // nolint: errcheck
+	defer common.LogIfError(ctx, rows.Close(), "rows.close() failed")
```
Kegsay: Contextual error messages would help here e.g by embedding the name of the calling function:
```
	defer common.LogIfError(ctx, rows.Close(), "insertAccountData: rows.close() failed")
```

prateek2211: everywhere or just this one?

common/log.go
```diff
@@ -156,3 +158,9 @@ func setupFileHook(hook config.LogrusHook, level logrus.Level, componentName str
 		),
 	})
 }
+
+func LogIfError(ctx context.Context, err error, message string) {
+	if err != nil {
+		util.GetLogger(ctx).WithError(err).Error(message)
```
Kegsay: `util.GetLogger` will not accept a `nil` Context. We don't happen to pass in a nil context anywhere I can see, but it might be nice to sanity check with something like:
```
if ctx == nil {
    ctx = context.TODO()
}
```
which will save people a potential headache.

clientapi/auth/storage/accounts/postgres/account_data_table.go
Kegsay: Everywhere please! (sorry I know it's a pain)

prateek2211:
@Kegsay made the changes as you suggested. Could you please review it

prateek2211:
yeah, got it

Kegsay:
Alternatively, it might be easier and nicer to change the signature to:
```
func LogIfError(ctx context.Context, closer io.Closer, message string)
```
and then call `err := closer.Close()` in the body of `LogIfError. This saves you having to bloat all the callsites. You'd then pass in `rows` instead of `rows.Close()`.

PR #921, opened 2020-03-18
Implement QueryRoomVersionForRoom internal API

neilalexander:
This implements a `QueryRoomVersionForRoom` API, since there are a couple of places where we need to know the room version ahead of the room server:

1. The client API, when building a new event in response to a `/send` or similar
1. The federation API, when receiving events from a room over federation and performing auth checks etc

This also uses the `gomatrixserverlib.RoomVersion` type in `QueryRoomVersionCapabilities`.

### DIFF
--- a/roomserver/api/query.go
+++ b/roomserver/api/query.go
@@ -251,8 +251,18 @@ type QueryRoomVersionCapabilitiesRequest struct{}
 
 // QueryRoomVersionCapabilitiesResponse is a response to QueryServersInRoomAtEventResponse
 type QueryRoomVersionCapabilitiesResponse struct {
-	DefaultRoomVersion    string            `json:"default"`
-	AvailableRoomVersions map[string]string `json:"available"`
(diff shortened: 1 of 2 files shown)


Kegsay:
In an attempt to not sprout more APIs:

> The client API, when building a new event in response to a /send or similar

Can't we just add an extra field for the response to `QueryLatestEventsAndState`? It's reasonable to include that I think in that we include other room state that is strictly not state events such as the `depth+1` in that response, and whether the room exists at all.

> The federation API, when receiving events from a room over federation and performing auth checks etc

Likewise, I would propose adding a field to `QueryStateAfterEvents` which again already returns metadata like whether the room exists at all.

Kegsay:
Basically, anywhere where the caller is unsure if the room exists should be able to say: "yep, and it's v3", so that would be:
 - `QueryStateAndAuthChainResponse`
 - `QueryLatestEventsAndStateResponse`
 - `QueryStateAfterEventsResponse`

neilalexander:
Closing this for now.

PR #922, opened 2020-03-18
HeaderedEvents in sync API

neilalexander:
This PR updates the sync API to use the `HeaderedEvent` pretty much throughout, including persisting it to the database. This means that the sync API, for the most part, doesn't have any reason to care about the room versions.

The only places that it will care are seemingly in `syncapi/routing/messages.go` in `backfill()`. It stands to reason that the roomserver would know the room version by then, so either we can just store it in the sync API in response to a `m.room.create` event in the room server, or we could use an internal API to query it.

It also adds `RoomVersion` entries into some of the existing API responses to allow spreading this information further.

### DIFF
--- a/roomserver/api/query.go
+++ b/roomserver/api/query.go
@@ -255,7 +255,7 @@ type QueryServersInRoomAtEventResponse struct {
 // QueryRoomVersionCapabilities asks for the default room version
 type QueryRoomVersionCapabilitiesRequest struct{}
 
-// QueryRoomVersionCapabilitiesResponse is a response to QueryServersInRoomAtEventResponse
+// QueryRoomVersionCapabilitiesResponse is a response to QueryRoomVersionCapabilitiesRequest
 type QueryRoomVersionCapabilitiesResponse struct {
 	DefaultRoomVersion    gomatrixserverlib.RoomVersion            `json:"default"`
 	AvailableRoomVersions map[gomatrixserverlib.RoomVersion]string `json:"available"`
--- a/roomserver/api/query.go
+++ b/roomserver/api/query.go
@@ -43,6 +43,8 @@ type QueryLatestEventsAndStateResponse struct {
 	// Does the room exist?
 	// If the room doesn't exist this will be false and LatestEvents will be empty.
 	RoomExists bool `json:"room_exists"`
+	// The room version of the room.
+	RoomVersion gomatrixserverlib.RoomVersion `json:"room_version"`
 	// The latest events in the room.
 	// These are used to set the prev_events when sending an event.
 	LatestEvents []gomatrixserverlib.EventReference `json:"latest_events"`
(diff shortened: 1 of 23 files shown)


Kegsay:
pq tests are failing.

roomserver/api/query.go
```diff
@@ -251,8 +257,8 @@ type QueryRoomVersionCapabilitiesRequest struct{}
 
 // QueryRoomVersionCapabilitiesResponse is a response to QueryServersInRoomAtEventResponse
 type QueryRoomVersionCapabilitiesResponse struct {
```
Kegsay: Docs make no sense. It's a response to a response..? Why do we have this response struct at all? I thought we were planning on having versions threaded to other responses like we have done in this PR?

neilalexander: Typo in comments, will fix. Also this is a different request - the one that allows asking the roomserver for *supported* room versions for `capabilities` requests.

syncapi/storage/postgres/current_room_state_table.go
```diff
@@ -223,17 +223,22 @@ func (s *currentRoomStateStatements) upsertRoomState(
...
+	_, err = stmt.ExecContext(
 		ctx,
 		event.RoomID(),
 		event.EventID(),
 		event.Type(),
 		event.Sender(),
 		containsURL,
 		*event.StateKey(),
-		event.JSON(),
+		headeredJSON,
```
Kegsay: For claritys sake, I would like the column name to be renamed to something like this instead of its current `event_json` which is now ambiguous and has incorrect SQL-level comments. Please update.

syncapi/storage/sqlite3/current_room_state_table.go
```diff
@@ -242,16 +242,16 @@ func (s *currentRoomStateStatements) selectEventsWithEventIDs(
...
 	for rows.Next() {
 		var eventBytes []byte
 		if err := rows.Scan(&eventBytes); err != nil {
 			return nil, err
 		}
 		// TODO: Handle redacted events
-		ev, err := gomatrixserverlib.NewEventFromTrustedJSON(eventBytes, false)
-		if err != nil {
+		var ev gomatrixserverlib.HeaderedEvent
+		if err := json.Unmarshal(eventBytes, &ev); err != nil {
```
Kegsay: So I've seen this a lot where we are no longer using `NewEventFromTrustedJSON` but instead unmarshalling directly into a `HeaderedEvent`. What are the implications of this? Does the default `Event.UnmarshalJSON` do it trusted or untrusted? Performance impacts?

roomserver/api/query.go
Kegsay: Right, gotcha!

syncapi/storage/postgres/current_room_state_table.go
neilalexander: Have renamed the columns to `headered_event_json` and updated the SQL-level comments.

syncapi/storage/sqlite3/current_room_state_table.go
neilalexander: `Event.UnmarshalJSON` delegates to `NewEventFromUntrustedJSON`. The net result of this is that we will double-check the event hashes, and if the hash doesn't match on the first pass, to run the redaction algorithm and try again. Since events from the outside world will have already been through this path before persistence, the event hashes will be re-checked but redaction won't be done.

There are probably ways that we can optimise this but I'd rather get it working first.

PR #930, opened 2020-03-19
Room server changes for room versions

neilalexander:
Most of the changed lines here are me just moving things around in the `state` package, however it also adds support for looking up the room version and passing them (or room NIDs) around enough for them to be useful enough to pick a state resolution algorithm.

This still pins everything to room version 1 for now. The next PR will likely move to a newer version of gomatrixserverlib with room version 3 and onward support.


roomserver/state/state.go
```diff
@@ -19,63 +19,1068 @@ package state
...
 )
 
-type StateResolutionVersion int
+type StateResolution struct {
+	db database.RoomStateDatabase
+}
+
+func Prepare(db database.RoomStateDatabase) StateResolution {
+	return StateResolution{
+		db: db,
```
Kegsay: It's not "preparing" anything (I intuitively expected to see SQL prepares here), please call it what it is, `NewStateResolution` or whatever.

roomserver/state/state.go
```diff
@@ -19,63 +19,1068 @@ package state
...
+	// Stable sort so that the most recent entry for each state key stays
+	// remains later in the list than the older entries for the same state key.
+	sort.Stable(stateEntryByStateKeySorter(fullState))
+	// Unique returns the last entry and hence the most recent entry for each state key.
+	fullState = fullState[:util.Unique(stateEntryByStateKeySorter(fullState))]
+	return fullState, nil
+}
+
+// LoadStateAtEvent loads the full state of a room at a particular event.
+func (v StateResolution) LoadStateAtEvent(
```
Kegsay: Clarity on whether this is before and after said event has been applied would be great.

roomserver/state/state.go
```diff
@@ -19,63 +19,1068 @@ package state
...
+	if err != nil {
+		return nil, err
+	}
+
+	return stateEntries, nil
+}
+
+// LoadCombinedStateAfterEvents loads a snapshot of the state after each of the events
+// and combines those snapshots together into a single list.
+func (v StateResolution) LoadCombinedStateAfterEvents(
```
Kegsay: Clarity on whether there are duplicate state for a given `type,state_key` or not would be good, given it's a list so there's nothing stopping it from happening.

Clarity on the combination logic would be nice: this is doing state res v2?

roomserver/state/v1/state_test.go
```diff
@@ -1,58 +0,0 @@
-// Copyright 2017 Vector Creations Ltd
```
Kegsay: I don't see where these tests have landed. Where are they now?

roomserver/state/state.go
neilalexander: Done

roomserver/state/v1/state_test.go
neilalexander: They're gone forever! (I accidentally forgot to resurrect them, they're back now.)

roomserver/state/state.go
neilalexander: Updated the comments about duplicate tuples. However, at this point state res isn't running - that happens later in `CalculateAndStoreStateAfterEvents`.

roomserver/state/state.go
neilalexander: Done

PR #931, opened 2020-03-19
Handle the case of user leaving the room in rooms/{roomID}/state API

### LINKED ISSUE #287: Add auth to /rooms/{roomID}/state... APIs
The API also needs to handle the case of the user having left the room, where we should then return the state of the room at that point as per: https://matrix.org/docs/spec/client_server/r0.2.0.html#get-matrix-client-r0-rooms-roomid-state

prateek2211:
Fixes #287 
### Pull Request Checklist

Still trying to figure out for `rooms/{roomID}/state/{type}/{stateKey}`


roomserver/api/query.go
```diff
@@ -255,6 +255,13 @@ type QueryRoomVersionCapabilitiesResponse struct {
 	AvailableRoomVersions map[string]string `json:"available"`
 }
 
+type QueryFullStateAtEventRequest struct {
+	EventID string `json:"event_id"`
+}
+type QueryFullStateAtEventResponse struct {
```
Kegsay, line 261: As of https://github.com/matrix-org/dendrite/pull/962 you can use that query API instead of adding another.

roomserver/api/query.go
```diff
@@ -548,3 +564,16 @@ func (h *httpRoomserverQueryAPI) QueryRoomVersionCapabilities(
...
 	return commonHTTP.PostJSON(ctx, span, h.httpClient, apiURL, request, response)
 }
+
+// QueryServersInRoomAtEvent implements RoomServerQueryAPI
+func (h *httpRoomserverQueryAPI) QueryFullStateAtEvent(
+	ctx context.Context,
+	request *QueryFullStateAtEventRequest,
+	response *QueryFullStateAtEventResponse,
+) error {
+	span, ctx := opentracing.StartSpanFromContext(ctx, "QueryRoomVersionCapabilities")
```
Kegsay, line 574: No it isn't :)

prateek2211:
@Kegsay It seems to me that in order to get complete state after a given event, `queryAPI.QueryStateAfterEvents` should be used(as the doc comment suggest), but seemingly it doesn't work as expected.
Probably it isn't meant to work this way. 
So, I guess queryAPI I created could still be used

Kegsay:
Be more descriptive. In what way doesn't it work? What did you expect?

prateek2211:
I  need complete state of a room at a given eventId.
So, I used `queryAPI.QueryStateAfterEvents`, using `prevEventId` as event id of event in which membership of user changes from join to ban or leave, and `stateToFetch` as nil.
But the response is very different from expected.
 Also if I pass a state tuple to `stateToFetch`, even then the response state is not necessarily the same type as in tthe tuple.

According  to me , the state at an  event could be found from snapshot table,  which I tried to do the same in the API I created.

Kegsay:
Please bring this PR up-to-date and I can look into taking this over.

Kegsay:
Please re-open if you bring this up-to-date.

PR #938, opened 2020-03-24
Implement backfill over federation

### LINKED ISSUE #910: Backfill over federation doesn't seem to work
Could just be a p2p thing though?

Kegsay:
Previously, backfill did not work correctly over federation. This PR adds support for this with accompanying passing sytests.

This PR fixes a number of existing backfill bugs:
 - When receiving federated backfill requests, do not send the event that was in the original request.
 - When storing backfill results, correctly update the backwards extremity for the room.
 - When judging whether or not the server can see an event, use the room history visibility rules rather than always assuming that they cannot unless joined.

There's a number of improvements architecturally that can be made:
 - The room history visibility auth logic should be in `gomatrixserverlib`.
 - The storage code for sqlite/postgres is identical. We should do what [naffka does](https://github.com/matrix-org/naffka/blob/master/sqldatabase-sqlite.go) and just define the SQL commands to run in each dialect, and share the Go impl.

There's an outstanding hack added in this PR:
 - Backfilled events are never sent to the room server. Before this was okay because the sync API had all the events in its own database so could satisfy `/messages` requests without a problem. In federation though this doesn't work. We should be telling the room server of these old events so the room server can store them for later querying, particularly for `QueryServersInRoomAtEvent` which is currently hacked around.

Fixes #910


syncapi/storage/postgres/backward_extremities_table.go
```diff
@@ -68,27 +82,22 @@ func (s *backwardExtremitiesStatements) prepare(db *sql.DB) (err error) {
...
 	}
 	return
 }
 
 func (s *backwardExtremitiesStatements) insertsBackwardExtremity(
-	ctx context.Context, roomID, eventID string,
+	ctx context.Context, txn *sql.Tx, roomID, eventID string, prevEventID string,
 ) (err error) {
-	_, err = s.insertBackwardExtremityStmt.ExecContext(ctx, roomID, eventID)
+	_, err = txn.Stmt(s.insertBackwardExtremityStmt).ExecContext(ctx, roomID, eventID, prevEventID)
```
neilalexander: This can result in a panic if `txn` is `nil` for some reason?

syncapi/storage/postgres/backward_extremities_table.go
```diff
@@ -107,16 +116,9 @@ func (s *backwardExtremitiesStatements) selectBackwardExtremitiesForRoom(
...
-	err = s.isBackwardExtremityStmt.QueryRowContext(ctx, roomID, eventID).Scan(&isBE)
-	return
-}
-
 func (s *backwardExtremitiesStatements) deleteBackwardExtremity(
-	ctx context.Context, roomID, eventID string,
+	ctx context.Context, txn *sql.Tx, roomID, knownEventID string,
 ) (err error) {
-	_, err = s.insertBackwardExtremityStmt.ExecContext(ctx, roomID, eventID)
+	_, err = txn.Stmt(s.deleteBackwardExtremityStmt).ExecContext(ctx, roomID, knownEventID)
```
neilalexander: Same here (and so forth).

Kegsay: No, all the call-sites always pass non-nil txns for this.

Kegsay:
CI fails because of build kite, not the tests.

PR #950, opened 2020-03-31
Pass pointers to events rather than copying them

neilalexander:
We make *a lot* of copies of `Event` and `HeaderedEvent` by passing things as values and using value receivers a lot. This PR switches things around so that we reduce the number of copies and instead pass pointers around.

This also fixes one or two minor things, sets the default room version to v4 and updates gomatrixserverlib - merge matrix-org/gomatrixserverlib#161 first and then update `go.mod`/`go.sum` with the squash commit ID.


Kegsay:
I'd like to resurrect talks on this at some point. To that end, I'd like to know how big the cost is having to do these copies vs having to indirect via pointers all the time. We made some early optimisations on state res by profiling and moving to pointers, perhaps we can do something similar?

neilalexander:
I suspect this specific PR will be difficult to fast-forward now so it may be easier to start a new one, but yes, the best thing to do probably is just to try it and to compare profiler results. The certain positives would be a substantial reduction in allocations and probably reducing the resident set size as a result, since Go doesn't like giving up memory it's once had.

Kegsay:
> the best thing to do probably is just to try it and to compare profiler results.

If we run `go test -trace` running sytest we will get cpu/mem usage, so I propose we do that with/without pointers to make an informed decision. I'll look into setting this up.

Kegsay:
When I'm out of E2E hell I'll look into setting up sytest with tracing enabled and then see how the two compare.

Kegsay:
We might be able to use the R&J benchmark here instead?

neilalexander:
Superseded by #1583.

PR #952, opened 2020-04-02
Invites v2 endpoint

neilalexander:
This PR nukes the `/_matrix/federation/v1/invite` endpoint and replaces it with the new `/_matrix/federation/v2/invite` endpoint, along with handling the new format structures from matrix-org/gomatrixserverlib#163 and all of the logic that spreads room versions throughout.

It also actually tries to send invite events through the federation sender now, which is a definite improvement on merely blackholing them.

Currently actually accepting the invite doesn't work, and seemingly Riot still doesn't see the name etc of the room (even though that should be encapsulated in the `invite_room_state`) so there is still some work to do here.


federationsender/queue/destinationqueue.go
```diff
@@ -51,59 +53,62 @@ func (oq *destinationQueue) sendEvent(ev *gomatrixserverlib.HeaderedEvent) {
...
+// backgroundSend is the worker goroutine for sending events.
 func (oq *destinationQueue) backgroundSend() {
+	oq.running.Store(true)
+	defer oq.running.Store(false)
+
 	for {
-		t := oq.next()
-		if t == nil {
+		transaction, invites := oq.nextTransaction(), oq.nextInvites()
+		if !transaction && !invites {
```
Kegsay, line 92: This feels like this should be possible and more easily read if it used channels instead of mutexes and shared memory. Not a task for now but thoughts welcome.

neilalexander: I actually almost did this but I figured it might end up spiralling the PR a bit further than necessary. I also figured that it would need some thought on channel buffer size vs what might happen if a channel send blocks for some reason. I'll create a separate issue for looking into this!

Kegsay: Thanks!

PR #954, opened 2020-04-03
Federation for v3/v4 rooms

neilalexander:
This PR enables v3 and v4 rooms and fixes a couple of related federation bugs:

- perform state resolution when handling a `send_join` so that we actually auth the event properly and don't just allow room joins by total fluke
- limits the number of `auth_events` and `prev_events` as per matrix-org/matrix-doc#2307
- updates gomatrixserverlib so that it tries harder to prevent `null` values in `auth_events` and `prev_events` (merge matrix-org/gomatrixserverlib#162 first and then update `go.mod`/`go.sum` before merging this PR)
- fixes `/state` so that it actually [does the right thing](https://matrix.org/docs/spec/server_server/r0.1.3#get-matrix-federation-v1-state-roomid) and looks up state at the given event, rather than at the given event's `prev_events` resulting in state type-key duplicates


roomserver/api/query.go
```diff
@@ -203,6 +203,8 @@ type QueryStateAndAuthChainRequest struct {
 	PrevEventIDs []string `json:"prev_event_ids"`
 	// The list of auth events for the event. Used to calculate the auth chain
 	AuthEventIDs []string `json:"auth_event_ids"`
+	// Should state resolution be ran on the result events?
+	ResolveState bool `json:"resolve_state"`
```
Kegsay, line 207: When would we *not* want to run state resolution?

federationapi/routing/state.go
```diff
@@ -107,7 +107,7 @@ func getState(
 		return nil, &util.JSONResponse{Code: http.StatusNotFound, JSON: nil}
 	}
 
-	prevEventIDs := getIDsFromEventRef(event.PrevEvents())
+	prevEventIDs := []string{eventID}
```
Kegsay, line 110: Why aren't we using `event.PrevEvents()` anymore? Variable name indicates that's the correct behaviour...

roomserver/api/query.go
neilalexander: Without examining things that call `QueryStateAndAuthChain` I am not sure, so I've taken the non-breaking approach for now.

sytest-blacklist
```diff
@@ -31,4 +31,10 @@ Alias creators can delete canonical alias with no ops
 
 # Blacklisted because we need to implement v2 invite endpoints for room versions
 # to be supported (currently fails with M_UNSUPPORTED_ROOM_VERSION) 
-Inbound federation rejects invites which are not signed by the sender
\ No newline at end of file
+Inbound federation rejects invites which are not signed by the sender
+
+# Blacklisted due to flakiness
+Regular users can add and delete aliases when m.room.aliases is restricted
```
Kegsay, line 37: Why now? Postgres tests seem pretty consistently happy prior to this.

federationapi/routing/state.go
neilalexander: I'll need to rename the variable but it looks like using `event.PrevEvents()` was actually incorrect behaviour, [the spec](https://matrix.org/docs/spec/server_server/r0.1.3#get-matrix-federation-v1-state-roomid) states it should return "at" a given event.

roomserver/api/query.go
Kegsay: Please flag it as such then e.g `// TODO: check call sites and remove if we always want to do state res`

sytest-blacklist
neilalexander: I think this was because of SQLite, before we made SQLite non-mandatory in the pipeline. I can try removing it again.

roomserver/state/state.go
```diff
@@ -681,6 +681,60 @@ func (v StateResolution) calculateStateAfterManyEvents(
 	return
 }
 
+// ResolveConflictsAdhoc is a helper function to assist the query API in
+// performing state resolution when requested. This is a different code
+// path to the rest of state.go because this assumes you already have
+// gomatrixserverlib.Event objects and not just a bunch of NIDs like
+// elsewhere in the state resolution.
```
Kegsay, line 688: Good explanation 👍

roomserver/state/state.go
```diff
@@ -681,6 +681,60 @@ func (v StateResolution) calculateStateAfterManyEvents(
...
+// TODO: Some of this can possibly be deduplicated
+func ResolveConflictsAdhoc(
+	version gomatrixserverlib.RoomVersion,
+	events []gomatrixserverlib.Event,
+	authEvents []gomatrixserverlib.Event,
+) ([]gomatrixserverlib.Event, error) {
```
Kegsay, line 694: Unclear what the return arg is. `resolved` I guess is a strict subset of `events`? It's guaranteed to only be state events?

roomserver/state/state.go
```diff
@@ -681,6 +681,60 @@ func (v StateResolution) calculateStateAfterManyEvents(
...
+	type stateKeyTuple struct {
+		Type     string
+		StateKey *string
+	}
+
+	eventMap := make(map[stateKeyTuple][]gomatrixserverlib.Event)
```
Kegsay, line 700: Fairly sure this doesn't work because `*string` won't be dereferenced when checking key equality. See https://play.golang.org/p/peYVWfZLtA7

roomserver/state/state.go
```diff
@@ -681,6 +681,60 @@ func (v StateResolution) calculateStateAfterManyEvents(
...
+	}
+	switch stateResAlgo {
+	case gomatrixserverlib.StateResV1:
+		resolved = gomatrixserverlib.ResolveStateConflicts(conflicted, authEvents)
+		resolved = append(resolved, notConflicted...)
+	case gomatrixserverlib.StateResV2:
+		// TODO: auth difference here?
+		resolved = gomatrixserverlib.ResolveStateConflictsV2(conflicted, notConflicted, authEvents, authEvents)
+	default:
+		return nil, errors.New("unsupported state resolution algorithm")
```
Kegsay, line 732: Please give more contextual error messages. State what value you got.

sytest-blacklist
Kegsay: Yeah please don't blacklist on sqlite's bugs - I will aim to go through those bugs this week to resolve this.

roomserver/state/state.go
neilalexander: You're right, that's a mistake.

roomserver/state/state.go
neilalexander: Done.

roomserver/state/state.go
neilalexander: I've added a comment, but yes, it's the resolved state plus unconflicted state, and they will only be state events.

federationapi/routing/state.go
neilalexander: I've moved that inline a couple of lines down so the variable name is no longer confusing.

roomserver/api/query.go
neilalexander: Done.

roomserver/state/state.go
neilalexander: Fixed now and added comments.

sytest-blacklist
neilalexander: I've unblacklisted it now.

PR #957, opened 2020-04-09
Set default room version to 4

neilalexander:
This sets the default room version to 4.


roomserver/version/version.go
```diff
@@ -59,7 +59,7 @@ var roomVersions = map[gomatrixserverlib.RoomVersion]RoomVersionDescription{
 // DefaultRoomVersion contains the room version that will, by
 // default, be used to create new rooms on this server.
 func DefaultRoomVersion() gomatrixserverlib.RoomVersion {
-	return gomatrixserverlib.RoomVersionV2
+	return gomatrixserverlib.RoomVersionV4
```
Kegsay, line 62: Have you tried this on v3 to see if tests still fail?

neilalexander: Yeah - same tests fail, which is to be expected really as the only thing that changes between v3 and v4 is URL-safe encoding of event IDs.

neilalexander:
Only failing test now is `Message history can be paginated`, which should be fixed by matrix-org/sytest#851.

Issue #893, opened 2020-03-11
General arch: data propagation


Kegsay:
Draw up the arrows for topic consumer/producers AND internal APIs.
 - Do they make sense? 
 - Can we cut them down anywhere? 
 - Then write it up with justifications for each arrow.

gedw99:
The implementation uses Kafka it seems.

I would be interested in adding. NATS / liftbridge solution as well. So 100% golang.

Kegsay:
We already use https://github.com/matrix-org/naffka for when we run in monolith mode, which is 100% golang. That's not the point of this issue.

PR #959, opened 2020-04-14
Set room visibility on room creation

### LINKED ISSUE #603: Save room details in the publicroomsapi DB if created with the "public" visibility preset
If a room is created using `"visibility":"public"`, save its details in the publicroomapi component's database so it's listed in responses for `GET /publicRooms`.

Not sure about the best way to do it, since the publicroomapi component is pretty independent and handles its own endpoints. The easy solution would be to make the clientapi component hit `PUT /_matrix/client/r0/directory/list/room/{roomId}` when creating a room with that preset, but that doesn't sound great performance-wise. We could also create a kafka topic the clientapi component would publish to and the publicroomsapi component would consume from.

That's just my thoughts about it, feel free to shout if I missed something.

prateek2211:
fixed #603 
### Pull Request Checklist


prateek2211:
I need to get hostname with port no of the server. I tried to get it from incoming request, but it seems to be impossible to do so, because golang determines host from request header which might not be always correct.
Another way could be to add address of the server in dendrite config, and update it on runtime. 
@Kegsay Should I do that?

Cnly:
I think we're actually talking about an internal API, not the client-server API endpoint? Or at least @anoadragon453 and I thought so. Perhaps there's a typo in #603's description.

Kegsay:
We purposefully want to hit the client-server API endpoint, rather than add yet another internal API with the same semantics and behaviour.

It's frustrating that this is the thing that trips us up though. I don't see a nice way to do this - the internal HTTP API hit is probably going to be the clearest, only because we don't know exactly where to hit if we go externally. Bah.

Cnly:
@Kegsay Oh I see, I think that's a good point. Maybe we should have an issue in gomatrixserverlib, or somewhere else, for this?

Kegsay:
It's a dendrite-specific architectural question really. We haven't (until now) needed to do something that involves the same code paths as an existing CS API hit. This falls under the remit of https://github.com/matrix-org/dendrite/issues/893

Kegsay:
Superseded by https://github.com/matrix-org/dendrite/pull/1174 which replaces `publicroomsapi` entirely.

PR #972, opened 2020-04-18
Ensure room version tests can be grouped whether failed or passed

Cnly:
`are-we-synapse-yet.py` on master errors because of `Exception: The test 'User can create and send/receive messages in a room with version 3 (2 subtests)' doesn't have a group`, which is in turn because SyTest outputs slightly different test descriptions for failed and passed tests. This PR is a quick fix for that.

### Pull Request Checklist

* [ ] `N/A` ~I have added any new tests that need to pass to `testfile` as specified in [docs/sytest.md](https://github.com/matrix-org/dendrite/blob/master/docs/sytest.md)~

### DIFF
--- a/are-we-synapse-yet.list
+++ b/are-we-synapse-yet.list
@@ -405,6 +405,7 @@ syn Unnamed room comes with a name summary
 syn Named room comes with just joined member count summary 
 syn Room summary only has 5 heroes 
 syn Room summary counts change when membership changes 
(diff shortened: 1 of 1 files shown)


Kegsay:
LGTM, thanks! I grabbed the `results.tap` from docker run of sytest, so I'm surprised I didn't hit this.

Issue #410, opened 2018-03-04
Write tests for login methods


APwhitehat:
spec url: https://matrix.org/docs/spec/client_server/r0.3.0.html#login
Currently there are no tests for login methods.
Now password login is supported (+ token based if #408 is merged)
Tests for these would help subsequent development.

#### Tests for:
- [ ] "m.login.password"
- [ ] "m.login.token"

Kegsay:
We rely on sytest for integration testing, so this would just be a matter of uncommenting out those tests.

Kegsay:
```
not ok 26 POST /login can log in as a user
# Started: 2020-04-14 16:28:33.130
# Ended: 2020-04-14 16:28:33.197
# HTTP Request failed ( 400 Bad Request https://localhost:8800/_matrix/client/r0/login ) from POST https://localhost:8800/_matrix/client/r0/login?...
# {"errcode":"M_BAD_JSON","error":"login identifier '' not supported"}
not ok 27 (expected fail) POST /login returns the same device_id as that in the request # TODO expected fail
# Started: 2020-04-14 16:28:33.198
# Ended: 2020-04-14 16:28:33.201
# HTTP Request failed ( 400 Bad Request https://localhost:8800/_matrix/client/r0/login ) from POST https://localhost:8800/_matrix/client/r0/login?...
# {"errcode":"M_BAD_JSON","error":"login identifier '' not supported"}
not ok 28 POST /login can log in as a user with just the local part of the id
# Started: 2020-04-14 16:28:33.202
# Ended: 2020-04-14 16:28:33.204
# HTTP Request failed ( 400 Bad Request https://localhost:8800/_matrix/client/r0/login ) from POST https://localhost:8800/_matrix/client/r0/login?...
# {"errcode":"M_BAD_JSON","error":"login identifier '' not supported"}
not ok 29 POST /login as non-existing user is rejected
# Started: 2020-04-14 16:28:33.205
# Ended: 2020-04-14 16:28:33.208
# HTTP Request failed ( 400 Bad Request https://localhost:8800/_matrix/client/r0/login ) from POST https://localhost:8800/_matrix/client/r0/login?...
# {"errcode":"M_BAD_JSON","error":"login identifier '' not supported"}
not ok 30 POST /login wrong password is rejected
# Started: 2020-04-14 16:28:33.209
# Ended: 2020-04-14 16:28:33.212
[...]

PR #974, opened 2020-04-20
Add login sytests

Kegsay:
Now that https://github.com/matrix-org/sytest/pull/853 has landed.

Closes https://github.com/matrix-org/dendrite/issues/410


Kegsay:
`run-tests PASSED`

it's the upload to buildkite which failed.

PR #975, opened 2020-04-20
Perspective key fetching, some federation room join fixes

neilalexander:
This PR does a couple things:

- Adds support for perspective key fetchers, which were somewhat useful in diagnosing federation issues (you can ask a known server like `matrix.org` for keys which reduces the chance that offline servers will break authing a room join, for example)
- No longer tries to start new join-by-servers when the original context has expired
- Whacks the HTTP server timeout up to 5 minutes to deal with current slowness
- Automatically tries to fetch missing auth events during a room join if they were not supplied to us as part of the `send_join` response (this really feels like a synapse bug is involved but I need to look into that more)

### DIFF
--- a/clientapi/routing/joinroom.go
+++ b/clientapi/routing/joinroom.go
@@ -365,8 +365,46 @@ func (r joinRoomReq) joinRoomUsingServer(roomID string, server gomatrixserverlib
 		return nil, fmt.Errorf("r.federation.SendJoin: %w", err)
 	}
 
+	// A list of events that we have retried, if they were not included in
+	// the auth events supplied in the send_join.
+	retries := map[string]bool{}
+
+retryCheck:
+	fmt.Println("STATE EVENTS:")
+	for _, e := range respSendJoin.StateEvents {
+		fmt.Println("*", e.EventID())
+	}
+	fmt.Println("AUTH EVENTS:")
+	for _, e := range respSendJoin.AuthEvents {
+		fmt.Println("*", e.EventID())
+	}
 	if err = respSendJoin.Check(r.req.Context(), r.keyRing, event); err != nil {
-		return nil, fmt.Errorf("respSendJoin: %w", err)
+		switch e := err.(type) {
+		case gomatrixserverlib.MissingAuthEventError:
+			// Check that we haven't already retried for this event, prevents
+			// us from ending up in endless loops
+			if _, ok := retries[e.AuthEventID]; !ok {
+				// Ask the server that we're talking to right now for the event
+				tx, txerr := r.federation.GetEvent(r.req.Context(), server, e.AuthEventID)
+				if txerr != nil {
+					return nil, fmt.Errorf("r.federation.GetEvent: %w", txerr)
+				}
+				// For each event returned, add it to the auth events.
+				for _, pdu := range tx.PDUs {
+					ev, everr := gomatrixserverlib.NewEventFromUntrustedJSON(pdu, respMakeJoin.RoomVersion)
+					if everr != nil {
+						return nil, fmt.Errorf("gomatrixserverlib.NewEventFromUntrustedJSON: %w", everr)
+					}
+					respSendJoin.AuthEvents = append(respSendJoin.AuthEvents, ev)
+				}
+				// Mark the event as retried and then give the check another go.
+				retries[e.AuthEventID] = true
+				goto retryCheck
+			}
+			return nil, fmt.Errorf("respSendJoin (after retries): %w", e)
+		default:
+			return nil, fmt.Errorf("respSendJoin: %w", err)
+		}
 	}
 
 	if err = r.producer.SendEventWithState(
--- a/clientapi/routing/joinroom.go
+++ b/clientapi/routing/joinroom.go
@@ -15,6 +15,7 @@
 package routing
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 	"strings"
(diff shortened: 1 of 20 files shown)


cmd/dendrite-monolith-server/main.go
```diff
@@ -55,7 +55,7 @@ func main() {
 	deviceDB := base.CreateDeviceDB()
 	keyDB := base.CreateKeyDB()
 	federation := base.CreateFederationClient()
-	keyRing := keydb.CreateKeyRing(federation.Client, keyDB)
+	keyRing := keydb.CreateKeyRing(federation.Client, keyDB, cfg)
```
Kegsay, line 58: Please don't do this. I don't like this because it's not obvious what `CreateKeyRing` actually requires from `cfg`, which damages readability and damages the ability to independently test that component. I can clearly see in this diff why you are adding it, but once we lose that context on merge it'll be too late. Please yank out `KeyPerspectives` into a separate struct and require that instead.

common/config/config.go
```diff
@@ -99,6 +99,20 @@ type Dendrite struct {
 		// If set disables new users from registering (except via shared
 		// secrets)
 		RegistrationDisabled bool `yaml:"registration_disabled"`
+		// Perspective keyservers, to use as a backup when direct key fetch
+		// requests don't succeed
```
Kegsay, line 103: General query: does the spec state when to use perspective servers vs direct fetches? If not, presumably servers have the right to pick and choose a combination. Would it ideally make sense to sort servers by uptime/latency/etc and then optimistically try them first?

clientapi/routing/joinroom.go
```diff
@@ -365,17 +375,63 @@ func (r joinRoomReq) joinRoomUsingServer(roomID string, server gomatrixserverlib
...
+	retries := map[string]bool{}
+
+retryCheck:
 	if err = respSendJoin.Check(r.req.Context(), r.keyRing, event); err != nil {
-		return nil, fmt.Errorf("respSendJoin: %w", err)
+		switch e := err.(type) {
+		case gomatrixserverlib.MissingAuthEventError:
+			// Check that we haven't already retried for this event, prevents
+			// us from ending up in endless loops
+			if _, ok := retries[e.AuthEventID]; !ok {
```
Kegsay: Just do `if !retries[e.AuthEventID]` and rely on the zero value c.f

> Exploiting zero values
>
> It can be convenient that a map retrieval yields a zero value when the key is not present.
>For instance, a map of boolean values can be used as a set-like data structure (recall that the zero value for the boolean type is false). 
> ...

https://blog.golang.org/maps

clientapi/routing/joinroom.go
```diff
@@ -365,17 +375,63 @@ func (r joinRoomReq) joinRoomUsingServer(roomID string, server gomatrixserverlib
...
+				// For each event returned, add it to the auth events.
+				for _, pdu := range tx.PDUs {
+					ev, everr := gomatrixserverlib.NewEventFromUntrustedJSON(pdu, respMakeJoin.RoomVersion)
+					if everr != nil {
+						return nil, fmt.Errorf("gomatrixserverlib.NewEventFromUntrustedJSON: %w", everr)
+					}
+					respSendJoin.AuthEvents = append(respSendJoin.AuthEvents, ev)
+				}
+				// Mark the event as retried and then give the check another go.
+				retries[e.AuthEventID] = true
```
Kegsay: This made me think of a possible optimisation which I'd like to run by you. It would be nice if `respSendJoin.Check` returned a *list* of missing auth event IDs so we can fetch them all before redoing the check. Does this sound reasonable? If so, pop a TODO in please.

clientapi/routing/joinroom.go
```diff
@@ -365,17 +375,63 @@ func (r joinRoomReq) joinRoomUsingServer(roomID string, server gomatrixserverlib
...
+	util.GetLogger(r.req.Context()).WithFields(logrus.Fields{
+		"room_id":          roomID,
+		"num_auth_events":  len(respSendJoin.AuthEvents),
+		"num_state_events": len(respSendJoin.StateEvents),
+	}).Info("Room join signature and auth verification passed")
+
+	// By this point we've verified all of the signatures, retrieved all of the
+	// missing auth events and verified that everything checks out. Nothing
+	// *should* go wrong in the roomserver after this point, so rather than have
+	// the client block on the roomserver taking in all of the new events, we
```
Kegsay, line 391: I'm surprised that `r.producer.SendEventWithState` blocks. In polylith mode, that would just be writing to Kafka, and not waiting for any consumers. How does naffka handle this? If it is doing it synchronously, then we should fix *that* so we are consistent between mono/polylith modes.

cmd/dendrite-monolith-server/main.go
neilalexander: Split it out.

clientapi/routing/joinroom.go
neilalexander: Done.

clientapi/routing/joinroom.go
neilalexander: I've added a TODO. I'd need to look at the code path more closely to see where it gives up.

clientapi/routing/joinroom.go
neilalexander: It definitely blocks in Naffka mode, I haven't tested it in polylith. I think we wait for errors.

common/config/config.go
neilalexander: I don't think the spec states specifically. In our case we attempt direct fetching first, followed by the perspectives in the order they are supplied.

PR #976, opened 2020-04-22
LRU cache for room versions in RS query API

neilalexander:
This PR creates a new caching interface, which we can add things to selectively, and a bounded LRU implementation for the room version cache.

Currently one of the biggest offenders for internal API hits is asking for room versions via the roomserver query API, which is especially bad when handling transactions over federation and sending outbound via destination queues, so this will reduce the number of API hits significantly and hopefully speed things up quite a bit. (It's in both `api` and `query` so that it reduces both API and database hits both in polylith and monolith mode.)

The caching interface exposes functions specific to the things we want to cache to make it a *deliberate* decision to cache something, rather than something we do by accident, to prevent too much premature optimisation. It also allows us to swap out other implementations in the future, e.g. if we wanted shared Redis caches in polylith deployments.

I don't know if 128 cache items for room versions is a sensible number. I pulled it completely out of the air.


common/caching/inmemorylru.go
```diff
@@ -0,0 +1,43 @@
+package caching
+
+import (
+	"sync"
+
+	"github.com/golang/groupcache/lru"
```
Kegsay, line 6: Any reason why we aren't using `github.com/hashicorp/golang-lru` which is what we do for the KeyDB? The Hashicorp version is based on groupcache but also does locking for you so you can remove all the mutex stuff.

common/caching/cache.go
```diff
@@ -0,0 +1,8 @@
+package caching
+
+import "github.com/matrix-org/gomatrixserverlib"
+
+type Cache interface {
+	GetRoomVersion(roomId string) (gomatrixserverlib.RoomVersion, bool)
+	StoreRoomVersion(roomId string, roomVersion gomatrixserverlib.RoomVersion)
+}
```
Kegsay, line 8: Feels like we're prematurely interfacing this just in case we have caches that are not just in-memory. I'd prefer it if we didnt until we actually wanted another cache mechanism.

common/basecomponent/base.go
```diff
@@ -87,6 +89,7 @@ func NewBaseDendrite(cfg *config.Dendrite, componentName string) *BaseDendrite {
 		componentName: componentName,
 		tracerCloser:  closer,
 		Cfg:           cfg,
+		Cache:         caching.NewInMemoryLRUCache(), // TODO: make configurable
```
Kegsay, line 92: I'd rather we hid the fact that we're using an LRU cache as an implementation detail, and not yank it up to the top unless we really feel we have to. We already have an LRU cache for the key database, and that puts the entire caching code in just one file: https://github.com/matrix-org/dendrite/blob/c019ad708669a4c269072bea933f0a520a78ca7c/common/keydb/sqlite3/server_key_table.go#L23 

I find this preferable as it's less stuff to worry about when perusing the codebase.

neilalexander: The main motivation for taking it up to the `BaseComponent` was that it allows both the roomserver `api` and `query` packages to use the same cache when in monolith mode, rather than duplicating it. In polylith mode it will mean that each component predictably gets exactly one cache.

common/caching/inmemorylru.go
neilalexander: Not particularly, I've updated it now to use the hashicorp version and removed the mutex stuff.

common/caching/cache.go
neilalexander: Honestly a bigger part of this is because I wanted one single place where we can look to see exactly what we have caching functions for.

common/caching/inmemorylru.go
neilalexander: Incidentally, I really don't like that the KeyDB has LRU caching as far down as the database-specific statement functions. At the deepest level it should be at the `Database` layer, and probably sharing the same `InMemoryLRUCache` as room versions.

common/basecomponent/base.go
neilalexander: Incidentally, I really don't like that the KeyDB has LRU caching as far down as the database-specific statement functions. At the deepest level it should be at the Database function layer, and probably sharing the same InMemoryLRUCache as room versions.

cmd/roomserver-integration-tests/main.go
```diff
@@ -270,7 +270,7 @@ func testRoomserver(input []string, wantOutput []string, checkQueries func(api.R
 	cmd.Args = []string{"dendrite-room-server", "--config", filepath.Join(dir, test.ConfigFile)}
 
 	gotOutput, err := runAndReadFromTopic(cmd, cfg.RoomServerURL()+"/metrics", doInput, outputTopic, len(wantOutput), func() {
-		queryAPI, _ := api.NewRoomserverQueryAPIHTTP("http://"+string(cfg.Listen.RoomServer), &http.Client{Timeout: timeoutHTTP})
+		queryAPI, _ := api.NewRoomserverQueryAPIHTTP("http://"+string(cfg.Listen.RoomServer), &http.Client{Timeout: timeoutHTTP}, nil)
```
Kegsay, line 273: Won't `nil` be a problem on https://github.com/matrix-org/dendrite/pull/976/files#diff-938f4502716184de911a903a831e2617R590 ?

common/caching/immutableinmemorylru.go
```diff
@@ -0,0 +1,47 @@
...
+		return nil, rvErr
+	}
+	return &ImmutableInMemoryLRUCache{
+		roomVersions: roomVersionCache,
+	}, nil
+}
+
+func checkForInvalidMutation(cache *lru.Cache, key string, value interface{}) {
+	if peek, ok := cache.Peek(key); ok && peek != value {
+		panic("invalid use of immutable cache tries to mutate existing value")
```
Kegsay, line 24: Please log the `key` used as well.

common/caching/immutableinmemorylru.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+}
+
+func (c *ImmutableInMemoryLRUCache) GetRoomVersion(roomID string) (gomatrixserverlib.RoomVersion, bool) {
+	if c == nil || !RoomVersionCachingEnabled {
```
Kegsay, line 29: Right so this is where we do `nil` guards. I personally think it's a little premature to add options for enabling/disabling certain caches at this stage, as we don't yet have use cases for it (eg maybe doing it at compile time like this won't be what we want).

neilalexander: I've removed the enable flags.

common/caching/immutableinmemorylru.go
neilalexander: Done.

cmd/roomserver-integration-tests/main.go
neilalexander: Yeah, as you found there are `nil` guards, although maybe `nil` guards are a bit overkill.

PR #979, opened 2020-04-24
More invite support

neilalexander:
This PR does a few things:

- tries to pull invite room state from federation requests and stores it in `unsigned`, otherwise tries to creates it from local knowledge in the roomserver if it's available
- the sync API now splices the invite room state into the right key
- the roomserver now has a concept of "stub rooms" in the storage layer, as we need this to be able to look up the room version before we've actually accepted the invite
- updated the query API so that it doesn't try to satisfy requests for stub rooms

So far:
- local invites and accepting them works
- federated invites get sent/received, accepting them doesn't quite work yet
- need to still do rejecting/withdrawing invites

There does seem to be a problem with handling synapse-originated invites, which for some reason crashes Riot, see vector-im/riot-web#13324.


roomserver/input/events.go
```diff
@@ -196,8 +197,22 @@ func processInviteEvent(
...
+		// most likely to be if the event came in over federation) then use
+		// that.
+		if err = event.SetUnsignedField("invite_room_state", input.InviteRoomState); err != nil {
+			return err
+		}
+	} else {
+		// There's no invite room state, so let's have a go at building it
+		// up from local data. If we know about the room then we can insert
+		// the invite room state, if we don't then we just fail quietly.
+		if irs, ierr := buildInviteStrippedState(ctx, db, input); ierr == nil {
```
Kegsay, line 211: When wouldn't we be able to build the invite room state? It seems like rarely, in which case a log line to say what went wrong would still be useful I think?

roomserver/input/events.go
```diff
@@ -212,3 +227,50 @@ func processInviteEvent(
...
 }
+
+func buildInviteStrippedState(
+	ctx context.Context,
+	db storage.Database,
+	input api.InputInviteEvent,
+) ([]gomatrixserverlib.InviteV2StrippedState, error) {
+	roomNID, err := db.RoomNID(ctx, input.Event.RoomID())
+	if err != nil || roomNID == 0 {
+		return nil, errors.New("room unknown")
```
Kegsay, line 238: As always, please give context to your errors. Please add the room ID.

roomserver/storage/interface.go
```diff
@@ -83,4 +83,5 @@ type Database interface {
 	GetMembershipEventNIDsForRoom(ctx context.Context, roomNID types.RoomNID, joinOnly bool) ([]types.EventNID, error)
 	EventsFromIDs(ctx context.Context, eventIDs []string) ([]types.Event, error)
 	GetRoomVersionForRoom(ctx context.Context, roomID string) (gomatrixserverlib.RoomVersion, error)
+	IsRoomStub(ctx context.Context, roomNID types.RoomNID) (bool, error)
```
Kegsay, line 86: Please give API documentation.

syncapi/types/types.go
```diff
@@ -247,14 +248,17 @@ func NewJoinResponse() *JoinResponse {
...
 }
 
 // NewInviteResponse creates an empty response with initialised arrays.
-func NewInviteResponse() *InviteResponse {
+func NewInviteResponse(event gomatrixserverlib.HeaderedEvent) *InviteResponse {
 	res := InviteResponse{}
-	res.InviteState.Events = make([]gomatrixserverlib.ClientEvent, 0)
+	res.InviteState.Events = json.RawMessage{'[', ']'}
+	if inviteRoomState := gjson.GetBytes(event.Unsigned(), "invite_room_state"); inviteRoomState.Exists() {
+		res.InviteState.Events = json.RawMessage(inviteRoomState.Raw)
```
Kegsay, line 260: There's a risk that if a federated server provides us with bogus invite state (e.g an object instead of an array) then we will pass this through to the client. We should probably do better than that.

clientapi/routing/membership.go
```diff
@@ -104,23 +106,39 @@ func SendMembership(
...
-	}
-
 	var returnData interface{} = struct{}{}
 
-	// The join membership requires the room id to be sent in the response
-	if membership == gomatrixserverlib.Join {
+	switch membership {
+	case gomatrixserverlib.Invite:
+		// Invites need to be handled specially
+		err = producer.SendInvite(
```
Kegsay, line 114: We still fall through to `producer.SendEvents(` afterwards which feels like a bug?

neilalexander: At the moment this is deliberate bceause `producer.SendInvites()` only notifies the federation sender/sync API about the invites but doesn't update the actual room state itself - I am still thinking about the best way to solve this so we don't need to do both.

syncapi/types/types.go
neilalexander: Yeah, could do with some verification at least. I'll look into it.

roomserver/input/events.go
neilalexander: Done.

roomserver/input/events.go
neilalexander: The CS API deliberately passes `nil` here to get the roomserver to build the state up for us. I've updated the comments.

neilalexander:
> A few things. I don't really like the whole `IsRoomStub` thing. Why are we creating room entries at all at this stage? Can we split that into another table perhaps?

So having looked into this some more, the room stubs are important for allowing us to reuse the same `InputRoomEvent` code paths for notifying other users in the room that the state has changed and that the user was invited. The stub is created automatically by the roomserver as a part of the regular event handling.

I’ve modified the API so that rather than having a separate check for whether the room is a stub or not, the query API can instead use `RoomNIDExcludingStubs` which will automatically make it look like a room stub is a non-existent room. This allows all of the other code paths through the event handling in the roomserver to remain the same but prevents the query API from trying to service requests for stub rooms.

The one thing that could probably use some work is that input invite events don’t currently generate input room events to notify other room participants about the invite, so that’s an area for improvement. That’s why the client API currently sends both an invite and a room event to the roomserver.

PR #981, opened 2020-04-27
Loopback room events for invites

neilalexander:
In the previous invites PR, the client API needed to make two input calls to the roomserver for invites—one for the invite and one for the room event telling other room users about the invite. 

This PR updates the `processInviteEvent` function to return a loopback input room event and the roomserver will automatically take this and process it with the other room events afterward, so that the client API doesn't need to send two separate input events.

### DIFF
--- a/clientapi/routing/membership.go
+++ b/clientapi/routing/membership.go
@@ -127,18 +127,18 @@ func SendMembership(
 		returnData = struct {
 			RoomID string `json:"room_id"`
 		}{roomID}
+		fallthrough
 	default:
-	}
-
-	_, err = producer.SendEvents(
-		req.Context(),
-		[]gomatrixserverlib.HeaderedEvent{event.Headered(verRes.RoomVersion)},
-		cfg.Matrix.ServerName,
-		nil,
-	)
-	if err != nil {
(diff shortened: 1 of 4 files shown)


clientapi/routing/membership.go
```diff
@@ -127,18 +127,18 @@ func SendMembership(
 		returnData = struct {
 			RoomID string `json:"room_id"`
 		}{roomID}
+		fallthrough
```
Kegsay, line 130: Is it ok that `returnData` is empty for invites?

neilalexander: Yep, expected response for `/invite` is a 200 and `{}` body.

roomserver/roomserver.go
```diff
@@ -42,6 +42,7 @@ func SetupRoomServerComponent(
 
 	inputAPI := input.RoomserverInputAPI{
 		DB:                   roomserverDB,
+		Cfg:                  base.Cfg,
```
Kegsay, line 45: We're only doing this to pull out the server name - it would be nicer if you just gave that here so we don't have to wonder what else this is doing with `Cfg`.

neilalexander:
Replaced by #982.

PR #982, opened 2020-04-27
Loopback event from invite response

neilalexander:
This PR:

- adds a producer to the federation sender so that it can...
- use the response of a `/v2/invite` request (which is properly signed) instead of using the original pre-federation event (which isn't) for notifying other room users and updating room state

Needs matrix-org/gomatrixserverlib#180 first and then `go.mod` updating.

### DIFF
--- a/clientapi/routing/membership.go
+++ b/clientapi/routing/membership.go
@@ -128,17 +128,16 @@ func SendMembership(
 			RoomID string `json:"room_id"`
 		}{roomID}
 	default:
-	}
-
-	_, err = producer.SendEvents(
-		req.Context(),
-		[]gomatrixserverlib.HeaderedEvent{event.Headered(verRes.RoomVersion)},
-		cfg.Matrix.ServerName,
-		nil,
-	)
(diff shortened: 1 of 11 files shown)


clientapi/routing/membership.go
```diff
@@ -128,17 +128,16 @@ func SendMembership(
...
-	_, err = producer.SendEvents(
-		req.Context(),
-		[]gomatrixserverlib.HeaderedEvent{event.Headered(verRes.RoomVersion)},
-		cfg.Matrix.ServerName,
-		nil,
-	)
-	if err != nil {
-		util.GetLogger(req.Context()).WithError(err).Error("producer.SendEvents failed")
-		return jsonerror.InternalServerError()
+		_, err = producer.SendEvents(
```
Kegsay, line 131: You're missing the `fallthrough` you had before for `case gomatrixserverlib.Join`

neilalexander: Ah whoops, well spotted!

PR #983, opened 2020-04-27
Implement backfill in the roomserver

Kegsay:
This reasonably chunky PR:
 - Removes `QueryServersInRoomAtEvent` from `roomserver`.
 - Passes the relevant structs to `roomserver` so it can verify events and make federation requests.
 - Modifies the `QueryBackfill` endpoint to request backfilled events via Federation if the entity requesting the backfill is us (same `ServerName`). This implements `gomatrixserverlib.BackfillRequester`.
 - Removes the code which did backfilling in `syncapi`, instead calling `QueryBackfill`.

There's still the following to do:
 - Not persist events directly from the `QueryBackfill` response. Instead, we should be waiting for the events to come down Kafka and then we can store -ve stream tokens for them. This way is preferred so if ANY `syncapi` requests a backfill, ALL `syncapis` will know about the new values, not just the one requesting it.
 - Make `roomserver` query its own database when backfilling before resorting to federation, similar to how `syncapi` does currently.
 - Move the `backwards_extremitites_table.go` from `syncapi` to `roomserver`, so `syncapi` is ignorant to topological orderings entirely, and just knows stream ordering (from Kafka)

The net result is that backfill is now reliable as it no longer introspects the event sender to work out the servers in the room.


roomserver/query/backfill.go
```diff
@@ -0,0 +1,217 @@
...
+	foundEvent := false   // true if we found a (type, state_key) match
+	// find which state ID to replace, if any
+	for i, id := range newStateIDs {
+		ev, ok := b.eventIDMap[id]
+		if !ok {
+			missingState = true
+			continue
+		}
+		// The state IDs BEFORE the target event are the state IDs BEFORE the prev_event PLUS the prev_event itself
+		if ev.Type() == prevEvent.Type() && ev.StateKey() != nil && ev.StateKey() == prevEvent.StateKey() {
```
neilalexander, line 98: I think you mean `*ev.StateKey() == *prevEvent.StateKey()` here, otherwise you are comparing pointers.

Kegsay: Duh, thanks!

Kegsay: So the reason this got through was because sytest just sends message events and then backfills, so this code is never exercised. :/

PR #985, opened 2020-04-28
sql/backwards_extremities: Shift to table format and share code

Kegsay:
This is an initial cut to reduce boilerplate at the storage layer.
It removes the need for 2x `_table.go` files, one for each DB engine,
replacing it with a single struct which has an interface which
implements the raw SQL statements.

The actual impl sits alongside the interface declaration which is
generally regarded as best practice (though no canonical sources).
Especially in this case where the impl is tiny (functions returning
strings) and relies heavily on the function signatures of the
table struct (for parameters), having the context in the same file
is useful.

---

Assuming we're happy with this structure, I'll go ahead and convert the rest of `syncapi` to use this new format (the mapping is fairly mechanical and boring) as it's annoying to have a mix of both formats in a component.

The main victory will be to share as much of the `Database` implementation as possible, which will hopefully be possible once the tables have been ported over. Crucially, nothing stops us from breaking out of this structure should the tables/statements be vastly different between the two DB engines. We should only be making files in the `tables` package if there is genuine overlap.


Kegsay:
Yeah this is where we will still need different files for different DB engines. That should hopefully be the exception though, not the rule!

PR #988, opened 2020-04-28
Federation sender API remodel

neilalexander:
This is a WIP to define the federation sender input API, and to wire it into the roomserver input API. This will let us:

- Start concentrating federation requests in the federation sender, rather than sprawled across everywhere else (like the client API)
- Move some intelligence from the client API to the roomserver
- Allow the roomserver to delegate federation stuff onto the federation sender directly

Note to self: fix the copyrights.


federationsender/query/api.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"github.com/matrix-org/dendrite/common"
+	"github.com/matrix-org/dendrite/federationsender/api"
+	"github.com/matrix-org/dendrite/federationsender/types"
+	rsAPI "github.com/matrix-org/dendrite/roomserver/api"
+	"github.com/matrix-org/util"
+)
+
+// FederationSenderQueryDatabase has the APIs needed to implement the query API.
+type FederationSenderQueryDatabase interface {
```
Kegsay, line 16: We already have a storage interface, can we just use that? The only reason why we might want this would be for testing internal APIs in isolation without a database more easily (fewer things to implement) but we don't have any tests like that and I'd rather we tested the actual SQL as well.

neilalexander: Yeah, don't see why not. This interface was there before. I'll remove it.

PR #989, opened 2020-04-29
Join room support in federation sender

neilalexander:
This PR adds join room support into the federation sender. This is part of an activity to concentrate federation actions within the federation components, rather than, e.g. the client API.

### DIFF
--- a/clientapi/routing/joinroom.go
+++ b/clientapi/routing/joinroom.go
@@ -27,12 +27,11 @@ import (
 	"github.com/matrix-org/dendrite/clientapi/producers"
 	"github.com/matrix-org/dendrite/common"
 	"github.com/matrix-org/dendrite/common/config"
-	"github.com/matrix-org/dendrite/roomserver/api"
+	federationSenderAPI "github.com/matrix-org/dendrite/federationsender/api"
 	roomserverAPI "github.com/matrix-org/dendrite/roomserver/api"
 	"github.com/matrix-org/gomatrix"
 	"github.com/matrix-org/gomatrixserverlib"
 	"github.com/matrix-org/util"
-	"github.com/sirupsen/logrus"
 )
 
 // JoinRoomByIDOrAlias implements the "/join/{roomIDOrAlias}" API.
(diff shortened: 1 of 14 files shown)


clientapi/routing/joinroom.go
```diff
@@ -326,72 +328,16 @@ func (r joinRoomReq) joinRoomUsingServers(
...
-		}, nil
-	}
-
-	event, err := respMakeJoin.JoinEvent.Build(
-		r.evTime, r.cfg.Matrix.ServerName, r.cfg.Matrix.KeyID,
-		r.cfg.Matrix.PrivateKey, respMakeJoin.RoomVersion,
-	)
-	if err != nil {
-		return nil, fmt.Errorf("respMakeJoin.JoinEvent.Build: %w", err)
+	fedJoinReq := federationSenderAPI.PerformJoinRequest{
```
Kegsay, line 331: The intention here is to eventually get to a point where clientapi -> roomserver right? So clientapi doesn't know or care if federation is involved?

federationsender/query/api.go
```diff
@@ -5,17 +5,37 @@ import (
...
-	RoomserverInputAPI rsAPI.RoomserverInputAPI
+	db         storage.Database
+	cfg        *config.Dendrite
+	producer   *producers.RoomserverProducer
+	federation *gomatrixserverlib.FederationClient
+	keyRing    *gomatrixserverlib.KeyRing
+}
+
+func NewFederationSenderInternalAPI(
+	db storage.Database, cfg *config.Dendrite,
```
Kegsay, line 27: I don't like passing in the whole `config`, but I see we need it for:
```
	r.cfg.Matrix.ServerName,
		r.cfg.Matrix.KeyID,
		r.cfg.Matrix.PrivateKey,
```
which is annoying as that's a lot of params to pass around. For now this will do, but we will have to think about best to handle this going forwards.

clientapi/routing/joinroom.go
neilalexander: Yep, that's the plan. The roomserver internal API will try to perform the local join, and if needs be, will reflect onto the federation sender.

PR #993, opened 2020-04-30
Return sooner when joining room over federation

neilalexander:
Right now, after we receive a response for `/send_join`, we block the client request on verifying the signatures and auth chain. This can take a really long time on big rooms and it's entirely possible that the original client request will time out before then.

This PR waits for `/send_join` to complete and then returns success on the client request, letting the verification (and updating the roomserver) happen in its own time afterwards. It's possible that will fail, but in that case, the room just won't appear in the sync stream which is about the best we can do until we have a three-way handshake on room joins.

This also tries to send `content` to the federation sender, which should include the display name and avatar URL hopefully.


neilalexander:
The sytests are failing right now because it doesn't appear that the sytests wait long enough for the roomserver to process the events. However, this fix did allow me to finally join the  (very large) Matrix HQ and TWIM channels on `dendrite.neilalexander.dev`, so we need to think about this.

neilalexander:
Closed due to #1001 rather dramatically changing the shape of this.

PR #994, opened 2020-04-30
Consolidation of roomserver APIs

neilalexander:
This consolidates the input, alias and query roomserver APIs into one internal API.

Some notes:
- commented out a call from the roomserver to the appservice API because why would we do that?
- the `roomserver/internal` folder could probably do with some cleaning up/restructuring
- `alias_test.go` is commented out due to the fact that the storage interface is bigger now and fails lint


cmd/dendrite-sync-api-server/main.go
```diff
@@ -28,9 +28,11 @@ func main() {
 	accountDB := base.CreateAccountsDB()
 	federation := base.CreateFederationClient()
 
-	_, _, query := base.CreateHTTPRoomserverAPIs()
+	fsAPI := base.CreateHTTPFederationSenderAPIs()
+	rsAPI := base.CreateHTTPRoomserverAPIs()
+	rsAPI.SetFederationSenderAPI(fsAPI)
```
Kegsay, line 33: I don't understand why we've sprouted an additional dependency on `fsAPI` here. Is this because one of `alias` or `input` APIs required it, and hence to make `rsAPI` we require it? So even though we don't actually invoke those methods, we need to supply them?

roomserver/internal/alias.go
```diff
@@ -91,44 +77,50 @@ func (r *RoomserverAliasAPI) SetRoomAlias(
...
-		// the appservice component
-		aliasReq := appserviceAPI.RoomAliasExistsRequest{Alias: request.Alias}
-		var aliasResp appserviceAPI.RoomAliasExistsResponse
-		if err = r.AppserviceAPI.RoomAliasExists(ctx, &aliasReq, &aliasResp); err != nil {
-			return err
-		}
+	/*
+		TODO: Why is this here? It creates an unnecessary dependency
+		from the roomserver to the appservice component, which should be
+		altogether optional.
```
Kegsay, line 95: The AS API allows lazy loading of new rooms. See https://matrix.org/docs/spec/application_service/r0.1.2#get-matrix-app-v1-rooms-roomalias

Basically, if the alias doesn't exist, an application service may be able to make it exist so we hit that linked API. Both federated users and local users can hit the alias, so it shouldn't live in `clientapi` or `federationapi`. I think this should probably sit with `publicroomsapi` to be honest, and rename `publicroomsapi` to something like `directoryapi` which:
 - Handles published rooms in the room directory
 - Handles aliases

cmd/dendrite-sync-api-server/main.go
neilalexander: I think in this instance I was just trying to be consistent but as far as I can tell, none of the sync API calls will call a path that requires the federation sender. I'll remove this.

PR #1001, opened 2020-05-01
Roomserver perform join

neilalexander:
This PR moves room join logic out of the client API and into the roomserver internal API. This will significantly reduce the number of API calls and will establish a pattern that we will reuse for leaving rooms and accepting/rejecting invites.

### DIFF
--- a/federationapi/routing/join.go
+++ b/federationapi/routing/join.go
@@ -62,9 +62,7 @@ func MakeJoin(
 	if !remoteSupportsVersion {
 		return util.JSONResponse{
 			Code: http.StatusBadRequest,
-			JSON: jsonerror.UnsupportedRoomVersion(
-				fmt.Sprintf("Joining server does not support room version %s", verRes.RoomVersion),
-			),
+			JSON: jsonerror.IncompatibleRoomVersion(verRes.RoomVersion),
 		}
 	}
 
--- a/federationapi/routing/join.go
+++ b/federationapi/routing/join.go
@@ -15,6 +15,7 @@
 package routing
 
 import (
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"time"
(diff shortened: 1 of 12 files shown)


federationapi/routing/join.go
```diff
@@ -234,20 +243,40 @@ func SendJoin(
...
+	alreadyJoined := false
+	for _, se := range stateAndAuthChainResponse.StateEvents {
+		if se.Type() == gomatrixserverlib.MRoomMember {
+			if se.StateKey() != nil && *se.StateKey() == *event.StateKey() {
+				var content map[string]interface{}
+				if err = json.Unmarshal(se.Content(), &content); err != nil {
+					continue
+				}
+				if membership, ok := content["membership"]; ok {
+					alreadyJoined = (membership == "join")
```
Kegsay, line 257: Just use `event.Membership()`.

roomserver/internal/perform_join.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+
+	// It is possible for the request to include some "content" for the
+	// event. We'll always overwrite the "membership" key, but the rest,
+	// like "display_name" or "avatar_url", will be kept if supplied.
+	if req.Content == nil {
+		req.Content = map[string]interface{}{}
+	}
+	req.Content["membership"] = "join"
+	if err = eb.SetContent(req.Content); err != nil {
```
Kegsay, line 121: This seems like new code which may cause a problem if the roomserver thinks: "oh this event has content, I will make a new event and set `prev_content` or something.

roomserver/internal/perform_join.go
```diff
@@ -0,0 +1,224 @@
...
+	if err != nil {
+		return fmt.Errorf("Alias %q is not in the correct format", req.RoomIDOrAlias)
+	}
+	req.ServerNames = append(req.ServerNames, domain)
+
+	// Check if this alias matches our own server configuration. If it
+	// doesn't then we'll need to try a federated join.
+	var roomID string
+	if domain != r.Cfg.Matrix.ServerName {
+		// The alias isn't owned by us, so we will eed to try joining using
```
Kegsay, line 55: > eed

roomserver/internal/perform_join.go
neilalexander: I think the only part of this code that's new is the bit where we create the `Content` if it is `nil`, since that would be an obviously invalid join event. The rest is just effectively lifted from the old `writeToBuilder`.

PR #1004, opened 2020-05-04
Roomserver perform leave

neilalexander:
Moves the leave logic into the new perform workflow between the roomserver and the federationsender, and out of the client API.

### DIFF
--- a/federationsender/internal/perform.go
+++ b/federationsender/internal/perform.go
@@ -169,7 +169,8 @@ func (r *FederationSenderInternalAPI) PerformLeave(
 		)
 		if err != nil {
 			// TODO: Check if the user was not allowed to leave the room.
-			return err
+			logrus.WithError(err).Warnf("r.federation.MakeLeave failed")
+			continue
 		}
 
 		// Set all the fields to be what they should be, this should be a no-op
--- a/federationsender/internal/perform.go
+++ b/federationsender/internal/perform.go
@@ -193,11 +193,8 @@ func (r *FederationSenderInternalAPI) PerformLeave(
 
 		// Work out if we support the room version that has been supplied in
 		// the make_leave response.
-		if respMakeLeave.RoomVersion == "" {
-			respMakeLeave.RoomVersion = gomatrixserverlib.RoomVersionV1
-		}
 		if _, err = respMakeLeave.RoomVersion.EventFormat(); err != nil {
-			return fmt.Errorf("respMakeLeave.RoomVersion.EventFormat: %w", err)
+			return gomatrixserverlib.UnsupportedRoomVersionError{}
 		}
 
 		// Build the leave event.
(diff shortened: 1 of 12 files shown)


federationsender/internal/perform.go
```diff
@@ -153,5 +153,79 @@ func (r *FederationSenderInternalAPI) PerformLeave(
...
+		// request.
+		respMakeLeave, err := r.federation.MakeLeave(
+			ctx,
+			serverName,
+			request.RoomID,
+			request.UserID,
+		)
+		if err != nil {
+			// TODO: Check if the user was not allowed to leave the room.
+			return err
```
Kegsay, line 172: Or try the next server? Justify why returning is the right thing here.

roomserver/storage/sqlite3/event_state_keys_table.go
```diff
@@ -136,7 +136,7 @@ func (s *eventStateKeyStatements) bulkSelectEventStateKey(
 	for k, v := range eventStateKeyNIDs {
 		iEventStateKeyNIDs[k] = v
 	}
-	selectOrig := strings.Replace(bulkSelectEventStateKeyNIDSQL, "($1)", common.QueryVariadic(len(eventStateKeyNIDs)), 1)
+	selectOrig := strings.Replace(bulkSelectEventStateKeySQL, "($1)", common.QueryVariadic(len(eventStateKeyNIDs)), 1)
```
Kegsay, line 139: Please think of your future self and rename the constant! `bulkSelectEventStateKeyNIDSQL` should be `bulkSelectEventStateKeySQL` and vice versa.

roomserver/internal/perform_leave.go
```diff
@@ -0,0 +1,208 @@
...
+
+	// Let's see if we have an event active for the user in the room. If
+	// we do then it will contain a server name that we can direct the
+	// send_leave to.
+	senderUserNIDs, err := r.DB.GetInvitesForUser(ctx, roomNID, targetUserNID)
+	if err != nil {
+		return "", fmt.Errorf("r.DB.GetInvitesForUser: %w", err)
+	}
+	if len(senderUserNIDs) == 0 {
+		return "", fmt.Errorf("no senderUserNIDs")
```
Kegsay, line 190: Doesn't this just mean `isInvitePending` == `false`?

roomserver/storage/sqlite3/event_state_keys_table.go
neilalexander: Aren't they right now? `bulkSelectEventStateKeySQL` for `bulkSelectEventStateKey` (as it returns a state key), `bulkSelectEventStateKeyNIDSQL` in `bulkSelectEventStateKeyNID` (as it returns a NID)?

PR #1005, opened 2020-05-05
Fix RespState/RespSendJoin

neilalexander:
This updates gomatrixserverlib following matrix-org/gomatrixserverlib#191 and fixes a compile error due to the changing `RespSendJoin` form.

### DIFF
--- a/federationapi/routing/join.go
+++ b/federationapi/routing/join.go
@@ -275,6 +275,7 @@ func SendJoin(
 		}
 	}
 
+	// https://matrix.org/docs/spec/server_server/latest#put-matrix-federation-v1-send-join-roomid-eventid
 	return util.JSONResponse{
 		Code: http.StatusOK,
 		JSON: gomatrixserverlib.RespSendJoin{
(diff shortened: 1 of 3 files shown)


federationapi/routing/join.go
```diff
@@ -278,11 +278,9 @@ func SendJoin(
 	return util.JSONResponse{
 		Code: http.StatusOK,
 		JSON: gomatrixserverlib.RespSendJoin{
-			RespState: gomatrixserverlib.RespState{
-				StateEvents: gomatrixserverlib.UnwrapEventHeaders(stateAndAuthChainResponse.StateEvents),
-				AuthEvents:  gomatrixserverlib.UnwrapEventHeaders(stateAndAuthChainResponse.AuthChainEvents),
-			},
-			Origin: cfg.Matrix.ServerName,
+			StateEvents: gomatrixserverlib.UnwrapEventHeaders(stateAndAuthChainResponse.StateEvents),
```
Kegsay, line 281: We might have spotted this earlier if we had the spec readily to hand. Please add a link to the spec for this response format before the `return`.

PR #1007, opened 2020-05-05
Improve federation sender performance, implement backoff and blacklisting, fix up invites a bit

neilalexander:
This PR aims to dramatically improve the performance of the federation sender by improving the parallelism and reducing mutex contention in the federation sender.

It also implements exponential backoff for servers that return errors, and blacklisting for servers that continuously return errors.

It also fixes some stuff related to invites, because apparently that was upsetting CI.

### DIFF
--- a/federationsender/queue/destinationqueue.go
+++ b/federationsender/queue/destinationqueue.go
@@ -124,9 +124,11 @@ func (oq *destinationQueue) sendEvent(ev *gomatrixserverlib.HeaderedEvent) {
 		// If the destination is blacklisted then drop the event.
 		return
 	}
+	fmt.Println("Queuing event", ev.EventID())
 	oq.runningMutex.Lock()
 	oq.pendingPDUs = append(oq.pendingPDUs, ev)
 	oq.runningMutex.Unlock()
+	fmt.Println("Queued event", ev.EventID())
 	oq.wake()
 }
 
@@ -204,8 +206,8 @@ func (oq *destinationQueue) backgroundSend() {
 			// the pending events and EDUs.
 			if transaction {
 				oq.runningMutex.Lock()
-				oq.pendingPDUs = oq.pendingPDUs[:0]
-				oq.pendingEDUs = oq.pendingEDUs[:0]
+				oq.pendingPDUs = oq.pendingPDUs[len(pendingPDUs):]
+				oq.pendingEDUs = oq.pendingEDUs[len(pendingEDUs):]
 				oq.runningMutex.Unlock()
 			}
 		}
@@ -224,7 +226,7 @@ func (oq *destinationQueue) backgroundSend() {
 			// the pending invites.
 			if invites {
 				oq.runningMutex.Lock()
-				oq.pendingInvites = oq.pendingInvites[:0]
+				oq.pendingInvites = oq.pendingInvites[len(pendingInvites):]
 				oq.runningMutex.Unlock()
 			}
 		}
--- a/federationsender/queue/destinationqueue.go
+++ b/federationsender/queue/destinationqueue.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math"
 	"sync"
 	"time"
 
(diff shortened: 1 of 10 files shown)


federationsender/queue/destinationqueue.go
```diff
@@ -186,7 +277,7 @@ func (oq *destinationQueue) nextInvites() bool {
 				"state_key":   ev.StateKey(),
 				"destination": oq.destination,
 			}).WithError(err).Error("failed to send invite")
-			continue
+			return false, err
```
Kegsay, line 280: This seems like it could be a problem? It was `continue` before.

neilalexander: I think this possibly highlights another issue, which is that if *any* of the invites fail, we'll retry all of them on the next run of the destination queue. I don't think this is causing the problem though.

federationsender/queue/queue.go
```diff
@@ -19,18 +19,20 @@ import (
...
-	origin     gomatrixserverlib.ServerName
-	client     *gomatrixserverlib.FederationClient
-	// The queuesMutex protects queues
-	queuesMutex sync.Mutex
+	rsProducer  *producers.RoomserverProducer
+	origin      gomatrixserverlib.ServerName
+	client      *gomatrixserverlib.FederationClient
+	statistics  *types.Statistics
+	queuesMutex sync.RWMutex // protects the below
 	queues      map[gomatrixserverlib.ServerName]*destinationQueue
```
Kegsay, line 36: I would strongly advise you to make helper functions to:
 - Read from `queues`
 - Set if not exists a `destinationQueue`
and have them all be in self-contained functions where it's clear and obvious when you're locking.

federationsender/queue/queue.go
```diff
@@ -68,21 +72,24 @@ func (oqs *OutgoingQueues) SendEvent(
...
 				client:      oqs.client,
+				statistics:  oqs.statistics.ForServer(destination),
 			}
+			oqs.queuesMutex.Lock()
 			oqs.queues[destination] = oq
+			oqs.queuesMutex.Unlock()
 		}
 
-		oq.sendEvent(ev)
+		go oq.sendEvent(ev)
```
Kegsay, line 92: Doesn't this kind of making queuing pointless if you're just going to fork it off to a goroutine? What guarantees ordering?

federationsender/queue/destinationqueue.go
```diff
@@ -33,92 +34,196 @@ import (
...
+}
+
+// Start the destination queue if it needs to be started, or
+// otherwise signal to it that it should wake up from sleep.
+func (oq *destinationQueue) wake() {
+	if !oq.running.Load() {
+		go oq.backgroundSend()
+	} else {
+		select {
+		case oq.wakeup <- true:
```
Kegsay, line 59: Why is this in a `select`?

federationsender/queue/queue.go
Kegsay: Additionally, I see no point in doing this in a goroutine, as `sendEvent` just chucks it onto a queue then wakes up the worker.

federationsender/queue/destinationqueue.go
```diff
@@ -33,92 +34,196 @@ import (
...
-	if !oq.running.Load() {
-		go oq.backgroundSend()
+func (oq *destinationQueue) sendEDU(ev *gomatrixserverlib.EDU) {
+	if oq.statistics.Blacklisted() {
+		// If the destination is blacklisted then drop the event.
+		return
 	}
+	oq.runningMutex.Lock()
+	oq.pendingEDUs = append(oq.pendingEDUs, ev)
+	oq.runningMutex.Unlock()
```
Kegsay, line 89: I was really hoping that we would replace this logic with a `chan` since after all this is exactly what we're emulating here, complete with an unbounded buffer.

neilalexander: The problem with `chan` is that a buffered chan will occupy the memory of the full capacity, and we remove something from the `chan` by the nature of consuming it. We don't move things off the `pendingPDUs`, `pendingEDUs` until we have confirmed to have sent them, otherwise they remain there until the next retry.

federationsender/queue/destinationqueue.go
neilalexander: We don't actually know that the worker is sleeping at that point. I think we just want a best-effort "wake up" if it just happens to be rather than waiting for it.

federationsender/queue/queue.go
neilalexander: This was mostly just a test, I plan to undo that.

federationsender/queue/destinationqueue.go
```diff
@@ -33,92 +34,187 @@ import (
...
 		go oq.backgroundSend()
 	}
+	oq.incomingInvites <- ev
 }
 
 // backgroundSend is the worker goroutine for sending events.
+// nolint:gocyclo
 func (oq *destinationQueue) backgroundSend() {
+	// Mark the worker as running for its lifetime.
 	oq.running.Store(true)
```
Kegsay, line 98: This is still racey. Two calls to `sendEvent` might both see that `oq.running == false` and then spin up 2 goroutines, both of which would then store true and run. Read-modify-writes have to be done under a single lock. In all honesty, I dislike how we are using `go.uber.org/atomic` at all, but if we are we need to be using compare and swap - looks like it has https://godoc.org/go.uber.org/atomic#Bool.CAS

federationsender/types/statistics.go
```diff
@@ -0,0 +1,122 @@
...
+	failCounter    atomic.Uint32 // how many times have we failed?
+	successCounter atomic.Uint32 // how many times have we succeeded?
+}
+
+// Success updates the server statistics with a new successful
+// attempt, which increases the sent counter and resets the idle and
+// failure counters. If a host was blacklisted at this point then
+// we will unblacklist it.
+func (s *ServerStatistics) Success() {
+	s.successCounter.Add(1)
```
Kegsay, line 66: We do not need to have this be an `atomic.Uint32`. We _only_ read and write to this variable from the worker goroutine, so just keep them as normal `int`s.

federationsender/types/statistics.go
```diff
@@ -0,0 +1,122 @@
...
+	s.failCounter.Store(0)
```
Kegsay, line 67: Likewise, this variable is only accessed from the same goroutine, no point in using atomics!

federationsender/queue/destinationqueue.go
```diff
@@ -33,92 +34,187 @@ import (
...
 
 // Send event adds the event to the pending queue for the destination.
 // If the queue is empty then it starts a background goroutine to
 // start sending events to that destination.
 func (oq *destinationQueue) sendEvent(ev *gomatrixserverlib.HeaderedEvent) {
-	oq.runningMutex.Lock()
-	defer oq.runningMutex.Unlock()
-	oq.pendingEvents = append(oq.pendingEvents, ev)
+	if oq.statistics.Blacklisted() {
+		// If the destination is blacklisted then drop the event.
```
Kegsay, line 57: It's a shame we check if the destination is blacklisted here, because it means `ServerStatistics.blacklisted` is accessed from multiple goroutines.

federationsender/types/statistics.go
```diff
@@ -0,0 +1,122 @@
...
+		time.Now().Add(backoffSeconds),
+	)
+	return false
+}
+
+// BackoffDuration returns both a bool stating whether to wait,
+// and then if true, a duration to wait for.
+func (s *ServerStatistics) BackoffDuration() (bool, time.Duration) {
+	backoff, until := false, time.Second
+	if b, ok := s.backoffUntil.Load().(time.Time); ok {
```
Kegsay, line 104: Again, we only read and write to `backoffUntil` in the same goroutine, just drop the atomic stuff.

federationsender/queue/queue.go
```diff
@@ -154,35 +153,27 @@ func (oqs *OutgoingQueues) SendEDU(
...
 	return nil
 }
 
 // filterDestinations removes our own server from the list of destinations.
 // Otherwise we could end up trying to talk to ourselves.
-func filterDestinations(origin gomatrixserverlib.ServerName, destinations []gomatrixserverlib.ServerName) []gomatrixserverlib.ServerName {
-	var result []gomatrixserverlib.ServerName
-	for _, destination := range destinations {
-		if destination == origin {
+func filterDestinations(origin gomatrixserverlib.ServerName, destinations []gomatrixserverlib.ServerName) (
```
Kegsay, line 165: It doesn't _just_ do that now though, it also dedupes the destinations?

roomserver/internal/input.go
```diff
@@ -58,13 +59,21 @@ func (r *RoomserverInternalAPI) InputRoomEvents(
...
+	for i := range request.InputInviteEvents {
+		var loopback *api.InputRoomEvent
+		if loopback, err = processInviteEvent(ctx, r.DB, r, request.InputInviteEvents[i]); err != nil {
 			return err
 		}
+		// The processInviteEvent function can optionally return a
+		// loopback room event containing the invite, for local invites.
+		// If it does, we should process it with the room events below.
+		if loopback != nil {
+			fmt.Println("LOOPING BACK", string(loopback.Event.JSON()))
```
Kegsay, line 71: Debug line?

federationsender/types/statistics.go
neilalexander: We don't - it's also called from the `Perform` functions as a method of unblacklisting a blacklisted server in response to a successful `send_join` or `send_leave`. Otherwise servers stay blacklisted after a successful join takes place.

roomserver/internal/input.go
neilalexander: Yep, removed.

federationsender/queue/queue.go
neilalexander: Updated the name and comments.

federationsender/queue/destinationqueue.go
neilalexander: Fixed to use `CAS`.

PR #1011, opened 2020-05-06
only send new events to RS; add tests for /state_ids and /event

Kegsay:


clientapi/producers/roomserver.go
```diff
@@ -51,6 +51,31 @@ func (c *RoomserverProducer) SendEvents(
 	return c.SendInputRoomEvents(ctx, ires)
 }
 
+// SendEventWithKnownMissingState sends the missing state events followed by the new event to the roomserver with the given stateEventIDs.
+func (c *RoomserverProducer) SendEventWithKnownMissingState(
+	ctx context.Context, stateEventIDs []string, missingStateEvents []gomatrixserverlib.HeaderedEvent, event gomatrixserverlib.HeaderedEvent,
+) error {
+	var ires []api.InputRoomEvent
+	for _, outlier := range missingStateEvents {
```
neilalexander, line 59: Do we care about ordering here?

Kegsay: They are sent in the order provided in `missingStateEvents`.

We need to make sure that the auth events for these events are sent before the state events. We know we have the auth events as we mux them in from `/state_ids` when calling `/event`.

I guess the right thing to do is to make a `RespState`, call `Events()` on it then loop that slice and filter out events we already have. Will fix.

Kegsay: Done!

PR #1012, opened 2020-05-07
Direct messages

neilalexander:
This invite adds support for creating direct message rooms in Riot.

It also includes a fix for normal invites by including the invitee's member state in the `invite_stripped_state`, which might stop Riot from crashing.


neilalexander:
The only relevant test seems to be the SYN-205 test which is broken, so merging this for now and making a note to look into that test on Monday.

PR #1017, opened 2020-05-11
Reduce 500s

neilalexander:
This reduces the number of 500s we return by:
- returning 404 on media download errors
- trying to avoid returning 500s on `/send` at all, since we're really supposed to just return a 200 with a list of errored event IDs where possible

500s are bad because they cause the remote server to back off and stop talking to us, which I've seen a lot in practice on `dendrite.neilalexander.dev`.

### DIFF
--- a/federationapi/routing/send.go
+++ b/federationapi/routing/send.go
@@ -105,7 +105,7 @@ type txnFederationClient interface {
 	GetEvent(ctx context.Context, s gomatrixserverlib.ServerName, eventID string) (res gomatrixserverlib.Transaction, err error)
 }
 
-func (t *txnReq) processTransaction() *gomatrixserverlib.RespSend {
+func (t *txnReq) processTransaction() (*gomatrixserverlib.RespSend, error) {
 	results := make(map[string]gomatrixserverlib.PDUResult)
 
 	var pdus []gomatrixserverlib.HeaderedEvent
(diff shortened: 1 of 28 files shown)


Kegsay:
Please don't mess with `/send` too much, that file I'm hacking on currently.

neilalexander:
@Kegsay That's all I plan to do - it's just enough to stop us returning a sea of 500s.

Kegsay:
You can fix the merge conflicts now 😏

federationapi/routing/send.go
```diff
@@ -135,22 +124,28 @@ func (t *txnReq) processTransaction() (*gomatrixserverlib.RespSend, error) {
 		}
 		if err := json.Unmarshal(pdu, &header); err != nil {
 			util.GetLogger(t.context).WithError(err).Warn("Transaction: Failed to extract room ID from event")
-			return nil, unmarshalError{err}
+			continue
```
Kegsay, line 127: We should probably just add an error to the `results` map like we do elsewhere (in which case just factor out the for loop to a new function which returns an `error` and set that if it's non-nil.

neilalexander: The difficulty here is that we don't know the event ID until after `gomatrixserverlib.NewEventFromUntrustedJSON`.

Kegsay: oh hah. Fair enough, might be worth adding a quick comment about that.

PR #1020, opened 2020-05-11
Enforce `mediaIDRegex` to be only valid `mediaIDCharacters`

sakaru:
Enforce `mediaIDRegex` to be only valid `mediaIDCharacters`

Error messages indicate that:
> mediaId must be a non-empty string using only characters in `mediaIDCharacters`

However the regex used only required that some characters in the filename match
the restriction, not that the entire filename does. This commit ensures that
the filename must entirely fullfill the `mediaIDCharacters` restriction.

Signed-off-by: Sid Karunaratne <sid@karunaratne.net>

### Pull Request Checklist


Kegsay:
Waiting for CI - thanks for this!

PR #1022, opened 2020-05-12
Implement /get_missing_events

Kegsay:
This PR:
 - Will call `/get_missing_events` when receiving a transaction over Federation for which we do not have the `prev_events`. It will recurse exactly once (has to, sytests rely on it) then work out the state of the room at the earliest retrieved event if the hole has not been filled.
 - Working out the state of the room is done by fetching the state AFTER all the prev_events for the earliest event, then mixing in our known latest state (forwards extremities), then performing state resolution on the result.
 - Add a `/v1/send_join` endpoint (to make sytest happy)
 - Adds 2 unit tests to check `/get_missing_events` can completely fill holes as well as do state snapshots.
 - Enable a bunch more sytests.

Requires https://github.com/matrix-org/sytest/pull/865 to be merged before sytest will be happy.

Known issues:
 - We don't currently handle 'rejected' events very well and more sytests rely on it, so we need to do that next.
 - The main sytest has a race condition which I still haven't fully resolved. The problem is that we update the PL for the room, and sometimes Perl's Federation impl will use the old PL event as the auth_event (which then fails) rather than the new. This I'm 100% sure on based on logs.

### DIFF
--- a/federationapi/routing/send.go
+++ b/federationapi/routing/send.go
@@ -351,7 +351,8 @@ func (t *txnReq) processEventWithMissingState(e gomatrixserverlib.Event, roomVer
 	var states []*gomatrixserverlib.RespState
 	needed := gomatrixserverlib.StateNeededForAuth([]gomatrixserverlib.Event{*backwardsExtremity}).Tuples()
 	for _, prevEventID := range backwardsExtremity.PrevEventIDs() {
-		prevState, err := t.lookupStateAfterEvent(roomVersion, backwardsExtremity.RoomID(), prevEventID, needed)
+		var prevState *gomatrixserverlib.RespState
+		prevState, err = t.lookupStateAfterEvent(roomVersion, backwardsExtremity.RoomID(), prevEventID, needed)
 		if err != nil {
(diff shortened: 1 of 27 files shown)


federationapi/routing/send.go
```diff
@@ -306,49 +328,311 @@ func (t *txnReq) processEventWithMissingState(e gomatrixserverlib.Event, roomVer
...
+	}
+	latestEvents := make([]string, len(res.LatestEvents))
+	for i := range res.LatestEvents {
+		latestEvents[i] = res.LatestEvents[i].EventID
+	}
+	// this server just sent us an event for which we do not know its prev_events - ask that server for those prev_events.
+	missingResp, err := t.federation.LookupMissingEvents(t.context, t.Origin, e.RoomID(), gomatrixserverlib.MissingEvents{
+		Limit: 20,
+		// synapse uses the min depth they've ever seen in that room
+		MinDepth: int(res.Depth) - 20,
```
neilalexander, line 567: What happens here if `res.Depth` < 20?

Kegsay: It goes negative which in practice breaks nothing but should probably be clamped to 0.

PR #1027, opened 2020-05-13
Send display name and avatar URL in join content

neilalexander:
This was removed in #1001 and not re-added. Display names and such will now show again properly when joining rooms.


Kegsay:
Can you add sytests for this please, if we have any.

neilalexander:
The only relevant tests are these:
```
Test 81 POST /join/:room_id can join a room with custom content... OK
Test 82 POST /join/:room_alias can join a room with custom content... OK
```
... which both passed before because we were still using the custom create content, even though we weren't injecting profile stuff with it.

PR #1028, opened 2020-05-13
Fix some cases where accepting invites over federation doesn't work

neilalexander:
This should fix the problem whereby our join code *thinks* we know about a local room because we participated in it in the past, but we don't know about the new `invite` state because we are no longer in the room to find out about it. We will check the invites table to see if we should do a federated join.

### DIFF
--- a/roomserver/internal/input_membership.go
+++ b/roomserver/internal/input_membership.go
@@ -231,8 +231,7 @@ func updateToLeaveMembership(
 	return updates, nil
 }
 
-// membershipChanges pairs up the membership state changes from a sorted list
-// of state removed and a sorted list of state added.
+// membershipChanges pairs up the membership state changes.
 func membershipChanges(removed, added []types.StateEntry) []stateChange {
 	changes := pairUpChanges(removed, added)
 	var result []stateChange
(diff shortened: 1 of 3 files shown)


roomserver/internal/perform_join.go
```diff
@@ -121,6 +121,22 @@ func (r *RoomserverInternalAPI) performJoinRoomByID(
...
 
+	// First work out if this is in response to an existing invite.
+	// If it is then we avoid the situation where we might think we
+	// know about a room in the following section but don't know the
+	// latest state as all of our users have left.
+	isInvitePending, inviteSender, err := r.isInvitePending(ctx, req.RoomIDOrAlias, req.UserID)
+	if err == nil && isInvitePending {
+		// Add the server of the person who invited us to the server list,
+		// as they should be a fairly good bet.
+		if _, inviterDomain, ierr := gomatrixserverlib.SplitID('!', inviteSender); ierr == nil {
```
Kegsay, line 132: Shouldn't this be `@`?

neilalexander: Ha - yes, it should be.

roomserver/internal/input_membership.go
```diff
@@ -251,64 +250,39 @@ type stateChange struct {
...
-			})
-			ri++
+	tuples := make(map[types.StateKeyTuple]stateChange)
+	changes := []stateChange{}
+
+	// First, go through the newly added state entries.
+	for _, add := range added {
+		if change, ok := tuples[add.StateKeyTuple]; ok {
+			// If we already have an entry, update it.
+			change.addedEventNID = add.EventNID
```
Kegsay, line 262: Why are we replacing event NIDs here? Shouldn't we be adding this as an extra change? E.g in cases where there are 3 events with the same tuple.

neilalexander: Since the inputs here come from `u.removed, u.added, err = roomState.DifferenceBetweeenStateSnapshots()`, we should never have more than one `removed` or `added` per state key, so we shouldn't ever have a situation where we have to contend with more.

Issue #1029, opened 2020-05-13
Room version 3 event IDs aren't always url-safe


Kegsay:
sqlite had a failing test on https://github.com/matrix-org/dendrite/commit/32624697fd2d74d4a6e23549b647d323b210fb2a caused by:
```
not ok 432 User can invite remote user to room with version 3
# Started: 2020-05-11 17:32:06.525
# Ended: 2020-05-11 17:32:16.657
# Timed out waiting for test at ./run-tests.pl line 745.
# 0.122836: Registered new user @anon-20200511_172706-477:localhost:8802
# 0.127073: Registered new user @anon-20200511_172706-476:localhost:8800
# 0.380242: Invited user @anon-20200511_172706-477:localhost:8802 to !NI7IZkBI4S7mkLhy:localhost:8800
# {}
```

which was due to:

```
time="2020-05-11T17:32:06.912225116Z" level=info msg="sending invite" func="github.com/matrix-org/dendrite/federationsender/que
ue.(*destinationQueue).nextInvites" file="/src/federationsender/queue/destinationqueue.go:285" destination="localhost:8802" eve
nt_id="$09bbJvHoK8//DdCbQw8zt8FaiV7iCJ5sJI+DmSUdLsE" room_version=3
time="2020-05-11T17:32:06.912894067Z" level=info msg="Outgoing request" func="github.com/matrix-org/gomatrixserverlib.(*Client)
.DoHTTPRequest" file="/gopath/pkg/mod/github.com/matrix-org/gomatrixserverlib@v0.0.0-20200511154227-5cc71d36632b/client.go:379"
 context=missing out.req.ID=fBv2FXYIgJWH out.req.method=PUT out.req.uri="matrix://localhost:8802/_matrix/federation/v2/invite/%
21NI7IZkBI4S7mkLhy:localhost:8800/$09bbJvHoK8%2F%2FDdCbQw8zt8FaiV7iCJ5sJI+DmSUdLsE"
time="2020-05-11T17:32:06.913961762Z" level=info msg="Outgoing request returned" func="github.com/matrix-org/gomatrixserverlib.
(*Client).DoHTTPRequest" file="/gopath/pkg/mod/github.com/matrix-org/gomatrixserverlib@v0.0.0-20200511154227-5cc71d36632b/clien
t.go:393" context=missing out.req.ID=fBv2FXYIgJWH out.req.code=404 out.req.duration_ms=1 out.req.method=PUT out.req.uri="matrix
[...]

PR #1031, opened 2020-05-13
Factor out account data and events table

Kegsay:
Starting to see some gains now as we removed more code than we added :D


Kegsay:
postgres sytest failure is ` Test 432 User can invite remote user to room with version 3... FAIL` again which we know as #1029

PR #1034, opened 2020-05-14
Update INSTALL.md, move docs

neilalexander:
Updates the install doc to tidy it up a bit.


docs/INSTALL.md
```diff
@@ -2,38 +2,63 @@
...
 
+```bash
+git clone https://github.com/matrix-org/dendrite
+cd dendrite
+```
+
+Then build it:
+
+```bash
+./build.sh
```
Kegsay, line 42: I would rather just say:
```
go build ./cmd/dendrite-monolith-server
```
as `./build.sh`builds eeeeeeverytyhing.

neilalexander: True

docs/INSTALL.md
```diff
@@ -2,38 +2,63 @@
...
 
- - Go 1.13+
- - Postgres 9.5+
- - For Kafka (optional if using the monolith server):
-   - Unix-based system (https://kafka.apache.org/documentation/#os)
-   - JDK 1.8+ / OpenJDK 1.8+
-   - Apache Kafka 0.10.2+ (see [scripts/install-local-kafka.sh](scripts/install-local-kafka.sh) for up-to-date version numbers)
+* Go 1.13+
+* Postgres 9.5+ (if using Postgres databases)
+* Apache Kafka 0.10.2+ (optional if using the monolith server):
```
Kegsay, line 25: I would rather give minimum requirements and then a separate section for polylith deployments. Mainly because I know that people will skim this, see Postgres/Kafka under "Requirements" then whine about it. If instead it said:
```
Requirements
 - Go 1.13
 - sqlite3 (or Postgres 9.5+)

If you want to run in polylith mode, you also need Kafka (0.10.2+).
```

I would omit Kafka's requirements, as they will change and this will get outdated.

docs/INSTALL.md
```diff
@@ -78,42 +112,56 @@ Dendrite requires a postgres database engine, version 9.5 or later.
...
 ```bash
-# Generate a self-signed SSL cert for federation:
 test -f server.key || openssl req -x509 -newkey rsa:4096 -keyout server.key -out server.crt -days 3650 -nodes -subj /CN=localhost
+```
+
+Generate the server signing key:
 
-# generate ed25519 signing key
+```
 test -f matrix_key.pem || ./bin/generate-keys -private-key matrix_key.pem
```
Kegsay, line 128: I would include the `go build` step here as an aside.

PR #1036, opened 2020-05-14
Use gomatrixserverlib key caching

neilalexander:
This sets up Dendrite to satisfy the new gomatrixserverlib key caching interface from matrix-org/gomatrixserverlib#196.


neilalexander:
Closing in favour of #1038 instead.

PR #1037, opened 2020-05-14
Add Range

Kegsay:
This PR adds a `Range` struct which marries from/to/backwards together. This has a few benefits:
 - Bundling variables which always need to be passed around together.
 - Easier terminology on creation ("from" and "to").
 - Easier terminology on usage in SQL ("low" and "high").
 - Consistent place to document that `Range` means exclusive of low, inclusive of high.

### DIFF
--- a/syncapi/sync/requestpool.go
+++ b/syncapi/sync/requestpool.go
@@ -184,11 +184,20 @@ func (rp *RequestPool) appendAccountData(
 		return data, nil
 	}
 
+	r := types.Range{
+		From: req.since.PDUPosition(),
+		To:   currentPos,
+	}
+	// If both positions are the same, it means that the data was saved after the
+	// latest room event. In that case, we need to decrement the old position as
+	// results are exclusive of Low.
+	if r.Low() == r.High() {
+		r.From--
+	}
+
 	// Sync is not initial, get all account data since the latest sync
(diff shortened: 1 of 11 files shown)


syncapi/sync/requestpool.go
```diff
@@ -184,11 +184,20 @@ func (rp *RequestPool) appendAccountData(
...
 
+	r := types.Range{
+		From: req.since.PDUPosition(),
+		To:   currentPos,
+	}
+	// If both positions are the same, it means that the data was saved after the
+	// latest room event. In that case, we need to decrement the old position as
+	// results are exclusive of Low.
+	if r.Low() == r.High() {
+		r.From--
```
Kegsay, line 195: This was yanked from the account data table, but I worry this is actually applicable everywhere e.g invites table. Thoughts?

neilalexander: I'm not entirely sure I understand why this would happen?

Kegsay: The comment seems to suggest that there is some kind of race when updating `currentPos` - I don't know the full circumstances when it might fire. I would expect this kind of check to be done when notifiers get woken up e.g "wake up, is the current position > what I have as `since`".

I'll look into this a bit and either remove it or clarify in another PR.

PR #1038, opened 2020-05-14
Implement key caching directly

neilalexander:
This is an alternative to #1036. It doesn't require changes in gomatrixserverlib but does mean that we are effectively taking "caching" as a concept away from reuse in gomatrixserverlib.

### DIFF
--- a/common/keydb/cache/keydb.go
+++ b/common/keydb/cache/keydb.go
@@ -1,0 +1,69 @@
+package cache
+
+import (
+	"context"
+	"errors"
+
+	"github.com/matrix-org/dendrite/common/caching"
+	"github.com/matrix-org/dendrite/common/keydb"
+	"github.com/matrix-org/gomatrixserverlib"
+)
+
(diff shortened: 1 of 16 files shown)


neilalexander:
... which is going to be harder because of the import cycle. Damnit.

common/keydb/cache/keydb.go
```diff
@@ -0,0 +1,69 @@
...
+	"errors"
+
+	"github.com/matrix-org/dendrite/common/caching"
+	"github.com/matrix-org/dendrite/common/keydb"
+	"github.com/matrix-org/gomatrixserverlib"
+)
+
+// A Database implements gomatrixserverlib.KeyDatabase and is used to store
+// the public keys for other matrix servers.
+type Database struct {
```
Kegsay, line 14: Naming: would prefer `KeyDatabase` so it reads better - `cache.NewKeyDatabase(...)` as `cache.NewDatabase(...)` is vague.

PR #1042, opened 2020-05-15
Set canonical alias on room creation

Kegsay:
So it shows up in Riot.

It turns out I was the one who set the order up in the comment, and I'd rather do ACLs first before vanity stuff like canonical alias.


Kegsay:
`run-tests PASSED`

PR #1045, opened 2020-05-18
Tweaks to latest events updater

neilalexander:
This PR, aside from adding a whole bunch of comments, foundationally makes three subtle changes:

**Adds an `Overwrite` field to `types.StateAtEntry`**

When the roomserver input API receives a `KindNew` event that `HasState` then we:
  1. Set `Overwrite` to `true`
  2. Create a state snapshot based solely on the outliers specified in `StateEventIDs`
  3. We then set the state of the new event to the snapshot created in the previous step

**If `Overwrite` is set, then the latest events updater starts with an empty set of latest events**

This is effectively equivalent to discarding any previously-known forward extremities and updating the room so that it has exactly one new forward extremity: the new input event.

Note that we *don't* reset `oldStateNID` - this is deliberate, because we still secretly want to know the difference between the last state snapshot and our current one so that we can fast-forward changes.

That way we are still working strictly on a delta and we don't repeat any output events to other components that have been sent in the past.

**If `Overwrite` is set, then strip `ServerName` from the output event**

The other components still need these events to update their own internal state after working out the difference between our last-known state and the new state, but this makes sure that we don't send anything out over federation when fast-forwarding. They probably happened in the past.

Still to do, possibly in a separate PR:
- Make sure that we don't overwrite unnecessarily if we are already in the room
- Make sure that we correctly resolve the state between the `/send_join` response and the current room state if we are already in the room

### DIFF
--- a/roomserver/internal/input_latest_events.go
+++ b/roomserver/internal/input_latest_events.go
@@ -122,9 +122,19 @@ type latestEventsUpdater struct {
 
 func (u *latestEventsUpdater) doUpdateLatestEvents() error {
 	prevEvents := u.event.PrevEvents()
-	oldLatest := u.updater.LatestEvents()
 	u.lastEventIDSent = u.updater.LastEventIDSent()
 	u.oldStateNID = u.updater.CurrentStateSnapshotNID()
+
+	// If we are doing a regular event update then we will get the
+	// previous latest events to use as a part of the calculation. If
+	// we are overwriting the latest events because we have a complete
+	// state snapshot from somewhere else, e.g. a federated room join,
+	// then start with an empty set - none of the forward extremities
+	// that we knew about before matter anymore.
+	oldLatest := []types.StateAtEventAndReference{}
+	if !u.stateAtEvent.Overwrite {
+		oldLatest = u.updater.LatestEvents()
+	}
 
 	// If the event has already been written to the output log then we
 	// don't need to do anything, as we've handled it already.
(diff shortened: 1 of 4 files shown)


Kegsay:
> This is effectively equivalent to discarding any previously-known forward extremities and updating the room so that it has exactly one new forward extremity: the new input event.

This feels wrong, but I guess makes sense (and won't break the `/g_m_e` logic either as we mux in the previously-known forward extremities when calculating the new state). It feels wrong because there may be a legit reason not to clean forward extremities e.g:
```
   A
 /   \
B     C
|     |
...  ...
|     |
D     E
```
I could see us wanting to keep the state of the two branches separate (though to be fair we don't actually do this). I dunno, maybe I'm just worried for no reason here.

roomserver/internal/input_latest_events.go
```diff
@@ -292,7 +345,12 @@ func (u *latestEventsUpdater) makeOutputNewRoomEvent() (*api.OutputEvent, error)
 	for _, entry := range u.stateBeforeEventAdds {
 		ore.StateBeforeAddsEventIDs = append(ore.StateBeforeAddsEventIDs, eventIDMap[entry.EventNID])
 	}
-	ore.SendAsServer = u.sendAsServer
+	// If we are overwriting the latest events and state then we don't
+	// want to send out any changes that happened as a result to servers
+	// over federation.
+	if !u.stateAtEvent.Overwrite {
+		ore.SendAsServer = u.sendAsServer
```
Kegsay, line 352: Can you put this with `:237-:239` for clarity please.

neilalexander:
My feeling is that there shouldn't be any extra forward extremities after joining a room that the server doesn't currently belong to, since it will be either:
1. A new room which hasn't had time to diverge
2. An existing room which we now have *fully resolved* state for, thanks to `/send_join` and the help of a resident server

We *definitely do* want to preserve forward extremities when additional people join the room from the same homeserver - my next PR will solve that problem.

roomserver/internal/input_latest_events.go
neilalexander: Do you mean just to mention the line numbers in the comment here?

neilalexander: Actually, these lines may not be necessary anyway because of `237-239`.

PR #1047, opened 2020-05-19
Use memberships to determine whether to reset latest events/state on room join

neilalexander:
This adds the mechanics needed for tracking whether a membership is for a local user or a remote one, and then uses this to determine whether or not we should reset the latest events and state upon joining a room (we'll do so when none of our own users are joined but not when they are).

### DIFF
--- a/roomserver/internal/input_events.go
+++ b/roomserver/internal/input_events.go
@@ -31,21 +31,13 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-// OutputRoomEventWriter has the APIs needed to write an event to the output logs.
-type OutputRoomEventWriter interface {
-	// Write a list of events for a room
-	WriteOutputEvents(roomID string, updates []api.OutputEvent) error
-}
-
 // processRoomEvent can only be called once at a time
 //
 // TODO(#375): This should be rewritten to allow concurrent calls. The
 // difficulty is in ensuring that we correctly annotate events with the correct
 // state deltas when sending to kafka streams
-func processRoomEvent(
+func (r *RoomserverInternalAPI) processRoomEvent(
 	ctx context.Context,
-	db storage.Database,
-	ow OutputRoomEventWriter,
 	input api.InputRoomEvent,
 ) (eventID string, err error) {
 	// Parse and validate the event JSON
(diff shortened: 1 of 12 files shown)


roomserver/types/types.go
```diff
@@ -172,7 +172,7 @@ type RoomRecentEventsUpdater interface {
 	MarkEventAsSent(eventNID EventNID) error
 	// Build a membership updater for the target user in this room.
 	// It will share the same transaction as this updater.
-	MembershipUpdater(targetUserNID EventStateKeyNID) (MembershipUpdater, error)
+	MembershipUpdater(targetUserNID EventStateKeyNID, targetLocal bool) (MembershipUpdater, error)
```
Kegsay, line 175: Please doc what `targetLocal` is supposed to mean.

roomserver/storage/interface.go
```diff
@@ -83,9 +83,9 @@ type Database interface {
 	GetAliasesForRoomID(ctx context.Context, roomID string) ([]string, error)
 	GetCreatorIDForAlias(ctx context.Context, alias string) (string, error)
 	RemoveRoomAlias(ctx context.Context, alias string) error
-	MembershipUpdater(ctx context.Context, roomID, targetUserID string, roomVersion gomatrixserverlib.RoomVersion) (types.MembershipUpdater, error)
+	MembershipUpdater(ctx context.Context, roomID, targetUserID string, targetLocal bool, roomVersion gomatrixserverlib.RoomVersion) (types.MembershipUpdater, error)
```
Kegsay, line 86: Please doc what `targetLocal` means.

Kegsay: aiui it means whether `targetUserID` is a local user, so `isTargetLocal` would be a tad clearer.

roomserver/internal/input_events.go
```diff
@@ -116,29 +106,32 @@ func processRoomEvent(
...
-	roomState := state.NewStateResolution(db)
+	roomState := state.NewStateResolution(r.DB)
 
 	if input.HasState {
-		// TODO: Check here if we think we're in the room already.
+		// Check here if we think we're in the room already.
 		stateAtEvent.Overwrite = true
+		var joinEventNIDs []types.EventNID
+		if joinEventNIDs, err = r.DB.GetMembershipEventNIDsForRoom(ctx, roomNID, true, true); err == nil {
+			stateAtEvent.Overwrite = len(joinEventNIDs) == 0
```
Kegsay, line 124: Explain the logic here please e.g "if we have no local joined users we don't know the current state anymore as we aren't in the room, so overwrite and set the current state with the state at this event"

roomserver/internal/input_membership.go
```diff
@@ -132,6 +131,15 @@ func updateMembership(
 	}
 }
 
+func (r *RoomserverInternalAPI) isLocalTarget(event *gomatrixserverlib.Event) bool {
+	targetLocal := false
```
Kegsay, line 135: Naming: I find all this localTarget targetLocal stuff confusing because I don't know if it is saying "the target is local" or "target local users" - the former is informational and the latter is an instruction. `isTargetLocalUser(event)` would be a bit clearer.

PR #1050, opened 2020-05-19
Server key component

neilalexander:
This breaks out the key DB into a separate component, so we don't have lots of components all competing for access to the same database directly.


cmd/dendrite-monolith-server/main.go
```diff
@@ -67,9 +67,15 @@ func main() {
...
 	deviceDB := base.CreateDeviceDB()
-	keyDB := base.CreateKeyDB()
 	federation := base.CreateFederationClient()
-	keyRing := keydb.CreateKeyRing(federation.Client, keyDB, cfg.Matrix.KeyPerspectives)
+
+	serverKeyAPI := serverkeyapi.SetupServerKeyAPIComponent(
+		base, federation,
+	)
+	if base.EnableHTTPAPIs {
+		serverKeyAPI = base.CreateHTTPServerKeyAPIs()
```
Kegsay, line 76: API wise I'd like to combine these into a single access point per internal API. At the moment, I never know if I should be using `base.CreateFoo` or `SetupFooAPI` etc. I appreciate we can't have one **function** since the HTTP form has no need to dependency inject but the actual concrete API impl does need to inject stuff.

If we just had:
```go
api.NewFooHTTP() *Foo
api.NewFoo(a,b,c Deps) *Foo
```
that would make this a lot nicer I think.

What are your thoughts? If we do want to do this, I'd propose we add it to the board.

internal/basecomponent/base.go
```diff
@@ -157,6 +154,20 @@ func (b *BaseDendrite) CreateHTTPFederationSenderAPIs() federationSenderAPI.Fede
 	return f
 }
 
+// CreateHTTPFederationSenderAPIs returns FederationSenderInternalAPI for hitting
+// the federation sender over HTTP
+func (b *BaseDendrite) CreateHTTPServerKeyAPIs() serverKeyAPI.ServerKeyInternalAPI {
```
Kegsay, line 159: Lying comments.

internal/basecomponent/base.go
```diff
@@ -157,6 +154,20 @@ func (b *BaseDendrite) CreateHTTPFederationSenderAPIs() federationSenderAPI.Fede
...
+	f, err := serverKeyAPI.NewServerKeyInternalAPIHTTP(
+		b.Cfg.FederationSenderURL(),
```
Kegsay, line 161: This is probably wrong?

internal/keydb/keydb.go
```diff
@@ -14,7 +14,7 @@
 
 // +build !wasm
 
-package keydb
+package storage
```
Kegsay, line 17: Please name the package the same as the directory.

cmd/dendrite-monolith-server/main.go
neilalexander: I definitely think we can improve the naming for all of this for all components, yes.

internal/basecomponent/base.go
neilalexander: Ah, yes it is. That was masked by the fact that we overwrite those all to be the same endpoint in monolith mode. Good catch.

internal/basecomponent/base.go
neilalexander: Fixed.

serverkeyapi/internal/api.go
```diff
@@ -0,0 +1,78 @@
...
+	"fmt"
+
+	"github.com/matrix-org/dendrite/internal/caching"
+	"github.com/matrix-org/dendrite/internal/config"
+	"github.com/matrix-org/dendrite/serverkeyapi/api"
+	"github.com/matrix-org/gomatrixserverlib"
+)
+
+type ServerKeyAPI struct {
+	api.ServerKeyInternalAPI
```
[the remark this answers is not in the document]
neilalexander: It's a hint to the compiler that the struct must implement the interface. It yells nicely if it doesn't comply, even if the interface isn't being used anywhere. Helpful when working on new interfaces.

PR #1051, opened 2020-05-20
sytest: Make 'Inbound federation can backfill events' pass

Kegsay:
This breaks 'Outbound federation can backfill events' because now
we are returning the right number of events, which the previous
test was relying on.

Previously, /messages was backfilling the membership event, causing
the test to pass. Now we are no longer backfilling the membership
event due to the change in this commit, causing the test to fail.

The test should instead be returning the membership event locally
from synacpis database, but it doesn't do it fast enough, resulting
in a no-op /sync response with a next_batch=s0_0 which will never
pick up the local membership event when it rolls in. The test
does attempt to retry, but doesn't take the new next_batch=s1_0
resulting in it missing from the /messages response.

This race condition is fixed in https://github.com/matrix-org/sytest/pull/874 so tests will fail until that is merged.

### DIFF
--- a/federationapi/routing/backfill.go
+++ b/federationapi/routing/backfill.go
@@ -69,9 +69,14 @@ func Backfill(
 
 	// Populate the request.
 	req := api.QueryBackfillRequest{
-		RoomID:            roomID,
-		EarliestEventsIDs: eIDs,
-		ServerName:        request.Origin(),
+		RoomID: roomID,
+		// we don't know who the successors are for these events, which won't
+		// be a problem because we don't use that information when servicing /backfill requests,
+		// only when making them. TODO: Think of a better API shape
+		BackwardsExtremities: map[string][]string{
+			"": eIDs,
+		},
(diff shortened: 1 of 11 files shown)


federationapi/routing/backfill.go
```diff
@@ -105,6 +110,12 @@ func Backfill(
 		eventJSONs = append(eventJSONs, e.JSON())
 	}
 
+	// sytest wants these in reversed order, similar to /messages, so reverse them now.
+	for i := len(eventJSONs)/2 - 1; i >= 0; i-- {
```
neilalexander, line 114: This feels a bit strange to do here - should the API just process `QueryBackfill` in reverse order?

Kegsay: The thing which demands it being in that order is the `/backfill` endpoint, not `QueryBackfill`, so I want to keep the code which is making it conform to the API as close as possible to the endpoint which demands it, if that makes sense.

PR #1053, opened 2020-05-20
Docker Hub

neilalexander:
This contains all of the necessary things to build and push Docker Hub images for working monolith and polylith deployments with dependencies.


docker/hub/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM docker.io/golang:1.13.7-alpine3.11 AS builder
+
+RUN apk --update --no-cache add bash build-base
+
+WORKDIR /build
+
+ADD . /build
+
+RUN mkdir -p bin
+RUN sh ./build.sh
```
Kegsay, line 10: I'd prefer we didn't do this as it makes it hard to know which `build.sh` will run (it depends where you run the dockerfile).

docker/hub/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM docker.io/golang:1.13.7-alpine3.11 AS builder
+
+RUN apk --update --no-cache add bash build-base
+
+WORKDIR /build
+
+ADD . /build
```
Kegsay, line 7: > In your Dockerfile, favor COPY instead of ADD. ADD automatically extracts zipped files and can copy files from URLs. COPY doesn’t have these capabilities. Whenever possible, avoid using ADD so you aren’t susceptible to attacks through remote URLs and Zip files.

https://towardsdatascience.com/top-20-docker-security-tips-81c41dd06f57

docker/hub/Dockerfile.eduserver
```diff
@@ -0,0 +1,13 @@
+FROM matrixdotorg/dendrite:latest AS base
+
+FROM alpine:latest
+
+COPY --from=base /build/bin/dendrite-edu-server /usr/bin
+
+WORKDIR /etc/dendrite
+
+ENTRYPOINT ["/usr/bin/dendrite-edu-server", "--config=dendrite.yaml"]
```
Kegsay, line 9: Unclear where `dendrite.yaml` comes from. We only copied the binary from `base` so uh...?

docker/hub/Dockerfile.monolith
```diff
@@ -0,0 +1,14 @@
+FROM matrixdotorg/dendrite:latest AS base
+
+FROM alpine:latest
+
+COPY --from=base /build/bin/dendrite-monolith-server /usr/bin
+
+WORKDIR /etc/dendrite
+
+ENTRYPOINT [ "/usr/bin/dendrite-monolith-server", "--tls-cert=server.crt", "--tls-key=server.key", "--config=dendrite.yaml" ]
```
Kegsay, line 9: Unclear where the crt/key are comimg from.

docker/hub/build.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+cd $(git rev-parse --show-toplevel)
```
Kegsay, line 3: Magic and voodoo :( I'd prefer if we can avoid needing to know the working directory at all, but I guess we need to to specify the dockerfiles later on or something...

docker/hub/build.sh
```diff
@@ -0,0 +1,12 @@
...
+
+# go run ./cmd/generate-keys --private-key=docker/hub/config/matrix_key.pem -tls-cert=docker/hub/config/server.crt -tls-key=docker/hub/config/server.key
+
+docker build -f docker/hub/Dockerfile -t matrixdotorg/dendrite:latest .
```
Kegsay, line 7: Hang on, does this mean we're running `docker build` inside a Dockerfile (as `/docker/hub/Dockerfile` runs `RUN sh ./build.sh`) - that sounds a little insane, why are we doing that?

docker/hub/build.sh
```diff
@@ -0,0 +1,12 @@
...
+
+COMPONENTS=$(ls docker/hub/Dockerfile.* | cut -d "." -f2)
+for NAME in $COMPONENTS; do
+	docker build -f docker/hub/Dockerfile.$NAME -t matrixdotorg/dendrite:$NAME .
```
Kegsay, line 11: This doesn't allow any versioning. Also, it's unclear that `Dockerfile.roomserver` maps to `matrixdotorg/dendrite:roomserver`, it's surprising behaviour.

Kegsay:
I would consider using `ARG` https://docs.docker.com/engine/reference/builder/#arg for specifying which component to build, that would dramatically simplify the codebase and make it clearer because something like:
```
docker build -t dendrite:roomserver --build-arg component=roomserver .
```
is a lot clearer than building the base docker image, running build.sh and scanning the directory for `Dockerfile.{componentName}` and then tagging that.

docker/hub/build.sh
neilalexander: Yeah, this is one of those annoying "`docker build` makes the build context out of the current working directory" things. :-(

docker/hub/build.sh
neilalexander: We're not. It's running `build.sh` from the root of the repo, not the `docker/hub` folder.

docker/hub/Dockerfile
neilalexander: Yeah - there's an unfortunate naming clash here too. Will think about that.

neilalexander: So this is relative to the `WORKDIR`, which is `/build`.

docker/hub/Dockerfile.eduserver
neilalexander: I've added a readme.

docker/hub/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Starting Dendrite
+
+Once in place, start the dependencies:
+
+```
+docker-compose -f docker-compose.deps.yml up
+```
+
+Wait a few seconds for Kafka and Postgres to finish starting up, and then start a monolith:
```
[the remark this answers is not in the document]
neilalexander: With the default config, yes. Obviously you could tweak the config and not start the kafka component if desired.

docker/hub/Dockerfile.component
```diff
@@ -0,0 +1,13 @@
+FROM matrixdotorg/dendrite:latest AS base
+
+FROM alpine:latest
+
+ARG component=monolith
+ENV entrypoint=${component}
```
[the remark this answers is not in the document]
neilalexander: Apparently this is required because of how Docker handles shell substitutions.

PR #1055, opened 2020-05-22
Adds support for adding a proxy to the HTTP Client from the config

dr-bonez:
Signed-off-by: Aiden McClelland me@drbonez.dev

### Pull Request Checklist


Kegsay:
Why is this desirable?

dr-bonez:
This is a step towards adding support for anonymous .onion homeservers. It is a relatively small change that would allow people to federate with homeservers run as tor hidden services by simply running the tor executable and pointing the socks5 proxy there. There is still other work to be done, such as not requiring ssl certificates for .onion domains, since they are self-authenticated (the address *is* an ed25519 pubkey), and it is impossible to get a CA signed ssl cert for these domains without paying for one. There is a relevant issue in synapse issue [here](https://github.com/matrix-org/synapse/issues/7088). 

I'm not a big golang dev, so I still need to figure out how to set up goimports but if there is a willingness to merge this, I will do so.

internal/config/config.go
```diff
@@ -264,6 +264,12 @@ type Dendrite struct {
 	// The config for logging informations. Each hook will be added to logrus.
 	Logging []LogrusHook `yaml:"logging"`
 
+	Proxy *struct {
```
Kegsay, line 267: Please add comments to explain what this will do.

Kegsay:
Thanks!

wi18b088:
Is this feature still supported? I am having trouble directing dendrite to a local sock5 proxy. There seems to be no traffic at all coming from dendrite to the local socks proxy. My configuration:

```
proxy_outbound:
  enabled: true
  protocol: socks5
  host: localhost
  port: 9050
```

PR #1057, opened 2020-05-22
Use HTTP APIs when -api specified

neilalexander:
This updates the monolith to use the full-fat HTTP APIs instead of short-circuiting when `-api` is specified. We can then use this flag as another pipeline in sytest so that the HTTP APIs get tested without the complexity of setting up a full polylith deployment.

### DIFF
--- a/cmd/dendrite-monolith-server/main.go
+++ b/cmd/dendrite-monolith-server/main.go
@@ -27,6 +27,7 @@ import (
 	"github.com/matrix-org/dendrite/federationsender"
 	"github.com/matrix-org/dendrite/internal"
 	"github.com/matrix-org/dendrite/internal/basecomponent"
+	"github.com/matrix-org/dendrite/internal/config"
 	"github.com/matrix-org/dendrite/internal/keydb"
 	"github.com/matrix-org/dendrite/internal/transactions"
 	"github.com/matrix-org/dendrite/keyserver"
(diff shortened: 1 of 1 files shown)


cmd/dendrite-monolith-server/main.go
```diff
@@ -61,22 +74,38 @@ func main() {
...
 	)
+	if base.EnableHTTPAPIs {
+		asAPI = base.CreateHTTPAppServiceAPIs()
+	}
+
 	fsAPI := federationsender.SetupFederationSenderComponent(
 		base, federation, rsAPI, &keyRing,
 	)
+	if base.EnableHTTPAPIs {
+		fsAPI = base.CreateHTTPFederationSenderAPIs()
```
Kegsay, line 99: Uhh can we batch this please? E.g
```go
if base.EnableHTTPAPIs {
    eduInputAPI = base.CreateHTTPEDUServerAPIs()
    asAPI = base.CreateHTTPAppServiceAPIs()
    fsAPI = base.CreateHTTPFederationSenderAPIs()
}
```

neilalexander: They are used as parameters when setting up other components so the ordering is important.

Kegsay: Oh :( sadness

PR #1068, opened 2020-05-27
Per-user-per-device sync streams

neilalexander:
Sync streams and notifiers were previously per-user only, which makes it difficult to implement send-to-device messaging. This updates them so that they are per-user-per-device, allowing you to wake up specific device streams if needed.


syncapi/sync/notifier_test.go
```diff
@@ -124,22 +126,33 @@ func TestNewEventAndJoinedToRoom(t *testing.T) {
...
-	stream := lockedFetchUserStream(n, bob)
+	stream := lockedFetchUserStream(n, bob, bobDev)
 	waitForBlocking(stream, 1)
 
 	n.OnNewEvent(&randomMessageEvent, "", nil, syncPositionAfter)
 
 	wg.Wait()
 }
 
+func TestCorrectStream(t *testing.T) {
```
Kegsay, line 145: Please add a test for 1 user with 2 devices to make sure waking up 1 device doesn't wake up the other.

syncapi/sync/notifier.go
```diff
@@ -37,8 +37,8 @@ type Notifier struct {
 	streamLock *sync.Mutex
 	// The latest sync position
 	currPos types.StreamingToken
-	// A map of user_id => UserStream which can be used to wake a given user's /sync request.
-	userStreams map[string]*UserStream
+	// A map of user_id => device_id => UserStream which can be used to wake a given user's /sync request.
+	userDeviceStreams map[string]map[string]*UserDeviceStream
```
Kegsay, line 41: I'd be tempted to mux user ID / device ID together given the restricted grammar they use, it'll make it easier to work with than nested maps.

neilalexander: It's a map like this so that we can wake up all user devices easily by just looping over the second map.

syncapi/sync/notifier_test.go
neilalexander: Done.

PR #1072, opened 2020-05-28
Send-to-device support

neilalexander:
This PR adds support for send-to-device messages as a prerequisite for E2EE.


syncapi/consumers/eduserver_sendtodevice.go
```diff
@@ -0,0 +1,113 @@
+// Copyright 2019 Alex Chen
```
ara4n, line 1: please ensure that we have M.org copyright on files we touch :)

clientapi/routing/routing.go
```diff
@@ -274,6 +274,29 @@ func Setup(
...
+			vars, err := internal.URLDecodeMapValues(mux.Vars(req))
+			if err != nil {
+				return util.ErrorResponse(err)
+			}
+			txnID := vars["txnID"]
+			return SendToDevice(req, device, eduProducer, transactionsCache, vars["eventType"], &txnID)
+		}),
+	).Methods(http.MethodPut, http.MethodOptions)
+
+	// TODO: Remove this if/when sytest is fixed!
```
Kegsay, line 288: Please state what is wrong e.g using `/unstable` so it's clear what you're talking about.

cmd/dendrite-edu-server/main.go
```diff
@@ -29,8 +29,9 @@ func main() {
 			logrus.WithError(err).Warn("BaseDendrite close failed")
 		}
 	}()
+	deviceDB := base.CreateDeviceDB()
```
Kegsay, line 32: Doesn't this mean more than one component is now touching the device DB?

cmd/dendritejs/main.go
```diff
@@ -209,7 +209,7 @@ func main() {
 	p2pPublicRoomProvider := NewLibP2PPublicRoomsProvider(node)
 
 	rsAPI := roomserver.SetupRoomServerComponent(base, keyRing, federation)
-	eduInputAPI := eduserver.SetupEDUServerComponent(base, cache.New())
+	eduInputAPI := eduserver.SetupEDUServerComponent(base, cache.New(), deviceDB)
```
Kegsay, line 212: You need to also initialise a `output_send_to_device_event` topic name please.

eduserver/input/input.go
```diff
@@ -90,6 +112,60 @@ func (t *EDUServerInputAPI) sendEvent(ite *api.InputTypingEvent) error {
...
+	if domain == t.ServerName && ise.DeviceID == "*" {
+		devs, err := t.DeviceDB.GetDevicesByLocalpart(context.TODO(), localpart)
+		if err != nil {
+			return err
+		}
+		for _, dev := range devs {
+			devices = append(devices, dev.ID)
+		}
+	} else {
+		devices = append(devices, ise.DeviceID)
```
Kegsay, line 131: What happens in scenarios where the domain doesn't match and the device ID is `*`? Need commentary on why bundling `if domain == t.ServerName` makes sense at all.

federationapi/routing/send.go
```diff
@@ -265,6 +265,21 @@ func (t *txnReq) processEDUs(edus []gomatrixserverlib.EDU) {
...
+			var directPayload gomatrixserverlib.ToDeviceMessage
+			if err := json.Unmarshal(e.Content, &directPayload); err != nil {
+				util.GetLogger(t.context).WithError(err).Error("Failed to unmarshal send-to-device events")
+				continue
+			}
+			for userID, byUser := range directPayload.Messages {
+				for deviceID, message := range byUser {
+					// TODO: check that the user and the device actually exist here
+					if err := t.eduProducer.SendToDevice(t.context, directPayload.Sender, userID, deviceID, directPayload.Type, message); err != nil {
+						util.GetLogger(t.context).WithError(err).Error("Failed to send send-to-device event to edu server")
```
Kegsay, line 279: Please log the user ID and device ID as well.

eduserver/cache/cache.go
```diff
@@ -109,6 +113,17 @@ func (t *EDUCache) AddTypingUser(
 	return t.GetLatestSyncPosition()
 }
 
+// AddSendToDeviceMessage increases the sync position for
+// send-to-device updates.
+// Returns the latest sync position for typing after update.
```
Kegsay, line 118: Unsure if this is a typo or not given it does appear to increment the same counter. Needs additional commentary.

eduserver/api/output.go
```diff
@@ -32,3 +40,12 @@ type TypingEvent struct {
 	UserID string `json:"user_id"`
 	Typing bool   `json:"typing"`
 }
+
+// OutputTypingEvent is an entry in typing server output kafka log.
+// This contains the event with extra fields used to create 'm.typing' event
+// in clientapi & federation.
+type OutputSendToDeviceEvent struct {
```
Kegsay, line 47: Comments lie

syncapi/storage/interface.go
```diff
@@ -56,9 +57,9 @@ type Database interface {
 	// transaction IDs associated with the given device. These transaction IDs come
 	// from when the device sent the event via an API that included a transaction
 	// ID.
-	IncrementalSync(ctx context.Context, device authtypes.Device, fromPos, toPos types.StreamingToken, numRecentEventsPerRoom int, wantFullState bool) (*types.Response, error)
+	IncrementalSync(ctx context.Context, res *types.Response, device authtypes.Device, fromPos, toPos types.StreamingToken, numRecentEventsPerRoom int, wantFullState bool) (*types.Response, error)
 	// CompleteSync returns a complete /sync API response for the given user.
-	CompleteSync(ctx context.Context, device authtypes.Device, numRecentEventsPerRoom int) (*types.Response, error)
+	CompleteSync(ctx context.Context, res *types.Response, device authtypes.Device, numRecentEventsPerRoom int) (*types.Response, error)
```
Kegsay, line 62: Need to update the comments here. Should `res` be initialised or will the function do that for us? Not clear.

syncapi/storage/interface.go
```diff
@@ -104,4 +105,16 @@ type Database interface {
 	StreamEventsToEvents(device *authtypes.Device, in []types.StreamEvent) []gomatrixserverlib.HeaderedEvent
 	// SyncStreamPosition returns the latest position in the sync stream. Returns 0 if there are no events yet.
 	SyncStreamPosition(ctx context.Context) (types.StreamPosition, error)
+	// AddSendToDevice increases the EDU position in the cache and returns the stream position.
+	AddSendToDevice() types.StreamPosition
```
Kegsay, line 109: It seems like a weird API shape to have this decoupled from `StoreNewSendForDeviceMessage` which would surely increment the stream position..? Why does it have this shape?

syncapi/storage/interface.go
```diff
@@ -104,4 +105,16 @@ type Database interface {
...
+	// SendToDeviceUpdatesForSync returns a list of send-to-device updates, after having completed
+	// updates and deletions for previous events. The sync token should be supplied to this function so
+	// that we can clean up old events properly.
+	SendToDeviceUpdatesForSync(ctx context.Context, userID, deviceID string, token types.StreamingToken) ([]types.SendToDeviceEvent, []types.SendToDeviceNID, []types.SendToDeviceNID, error)
```
Kegsay, line 113: Unclear which token should be supplied, the old token the client gave us or the new calculated token?

syncapi/storage/interface.go
```diff
@@ -104,4 +105,16 @@ type Database interface {
...
+	// StoreNewSendForDeviceMessage stores a new send-to-device event for a user's device.
+	StoreNewSendForDeviceMessage(ctx context.Context, streamPos types.StreamPosition, userID, deviceID string, event gomatrixserverlib.SendToDeviceEvent) (types.StreamPosition, error)
+	// CleanSendToDeviceUpdates will update or remove any send-to-device updates based on the given sync.
+	CleanSendToDeviceUpdates(ctx context.Context, toUpdate, toDelete []types.SendToDeviceNID, token types.StreamingToken) (err error)
```
Kegsay, line 117: Unclear which token should be supplied, the old token the client gave us or the new calculated token?

syncapi/storage/interface.go
```diff
@@ -104,4 +105,16 @@ type Database interface {
...
+	// SendToDeviceUpdatesWaiting returns true if there are send-to-device updates waiting to be sent.
+	SendToDeviceUpdatesWaiting(ctx context.Context, txn *sql.Tx, userID, deviceID string) (bool, error)
```
Kegsay, line 119: This is the first time ever that we have `sql.Tx` leaking across the storage interface. Why?

syncapi/storage/postgres/send_to_device_table.go
```diff
@@ -0,0 +1,170 @@
...
+	id BIGINT PRIMARY KEY DEFAULT nextval('syncapi_send_to_device_id'),
+	-- The user ID to send the message to.
+	user_id TEXT NOT NULL,
+	-- The device ID to send the message to.
+	device_id TEXT NOT NULL,
+	-- The event content JSON.
+	content TEXT NOT NULL,
+	-- The sync token that was supplied when we tried to send the message,
+	-- or NULL if we haven't tried to send it yet.
+	sent_by_token TEXT
```
Kegsay, line 43: Unsure what this is supposed to mean. Please go into more detail about how you expect this table to function, the more detail the better. E.g something like https://github.com/matrix-org/dendrite/blob/master/syncapi/storage/tables/interface.go#L70 would be fab.

syncapi/sync/requestpool.go
```diff
@@ -54,17 +57,27 @@ func (rp *RequestPool) OnIncomingSyncRequest(req *http.Request, device *authtype
...
+		"timeout":   syncReq.timeout,
+		"limit":     syncReq.limit,
 	})
 
 	currPos := rp.notifier.CurrentPosition()
 
-	if shouldReturnImmediately(syncReq) {
+	returnImmediately := shouldReturnImmediately(syncReq)
+	if !returnImmediately {
+		if waiting, werr := rp.db.SendToDeviceUpdatesWaiting(
```
Kegsay, line 73: It feels like this check should be in `shouldReturnImmediately` so we don't have this strange control flow where we then check additional places.

syncapi/sync/requestpool.go
```diff
@@ -134,13 +146,51 @@ func (rp *RequestPool) OnIncomingSyncRequest(req *http.Request, device *authtype
...
+	// See if we have any new tasks to do for the send-to-device messaging.
+	events, updates, deletions, err := rp.db.SendToDeviceUpdatesForSync(req.ctx, req.device.UserID, req.device.ID, since)
+	if err != nil {
+		return nil, err
+	}
+
+	// Before we return the sync response, make sure that we take action on
+	// any send-to-device database updates or deletions that we need to do.
+	// Then add the updates into the sync response.
+	defer func() {
```
Kegsay, line 165: Why are we doing this in a `defer`?

internal/sql.go
```diff
@@ -107,3 +111,54 @@ type DbProperties interface {
...
+	f    func(txn *sql.Tx) error
+	wait chan error
+}
+
+// Do queues a task to be run by a TransactionWriter. The function
+// provided will be ran within a transaction as supplied by the
+// database parameter. This will block until the task is finished.
+func (w *TransactionWriter) Do(db *sql.DB, f func(txn *sql.Tx) error) error {
+	if w.todo == nil {
+		w.todo = make(chan transactionWriterTask)
```
Kegsay, line 135: We can't do this here. Make a constructor for it please.

Having it here means we will race on the first calls to `Do`: both threads may see `todo == nil` then make the channel, resulting in a dropped message.

syncapi/storage/interface.go
Kegsay: Also unclear what return order the `[]types.SendToDeviceNID` should be - use named return args here for clarity.

syncapi/types/types.go
```diff
@@ -296,13 +296,14 @@ type Response struct {
...
+	} `json:"to_device"`
 }
 
 // NewResponse creates an empty response with initialised maps.
-func NewResponse(token StreamingToken) *Response {
-	res := Response{
-		NextBatch: token.String(),
-	}
+func NewResponse() *Response {
+	res := Response{}
```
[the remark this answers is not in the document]
neilalexander: Because at this point we don't actually know the definitive `NextBatch`, since the send-to-device messages can increase the EDU position.

cmd/dendrite-edu-server/main.go
neilalexander: Just about all of the components touch the device DB in some way or other already - it really needs the same kind of treatment as the server key API got.

eduserver/input/input.go
neilalexander: The idea is that, if it's targeted locally, we expand `*` into individual device names. If it isn't targeted locally then we just want to send the `*` unmodified, and the federation sender will (eventually, in a future PR) pick that up and shift it as-is.

syncapi/storage/interface.go
neilalexander: That's a mistake from where I've copied a function signature. I'll remove that parameter.

eduserver/input/input.go
Kegsay: Please add exactly this commentary to the code :)

syncapi/sync/requestpool.go
neilalexander: Mostly from when I was shuffling things around - it is a defer no longer.

syncapi/storage/interface.go
```diff
@@ -107,14 +108,24 @@ type Database interface {
...
 	AddSendToDevice() types.StreamPosition
-	// SendToDeviceUpdatesForSync returns a list of send-to-device updates, after having completed
-	// updates and deletions for previous events. The sync token should be supplied to this function so
-	// that we can clean up old events properly.
-	SendToDeviceUpdatesForSync(ctx context.Context, userID, deviceID string, token types.StreamingToken) ([]types.SendToDeviceEvent, []types.SendToDeviceNID, []types.SendToDeviceNID, error)
+	// SendToDeviceUpdatesForSync returns a list of send-to-device updates. It returns three lists:
+	// - "events": a list of send-to-device events that should be included in the sync
+	// - "changes": a list of send-to-device events that should be updated in the database by
+	//      CleanSendToDeviceUpdates
+	// - "defletions": a list of send-to-device events which have been confirmed as sent and
```
Kegsay, line 115: Typo :)

neilalexander: Curse my fingers!

PR #1074, opened 2020-05-29
p2p: de-dupe and shuffle public rooms; implement keydb

Kegsay:
- We don't want a serverKeyAPI as fetching keys doesn't need a DB.
- De-dupe rooms so we don't see them multiple times, but shuffle the
  alias we join via so we don't all flood a single server.

### DIFF
--- a/cmd/dendritejs/main.go
+++ b/cmd/dendritejs/main.go
@@ -37,7 +37,6 @@ import (
 	"github.com/matrix-org/dendrite/publicroomsapi"
 	"github.com/matrix-org/dendrite/publicroomsapi/storage"
 	"github.com/matrix-org/dendrite/roomserver"
-	"github.com/matrix-org/dendrite/serverkeyapi"
 	"github.com/matrix-org/dendrite/syncapi"
 	go_http_js_libp2p "github.com/matrix-org/go-http-js-libp2p"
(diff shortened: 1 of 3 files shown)


cmd/dendritejs/main.go
```diff
@@ -197,14 +196,12 @@ func main() {
...
 	federation := createFederationClient(cfg, node)
 
-	serverKeyAPI := serverkeyapi.SetupServerKeyAPIComponent(
-		base, federation,
-	)
+	fetcher := &libp2pKeyFetcher{}
 	keyRing := gomatrixserverlib.KeyRing{
 		KeyFetchers: []gomatrixserverlib.KeyFetcher{
-			&libp2pKeyFetcher{},
+			fetcher,
```
neilalexander, line 202: I think if you already have `fetcher` satisfying `KeyDatabase` then you probably don't need to specify any `KeyFetchers` at all.

Kegsay: Can't hurt to have it?

neilalexander: Won't do any harm, nope. It'll probably never get hit.

Kegsay: That's what she said

PR #1076, opened 2020-06-01
Fix pg user already exists

### LINKED ISSUE #1002: Trying to register an account that already exists returns useless error
`failed to create account: pq: Could not complete operation in a failed transaction`

S7evinK:
Fixes #1002 

### Pull Request Checklist


clientapi/auth/storage/accounts/postgres/storage.go
```diff
@@ -164,7 +164,7 @@ func (d *Database) createAccount(
 	}
 	if err := d.profiles.insertProfile(ctx, txn, localpart); err != nil {
 		if internal.IsUniqueConstraintViolationErr(err) {
-			return nil, nil
+			return nil, internal.ErrUserExists
```
Kegsay, line 167: You need to update `sqlite` as well please to return this sentinel error.

S7evinK:
I've add a `sqlite3.ErrConstraint` check in `internal.IsUniqueConstraintViolationErr`, even though the file is name `postgres.go`.

Also, maybe i should move the new error to the interface file?

internal/postgres.go
```diff
@@ -16,10 +16,22 @@
...
-import "github.com/lib/pq"
+import (
+	"errors"
 
-// IsUniqueConstraintViolationErr returns true if the error is a postgresql unique_violation error
+	"github.com/lib/pq"
+	"github.com/mattn/go-sqlite3"
+)
+
+// IsUniqueConstraintViolationErr returns true if the error is a postgresql unique_violation or sqlite3.ErrConstraint
```
Kegsay, line 26: We can't mix sqlite and postgres in the same file otherwise dendritejs will fail to compile (pq cannot be compiled to WASM). Keep the check contained to `sqlite` packages for now please.

Kegsay:
Please add the function comments to `clientapi/auth/storage/accounts/interface.go` instead of the concrete impls. Thanks!

internal/postgres.go
S7evinK: I thought since there's also a `postgres_wasm.go` (with `+build wasm`) that this file would be used when compiling for wasm.
Nevertheless, i've reverted the change and moved the check directly to the `sqlite` package.

Kegsay:
LGTM thanks!

PR #1077, opened 2020-06-01
Add a way to force federationsender to retry sending transactions

Kegsay:
And use it in P2P mode when we pick up new nodes.

Entirely untested.


federationsender/queue/destinationqueue.go
```diff
@@ -47,6 +47,23 @@ type destinationQueue struct {
...
+// retry will clear the blacklist state and attempt to send built up events to the server,
+// resetting and interrupting any backoff timers.
+func (oq *destinationQueue) retry() {
+	// TODO: We don't send all events in the case where the server has been blacklisted as we
+	// drop events instead then. This means we will send the oldest N events (chan size, currently 128)
+	// and then skip ahead a lot which feels non-ideal but equally we can't persist thousands of events
+	// in-memory to maybe-send it one day. Ideally we would just shove these pending events in a database
+	// so we can send a lot of events.
+	oq.statistics.Success()
+	oq.retryServerCh <- true
```
neilalexander, line 62: If the queue worker isn't actively backing off at this point, or it isn't running at all, then wouldn't this block indefinitely?

It might make sense to add another `atomic.Bool` alongside `running` (maybe call it `blocking` or something) and ensure that it's only `true` when we're *really* backing off.

federationsender/queue/queue.go
```diff
@@ -66,6 +72,7 @@ func (oqs *OutgoingQueues) getQueue(destination gomatrixserverlib.ServerName) *d
 			incomingPDUs:    make(chan *gomatrixserverlib.HeaderedEvent, 128),
 			incomingEDUs:    make(chan *gomatrixserverlib.EDU, 128),
 			incomingInvites: make(chan *gomatrixserverlib.InviteV2Request, 128),
+			retryServerCh:   make(chan bool),
```
neilalexander, line 75: We don't seem to `close` this anywhere - I suspect the rate of leakage is low, but still not ideal. If we have a way to check if the worker is really blocked (as per the other comment) then we can `defer` a closure of this at the beginning of `backgroundSend()`.

Kegsay: We don't close any of the other channels so I don't see a problem here.

neilalexander: You make a good point.

federationsender/queue/destinationqueue.go
Kegsay: Hmm, yes this will block forever. I'll have a think.

PR #1087, opened 2020-06-03
Enable room version 6

neilalexander:
This enables room version 6 and tries to return more useful errors in some places.

It doesn't set room version 6 as the default yet - we can do this when Synapse does.


Kegsay:
You'll need to enable https://github.com/matrix-org/sytest/blob/dd51dcbcc8afeb997886fe10bc1a312165f1e2db/tests/80torture/20json.pl#L32 too if you update GMSL

PR #1100, opened 2020-06-05
Use AuthChainProvider to try and speed up federated joins

neilalexander:
This uses the `AuthChainProvider`, as in matrix-org/gomatrixserverlib#205, to try and speed up federated room joins. We'll now supply a `AuthChainProvider` so that when we encounter a missing auth event (which does happen, annoyingly), we don't need to retry the entire verification phase (including signatures and all) for each missing event.

### DIFF
--- a/federationsender/internal/perform/join.go
+++ b/federationsender/internal/perform/join.go
@@ -38,7 +38,7 @@ func (r joinContext) CheckSendJoinResponse(
 	// Define a function which we can pass to Check to retrieve missing
 	// auth events inline. This greatly increases our chances of not having
 	// to repeat the entire set of checks just for a missing event or two.
-	missingAuth := func(eventID string) (*gomatrixserverlib.Event, error) {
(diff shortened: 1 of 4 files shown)


federationsender/internal/perform/join.go
```diff
@@ -33,38 +33,58 @@ func (r joinContext) CheckSendJoinResponse(
...
+			tx, txerr := r.federation.GetEvent(ctx, server, eventID)
+			if txerr != nil {
+				return nil, fmt.Errorf("missingAuth r.federation.GetEvent: %w", txerr)
+			}
+
+			// For each event returned, add it to the set of return events. We
+			// also will populate the retries, in case someone asks for this
+			// event ID again.
+			for _, pdu := range tx.PDUs {
+				ev, everr := gomatrixserverlib.NewEventFromUntrustedJSON(pdu, roomVersion)
```
Kegsay, line 71: Don't you also need to check the signature?

neilalexander: Hm, yes, good point - by this time the `RespSendJoin` will have had signature checks but new events that we've provided won't have. I'll add signature checking.

PR #1101, opened 2020-06-05
More flexible caching

neilalexander:
This PR makes the following changes to the caching code:

- Caches now have the ability to be either mutable or immutable
- Code for setting up in-memory LRUs is deduped now, for things like setting up Prometheus metrics and building the LRU cache itself
- A number of interfaces now expose specific behaviours, allowing implementations to take creative license if needed on how they serialise data
- Functions which cast types, like `GetRoomVersion` etc, no longer need to be duplicated per-implementation

Net effects are:

- Nothing should escape the watchful eye of Prometheus
- Server keys being updated for new validity periods will no longer panic due to immutability
- Adding new implementations (Redis?) in the future should be dead simple
- We can change *everything* about this as long as we maintain the `RoomVersionCache` and `ServerKeyCache` types, without it being particularly breaking


neilalexander:
On the contrary, if we *don't* make some effort to track caches centrally then we just add them in places that suit us (like with `transactions.Cache`, `EDUCache` and so on), they get forgotten about and, *worse*, they don't get reported on because they probably were never wired into Prometheus.

How big is a `transactions.Cache` or an `EDUCache`? I have no idea. I don't know how often they are used, how much data they store nor do I know how many of them there even are. There are probably lots of little inline caches in other places too - anything with longevity in memory really should be managed by something.

We might decide to rip up caching and rethink it in the future, and that's fine, but at least if we tracked them centrally we'd know what we're touching when we do.

Kegsay:
Overall conclusion was \*shrug\* we need to have this, even though it's opening a can of worms. #1102 filed.

PR #1103, opened 2020-06-05
WIP base refactor

Kegsay:
This PR:
 - Renames `basecomponent.BaseDendrite` to `setup.Base`
 - Automatically creates account/device DBs and the federation client on instantiation.
 - Automatically sets up internal HTTP clients if `useHTTPAPIs == true`.
 - Exposes the clients/dbs as public fields.
 - Makes all `SetupFooComponent` primarily just take a `setup.Base`, and pull out the fields which they need to use.

This means most polylith components are now reduced to:
```go
func main() {
	cfg := setup.ParseFlags(false)
	base := setup.NewBase(cfg, "FederationSender", true)
	defer base.Close() // nolint: errcheck
	federationsender.SetupFederationSenderComponent(base)
	base.SetupAndServeHTTP(string(base.Cfg.Bind.FederationSender), string(base.Cfg.Listen.FederationSender))
}
```

The monolith doesn't need HTTP clients if `-api` is not provided, so in those circumstances there are now `SetFoo(fooAPI)` functions which replace the public fields. Accessing `Foo()` when it is unset causes a panic early.

There's a few problems that I still need to work on:
 - `dendrite-demo-libp2p` replaces the public rooms DB which is no longer possible.
 - `dendritejs` now needs to implement `ServerKeyAPI` so it can pull out the key ring. Given no one seems to actually want the server key API itself I wonder how useful this struct actually is (they all immediately call `KeyRing()`).
 - We've lost the ability to identify dependencies at compile time, making the ordering in the monolith code very important. I need to think about how to better structure this.

### DIFF
--- a/appservice/appservice.go
+++ b/appservice/appservice.go
@@ -32,24 +32,19 @@ import (
 	"github.com/matrix-org/dendrite/clientapi/auth/storage/accounts"
 	"github.com/matrix-org/dendrite/clientapi/auth/storage/devices"
 	"github.com/matrix-org/dendrite/internal"
-	"github.com/matrix-org/dendrite/internal/basecomponent"
 	"github.com/matrix-org/dendrite/internal/config"
+	"github.com/matrix-org/dendrite/internal/setup"
 	"github.com/matrix-org/dendrite/internal/transactions"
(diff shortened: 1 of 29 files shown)


Kegsay:
I'm unhappy with this in its current form. :/ Closing.

PR #1105, opened 2020-06-07
WIP: Adding MySQL database support

Murazaki:
### Pull Request Checklist


Murazaki:
This merge request depends on this one in matrix-org/naffka : https://github.com/matrix-org/naffka/pull/15

Kegsay:
Thank you for the effort you've put into this PR. Unfortunately, we do not wish to add MySQL support at this time for the following reasons:
 - MySQL has no specific niche to fill. Large scale deployments can use Postgres, and small scale/embedded deployments can use SQLite. It's unclear what benefit MySQL would add.
 - Every database engine we add involves significant amounts of ongoing work to maintain. Each table we add we have to write code for each database engine and the work to implement and test this is nontrivial. It means we cannot iterate as quickly.

When performing large changes (9k added lines of code!) it would help to ask us _before_ you begin the work to avoid wasted effort. Thanks.

PR #1108, opened 2020-06-10
Yggdrasil-based P2P demo

neilalexander:
This includes an early Yggdrasil-based P2P demo, where federation traffic is carried over Yggdrasil sessions which are multiplexed using Yamux. 

To install and run:
```
go install ./cmd/dendrite-demo-yggdrasil
~/go/bin/dendrite-demo-yggdrasil -peer tcp://x.x.x.x:xxxxx
```
... where a peer is selected from [public peers](http://github.com/yggdrasil-network/public-peers) (with the caveat that not all of the peers listed there are still up). Look for `Connected TCP:` lines in the output for confirmation.

Then use Riot to register and log into `http://localhost:8008`.

Then try to join `#beachparty:a720a2ce30bafe82623b6b784587760b184232b0fe32cad9f51801c18470b414` as an example.


cmd/dendrite-demo-yggdrasil/main.go
```diff
@@ -0,0 +1,221 @@
...
+	defer base.Close() // nolint: errcheck
+
+	accountDB := base.CreateAccountsDB()
+	deviceDB := base.CreateDeviceDB()
+	federation := createFederationClient(base, ygg)
+
+	serverKeyAPI := serverkeyapi.NewInternalAPI(
+		base.Cfg, federation, base.Caches,
+	)
+	if base.UseHTTPAPIs {
```
Kegsay, line 138: This is always `false` so remove?

PR #1110, opened 2020-06-10
Use ServerKeyAPI as keyring

neilalexander:
This should fix cases where the monolith doesn't push key fetch and database operations through the server key API.

### DIFF
--- a/serverkeyapi/internal/api.go
+++ b/serverkeyapi/internal/api.go
@@ -88,5 +88,5 @@ func (s *ServerKeyAPI) FetchKeys(
 }
 
 func (s *ServerKeyAPI) FetcherName() string {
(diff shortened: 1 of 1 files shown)


neilalexander:
Will merge if Sytest is happy, but the general rule of thumb is that the right one to use is always what `ServerKeyInternalAPI.KeyRing()` returns. 

If the `ServerKeyInternalAPI` is actually a `httpServerKeyInternalAPI` (like in polylith mode) then it returns a keyring that proxies the requests through the HTTP interface via the functions in `serverkeyapi/inthttp/client.go`. 

If the `ServerKeyInternalAPI` is actually a `ServerKeyAPI` (like in monolith mode) then it returns a keyring that wraps gomatrixserverlib and proxies the requests through the functions in `serverkeyapi/internal/api.go`.

PR #1116, opened 2020-06-11
More key tweaks

neilalexander:
This makes, yet again, more tweaks to the key fetching behaviour:

- The `ServerKeys` cache will no longer return a result that isn't valid now (and will actively evict such cache entries if it finds them)
- The `ServerKeyAPI` and `httpServerKeyInternalAPI` keyrings no longer returns itself as a `KeyFetcher` (as everything they do is already covered by the fact they return themselves as the `KeyDatabase`)
- Use `gomatrixserverlib.WasValidAt` for consistency
- Adds a missing validity check in `ServerKeyAPI`
- Removes one or two redundant checks from other places where a validity check has already been done

### DIFF
--- a/internal/caching/cache_serverkeys.go
+++ b/internal/caching/cache_serverkeys.go
@@ -2,6 +2,7 @@ package caching
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/matrix-org/gomatrixserverlib"
 )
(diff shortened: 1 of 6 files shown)


neilalexander:
Repeatedly making changes and seeing whether it gets better is what iterative development is all about. ;-)

I'm happy to add a todo issue to make further tests for this in the future but I don't see that this should block merging this fix, especially now that it seems to have resolved the problem.

Issue #1124, opened 2020-06-12
Add a User API


Kegsay:
Which contains the account/device DBs with an internal API for access. Currently it's a free-for-all with 5-6 components accessing the same DB. This is bad for SQLite in particular which should only have 1 writer per DB.

Kegsay:
https://github.com/matrix-org/dendrite/pull/1141
https://github.com/matrix-org/dendrite/pull/1139
https://github.com/matrix-org/dendrite/pull/1138
https://github.com/matrix-org/dendrite/pull/1136
https://github.com/matrix-org/dendrite/pull/1135
https://github.com/matrix-org/dendrite/pull/1133
https://github.com/matrix-org/dendrite/pull/1127

PR #1125, opened 2020-06-12
Current wiring

### LINKED ISSUE #893: General arch: data propagation

Kegsay:
Closes #893 as we now justify each arrow of communication between components.

There are major rooms for improvement here, notably:
 - Add a `UserAPI` to guard access to device/user DB: #1124
 - Add internal API for `UserAPI` to expose profile info, moving arrows (Fed|ClientAPI->AS) to (Fed|ClientAPI->UserAPI), allowing us to remove the internal APIs from AS entirely.
 - Reconsider Kafka log (Roomserver -> ClientAPI) - we could expose a "all joined rooms for this user" endpoint on Roomserver which means ClientAPI doesn't need to have an entire database devoted to this just for profile updates.

### DIFF
--- a/clientapi/clientapi.go
+++ b/clientapi/clientapi.go
@@ -41,7 +41,6 @@ func AddPublicRoutes(
 	deviceDB devices.Database,
 	accountsDB accounts.Database,
 	federation *gomatrixserverlib.FederationClient,
-	keyRing *gomatrixserverlib.KeyRing,
 	rsAPI roomserverAPI.RoomserverInternalAPI,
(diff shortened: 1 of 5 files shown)


Kegsay:
That's what I thought, but when adding this line actually it's just `federationapi`, `federationsender`(for obvious reasons) and `roomserver` (for checking backfilled events) - so adding these lines in.

PR #1126, opened 2020-06-12
Don't include current state in processEventWithMissingState

neilalexander:
This PR changes `processEventWithMissingState` so that it no longer includes the current room state. I don't *think* that makes sense to do, since what we're trying to do is to work out what the room state was *before* that event (that is, *after* the event's `prev_events`) so that we can send it to the roomserver for a state snapshot to be built.

In some cases this can result in an old event failing to be allowed because the current room state wins over the state before the event.

I'm guessing that the intention here was to make sure that we don't overwrite the current room state with old state, but that should be the responsibility of the roomserver to do - probably we should be building a new state snapshot and resolve the state against the current room state at that point, but not before.


Kegsay:
https://github.com/matrix-org/synapse/blob/229eb81498b0fe1da81e9b5b333a0285acde9446/synapse/handlers/federation.py#L325 is relevant synapse info

Kegsay:
> Say you have a forward extremity F with state you trust, and a malicious event M comes in which points to a prev event P, then if you take the state at M you don't do any state res. If you fetch state for P then to get the new current state of the room you have to do state res across F and P. and if P is an old state or is trying to do something dodgy then state res will come to a vaguely sane conclusion, as it can see that new stuff has happened in F

was from Erik a while ago.

Kegsay:
It's annoying we have no sytest seemingly to test this scenario, or maybe there is and there is a silly reason why it is failing - I'll check this week.

Kegsay:
I checked and there appears to be none. I brought it up with the Synapse crew, pending feedback.

neilalexander:
> I checked and there appears to be none. I brought it up with the Synapse crew, pending feedback.

My feeling is that this should actually be handled in two separate steps:

1. **State at event:** Resolve the state *before the event* without including current state, as this PR is proposing. Otherwise it’s possible for a new state event that happened afterwards to prevent us from allowing a state change that was otherwise valid at the time of the event;
2. **Current state:** Fast-forward state snapshots that have taken place since that event with the new change, all the way up to the current state, including both the newly-inserted state before the event and the current state snapshot that is being updated, performing state resolution, to get newly resolved state.

Currently we only do step 1 and try to update the current room state directly from that, which is where we run into problems. We don’t have a concept in Dendrite yet of fast-forwarding state changes but ultimately we need that to ensure that the state before each event in the timeline is correct after backfilling/getting missing events.

Kegsay:
Basically it should be more like:
```
 A         A    
 |        / \    
 B       B   D   
 |       |   |   
 C--.    C   E  
 |  |    |   |  
... |    |  ...  
 |  |    |   |  
 X  |    |   X   
 |  |    |   |  
 Y  |    |   Y  
 |  |    |   |   
 Z--`    `---Z  
```

and less like:
```
 A         A     
 |        / \      
 B       B   D  
 |       |   |   
 C--.    C   E    
 |  |    |   |    
... |    |  ...  
 |  |    |   |   
 X--`     `--X   
 |           |   
 Y           Y   
 |           |  
 Z           Z 
```
which is what we currently have.

Kegsay:
Which we may already have, depending on how intelligent the roomserver is. `QueryLatestEventsAndStateResponse` already has `LatestEvents []gomatrixserverlib.EventReference `json:"latest_events"`` so we just need to make sure that we have 2 forward extremities when we do /gme. Assuming we do in fact do this, this PR is indeed the fix.

Kegsay:
It looks like it does this correctly because it looks for prev_events, and if they are missing it assigns the event as a forward extremity. So in other words, LGTM

PR #1127, opened 2020-06-12
Add bare bones user API

Kegsay:
with tests! It's not hooked up to anything yet.

### DIFF
--- a/userapi/internal/api.go
+++ b/userapi/internal/api.go
@@ -1,0 +1,53 @@
+// Copyright 2020 The Matrix.org Foundation C.I.C.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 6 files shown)


userapi/internal/api.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"github.com/matrix-org/dendrite/clientapi/auth/storage/accounts"
+	"github.com/matrix-org/dendrite/clientapi/auth/storage/devices"
+	"github.com/matrix-org/dendrite/userapi/api"
+	"github.com/matrix-org/gomatrixserverlib"
+)
+
+type UserInternalAPI struct {
+	AccountDB  accounts.Database
+	DeviceDB   devices.Database
```
neilalexander, line 30: Is the plan to do device management here too?

Kegsay: Yes. The way these DBs are structured, accounts=profile info and devices=access tokens. Every public-facing component needs to check access tokens, and the majority of them also need access to profile info (client/fed api to service /profile requests, syncapi for access_token + profile data when `since` is empty, etc)

Issue #1119, opened 2020-06-12
Server key API tests


neilalexander:
It took a little while to get the server key API to behave properly with the key renewal (see #1116, #1110, #1101, #1094, matrix-org/gomatrixserverlib#206).

It would be good to add some tests to make sure that the server key API behaves as expected, both using `internal.ServerKeyAPI` and `inthttp.httpServerKeyInternalAPI`, verifying also that the caches behave as expected.

PR #1129, opened 2020-06-15
More server key updates, tests

neilalexander:
This introduces some further server key API changes:

- The server key API now will return old results from the database and will even accept them from federation for satisfying key requests that are because of old events (this happens when a server goes offline), but should be more responsible with working out validites
- The server key cache now returns results that were valid for the request, rather than the current timestamp (which sometimes did the wrong thing)
- The server key cache takes a new setup parameter to determine whether to set up Prometheus logging or not (enables testing with multiple caches)
- The server key API no longer stores our own keys in the database at all, but instead intercepts fetch requests for lookups for our own keys (the added rationale for this is that no server admin should need to touch the server key database to manage their own keys!)
- The server key API now decides whether to store fetched results separately to the results that we return (there's no use in storing results unless they are newer than what we already have)
- The server key API now more robustly checks that the requests were satisfied properly, rather than trying to obscurely count keys
- Outgoing key fetcher requests are now limited to 20 seconds

There are now also some tests for checking caching and renewal behaviour, fixing #1119.


internal/caching/cache_serverkeys.go
```diff
@@ -16,22 +15,21 @@ const (
 // ServerKeyCache contains the subset of functions needed for
 // a server key cache.
 type ServerKeyCache interface {
-	GetServerKey(request gomatrixserverlib.PublicKeyLookupRequest) (response gomatrixserverlib.PublicKeyLookupResult, ok bool)
+	GetServerKey(request gomatrixserverlib.PublicKeyLookupRequest, timestamp gomatrixserverlib.Timestamp) (response gomatrixserverlib.PublicKeyLookupResult, ok bool)
```
Kegsay, line 18: It is not entirely clear what `timestamp` means - the current time? Can we opt out of checks by passing `0`? Please document.

Kegsay: Also, please document the logic that implementations of this interface need to make e.g what happens if things are not found / found but expired / etc.

serverkeyapi/api/api.go
```diff
@@ -22,6 +22,12 @@ type ServerKeyInternalAPI interface {
 		request *QueryPublicKeysRequest,
 		response *QueryPublicKeysResponse,
 	) error
+
+	QueryLocalKeys(
```
Kegsay, line 26: I don't know what local means in this context. We already have a query keys method - `QueryPublicKeys` - why do we need another?

serverkeyapi/internal/api.go
```diff
@@ -2,20 +2,52 @@ package internal
...
 type ServerKeyAPI struct {
 	api.ServerKeyInternalAPI
 
+	Cfg        *config.Dendrite
 	OurKeyRing gomatrixserverlib.KeyRing
 	FedClient  *gomatrixserverlib.FederationClient
 }
 
+func (s *ServerKeyAPI) QueryLocalKeys(ctx context.Context, request *api.QueryLocalKeysRequest, response *api.QueryLocalKeysResponse) error {
+	publicKey := s.Cfg.Matrix.PrivateKey.Public().(ed25519.PublicKey)
```
Kegsay, line 25: I don't know why this is A Thing. We pass the `config` around all components, and if they need the local key then they can just grab it? Why put it here, this seems like more work?

Kegsay: Let's put it another way: I would rather have a 'startup time' dep (needing the key in a component's config) than a runtime dep (needing server key API).

serverkeyapi/internal/api.go
```diff
@@ -33,10 +65,12 @@ func (s *ServerKeyAPI) StoreKeys(
 	// Run in a background context - we don't want to stop this work just
 	// because the caller gives up waiting.
 	ctx := context.Background()
+
 	// Store any keys that we were given in our database.
 	return s.OurKeyRing.KeyDatabase.StoreKeys(ctx, results)
 }
 
+// nolint:gocyclo
```
Kegsay, line 73: Uhh, this is a 150 line function. Large functions make it harder to read because:
 - There's inevitably tons of local variables flying around, making it harder to grok.
 - It's easy to lose track of the point of all the code, as you read the function signature pages and pages ago.
 - It encourages long control loop bodies which become confusing to read (e.g 10s of lines `if` statements make it hard to realise where the `else` block pairs up with, and what the original branch condition was).

Please break it up!

serverkeyapi/internal/api.go
```diff
@@ -45,51 +79,153 @@ func (s *ServerKeyAPI) FetchKeys(
...
+					VerifyKey:    verifyKeys.VerifyKey,
+					ExpiredTS:    verifyKeys.ExpiredTS,
+					ValidUntilTS: gomatrixserverlib.PublicKeyNotValid,
+				}
+			}
+		}
+	}
+
+	// Then consult our local database and see if we have the requested
 	// keys. These might come from a cache, depending on the database
```
Kegsay, line 125: Might it come from a cache? I thought the server key api did the caching..?

serverkeyapi/internal/api.go
```diff
@@ -45,51 +79,153 @@ func (s *ServerKeyAPI) FetchKeys(
...
-		// We successfully got some keys. Add them to the results and
-		// remove them from the request list.
+		// We successfully got some keys. Add them to the results.
 		for req, res := range dbResults {
-			if !res.WasValidAt(now, true) {
-				// We appear to be past the key validity. Don't return this
-				// key with the results.
-				continue
-			}
 			results[req] = res
```
Kegsay, line 130: We're now adding invalid keys to `results`. Why?

serverkeyapi/internal/api.go
```diff
@@ -45,51 +79,153 @@ func (s *ServerKeyAPI) FetchKeys(
...
 	for _, fetcher := range s.OurKeyRing.KeyFetchers {
+		// If there are no more keys to look up then stop.
 		if len(requests) == 0 {
 			break
 		}
-		if fetcherResults, err := fetcher.FetchKeys(ctx, requests); err == nil {
-			// We successfully got some keys. Add them to the results and
-			// remove them from the request list.
+		fetcherCtx, fetcherCancel := context.WithTimeout(ctx, time.Second*30)
+		defer fetcherCancel()
```
Kegsay, line 150: This won't behave as you expect. Put the body of the `for` into a separate function if you wish to defer cancel.

serverkeyapi/internal/api.go
```diff
@@ -45,51 +79,153 @@ func (s *ServerKeyAPI) FetchKeys(
...
+				}).Errorf("Failed to store keys in the database")
 				return nil, fmt.Errorf("server key API failed to store retrieved keys: %w", err)
 			}
+
+			if len(storeResults) > 0 {
+				logrus.WithFields(logrus.Fields{
+					"fetcher_name": fetcher.FetcherName(),
+				}).Infof("Updated %d of %d key(s) in database", len(storeResults), len(results))
+			}
+		} else {
```
Kegsay, line 207: This is what I mean about the whole "10s of lines for if statements makes it hard to grok what the else part means". I have to literally carefully scan back up the page to try to find it, 52 lines above.

Kegsay: I'd just invert then outdent, and then return (assuming you put the body of the for loop in a separate function) e.g

```go
if fetcherResults, err := fetcher.FetchKeys(fetcherCtx, requests); err != nil {
    logrus.WithError(err).WithFields(logrus.Fields{
        "fetcher_name": fetcher.FetcherName(),
    }).Errorf("Failed to retrieve %d key(s)", len(requests))
    return
}
// rest of body
```

serverkeyapi/serverkeyapi_test.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+
+	// Now that we have built our server key APIs, start the
+	// rest of the tests.
+	os.Exit(m.Run())
+}
+
+type MockRoundTripper struct{}
+
+func (m *MockRoundTripper) RoundTrip(req *http.Request) (res *http.Response, err error) {
```
Kegsay, line 101: Please check the request URL path - it helps for documentation purposes as well as correctness as I don't actually know which request we're intercepting...

serverkeyapi/serverkeyapi_test.go
```diff
@@ -0,0 +1,313 @@
...
+
+	"github.com/matrix-org/dendrite/internal/caching"
+	"github.com/matrix-org/dendrite/internal/config"
+	"github.com/matrix-org/dendrite/serverkeyapi/api"
+	"github.com/matrix-org/gomatrixserverlib"
+)
+
+type server struct {
+	name      gomatrixserverlib.ServerName
+	validity  time.Duration
```
Kegsay, line 24: Is this "valid for duration from now" or "valid for duration from key creation"?

serverkeyapi/internal/api.go
neilalexander: The next line of the comment says it can come from a cache :-) In our case yes, it does, as the server key API sets up a cache wrapper around the inner database.

serverkeyapi/internal/api.go
neilalexander: The logic in this PR aims to make a best-effort attempt to find the latest keys, but if it can't find newer keys, it will return the ones that we know about. This the only way to satisfy verifications for when the event happened in the past and the key was never renewed (i.e. the origin server is now dead).

serverkeyapi/internal/api.go
neilalexander: I've rearranged this all now, but I'm aware that the `defer` wouldn't have ran until the function ended. It was mostly to avoid leaky contexts.

serverkeyapi/internal/api.go
neilalexander: `FetchKeys` is broken up now.

serverkeyapi/serverkeyapi_test.go
neilalexander: Well, it's effectively just writing to `config.Matrix.KeyValidityPeriod`.

The behaviour around `/_matrix/key/v2/server` has always been that we will return our key with a validity of `time.Now() + validity`. The remote server can re-request our key as often as they want but will normally do so when they reach the validity.

serverkeyapi/internal/api.go
neilalexander: It's broken up.

serverkeyapi/internal/api.go
neilalexander: It seemed like the sensible thing to do, given that I'm not really sure why the federation API would own control of which keys we hand out to other servers otherwise.

It also means that the server key API tests are testing the *same* key retrieval behaviour that `/matrix/key/v2/server` does in the real world, which means that we haven't duplicated code and some off-hand change in either the server key API or the federation API won't render the test useless and/or break federation without us realising it.

Kegsay: We rely on sytests to test endpoints to ensure we don't break federation without us realising it. There's currently a fair amount of different code paths for local vs non-local keys so I can't really see how it saves much on code duplication either. For example, `handleLocalKeys` vs `handleDatabaseKeys`, and several places in `handleFetcherKeys` where we conditionally check if it's our keys to make sure we don't `storeResults`.

neilalexander: I'll put it back, but you might understand my hesitation to rely much on Sytest here given that it hasn't picked up on a single issue with any of the key behaviours and federation has still died nightly in many cases as a result.

internal/caching/cache_serverkeys.go
neilalexander: Updated the comments here.

serverkeyapi/serverkeyapi_test.go
Kegsay: So this is the "valid for duration from now"? Please comment to that effect, thanks!

serverkeyapi/internal/api.go
```diff
@@ -44,56 +52,197 @@ func (s *ServerKeyAPI) FetchKeys(
...
+		if res.WasValidAt(now, true) {
+			delete(requests, req)
+		}
+	}
+	return nil
+}
+
+// handleDatabaseKeys handles cases where a fetcher can satisfy
+// the remaining requests.
+func (s *ServerKeyAPI) handleFetcherKeys(
```
Kegsay, line 174: Typo in comments

serverkeyapi/internal/api.go
Kegsay: Please add a comment to this effect somewhere, I can't see it anywhere. It's mainly relevant for `handleDatabaseKeys`

PR #1133, opened 2020-06-16
Make userapi responsible for checking access tokens

Kegsay:
There's still plenty of dependencies on account/device DBs, but this
is a start. This is a breaking change as it adds a required config
value `listen.user_api`.

Part of #1124

Specifically, this PR:
 - Renames `authtypes.Device` to `api.Device` (in `userapi/api`)
 - Adds `QueryAccessToken` and uses it in `VerifyUserFromRequest` which gets called in `MakeAuthAPI`
 - Remove `auth.Data` and replace with `api.UserInternalAPI`


userapi/internal/api.go
```diff
@@ -51,3 +56,68 @@ func (a *UserInternalAPI) QueryProfile(ctx context.Context, req *api.QueryProfil
 	res.DisplayName = prof.DisplayName
 	return nil
 }
+
+func (a *UserInternalAPI) QueryAccessToken(ctx context.Context, req *api.QueryAccessTokenRequest, res *api.QueryAccessTokenResponse) error {
+	appServiceDevice, err := a.queryAppServiceToken(ctx, req.AccessToken, req.AppServiceUserID)
```
neilalexander, line 61: If `req.AppServiceUserID` is optional then please add a `req.AppServiceUserID != ""` guard around the relevant bits here.

clientapi/auth/auth.go
```diff
@@ -72,105 +61,23 @@ func VerifyUserFromRequest(
...
-
-// verifyAccessToken verifies that an access token was supplied in the given HTTP request
-// and returns the device it corresponds to. Returns resErr (an error response which can be
-// sent to the client) if the token is invalid or there was a problem querying the database.
-func verifyAccessToken(req *http.Request, deviceDB DeviceDatabase) (device *authtypes.Device, resErr *util.JSONResponse) {
-	token, err := ExtractAccessToken(req)
+	var res api.QueryAccessTokenResponse
+	err = userAPI.QueryAccessToken(req.Context(), &api.QueryAccessTokenRequest{
+		AccessToken:      token,
+		AppServiceUserID: req.URL.Query().Get("user_id"),
```
neilalexander, line 67: It took me a bit of digging through the CS and AS specs to realise what's going on here and stumbled across https://matrix.org/docs/spec/application_service/r0.1.2#using-sync-and-events - might be useful to link that here and explain as it's not obvious that this ends up masquerading the user.

PR #1134, opened 2020-06-16
Fix media API for demos and possibly Synapse

neilalexander:
This should fix the media API for the libp2p and Yggdrasil demos by creating a `gomatrixserverlib.Client` that uses the same round-trippers as the `gomatrixserverlib.FederationClient`s do. In my local testing with `dendrite-demo-libp2p` this seems to work.

This also adds the media `/download` endpoint into `v1` as well as `r0`, because apparently that's what Synapse (incorrectly) tries to hit when downloading media - see matrix-org/synapse#7707.


Kegsay:
Please investigate why these tests currently do not pass:
 - POST /media/v1/upload can create an upload
 - GET /media/v1/download can fetch the value again
 - Can upload with Unicode file name
 - Can download with Unicode file name locally

They should I think?

PR #1136, opened 2020-06-16
Make syncapi use userapi

Kegsay:
As an aside, we probably want to funnel account data through the user API for syncapi updates. Currently the syncapi stores a copy of account data in the table:
```
type AccountData interface {
	InsertAccountData(ctx context.Context, txn *sql.Tx, userID, roomID, dataType string) (pos types.StreamPosition, err error)
	// SelectAccountDataInRange returns a map of room ID to a list of `dataType`.
	SelectAccountDataInRange(ctx context.Context, userID string, r types.Range, accountDataEventFilter *gomatrixserverlib.EventFilter) (data map[string][]string, err error)
	SelectMaxAccountDataID(ctx context.Context, txn *sql.Tx) (id int64, err error)
}
```

and has no good way to correlate initial syncs to a position in this table.

### DIFF
--- a/cmd/dendrite-sync-api-server/main.go
+++ b/cmd/dendrite-sync-api-server/main.go
@@ -25,12 +25,11 @@ func main() {
 	defer base.Close() // nolint: errcheck
 
 	userAPI := base.UserAPIClient()
-	accountDB := base.CreateAccountsDB()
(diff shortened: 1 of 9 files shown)


Kegsay:
Remaining failures are media related.

PR #1138, opened 2020-06-16
BREAKING: Make eduserver/appservice use userapi

Kegsay:
This is a breaking change because this PR restructures how the AS API
tracks its position in Kafka streams. Previously, it used the account DB
to store partition offsets. However, this is also being used by `clientapi`
for the same purpose, which is bad (each component needs to store offsets
independently or else you might lose messages across restarts). This PR
changes this behaviour to now store partition offsets in the `appservice`
database.

This means that:
 - Upon restart, the `appservice` component will attempt to replay all
   room events from the beginning of time.
 - An additional table will be created in the appservice database, which
   in and of itself is backwards compatible.

### DIFF
--- a/appservice/appservice.go
+++ b/appservice/appservice.go
@@ -16,7 +16,6 @@ package appservice
 
 import (
 	"context"
-	"errors"
 	"net/http"
 	"sync"
 	"time"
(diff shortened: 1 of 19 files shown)


Kegsay:
Test failures are related to media not this.

PR #1139, opened 2020-06-17
Make userapi control account creation entirely

Kegsay:
This makes a chokepoint with which we can finally fix
'database is locked' errors on sqlite during account creation


Kegsay:
Failures due to media api.

PR #1146, opened 2020-06-17
Return the correct /joined_members response and allow ?format=event

Kegsay:
Yay quick wins.

### DIFF
--- a/clientapi/routing/memberships.go
+++ b/clientapi/routing/memberships.go
@@ -15,6 +15,7 @@
 package routing
 
 import (
+	"encoding/json"
 	"net/http"
 
 	"github.com/matrix-org/dendrite/userapi/storage/accounts"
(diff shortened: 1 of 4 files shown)


clientapi/routing/memberships.go
```diff
@@ -35,6 +36,16 @@ type getJoinedRoomsResponse struct {
 	JoinedRooms []string `json:"joined_rooms"`
 }
 
+// https://matrix.org/docs/spec/client_server/r0.6.0#get-matrix-client-r0-rooms-roomid-joined-members
```
neilalexander, line 39: gomatrix has this struct already: https://github.com/matrix-org/gomatrix/blob/9e7906b6766dd084ca15530f16ef33812ae30da5/responses.go#L64

Kegsay: It makes it a pain to unmarshal the content into due to that form being inlined. Also, I'd like to slowly remove our dependency on gomatrix as it's unmaintained (or otherwise actually maintain it).

PR #1150, opened 2020-06-18
Refactor account data

neilalexander:
This refactors account data so that it is:

1. Fully stored and handled internally by the User API as `json.RawMessage`
2. Sync API turns account data into `gomatrixserverlib.ClientEvent` only when needed

### DIFF
--- a/userapi/api/api.go
+++ b/userapi/api/api.go
@@ -36,7 +36,7 @@ type UserInternalAPI interface {
 type InputAccountDataRequest struct {
 	UserID      string          // required: the user to set account data for
 	RoomID      string          // optional: the room to associate the account data with
-	DataType    string          // required: the data type of the data
+	DataType    string          // optional: the data type of the data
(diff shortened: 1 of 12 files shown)


userapi/api/api.go
```diff
@@ -30,6 +32,18 @@ type UserInternalAPI interface {
 	QueryAccountData(ctx context.Context, req *QueryAccountDataRequest, res *QueryAccountDataResponse) error
 }
 
+// InputAccountDataRequest is the request for InputAccountData
+type InputAccountDataRequest struct {
+	UserID      string          // required: the user to set account data for
+	RoomID      string          // optional: the room to associate the account data with
+	DataType    string          // optional: the data type of the data
```
Kegsay, line 39: This is not optional according to the code

neilalexander: You're right - that's a mistake in the comment.

PR #1152, opened 2020-06-19
Bind build support, further Yggdrasil demo updates

neilalexander:
This adds support for building Dendrite using gomobile for embedding Dendrite as well as other miscellaneous changes for the Yggdrasil demo.


clientapi/routing/device.go
```diff
@@ -27,8 +27,10 @@ import (
 )
 
 type deviceJSON struct {
-	DeviceID string `json:"device_id"`
-	UserID   string `json:"user_id"`
+	DeviceID    string `json:"device_id"`
+	DisplayName string `json:"display_name"`
+	LastSeenIP  string `json:"last_seen_ip"`
+	LastSeenTS  uint64 `json:"last_seen_ts"`
```
Kegsay, line 33: Please link to the spec for this struct

syncapi/types/types.go
```diff
@@ -220,8 +223,8 @@ func NewTopologyTokenFromString(tok string) (token TopologyToken, err error) {
 		err = fmt.Errorf("token %s is not a topology token", tok)
 		return
 	}
-	if len(t.Positions) != 2 {
-		err = fmt.Errorf("token %s wrong number of values, got %d want 2", tok, len(t.Positions))
+	if len(t.Positions) < 2 {
+		err = fmt.Errorf("token %s wrong number of values, got %d want at least 2", tok, len(t.Positions))
```
Kegsay, line 227: Where is the extra value coming from? That shouldn't really be allowed as Topology tokens only have 2 params.

neilalexander: Don't ask me why but on occasion it seems Riot iOS tries to send us tokens that are longer than that.

PR #1157, opened 2020-06-23
Pass join errors through internal API boundaries

Kegsay:
Required for certain invite sytests. We will need to think of a
better way of handling this going forwards.


Kegsay:
Once https://github.com/matrix-org/sytest/pull/898 lands, the following tests will pass:
```
Can invite users to invite-only rooms
Uninvited users cannot join the room
Invited user can reject invite
Invited user can reject invite for empty room
Invited user can reject local invite after originator leaves
```

PR #1173, opened 2020-06-30
Persistent federation sender queues (PDUs)

neilalexander:
This implements persistent PDU queues in the federation sender. Rather than being held purely in memory, pending PDUs are written to the federation sender database. This way, they will survive process restarts and such.

We now only hold enough events in memory to build the *current* transaction and shouldn't drop events if we exceed that amount.

This PR does nothing yet with invites or EDUs - it's big enough as it is without going overboard 😄


federationsender/storage/postgres/queue_json_table.go
```diff
@@ -0,0 +1,112 @@
...
+	"context"
+	"database/sql"
+
+	"github.com/lib/pq"
+	"github.com/matrix-org/dendrite/internal"
+	"github.com/matrix-org/dendrite/internal/sqlutil"
+)
+
+const queueJSONSchema = `
+-- The queue_retry_json table contains event contents that
```
Kegsay, line 28: `federationsender_queue_json` you mean?

federationsender/storage/postgres/queue_pdus_table.go
```diff
@@ -0,0 +1,167 @@
+// Copyright 2017-2018 New Vector Ltd
```
Kegsay, line 1: 2020 only :)

federationsender/storage/postgres/queue_pdus_table.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/matrix-org/dendrite/internal/sqlutil"
+	"github.com/matrix-org/gomatrixserverlib"
+)
+
+const queuePDUsSchema = `
+CREATE TABLE IF NOT EXISTS federationsender_queue_pdus (
+    -- The transaction ID that was generated before persisting the event.
+	transaction_id TEXT NOT NULL,
+    -- The domain part of the user ID the m.room.member event is for.
+	server_name TEXT NOT NULL,
```
Kegsay, line 32: Why are we talking about `m.room.member` events all of a sudden? What happens if it's not that type of event?

federationsender/storage/postgres/queue_pdus_table.go
```diff
@@ -0,0 +1,167 @@
...
+	-- The JSON NID from the federationsender_queue_pdus_json table.
+	json_nid BIGINT NOT NULL
```
Kegsay, line 34: This implies 1 txn == 1 event? Don't we want to batch?

federationsender/storage/postgres/queue_pdus_table.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func (s *queuePDUsStatements) selectQueueReferenceJSONCount(
+	ctx context.Context, txn *sql.Tx, jsonNID int64,
+) (int64, error) {
+	var count int64
+	stmt := sqlutil.TxStmt(txn, s.selectQueueReferenceJSONCountStmt)
+	err := stmt.QueryRowContext(ctx, jsonNID).Scan(&count)
+	if err == sql.ErrNoRows {
+		return -1, nil
```
Kegsay, line 140: Should doc this.

federationsender/storage/interface.go
```diff
@@ -19,10 +19,15 @@ import (
...
 )
 
 type Database interface {
 	internal.PartitionStorer
 	UpdateRoom(ctx context.Context, roomID, oldEventID, newEventID string, addHosts []types.JoinedHost, removeHosts []string) (joinedHosts []types.JoinedHost, err error)
 	GetJoinedHosts(ctx context.Context, roomID string) ([]types.JoinedHost, error)
+	StoreJSON(ctx context.Context, js string) (int64, error)
+	AssociatePDUWithDestination(ctx context.Context, transactionID gomatrixserverlib.TransactionID, serverName gomatrixserverlib.ServerName, nids []int64) error
+	GetNextTransactionPDUs(ctx context.Context, serverName gomatrixserverlib.ServerName, limit int) (gomatrixserverlib.TransactionID, []*gomatrixserverlib.HeaderedEvent, error)
+	CleanTransactionPDUs(ctx context.Context, serverName gomatrixserverlib.ServerName, transactionID gomatrixserverlib.TransactionID) error
```
Kegsay, line 32: Why "clean" over delete/remove?

federationsender/queue/destinationqueue.go
```diff
@@ -249,38 +313,68 @@ func (oq *destinationQueue) backgroundSend() {
...
+// cleanPendingPDUs cleans out the pending PDU buffer, removing
+// all references so that the underlying objects can be GC'd.
+func (oq *destinationQueue) cleanPendingPDUs() {
+	numPDUs := len(oq.pendingPDUs)
+	for i := 0; i < numPDUs; i++ {
+		oq.pendingPDUs[i] = nil
+	}
+	oq.pendingPDUs = append(
+		[]*gomatrixserverlib.HeaderedEvent{},
+		oq.pendingPDUs[numPDUs:]...,
```
Kegsay, line 331: You need to explain why you're doing this.

federationsender/queue/destinationqueue.go
```diff
@@ -128,57 +160,86 @@ func (oq *destinationQueue) backgroundSend() {
...
+				return
+			}
+			transactionID = txid
+			oq.pendingPDUs = append(oq.pendingPDUs, pdus...)
+		}
+
+		// If we haven't reached the PDU limit yet then rehydrate the
+		// PDU queue from the database.
+		if len(oq.pendingPDUs) < maxPDUsPerTransaction {
+			hydrate()
```
Kegsay, line 197: I'd prefer if we removed `oq.pendingPDUs` completely and just had function-level vars now. It's weird to have all these side-effects.

federationsender/queue/destinationqueue.go
neilalexander: This is a bit of a hangover from the previous sliding window approach, no longer needed so I've changed it.

federationsender/storage/postgres/queue_pdus_table.go
neilalexander: We do batch. More than one PDU entry can refer to a JSON NID.

Kegsay: It's not clear what's contained under a JSON nid then, I thought it was just an event.

PR #1179, opened 2020-07-03
Hydrate destination queues at startup

neilalexander:
This PR asks the database which servers we have pending events for at startup and then starts those queues. This is particularly important for the iOS P2P demo where the app doesn't survive in the background.

### DIFF
--- a/federationsender/queue/destinationqueue.go
+++ b/federationsender/queue/destinationqueue.go
@@ -251,8 +251,10 @@ func (oq *destinationQueue) backgroundSend() {
 					oq.cleanPendingInvites()
 					return
 				} else {
-					// We haven't been told to give up terminally yet
-					// but we still need to wake up the next iteration.
+					// We haven't been told to give up terminally yet but we still have
+					// PDUs waiting to be sent. By sending a message into the wake chan,
+					// the next loop iteration will try processing these PDUs again,
+					// subject to the backoff.
 					oq.wakeServerCh <- true
 				}
 			} else if transaction {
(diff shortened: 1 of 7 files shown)


federationsender/queue/queue.go
```diff
@@ -60,6 +60,13 @@ func NewOutgoingQueues(
 		signing:    signing,
 		queues:     map[gomatrixserverlib.ServerName]*destinationQueue{},
 	}
+	// Look up which servers we have pending items for and then rehydrate those queues.
+	if serverNames, err := db.GetPendingServerNames(context.TODO()); err == nil {
```
Kegsay, line 64: `Background` probably more suitable at startup.

Kegsay: We also drop the err to the floor?

neilalexander: I'll switch to `Background`. Also I can log the error, although it doesn't really matter since another attempt to send a federation event would restart the queue anyway.

federationsender/queue/destinationqueue.go
```diff
@@ -250,6 +250,10 @@ func (oq *destinationQueue) backgroundSend() {
 					oq.cleanPendingEDUs()
 					oq.cleanPendingInvites()
 					return
+				} else {
+					// We haven't been told to give up terminally yet
+					// but we still need to wake up the next iteration.
+					oq.wakeServerCh <- true
```
Kegsay, line 256: I don't understand the implications of doing this, and the comments don't help me.

federationsender/queue/queue.go
Kegsay: Knowing that an error has occurred is useful especially for sqlite where `database is locked` is a constant worry.

neilalexander: It's logging now.

federationsender/queue/destinationqueue.go
neilalexander: I've updated the comments.

PR #768, opened 2019-07-27
Implement event redaction

Cnly:
Fixes #620.

This PR also includes a minor database fix taken from #696.

This PR doesn't provide full check on redactor's power level yet, which should be fixed at some point later, together with Dendrite's other power level TODOs. Another aspect in which it can be improved is that some similar code appearing in both roomserver and syncapi may be able to get factored out somewhere.

Signed-off-by: Alex Chen <minecnly@gmail.com>

### Pull Request Checklist

* [x] I have added any new tests that need to pass to `testfile` as specified in [docs/sytest.md](https://github.com/matrix-org/dendrite/blob/master/docs/sytest.md)
* [x] Pull request includes a [sign off](https://github.com/matrix-org/dendrite/blob/master/CONTRIBUTING.md#sign-off)

PR #1182, opened 2020-07-03
Implement local redaction

Kegsay:
- Add `/redact` endpoint.
- Perform checks according to spec.
- Make redaction events.
- Send them to the roomserver.

Few minor refactors:
 - Add more wrapper functions for interacting with internal APIs (one liners are nice).
 - Return a headered event when building an event as we must know the room version by then.

Critically this does not:
 - Redact events in the roomserver database or mark them as redacted.
 - Redact events in the syncapi database or mark them as redacted.

See work by @Cnly at https://github.com/matrix-org/dendrite/pull/768 for what this might entail - and we have room versions now so we can finish it all up. Unknown if this works currently over federation.

### DIFF
--- a/clientapi/routing/redaction.go
+++ b/clientapi/routing/redaction.go
@@ -74,14 +74,16 @@ func SendRedaction(
 		if plEvent == nil {
 			return util.JSONResponse{
 				Code: 403,
-				JSON: jsonerror.Forbidden("cannot redact event, missing pl event"),
+				JSON: jsonerror.Forbidden("You don't have permission to redact this event, no power_levels event in this room."),
 			}
 		}
 		pl, err := plEvent.PowerLevels()
 		if err != nil {
 			return util.JSONResponse{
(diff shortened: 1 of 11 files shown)


clientapi/routing/redaction.go
```diff
@@ -0,0 +1,134 @@
...
+				Code: 403,
+				JSON: jsonerror.Forbidden("cannot redact event, malformed pl event"),
+			}
+		}
+		allowedToRedact = pl.UserLevel(device.UserID) >= pl.Redact
+	}
+	if !allowedToRedact {
+		return util.JSONResponse{
+			Code: 403,
+			JSON: jsonerror.Forbidden("cannot redact event"),
```
neilalexander, line 92: I suspect this error gets rendered by the client - can we make it a bit more descriptive? e.g. `You don't have permission to redact this event (power level %d is required)`

PR #1189, opened 2020-07-08
Finish implementing redactions

Kegsay:
- Handle `OutputTypeRedactedEvent` events in `syncapi` and modify event JSON accordingly. 
- Only send `m.room.redaction` events to clients when redactions have been validated as per spec:

> servers should not apply or send redactions to clients until both the redaction event and original event have been seen, and are valid. 

https://matrix.org/docs/spec/rooms/v3#authorization-rules-for-events


syncapi/consumers/roomserver.go
```diff
@@ -81,11 +81,18 @@ func (s *OutputRoomEventConsumer) onMessage(msg *sarama.ConsumerMessage) error {
 
 	switch output.Type {
 	case api.OutputTypeNewRoomEvent:
+		// Ignore redaction events. We will add them to the database when they are
+		// validated (when we receive OutputTypeRedactedEvent)
+		if output.NewRoomEvent.Event.Type() == gomatrixserverlib.MRoomRedaction && output.NewRoomEvent.Event.StateKey() == nil {
```
neilalexander, line 86: Is `output.NewRoomEvent.Event.StateKey() == nil` right here or can it also be `""` as well as `nil`?

Kegsay: It _has_ to be nil. Redactions are not state events so if we get one we should not handle it as one.

PR #1191, opened 2020-07-09
Reduce federation sender wedges

neilalexander:
This PR attempts to reduce wedges in the federation sender.

It does so by:
1. Removing the `pendingPDUs` atomic and replacing it with `waitForPDUs`, which is just overall easier to understand and less prone to fall out of sync;
2. Remove a `select` on the main background loop which is no longer necessary.

This *appears* to improve reliability on my local testing with the Yggdrasil demo quite substantially.

### DIFF
--- a/federationsender/queue/destinationqueue.go
+++ b/federationsender/queue/destinationqueue.go
@@ -32,6 +32,7 @@ import (
 )
 
 const maxPDUsPerTransaction = 50
+const queueIdleTimeout = time.Second * 30
 
 // destinationQueue is a queue of events for a single destination.
 // It is responsible for sending the events to the destination and
@@ -52,7 +53,6 @@ type destinationQueue struct {
 	transactionIDMutex sync.Mutex                              // protects transactionID
 	transactionID      gomatrixserverlib.TransactionID         // last transaction ID
 	transactionCount   atomic.Int32                            // how many events in this transaction so far
-	pendingPDUs        atomic.Int64                            // how many PDUs are waiting to be sent
 	pendingEDUs        []*gomatrixserverlib.EDU                // owned by backgroundSend
 	pendingInvites     []*gomatrixserverlib.InviteV2Request    // owned by backgroundSend
 	notifyPDUs         chan bool                               // interrupts idle wait for PDUs
@@ -68,7 +68,6 @@ func (oq *destinationQueue) sendEvent(nid int64) {
 		log.Infof("%s is blacklisted; dropping event", oq.destination)
 		return
 	}
-	oq.wakeQueueIfNeeded()
 	// Create a transaction ID. We'll either do this if we don't have
 	// one made up yet, or if we've exceeded the number of maximum
 	// events allowed in a single tranaction. We'll reset the counter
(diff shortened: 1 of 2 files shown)


federationsender/queue/destinationqueue.go
```diff
@@ -138,26 +139,33 @@ func (oq *destinationQueue) wakeQueueIfNeeded() {
...
+	// channel. This will mean that the backgroundSend will not block.
+	if pendingPDUs > 0 {
+		ch := make(chan bool, 1)
+		close(ch)
+		return ch
+	}
+	// If there are no PDUs pending right now then instead we'll return
+	// the notify channel, so that backgroundSend can pick up normal
+	// notifications from sendEvent.
+	return oq.notifyPDUs
```
Kegsay, line 166: I don't understand why we want to do this.

federationsender/queue/destinationqueue.go
```diff
@@ -95,11 +94,13 @@ func (oq *destinationQueue) sendEvent(nid int64) {
...
-	// wake up if it's asleep. The return to the Add function will only
-	// be 1 if the previous value was 0, e.g. nothing was waiting before.
-	if oq.pendingPDUs.Add(1) == 1 {
-		oq.notifyPDUs <- true
+	// Wake up the queue if it's asleep.
+	oq.wakeQueueIfNeeded()
+	// If we're blocking on waiting PDUs then tell the queue that we
+	// have work to do.
+	select {
+	case oq.notifyPDUs <- true:
```
Kegsay, line 102: We always want to do `wakeQueueIfNeeded` and doing a soft-push onto the notify channel so I would fold this into a single function (and have a function args for whether this is a PDU/EDU/invite)

federationsender/queue/queue.go
```diff
@@ -136,7 +136,7 @@ func (oqs *OutgoingQueues) SendEvent(
 	}
 
 	for _, destination := range destinations {
-		oqs.getQueue(destination).sendEvent(nid)
+		go oqs.getQueue(destination).sendEvent(nid)
```
Kegsay: Why would `sendEvent` block in the new implementation? This could spawn a lot of goroutines, and even though Go supports millions of them this is not a good way to go about it I think.

federationsender/queue/destinationqueue.go
neilalexander: If there are no PDUs waiting in the database then returning `oq.notifyPDUs` allows `backgroundSend` to block for 30 seconds waiting for incoming PDUs. If `sendEvent` fires in the interim, then `oq.notifyPDUs` wakes up the `select`.

federationsender/queue/destinationqueue.go
neilalexander: When EDUs and invites are also following the same mechanism then that would probably make sense. For now it doesn't as `incomingEDUs`/`incomingInvites` are used instead.

federationsender/queue/queue.go
neilalexander: You're right - with the `select` it shouldn't block so I've taken the `go` out.

neilalexander:
The reason that I moved away from the `pendingPDUs` channel in the first place is precisely so that we don't have to store all of these things in memory, especially for queues that might never come back to life.

Kegsay:
> The reason that I moved away from the pendingPDUs channel in the first place is precisely so that we don't have to store all of these things in memory, especially for queues that might never come back to life.

When queues die they can close the channel to reclaim memory, and rehydrate from the database when they spin up again. I'd rather consume a bit more memory and grok this code better to be honest.

neilalexander:
> > The reason that I moved away from the pendingPDUs channel in the first place is precisely so that we don't have to store all of these things in memory, especially for queues that might never come back to life.
> 
> When queues die they can close the channel to reclaim memory, and rehydrate from the database when they spin up again. I'd rather consume a bit more memory and grok this code better to be honest.

It can take a very long time for that to happen though - currently the backoff will let us reach 18 hours before we get as far as nuking a dead queue. Plus I am already worried about memory usage on mobile as it is, especially given that nodes are much more likely to come and go in P2P settings.

I get that the shape is a little strange but I think I have reasoned fairly well about it in my head at the moment and once invites/EDUs are moved over too then it should make more sense. It's complicated right now because it is a hybrid either way.

PR #1193, opened 2020-07-09
Add User-Interactive Authentication 

Kegsay:
And use it when deleting a device. With tests.

Make all the device management APIs pass.


Kegsay:
Tests fail because of https://github.com/matrix-org/sytest/pull/910

PR #1200, opened 2020-07-14
Fix database lock errors in User API in SQLite mode

neilalexander:
This wraps the accounts, profiles and three PID tables in `TransactionWriter`s so that we don't hit locking issues in SQLite when registering and modifying accounts.


neilalexander:
Closing this as I don't like the direction of it.

PR #1201, opened 2020-07-14
Protect User API SQLite calls with mutexes

neilalexander:
This PR replaces #1200, which probably would have been more correct ordinarily, but it meant that we ended up taking three separate database transactions to create a user account and pretty much broke any possibility of being able to roll back some or all of the changes if something went wrong.

### DIFF
--- a/userapi/storage/accounts/sqlite3/storage.go
+++ b/userapi/storage/accounts/sqlite3/storage.go
@@ -40,7 +40,10 @@ type Database struct {
 	threepids    threepidStatements
 	serverName   gomatrixserverlib.ServerName
 
-	createAccountMu sync.Mutex
+	accountsMu     sync.Mutex
+	profilesMu     sync.Mutex
+	accountDatasMu sync.Mutex
+	threepidsMu    sync.Mutex
 }
 
 // NewDatabase creates a new accounts and profiles database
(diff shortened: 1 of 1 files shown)


userapi/storage/accounts/sqlite3/storage.go
```diff
@@ -281,6 +322,8 @@ func (d *Database) RemoveThreePIDAssociation(
 func (d *Database) GetLocalpartForThreePID(
 	ctx context.Context, threepid string, medium string,
 ) (localpart string, err error) {
+	d.threepidsMu.Lock()
+	defer d.threepidsMu.Unlock()
 	return d.threepids.selectLocalpartForThreePID(ctx, nil, threepid, medium)
```
Kegsay, line 327: Read transactions do not need protection. They will never encounter or cause 'database is locked' errors. Only write transactions can. We don't need to mutex here and in other SELECT places.

userapi/storage/accounts/sqlite3/storage.go
```diff
@@ -124,8 +143,12 @@ func (d *Database) CreateGuestAccount(ctx context.Context) (acc *api.Account, er
...
 	// race with CreateAccount
 	// We know we'll be the only process since this is sqlite ;) so a lock here will be all that is needed.
-	d.createAccountMu.Lock()
-	defer d.createAccountMu.Unlock()
+	d.profilesMu.Lock()
+	d.accountDatasMu.Lock()
+	d.accountsMu.Lock()
+	defer d.profilesMu.Unlock()
+	defer d.accountDatasMu.Unlock()
+	defer d.accountsMu.Unlock()
```
Kegsay, line 151: I don't like how we have so many mutexes. We run the risk of deadlocking if we lock in the wrong order for `CreateGuestAccount` and  `CreateAccount`, we should at least comment that here.

neilalexander: The alternative is that we just have a single lock for all write operations?

userapi/storage/accounts/sqlite3/storage.go
neilalexander: fixed

userapi/storage/accounts/sqlite3/storage.go
Kegsay: That would work, but this works too.

PR #1202, opened 2020-07-15
Implement key uploads

Kegsay:
Which makes another sytest pass.

Adds a new database and asserts it in the config - i.e it's a breaking change. Backwards compatible though as we don't modify tables.

### DIFF
--- a/clientapi/routing/keys.go
+++ b/clientapi/routing/keys.go
@@ -15,8 +15,13 @@
 package routing
 
 import (
+	"encoding/json"
 	"net/http"
 
+	"github.com/matrix-org/dendrite/clientapi/httputil"
+	"github.com/matrix-org/dendrite/clientapi/jsonerror"
+	"github.com/matrix-org/dendrite/keyserver/api"
+	userapi "github.com/matrix-org/dendrite/userapi/api"
 	"github.com/matrix-org/util"
 )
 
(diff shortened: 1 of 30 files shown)


clientapi/routing/keys.go
```diff
@@ -32,9 +37,55 @@ func QueryKeys(
 	}
 }
 
-func UploadKeys(req *http.Request) util.JSONResponse {
+type uploadKeysRequest struct {
```
neilalexander, line 40: Ultimately this is fine for now but I'd really love to see more of these types going into matrix-org/gomatrix eventually.

keyserver/internal/internal.go
```diff
@@ -25,7 +25,7 @@ import (
 )
 
 type KeyInternalAPI struct {
-	db storage.Database
+	DB storage.Database
```
neilalexander, line 28: It's not immediately obvious to me why this is exported?

Kegsay: Because a different package creates the struct (`keyserver/keyserver.go`)

clientapi/routing/keys.go
Kegsay: Ideally we'd automate mapping swagger -> Go structs, but we're nowhere near that yet.

keyserver/internal/internal.go
neilalexander: Ah, yes, okay!

Kegsay:
Failure is because of:
```
ERROR: The following tests passed but are not present in /src/sytest-whitelist. Please append them to the file:

Device list doesn't change if remote server is down
```

Blacklisted.

PR #1214, opened 2020-07-22
Persistent federation sender blacklist

neilalexander:
This PR persists the blacklist into the database, so that after a server restart, we don't try to flood nodes that are assumed dead with federation traffic. This is particularly important for the P2P demos where nodes spend more time offline than they do online.


federationsender/statistics/statistics.go
```diff
@@ -1,27 +1,28 @@
...
 // wrapper.
 type Statistics struct {
+	DB      storage.Database
 	servers map[gomatrixserverlib.ServerName]*ServerStatistics
 	mutex   sync.RWMutex
+
+	// How many times should we tolerate consecutive failures before we
+	// just blacklist the host altogether? Bear in mind that the backoff
+	// is exponential, so the max time here to attempt is 2**failures.
+	FailuresUntilBlacklist uint32
```
Kegsay, line 25: 2**failures seconds or minutes or what? Units please.

federationsender/statistics/statistics.go
```diff
@@ -65,7 +77,11 @@ type ServerStatistics struct {
 func (s *ServerStatistics) Success() {
 	s.successCounter.Add(1)
 	s.failCounter.Store(0)
-	s.blacklisted.Store(false)
+	if err := s.statistics.DB.RemoveServerFromBlacklist(s.serverName); err != nil {
+		logrus.WithError(err).Errorf("Failed to remove %q from blacklist", s.serverName)
```
Kegsay, line 81: We should still do `s.blacklisted.Store(false)` in this case right?

federationsender/statistics/statistics.go
```diff
@@ -77,12 +93,16 @@ func (s *ServerStatistics) Failure() bool {
...
 	// Check that we haven't failed more times than is acceptable.
-	if failCounter >= FailuresUntilBlacklist {
+	if failCounter >= s.statistics.FailuresUntilBlacklist {
 		// We've exceeded the maximum amount of times we're willing
 		// to back off, which is probably in the region of hours by
 		// now. Mark the host as blacklisted and tell the caller to
 		// give up.
-		s.blacklisted.Store(true)
+		if err := s.statistics.DB.AddServerToBlacklist(s.serverName); err != nil {
+			logrus.WithError(err).Errorf("Failed to add %q to blacklist", s.serverName)
```
Kegsay, line 102: Likewise

federationsender/storage/sqlite3/blacklist_table.go
```diff
@@ -0,0 +1,110 @@
+// Copyright 2017-2018 New Vector Ltd
+// Copyright 2019-2020 The Matrix.org Foundation C.I.C.
```
Kegsay, line 2: Not for a brand new file.

federationsender/storage/sqlite3/blacklist_table.go
```diff
@@ -0,0 +1,110 @@
...
+func (s *blacklistStatements) SelectBlacklist(
+	ctx context.Context, txn *sql.Tx, serverName gomatrixserverlib.ServerName,
+) (bool, error) {
+	stmt := sqlutil.TxStmt(txn, s.selectBlacklistStmt)
+	res, err := stmt.QueryContext(ctx, serverName)
+	if err != nil {
+		return false, err
+	}
+	defer res.Close() // nolint:errcheck
+	return res.Next(), nil
```
Kegsay, line 97: Took me a little while to work out why this would work. Probably worth a commen that we don't care about the data returned. and that this will return true if the server is blacklisted.

PR #1215, opened 2020-07-22
Handle inbound federation E2E key queries/claims

Kegsay:
This should in theory make E2E work now modulo device list updates.

### DIFF
--- a/cmd/dendrite-demo-yggdrasil/main.go
+++ b/cmd/dendrite-demo-yggdrasil/main.go
@@ -140,7 +140,7 @@ func main() {
 		RoomserverAPI:       rsAPI,
 		UserAPI:             userAPI,
 		StateAPI:            stateAPI,
-		KeyAPI:              keyserver.NewInternalAPI(base.Cfg, federation),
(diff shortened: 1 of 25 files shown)


cmd/dendrite-demo-yggdrasil/main.go
```diff
@@ -140,7 +140,7 @@ func main() {
 		RoomserverAPI:       rsAPI,
 		UserAPI:             userAPI,
 		StateAPI:            stateAPI,
-		KeyAPI:              keyserver.NewInternalAPI(base.Cfg, federation),
+		KeyAPI:              keyserver.NewInternalAPI(base.Cfg, federation, userAPI),
```
neilalexander, line 143: When you make changes to `cmd/dendrite-demo-yggdrasil`, please also make the same changes to `build/gobind`, as that's what the iOS demo is built from.

Kegsay: Can we build it as part of `build.sh`? Else I'll never remember.

neilalexander: Unfortunately it requires Xcode and gomobile, and takes ages to build. :-(

PR #1222, opened 2020-07-24
Modify /state/{eventType}/{stateKey} to return the event at the time the user left

Kegsay:
Or live, depending on their current state. Hopefully fixes some sytests!

### DIFF
--- a/clientapi/routing/state.go
+++ b/clientapi/routing/state.go
@@ -22,6 +22,7 @@ import (
 	"github.com/matrix-org/dendrite/clientapi/jsonerror"
 	"github.com/matrix-org/dendrite/roomserver/api"
 	"github.com/matrix-org/dendrite/syncapi/types"
+	userapi "github.com/matrix-org/dendrite/userapi/api"
 	"github.com/matrix-org/gomatrixserverlib"
 	"github.com/matrix-org/util"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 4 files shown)


clientapi/routing/state.go
```diff
@@ -99,40 +101,150 @@ func OnIncomingStateRequest(ctx context.Context, rsAPI api.RoomserverInternalAPI
...
+	// Then include the requested event type and state key, assuming it
+	// isn't for the same.
+	stateToFetch := []gomatrixserverlib.StateKeyTuple{
+		{
+			EventType: evType,
+			StateKey:  stateKey,
 		},
 	}
-	stateRes := api.QueryLatestEventsAndStateResponse{}
+	if evType != gomatrixserverlib.MRoomHistoryVisibility && stateKey != "" {
```
Kegsay, line 122: Why `&& stateKey != ""`?

clientapi/routing/state.go
```diff
@@ -99,40 +101,150 @@ func OnIncomingStateRequest(ctx context.Context, rsAPI api.RoomserverInternalAPI
...
+		stateToFetch = append(stateToFetch, gomatrixserverlib.StateKeyTuple{
+			EventType: gomatrixserverlib.MRoomHistoryVisibility,
+			StateKey:  "",
+		})
+	}
 
-	if err := rsAPI.QueryLatestEventsAndState(ctx, &stateReq, &stateRes); err != nil {
+	// First of all, get the latest state of the room. We need to do this
+	// so that we can look at the history visibility of the room. If the
+	// room is world-readable then we will always return the latest state.
```
Kegsay, line 131: History visibility is only for events at that point. A private visibility turned world-readable should not result in those private visibility messages being readable by anyone.

Kegsay:
Otherwise LGTM, thanks!

clientapi/routing/state.go
neilalexander: I should have commented this - it's basically just so that we don't add `MRoomHistoryVisibility` and `""` into the list of events to request twice, since we request it anyway, in the event that someone tries to use `/state` to find that.

clientapi/routing/state.go
neilalexander: In this case it doesn't, as it's only state events. It's basically to enforce the logic that if the room is world-readable right now, that you should be able to see the state right now, even after you leave.

Kegsay: Oh right, I see then!

PR #1224, opened 2020-07-26
Don't use more than 999 variables in SQLite querys.

### LINKED ISSUE #1223: Initial sync fails with SQLite as DB
With monolith server built from commit 7b862384a779f067f07ffeb2151856f89d372732. Running on Ubuntu server 18.04.
 
After creating account and joining a couple of rooms (#dendrite:matrix.org, #dendrite-dev:matrix.org and #go-lang:matrix.org), clear the client's cache or log out from the home server. When logging in again the initial sync fails. 

The log says:
`ERRO[2020-07-24T20:33:21.201032227Z] [github.com/matrix-org/dendrite@/syncapi/sync/requestpool.go:74] OnIncomingSyncRequest
         rp.currentSyncForUser failed                  device_id=XmAikzqI error="too many SQL variables" limit=20 req.id=GyfDiWaa0tAw req.method=GET req.path=/_matrix/client/r0/sync since=s0_0 timeout=0s user_id="<MATRIXID>"`

The source for this error is the function 
```Go
func (s *currentRoomStateStatements) SelectEventsWithEventIDs(
	ctx context.Context, txn *sql.Tx, eventIDs []string,
) ([]types.StreamEvent, error) {}
```
in syncapi/storage/sqlite3/current_room_state_table.go

The reason is that the query is expanded to something like.
`SELECT added_at, headered_event_json, 0 AS session_id, false AS exclude_from_sync, '' AS transaction_id FROM syncapi_current_room_state WHERE event_id IN ($1, $2, $3, $4, $5, $6, $7, ........ )`

With a large number of variables (in my case 1022) and the default limit of number of variables in SQLite is 999, (bullet number 9 in  [SQLite limits](https://www.sqlite.org/limits.html) ).

HenrikSolver:
Closes #1223

Signed-off-by: Henrik Sölver <henrik.solver@gmail.com>

### Pull Request Checklist

### DIFF
--- a/syncapi/storage/sqlite3/current_room_state_table.go
+++ b/syncapi/storage/sqlite3/current_room_state_table.go
@@ -243,6 +243,13 @@ func (s *currentRoomStateStatements) UpsertRoomState(
 	})
 }
 
+func minOfInts(a, b int) int {
+	if a <= b {
+		return a
+	}
+	return b
+}
+
 func (s *currentRoomStateStatements) SelectEventsWithEventIDs(
 	ctx context.Context, txn *sql.Tx, eventIDs []string,
 ) ([]types.StreamEvent, error) {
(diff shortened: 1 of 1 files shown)


neilalexander:
Thanks for looking into this! This is probably a bit of a band-aid fix and we might need to look at the effects of this elsewhere, so I'm going to keep the issue #1223 open for now and we can revisit it.

PR #1225, opened 2020-07-27
User directory

neilalexander:
This adds the ability to search local and known users in the room directory.

It adds API endpoints to both the user API and the current state API so that we can search local and known users. Currently full profile details are *not* retrieved for non-local users - this is still to be done.

### DIFF
--- a/currentstateserver/storage/postgres/current_room_state_table.go
+++ b/currentstateserver/storage/postgres/current_room_state_table.go
@@ -83,7 +83,9 @@ const selectJoinedUsersSetForRoomsSQL = "" +
 	" type = 'm.room.member' and content_value = 'join' GROUP BY state_key"
 
 const selectKnownUsersSQL = "" +
-	"SELECT DISTINCT state_key FROM currentstate_current_room_state WHERE type = 'm.room.member' AND state_key LIKE $1 LIMIT $2"
+	"SELECT DISTINCT state_key FROM currentstate_current_room_state WHERE room_id = ANY(" +
+	"  SELECT DISTINCT room_id FROM currentstate_current_room_state WHERE state_key=$1 AND TYPE='m.room.member' AND content_value='join'" +
+	") AND TYPE='m.room.member' AND content_value='join' AND state_key LIKE $2 LIMIT $3"
 
 type currentRoomStateStatements struct {
 	upsertRoomStateStmt              *sql.Stmt
@@ -304,8 +306,8 @@ func (s *currentRoomStateStatements) SelectBulkStateContent(
 	return strippedEvents, rows.Err()
 }
 
-func (s *currentRoomStateStatements) SelectKnownUsers(ctx context.Context, searchString string, limit int) ([]string, error) {
-	rows, err := s.selectKnownUsersStmt.QueryContext(ctx, fmt.Sprintf("%%%s%%", searchString), limit)
+func (s *currentRoomStateStatements) SelectKnownUsers(ctx context.Context, userID, searchString string, limit int) ([]string, error) {
+	rows, err := s.selectKnownUsersStmt.QueryContext(ctx, userID, fmt.Sprintf("%%%s%%", searchString), limit)
 	if err == sql.ErrNoRows {
 		return nil, nil
 	}
--- a/currentstateserver/storage/postgres/current_room_state_table.go
+++ b/currentstateserver/storage/postgres/current_room_state_table.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"database/sql"
 	"encoding/json"
+	"fmt"
 
 	"github.com/lib/pq"
 	"github.com/matrix-org/dendrite/currentstateserver/storage/tables"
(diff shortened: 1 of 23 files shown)


currentstateserver/storage/postgres/current_room_state_table.go
```diff
@@ -81,6 +82,11 @@ const selectJoinedUsersSetForRoomsSQL = "" +
 	"SELECT state_key, COUNT(room_id) FROM currentstate_current_room_state WHERE room_id = ANY($1) AND" +
 	" type = 'm.room.member' and content_value = 'join' GROUP BY state_key"
 
+const selectKnownUsersSQL = "" +
+	"SELECT DISTINCT state_key FROM currentstate_current_room_state WHERE room_id = ANY(" +
+	"  SELECT DISTINCT room_id FROM currentstate_current_room_state WHERE state_key=$1 AND TYPE='m.room.member' AND content_value='join'" +
+	") AND TYPE='m.room.member' AND content_value='join' AND state_key LIKE $2 LIMIT $3"
```
Kegsay, line 88: Comment explaining the subselect would be good. I _think_ it's so we only query joined users for rooms that the user is in?

currentstateserver/storage/postgres/current_room_state_table.go
```diff
@@ -295,3 +305,22 @@ func (s *currentRoomStateStatements) SelectBulkStateContent(
 	}
 	return strippedEvents, rows.Err()
 }
+
+func (s *currentRoomStateStatements) SelectKnownUsers(ctx context.Context, userID, searchString string, limit int) ([]string, error) {
+	rows, err := s.selectKnownUsersStmt.QueryContext(ctx, userID, fmt.Sprintf("%%%s%%", searchString), limit)
+	if err == sql.ErrNoRows {
```
Kegsay, line 311: `QueryContext` cannot return `sql.ErrNoRows` - only `QueryRows`.

currentstateserver/storage/postgres/current_room_state_table.go
```diff
@@ -295,3 +305,22 @@ func (s *currentRoomStateStatements) SelectBulkStateContent(
...
+		return nil, nil
+	}
+	if err != nil {
+		return nil, err
+	}
+	result := []string{}
+	for rows.Next() {
```
Kegsay, line 318: We never close `rows`

userapi/storage/accounts/postgres/profile_table.go
```diff
@@ -105,3 +113,21 @@ func (s *profilesStatements) setDisplayName(
...
+
+func (s *profilesStatements) selectProfilesBySearch(
+	ctx context.Context, searchString string, limit int,
+) ([]authtypes.Profile, error) {
+	var profiles []authtypes.Profile
+	rows, err := s.selectProfilesBySearchStmt.QueryContext(ctx, fmt.Sprintf("%%%s%%", searchString), limit)
+	if err != nil {
+		return nil, err
+	}
+	for rows.Next() {
```
Kegsay, line 125: Missing close

userapi/storage/accounts/postgres/profile_table.go
```diff
@@ -105,3 +113,21 @@ func (s *profilesStatements) setDisplayName(
 	_, err = s.setDisplayNameStmt.ExecContext(ctx, displayName, localpart)
 	return
 }
+
+func (s *profilesStatements) selectProfilesBySearch(
+	ctx context.Context, searchString string, limit int,
+) ([]authtypes.Profile, error) {
+	var profiles []authtypes.Profile
+	rows, err := s.selectProfilesBySearchStmt.QueryContext(ctx, fmt.Sprintf("%%%s%%", searchString), limit)
```
Kegsay, line 121: I don't understand how `%%%s%%` works. It's `%% %s %%` right? Why double `%`?

currentstateserver/storage/postgres/current_room_state_table.go
neilalexander: Good catch - thanks! I've added closes.

currentstateserver/storage/postgres/current_room_state_table.go
neilalexander: Copy-pasta mistake, thanks!

currentstateserver/storage/postgres/current_room_state_table.go
neilalexander: I've added comments. Yep, the goal is exactly that - that we return users based on the rooms that the searching user is joined to.

userapi/storage/accounts/postgres/profile_table.go
neilalexander: I've added comments to this, but it's pretty much because we're giving this parameter to `LIKE $1` in the SQL statement and we want to wildcard-search. `%%` escapes the `%` character, so it's basically turning into `LIKE %searchString%`.

PR #1241, opened 2020-08-06
Yggdrasil demo updates

neilalexander:
This makes some updates to the Yggdrasil demo:

- HTTP pooling is disabled (stops some blocking)
- QUIC sessions are now nuked when coordinates change (stops some blocking)
- Incoming QUIC sessions cause federation sender queue wake-up
- Timeouts adjusted
- Dead QUIC sessions can now be replaced with new ones
- Other tweaks


neilalexander:
This PR is equivalent to build 31 on Testflight.

Issue #1244, opened 2020-08-07
Sanity check what we log and what level we log at


Kegsay:
eg OPTIONS requests shouldn't log

PR #1248, opened 2020-08-07
Tweak log levels of some federation logging

neilalexander:
This reduces the logging level of some of the federation logging from `Info` to `Debug`. 

Associated with #1244.

### DIFF
--- a/federationapi/routing/send.go
+++ b/federationapi/routing/send.go
@@ -218,7 +218,9 @@ func (t *txnReq) processTransaction() (*gomatrixserverlib.RespSend, *util.JSONRe
 	}
 
 	t.processEDUs(t.EDUs)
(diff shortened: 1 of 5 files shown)


federationapi/routing/send.go
```diff
@@ -218,7 +218,9 @@ func (t *txnReq) processTransaction() (*gomatrixserverlib.RespSend, *util.JSONRe
 	}
 
 	t.processEDUs(t.EDUs)
-	util.GetLogger(t.context).Infof("Processed %d PDUs from transaction %q", len(results), t.TransactionID)
+	if c := len(results); c > 0 {
+		util.GetLogger(t.context).Infof("Processed %d PDUs from transaction %q", c, t.TransactionID)
```
Kegsay, line 222: Is this not the number of errors?

neilalexander: It's counting both errored and valid PDUs.

PR #1250, opened 2020-08-07
Backoff fixes

neilalexander:
This should fix some problems where the backoff could grow disproportionately if repeated failures happened quickly.


neilalexander:
I think there's possibly a bug in how this handles blacklisting - by the number of calls to `Failure` instead of expiring the backoff, so I will need to address that next week.

PR #1258, opened 2020-08-11
New sample config

neilalexander:
This tweaks various config things (ordering and such) and adds a new `dendrite-config.yaml` sample config file.

### DIFF
--- a/dendrite-config.yaml
+++ b/dendrite-config.yaml
@@ -1,0 +1,299 @@
+# This is the Dendrite configuration file.
+#
+# The configuration is split up into sections - each Dendrite component has a
+# configuration section, in addition to the "global" section which applies to
+# all components.
+#
+# At a minimum, to get started, you will need to update the settings in the
+# "global" section for your deployment, and you will need to check that the
+# database "connection_string" line in each component section is correct. 
+#
+# Each component with a "database" section can accept the following formats
+# for "connection_string":
+#   SQLite:     file:filename.db
+#               file:///path/to/filename.db
+#   PostgreSQL: postgresql://user:pass@hostname/database?params=...
+#
+# SQLite is embedded into Dendrite and therefore no further prerequisites are
+# needed for the database when using SQLite mode. However, performance with
+# PostgreSQL is significantly better and recommended for multi-user deployments.
+#
+# The "max_open_conns" and "max_idle_conns" settings configure the maximum 
+# number of open/idle database connections. The value 0 will use the database
+# engine default, and a negative value will use unlimited connections. The
+# "conn_max_lifetime" option controls the maximum length of time a database
+# connection can be idle in seconds - a negative value is unlimited.
+
+# The version of the configuration file. 
+version: 1
+
+# Global Matrix configuration. This configuration applies to all components.
+global:
+  # The domain name of this homeserver.
+  server_name: localhost
+
+  # The path to the signing private key file, used to sign requests and events.
+  private_key: matrix_key.pem
+
+  # A unique identifier for this private key. Must start with the prefix "ed25519:".
+  key_id: ed25519:auto
+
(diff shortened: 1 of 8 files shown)


dendrite-config.yaml
```diff
@@ -0,0 +1,299 @@
...
+#
+# Each component with a "database" section can accept the following formats
+# for "connection_string":
+#   SQLite:     file:filename.db
+#               file:///path/to/filename.db
+#   PostgreSQL: postgresql://user:pass@hostname/database?params=...
+#
+# SQLite is embedded into Dendrite and therefore no further prerequisites are
+# needed for the database when using SQLite mode. However, performance with
+# PostgreSQL is significantly better and recommended for multi-user deployments.
```
Kegsay, line 19: We should attempt to broadly quantify this so people can make an informed decision. We happen to know pq=16min and sqlite=20min on sytests for example, meaning sqlite is around 30% slower on low-user HSes, which will become exacerbated when the number of users increases.

dendrite-config.yaml
```diff
@@ -0,0 +1,299 @@
...
+  private_key: matrix_key.pem
+
+  # A unique identifier for this private key. Must start with the prefix "ed25519:".
+  key_id: ed25519:auto
+
+  # How long a remote server can cache our server signing key before requesting it
+  # again. Increasing this number will reduce the number of requests made by other
+  # servers for our key but increases the period that a compromised key will be
+  # considered valid by other homeservers.
+  key_validity_period: 168h0m0s
```
Kegsay, line 45: ❤️

dendrite-config.yaml
```diff
@@ -0,0 +1,299 @@
...
+  # Lists of domains that the server will trust as identity servers to verify third
+  # party identifiers such as phone numbers and email addresses.
+  trusted_third_party_id_servers:
+  - matrix.org
+  - vector.im
+
+  # Configuration for Kafka/Naffka.
+  kafka:
+    # List of Kafka addresses to connect to. This is not needed if using Naffka.
+    addresses: []
```
Kegsay, line 56: We should mention the word 'broker' somewhere, and maybe give an example using the default broker port.

internal/config/config_clientapi.go
```diff
@@ -12,24 +12,25 @@ type ClientAPI struct {
...
+	// If set disables new users from registering (except via shared
+	// secrets)
+	RegistrationDisabled bool `yaml:"registration_disabled"`
 	// If set, allows registration by anyone who also has the shared
 	// secret, even if registration is otherwise disabled.
 	RegistrationSharedSecret string `yaml:"registration_shared_secret"`
+
+	// Boolean stating whether catpcha registration is enabled
+	// and required
+	RecaptchaEnabled bool `yaml:"enable_registration_captcha"`
```
Kegsay, line 24: So this is exactly what I was trying to get at before when I mentioned about autogenning a config file from go source code. At the moment we have two descriptions of fields, this one and the one in `dendrite-config.yaml`. I would prefer it if:
 - Source of truth is the code comment above.
 - `generate-config` parses this file and writes out a YAML file with these comments.
 - We can have `Default: foo` and parse that to set the default when generating a config.
 - We check in the generated config.

WDYT?

neilalexander: Broadly I'm not against this at all, as it means that our code annotates the config itself, but that's probably a longer-term activity than something for this PR.

dendrite-config.yaml
neilalexander: We can probably safely set a default value here to `localhost:2181` - it should get safely ignored since naffka is enabled by default anyway.

internal/config/config_clientapi.go
Kegsay: Yep just running the idea past you. I've done this thing before so I can tackle it.

PR #1266, opened 2020-08-13
API setup refactoring

neilalexander:
This makes a number of changes:

- Removes `BaseMux` in favour of flexible HTTP listeners
- Breaks apart the public and internal muxes into CS API, SS API, media API, key API and internal API muxes
- Allows internal and externally facing APIs to be set up on different HTTP listeners
- Listen addresses are now configured for internal and external components separately in the config, e.g.
```
client_api:
  internal_api:
    listen: http://localhost:7771
    connect: http://localhost:7771
  external_api:
    listen: http://[::]:8071
```

### DIFF
--- a/build/gobind/monolith.go
+++ b/build/gobind/monolith.go
@@ -184,8 +184,8 @@ func (m *DendriteMonolith) Start() {
 	httpRouter.PathPrefix(httputil.ExternalMediaPathPrefix).Handler(base.ExternalMediaAPIMux)
 
 	yggRouter := mux.NewRouter()
-	httpRouter.PathPrefix(httputil.ExternalFederationPathPrefix).Handler(base.ExternalFederationAPIMux)
-	httpRouter.PathPrefix(httputil.ExternalMediaPathPrefix).Handler(base.ExternalMediaAPIMux)
+	yggRouter.PathPrefix(httputil.ExternalFederationPathPrefix).Handler(base.ExternalFederationAPIMux)
+	yggRouter.PathPrefix(httputil.ExternalMediaPathPrefix).Handler(base.ExternalMediaAPIMux)
 
 	// Build both ends of a HTTP multiplex.
 	m.httpServer = &http.Server{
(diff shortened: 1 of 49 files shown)


build/gobind/monolith.go
```diff
@@ -170,16 +171,22 @@ func (m *DendriteMonolith) Start() {
...
-		base.BaseMux,
-		base.PublicAPIMux,
-		base.InternalAPIMux,
-		&cfg.Global,
-		base.UseHTTPAPIs,
+	monolith.AddAllPublicRoutes(
+		base.ExternalClientAPIMux,
+		base.ExternalFederationAPIMux,
+		base.ExternalKeyAPIMux,
+		base.ExternalMediaAPIMux,
```
Kegsay, line 178: Maybe public instead of external? Given it's used for `AddAllPublicRoutes`?

internal/setup/base.go
```diff
@@ -266,37 +268,74 @@ func (b *BaseDendrite) CreateFederationClient() *gomatrixserverlib.FederationCli
...
+				if err := internalServ.ListenAndServe(); err != nil {
+					logrus.WithError(err).Fatal("failed to serve HTTP")
+				}
+			}
+			logrus.Infof("Stopped %s listener on %s", b.componentName, internalServ.Addr)
+		}()
 	}
 
-	logrus.Infof("Stopped %s server on %s", b.componentName, serv.Addr)
+	<-block
```
Kegsay, line 338: Why are you doing this? It's not safe how it is written currently as you have two goroutines which will call `close(block)` - the 2nd close will panic. Use a wait group instead.

neilalexander: The idea is that *either* listener dying will stop the process, since it's not a good place to be if either the internal or the external APIs are not working.

Kegsay: Definitely needs appropriate commentary then please.

neilalexander: It actually occurs to me that we're already using `logrus.Fatal` so instead I'll replace the channel read with a `select {}`.

build/gobind/monolith.go
neilalexander: Renamed `External` to `Public` throughout.

PR #1267, opened 2020-08-13
Don't send to ACL'd servers

neilalexander:
This PR prevents PDUs, invites and (sometimes) EDUs from being sent to servers that are prohibited by server ACLs.

The amount of calls to the current state API to achieve this is probably quite inefficient in polylith mode - we may want some in-component caching around that in the future.

### DIFF
--- a/federationsender/queue/queue.go
+++ b/federationsender/queue/queue.go
@@ -21,11 +21,11 @@ import (
 	"fmt"
 	"sync"
 
+	stateapi "github.com/matrix-org/dendrite/currentstateserver/api"
 	"github.com/matrix-org/dendrite/federationsender/statistics"
 	"github.com/matrix-org/dendrite/federationsender/storage"
 	"github.com/matrix-org/dendrite/roomserver/api"
 	"github.com/matrix-org/gomatrixserverlib"
-	"github.com/matrix-org/util"
 	log "github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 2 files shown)


federationsender/queue/queue.go
```diff
@@ -128,14 +131,33 @@ func (oqs *OutgoingQueues) SendEvent(
 		)
 	}
 
-	// Remove our own server from the list of destinations.
-	destinations = filterAndDedupeDests(oqs.origin, destinations)
-	if len(destinations) == 0 {
+	// Deduplicate destinations and remove the origin from the list of
+	// destinations just to be sure.
+	destmap := map[gomatrixserverlib.ServerName]struct{}{}
```
Kegsay, line 136: `map[gomatrixserverlib.ServerName]bool` would be slightly better? Reads nicer too.

federationsender/queue/queue.go
```diff
@@ -200,15 +235,44 @@ func (oqs *OutgoingQueues) SendEDU(
...
-		}).Info("Sending EDU event")
+	// There is absolutely no guarantee that the EDU will have a room_id
+	// field, as it is not required by the spec. However, if it *does*
+	// (e.g. typing notifications) then we should try to make sure we don't
+	// bother sending them to servers that are prohibited by the server
+	// ACLs.
+	var header struct {
+		RoomID string `json:"room_id"`
+	}
+	if err := json.Unmarshal(e.Content, &header); err == nil {
```
Kegsay, line 254: I imagine `gjson` would be faster than unmarshalling into a temp struct, it also reads nicer.

federationsender/queue/queue.go
neilalexander: I mostly picked an empty struct because they don't occupy memory.

federationsender/queue/queue.go
neilalexander: Good point. I'll update it to use gjson.

PR #1273, opened 2020-08-14
Synchronous invites

neilalexander:
This PR removes invite queues from the federation sender and instead wires up the APIs so that they are sent synchronously. The room server will send a federation request if it is required.

### DIFF
--- a/roomserver/api/perform.go
+++ b/roomserver/api/perform.go
@@ -105,8 +105,6 @@ type PerformInviteRequest struct {
 }
 
 type PerformInviteResponse struct {
-	// If non-nil, the invite request failed. Contains more information why it failed.
-	Error *PerformError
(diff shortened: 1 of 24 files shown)


roomserver/api/perform.go
```diff
@@ -105,8 +105,6 @@ type PerformInviteRequest struct {
 }
 
 type PerformInviteResponse struct {
-	// If non-nil, the invite request failed. Contains more information why it failed.
-	Error *PerformError
```
Kegsay, line 109: Why was this removed? The lack of a dedicated type means we can't return appropriate error codes to the client.

neilalexander: I did wonder about the API shape of that. I can put something back in.

PR #1276, opened 2020-08-17
Roomserver per-room input lock, database-wide TransactionWriters

neilalexander:
This updates the roomserver so that rather than locking globally when processing incoming events, we take a look per-room instead. 

This also updates all of the roomserver `TransactionWriter`s to be database-wide, since SQLite locks the entire database for a write.

Maybe this will cause some unpredictable ordering if lots of concurrent input calls have events for the same room - in that case it could be better to create per-room channels instead and queue in-order. I don't know yet if this will be an issue.

### DIFF
--- a/appservice/storage/sqlite3/appservice_events_table.go
+++ b/appservice/storage/sqlite3/appservice_events_table.go
@@ -67,7 +67,7 @@ const (
 
 type eventsStatements struct {
 	db                                     *sql.DB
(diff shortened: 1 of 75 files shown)


neilalexander:
Looks like this results in a deadlock in the sync API storage tests.

neilalexander:
Closing this PR as it's getting bigger and more complicated and not any closer to solving the issue.

PR #1284, opened 2020-08-19
Add FederationClient interface to federationsender

Kegsay:
- Use a shim struct in HTTP mode to keep the same API as `FederationClient`.
- Use `federationsender` instead of `FederationClient` in `keyserver`.

### DIFF
--- a/federationsender/internal/api.go
+++ b/federationsender/internal/api.go
@@ -12,7 +12,6 @@ import (
 	roomserverAPI "github.com/matrix-org/dendrite/roomserver/api"
 	"github.com/matrix-org/gomatrix"
 	"github.com/matrix-org/gomatrixserverlib"
-	"github.com/matrix-org/util"
 )
 
 // FederationSenderInternalAPI is an implementation of api.FederationSenderInternalAPI
@@ -82,7 +81,6 @@ func (a *FederationSenderInternalAPI) doRequest(
 ) (interface{}, error) {
 	stats, err := a.isBlacklistedOrBackingOff(s)
 	if err != nil {
-		util.GetLogger(ctx).Infof("isBlacklistedOrBackingOff %v", err)
 		return nil, err
 	}
 	res, err := request()
@@ -107,7 +105,6 @@ func (a *FederationSenderInternalAPI) GetUserDevices(
 	ctx context.Context, s gomatrixserverlib.ServerName, userID string,
 ) (gomatrixserverlib.RespUserDevices, error) {
 	ires, err := a.doRequest(ctx, s, func() (interface{}, error) {
-		util.GetLogger(ctx).Infof("GetUserDevices being called now")
 		return a.federation.GetUserDevices(ctx, s, userID)
 	})
 	if err != nil {
--- a/federationsender/internal/api.go
+++ b/federationsender/internal/api.go
@@ -71,7 +71,7 @@ func failBlacklistableError(err error, stats *statistics.ServerStatistics) (unti
 	if !ok {
 		return stats.Failure()
 	}
-	if mxerr.Code >= 500 || mxerr.Code < 600 {
+	if mxerr.Code >= 500 && mxerr.Code < 600 {
 		return stats.Failure()
 	}
 	return
--- a/federationsender/internal/api.go
+++ b/federationsender/internal/api.go
@@ -1,12 +1,17 @@
 package internal
 
 import (
+	"context"
+	"time"
+
+	"github.com/matrix-org/dendrite/federationsender/api"
 	"github.com/matrix-org/dendrite/federationsender/queue"
 	"github.com/matrix-org/dendrite/federationsender/statistics"
 	"github.com/matrix-org/dendrite/federationsender/storage"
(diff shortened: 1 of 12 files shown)


federationsender/internal/api.go
```diff
@@ -38,3 +44,98 @@ func NewFederationSenderInternalAPI(
...
+func failBlacklistableError(err error, stats *statistics.ServerStatistics) {
+	if err == nil {
+		return
+	}
+	mxerr, ok := err.(gomatrix.HTTPError)
+	if !ok {
+		stats.Failure()
+		return
+	}
+	if mxerr.Code == 500 {
```
neilalexander, line 80: This should probably check for any error in the 500s, rather than just 500 itself.

federationsender/internal/api.go
```diff
@@ -38,3 +44,98 @@ func NewFederationSenderInternalAPI(
...
+			Blacklisted: true,
+		}
+	}
+	// Call BackoffIfRequired with a closed channel to make it return immediately.
+	// It will return the duration to backoff for.
+	var duration time.Duration
+	interrupt := make(chan bool)
+	close(interrupt)
+	var bo atomic.Bool
+	duration, _ = stats.BackoffIfRequired(bo, interrupt)
```
neilalexander, line 61: At the moment, this call will affect the backoff intervals for the destination queue as it will increase the `backoffCount`. This will mean that the next time the destination queue calls this, it'll think that it has backed off more times than it really has and so the interval will go right up (possibly leading us to blacklist early).

The simple way around this is to make an amendment to `BackoffIfRequired` so that if the interrupt channel is already closed, we stop there before increasing the backoff count, e.g.
```
func (s *ServerStatistics) BackoffIfRequired(backingOff atomic.Bool, interrupt <-chan bool) (time.Duration, bool) {
	if started := s.backoffStarted.Load(); !started {
		return 0, false
	}

	// Check if the interrupt channel is already closed. If it is
	// then this call should have no side effects but still return
	// the current duration.
	select {
	case <-interrupt:
		count := s.backoffCount.Load()
		return time.Second * time.Duration(math.Exp2(float64(count))), true
	default:
	}

...
```

neilalexander (changes requested):
Overall the pattern is fine - just a couple of things that need tweaking.

federationsender/internal/api.go
Kegsay: Not sure on this one, a 502 Bad Gateway because of apache/nginx fails feels wrong? 503/504 though seems ok?

federationsender/internal/api.go
Kegsay: Thanks for this!

federationsender/internal/api.go
neilalexander: I would think that we should still backoff for 502, as if you are using an nginx reverse proxy, Cloudflare or similar (as I am) then a bad gateway error shouldn't result in us giving up.

neilalexander (approved):

federationsender/internal/api.go
Kegsay: This doesn't seem to have the intended behaviour I want. A sytest times out because we never actually ever mark the backoff as complete so always return "backoff for 1s" with this code...

federationsender/internal/api.go
```diff
@@ -38,3 +43,96 @@ func NewFederationSenderInternalAPI(
...
+	until, blacklisted := stats.BackoffInfo()
+	if blacklisted {
+		return stats, &api.FederationClientError{
+			Blacklisted: true,
+		}
+	}
+	now := time.Now()
+	if until != nil && now.Before(*until) {
+		return stats, &api.FederationClientError{
+			RetryAfter: until.Sub(now),
```
neilalexander, line 58: It's purely a readability thing but you can use `time.Until(until)` rather than `now := time.Now(); until.Sub(now)`.

federationsender/internal/api.go
```diff
@@ -38,3 +43,96 @@ func NewFederationSenderInternalAPI(
...
+	if err != nil {
+		return nil, err
+	}
+	res, err := request()
+	if err != nil {
+		until, blacklisted := failBlacklistableError(err, stats)
+		now := time.Now()
+		var retryAfter time.Duration
+		if until.After(now) {
+			retryAfter = until.Sub(now)
```
neilalexander, line 92: Same here.

federationsender/statistics/statistics_test.go
```diff
@@ -36,7 +36,10 @@ func TestBackoff(t *testing.T) {
 		// completes but we will find out how long the backoff should
 		// have been.
 		interrupt := make(chan bool, 1)
-		close(interrupt)
+		go func() {
```
neilalexander, line 39: Is this still needed here?

neilalexander (approved):
Looks fine - just wondering if we still need that 10ms delay in `statistics_test.go`.

federationsender/statistics/statistics_test.go
Kegsay: Nope!

PR #1285, opened 2020-08-19
Transaction writer changes, move roomserver writers

neilalexander:
This PR does a couple of things:

- Changes `sqlutil.TransactionWriter` to be an interface and implements `sqlutil.ExclusiveTransactionWriter` (today's implementation) and `sqlutil.DummyTransactionWriter` (which does very little and is suitable for use with Postgres)
- Moves the `TransactionWriter`s away from the statements in the roomserver and up into the storage shared functions, since that gives us much better control over when/why we use it

This doesn't fix all of the SQLite tests yet but it does reduce the number of `database is locked` errors in the logs substantially.

### DIFF
--- a/roomserver/storage/shared/latest_events_updater.go
+++ b/roomserver/storage/shared/latest_events_updater.go
@@ -66,9 +66,9 @@ func (u *LatestEventsUpdater) CurrentStateSnapshotNID() types.StateSnapshotNID {
 
 // StorePreviousEvents implements types.RoomRecentEventsUpdater
 func (u *LatestEventsUpdater) StorePreviousEvents(eventNID types.EventNID, previousEventReferences []gomatrixserverlib.EventReference) error {
-	return u.d.Writer.Do(u.d.DB, nil, func(txn *sql.Tx) error {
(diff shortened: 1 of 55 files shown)


roomserver/storage/shared/latest_events_updater.go
```diff
@@ -104,7 +112,9 @@ func (u *LatestEventsUpdater) HasEventBeenSent(eventNID types.EventNID) (bool, e
 
 // MarkEventAsSent implements types.RoomRecentEventsUpdater
 func (u *LatestEventsUpdater) MarkEventAsSent(eventNID types.EventNID) error {
-	return u.d.EventsTable.UpdateEventSentToOutput(u.ctx, u.txn, eventNID)
+	return u.d.Writer.Do(u.d.DB, nil, func(txn *sql.Tx) error {
+		return u.d.EventsTable.UpdateEventSentToOutput(u.ctx, u.txn, eventNID)
```
Kegsay, line 116: Won't this make a txn in a txn?

neilalexander: Good point, we should be giving `u.txn` to the `TransactionWriter` here.

Kegsay (approved):

Issue #903, opened 2020-03-11
Client API: mutex on (user_id, room_id)


Kegsay:
We should be able to make some guarantees about message ordering from a single client, insofar that if you send 1,2,3,4,5 it should be in the DAG as 1,2,3,4,5 without forks *created by your own messages*. We can do this by locking for a given room/user tuple, which assumes we don't run multiple client APIs which aren't sticky on the user.

This may be a cause of SyTest flakiness, as it rapidly sends events.

prateek2211:
I am not clear about what issues says, but as far as I understood, there shouldn't be race between goroutines sending events to roomserver . So for that wouldn't adding lock and unlock inside send messages routines work ?

anandv96:
Hello, I would like to work on this issue. I'm also unsure of what exactly needs to be done. I understand that with messages being sent very rapidly, there is a possibility that a newer message would be entered into the room than an older one.

Like @prateek2211 mentioned, shouldn't this be addressed at the place where the clientAPI sends the messages to the roomserver? (maybe in clientapi/producers/roomserver.go?)

Kegsay:
Yup

Kegsay:
Related: https://github.com/matrix-org/dendrite/issues/547 which has a mutex per room on the roomserver.

PR #1286, opened 2020-08-19
#903: Client API: mutex on (user_id, room_id)

anandv96:
Addresses a condition where if there are multiple messages sent by the clientAPI in quick succession, the order of receiving the messages in the roomserver is not guaranteed.

Added a mutex similar to the one in https://github.com/matrix-org/dendrite/pull/1276
Mutexes map key is (roomID+userID)

### Pull Request Checklist

Signed-off-by: Anand Vasudevan <anandv96@disroot.org>

### DIFF
--- a/clientapi/routing/sendevent.go
+++ b/clientapi/routing/sendevent.go
@@ -16,6 +16,7 @@ package routing
 
 import (
 	"net/http"
+	"sync"
 
 	"github.com/matrix-org/dendrite/clientapi/httputil"
 	"github.com/matrix-org/dendrite/clientapi/jsonerror"
@@ -34,6 +35,10 @@ import (
 type sendEventResponse struct {
 	EventID string `json:"event_id"`
 }
+
+var (
+	mutexes sync.Map // (roomID+userID) -> mutex. mutexes to ensure correct ordering of sendEvents
+)
 
 // SendEvent implements:
 //   /rooms/{roomID}/send/{eventType}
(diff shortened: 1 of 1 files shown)


clientapi/routing/sendevent.go
```diff
@@ -76,6 +81,12 @@ func SendEvent(
 		}
 	}
 
+	// create a mutex for the specific user in the specific room
```
neilalexander, line 84: Given that `generateSendEvent` just before this must hit internal APIs to get `prev_events`, I'm wondering if it makes sense for the mutex to come just before this call. This is probably the main place that flooding our own events could create forks.

It's probably also fine just to use `defer mutex.(*sync.Mutex).Unlock()` directly after taking out the lock, since then it's guaranteed to be released regardless of which code path we take or which errors we encounter. We may end up holding the lock for *marginally* longer that way, but on a per-user basis I wouldn't worry about that particularly.

clientapi/routing/sendevent.go
```diff
@@ -35,6 +36,10 @@ type sendEventResponse struct {
 	EventID string `json:"event_id"`
 }
 
+var (
+	mutexes sync.Map // (roomID+userID) -> mutex. mutexes to ensure correct ordering of sendEvents
```
neilalexander, line 40: Since this will ultimately become a global within the package, perhaps we can name it more descriptively to reduce the chance of collision? Maybe `userRoomSendMutexes` or something.

neilalexander (changes requested):
Thanks for this! Just a couple of minor things.

clientapi/routing/sendevent.go
```diff
@@ -76,6 +81,12 @@ func SendEvent(
 		}
 	}
 
+	// create a mutex for the specific user in the specific room
+	// this avoids a situation where events that are received in quick succession are sent to the roomserver in a jumbled order
+	userID := device.UserID
+	mutex, _ := mutexes.LoadOrStore(roomID+userID, &sync.Mutex{})
+	mutex.(*sync.Mutex).Lock()
+
```
Kegsay, line 89: Please `defer mutex.(*sync.Mutex).Unlock()` instead of unlocking before each `return` further down.

anandv96:
Updated with the suggestions from @Kegsay and @neilalexander

neilalexander (approved):

neilalexander:
Thank you for the contribution!

PR #1287, opened 2020-08-19
Fix more roomserver transactions/locks

neilalexander:
This PR *should* fix the remaining `database is locked` issues in the roomserver. That is, until we try and add per-room input mutexes back again, but we'll see.

### DIFF
--- a/roomserver/storage/shared/storage.go
+++ b/roomserver/storage/shared/storage.go
@@ -383,7 +383,7 @@ func (d *Database) StoreEvent(
 	err = d.Writer.Do(d.DB, nil, func(txn *sql.Tx) error {
 		if txnAndSessionID != nil {
 			if err = d.TransactionsTable.InsertTransaction(
-				ctx, nil, txnAndSessionID.TransactionID,
+				ctx, txn, txnAndSessionID.TransactionID,
 				txnAndSessionID.SessionID, event.Sender(), event.EventID(),
 			); err != nil {
 				return fmt.Errorf("d.TransactionsTable.InsertTransaction: %w", err)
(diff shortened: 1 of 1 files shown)


roomserver/storage/shared/storage.go
```diff
@@ -179,22 +179,19 @@ func (d *Database) RoomNIDExcludingStubs(ctx context.Context, roomID string) (ro
...
+	var eventNIDs []types.EventNID
+	eventNIDs, currentStateSnapshotNID, err = d.RoomsTable.SelectLatestEventNIDs(ctx, nil, roomNID)
+	if err != nil {
+		return
+	}
+	references, err = d.EventsTable.BulkSelectEventReference(ctx, nil, eventNIDs)
+	if err != nil {
+		return
+	}
+	depth, err = d.EventsTable.SelectMaxEventDepth(ctx, nil, eventNIDs)
```
Kegsay, line 191: These three selects need to be in a txn in order to snapshot the DB correctly, else you could end up with `eventNIDs` which are not in `EventsTable`, I think.

Kegsay (approved):

roomserver/storage/shared/storage.go
neilalexander: It looks like that's not the case in Postgres defaults:
> Read Committed is the default isolation level in PostgreSQL. When a transaction uses this isolation level, a SELECT query (without a FOR UPDATE/SHARE clause) sees only data committed before the query began; it never sees either uncommitted data or changes committed during query execution by concurrent transactions. In effect, a SELECT query sees a snapshot of the database as of the instant the query begins to run. However, SELECT does see the effects of previous updates executed within its own transaction, even though they are not yet committed. Also note that two successive SELECT commands can see different data, even though they are within a single transaction, if other transactions commit changes after the first SELECT starts and before the second SELECT starts.
I think we would need repeatable reads if we wanted that — https://www.postgresql.org/docs/9.5/transaction-iso.html

PR #1290, opened 2020-08-20
Component-wide TransactionWriters

### LINKED ISSUE #1252: Database is locked Error
Running dendrite-monolith-server with sqlite built from source (git pulled 20200807).
After joining a few rooms (#dendrite and #dendrite-dev) from element-web and also logging in to the same account with element-android, a lot of "database is locked" errors show up in the log. Can be seen in attached log.

[Monolith1.log.gz](https://github.com/matrix-org/dendrite/files/5045247/Monolith1.log.gz)

neilalexander:
This modifies the remaining components to use a global `TransactionWriter`. It also updates the partition offsets to use it too, so it *should* fix #1252.

### DIFF
--- a/appservice/storage/postgres/storage.go
+++ b/appservice/storage/postgres/storage.go
@@ -32,6 +32,7 @@ type Database struct {
 	events eventsStatements
 	txnID  txnStatements
 	db     *sql.DB
+	writer sqlutil.TransactionWriter
 }
 
 // NewDatabase opens a new database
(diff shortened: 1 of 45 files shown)


internal/sqlutil/writer_dummy.go
```diff
@@ -4,15 +4,15 @@ import (
...
-func NewDummyTransactionWriter() TransactionWriter {
-	return &DummyTransactionWriter{}
+func NewDummyWriter() Writer {
+	return &DummyWriter{}
 }
 
-func (w *DummyTransactionWriter) Do(db *sql.DB, txn *sql.Tx, f func(txn *sql.Tx) error) error {
-	if txn == nil {
+func (w *DummyWriter) Do(db *sql.DB, txn *sql.Tx, f func(txn *sql.Tx) error) error {
+	if db != nil && txn == nil {
```
Kegsay, line 15: When would `db` ever actually be `nil`?

PR #1291, opened 2020-08-20
Add support for running Complement on current working directories

Kegsay:
This will be used in the future by Buildkite to run on CI.


neilalexander (approved):
LGTM!

PR #1293, opened 2020-08-24
Don't overwrite global err before return

oliverpool:
_Closes #846_ 

This PR adds a helper to overwrite the error returned by a function if it was previously nil and the deferred function returned a non-nil error.

I looked for occurrences of `defer func() {` and found a couple of packages where I this this helper improves the code.
Since there is no `util` package, I copied/paste the function : I am open to suggestion to reduce this code-duplication.

In `sqlutil`, I adapted the function to work for `EndTransaction`.

* [ ] I didn't add any new test (since this is not adding any feature), but feel free to hint any test that this PR could add

### DIFF
--- a/appservice/workers/transaction_scheduler.go
+++ b/appservice/workers/transaction_scheduler.go
@@ -101,6 +101,9 @@ func worker(db storage.Database, ws types.ApplicationServiceWorkerState) {
 		// Backoff if the application service does not respond
 		err = send(client, ws.AppService, txnID, transactionJSON)
 		if err != nil {
+			log.WithFields(log.Fields{
+				"appservice": ws.AppService.ID,
+			}).WithError(err).Error("unable to send event")
 			// Backoff
 			backoff(&ws, err)
 			continue
@@ -207,7 +210,7 @@ func send(
 	appservice config.ApplicationService,
 	txnID int,
 	transaction []byte,
-) error {
+) (err error) {
 	// PUT a transaction to our AS
 	// https://matrix.org/docs/spec/application_service/r0.1.2#put-matrix-app-v1-transactions-txnid
 	address := fmt.Sprintf("%s/transactions/%d?access_token=%s", appservice.URL, txnID, url.QueryEscape(appservice.HSToken))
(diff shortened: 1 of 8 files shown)


appservice/workers/transaction_scheduler.go
```diff
@@ -207,7 +207,14 @@ func send(
 	appservice config.ApplicationService,
 	txnID int,
 	transaction []byte,
-) error {
+) (err error) {
+	defer func() {
+		if err != nil {
+			log.WithFields(log.Fields{
+				"appservice": appservice.ID,
+			}).WithError(err).Error("unable to send event")
```
Kegsay, line 215: I would rather log at the caller's side (the code which calls `send(...)`) than in a `defer` like this.

internal/sqlutil/sql.go
```diff
@@ -38,9 +38,18 @@ type Transaction interface {
...
 	} else {
-		return txn.Rollback() // nolint: errcheck
+		return txn.Rollback()
+	}
+}
+
+// CheckEndTransactionErr ends a transaction and overwrites the error pointer if its value was nil.
+// If the transaction succeeded then it is committed, otherwise it is rolledback.
+// Designed to be used with defer (see EndTransaction otherwise).
+func CheckEndTransactionErr(txn Transaction, succeeded *bool, err *error) {
```
Kegsay, line 50: Slighty confusing naming: the `Check` implies that you've already called `EndTransaction` and are now checking the result, but we're actually calling `EndTransaction` here. `EndTransactionAndCheck` would be clearer but still awkward.

Kegsay (changes requested):
Few minor things, otherwise LGTM thanks!

appservice/workers/transaction_scheduler.go
oliverpool: :+1: done

internal/sqlutil/sql.go
oliverpool: What do you think of `EndTransactionWithCheck`? (`With` is quite usual in the standard library)

Kegsay: Sounds good to me.

Kegsay (approved):
Next time please do not force push changes to PRs, as it makes it impossible for me to see the diff from when I last reviewed.

neilalexander (approved):

PR #1296, opened 2020-08-24
Use writer transaction in InsertKeyChange

neilalexander:
This might fix the final remaining `database is locked` errors from the SQLite sytest runs.

### DIFF
--- a/keyserver/storage/postgres/one_time_keys_table.go
+++ b/keyserver/storage/postgres/one_time_keys_table.go
@@ -21,7 +21,6 @@ import (
 	"time"
 
 	"github.com/matrix-org/dendrite/internal"
-	"github.com/matrix-org/dendrite/internal/sqlutil"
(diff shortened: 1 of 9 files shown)


keyserver/internal/device_list_update.go
```diff
@@ -154,7 +159,9 @@ func (u *DeviceListUpdater) mutex(userID string) *sync.Mutex {
 func (u *DeviceListUpdater) ManualUpdate(ctx context.Context, serverName gomatrixserverlib.ServerName, userID string) error {
 	mu := u.mutex(userID)
 	mu.Lock()
-	err := u.db.MarkDeviceListStale(ctx, userID, true)
+	err := u.writer.Do(nil, nil, func(_ *sql.Tx) error {
+		return u.db.MarkDeviceListStale(ctx, userID, true)
```
Kegsay, line 163: Why is the writer leaking outside the storage layer?

Kegsay (approved):

PR #1339, opened 2020-08-25
[readability] use event.StateKeyEquals where relevant and minor for-loop refactoring

oliverpool:
I am trying to understand how membership works (leave, join & co).

I saw that `StateKeyEquals` could be used in some places and make the code more readable.
And that some for-loops could be refactored to reduce their nesting (which IMHO improves readability).

### DIFF
--- a/clientapi/routing/membership.go
+++ b/clientapi/routing/membership.go
@@ -388,15 +388,20 @@ func checkMemberInRoom(ctx context.Context, stateAPI currentstateAPI.CurrentStat
 		e := jsonerror.InternalServerError()
 		return &e
 	}
-	ev, ok := membershipRes.StateEvents[tuple]
-	if !ok {
+	ev := membershipRes.StateEvents[tuple]
+	if ev == nil {
 		return &util.JSONResponse{
 			Code: http.StatusForbidden,
 			JSON: jsonerror.Forbidden("user does not belong to room"),
 		}
 	}
 	membership, err := ev.Membership()
-	if err != nil || membership != "join" {
+	if err != nil {
+		util.GetLogger(ctx).WithError(err).Error("Member event isn't valid")
+		e := jsonerror.InternalServerError()
+		return &e
+	}
+	if membership != gomatrixserverlib.Join {
 		return &util.JSONResponse{
 			Code: http.StatusForbidden,
 			JSON: jsonerror.Forbidden("user does not belong to room"),
--- a/clientapi/routing/sendtyping.go
+++ b/clientapi/routing/sendtyping.go
@@ -21,7 +21,6 @@ import (
 	"github.com/matrix-org/dendrite/eduserver/api"
 	userapi "github.com/matrix-org/dendrite/userapi/api"
 	"github.com/matrix-org/dendrite/userapi/storage/accounts"
-	"github.com/matrix-org/gomatrixserverlib"
 	"github.com/matrix-org/util"
 )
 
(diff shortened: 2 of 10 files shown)


clientapi/routing/sendtyping.go
```diff
@@ -46,46 +45,19 @@ func SendTyping(
...
-	}, &res)
-	if err != nil {
-		util.GetLogger(req.Context()).WithError(err).Error("QueryCurrentState failed")
-		return jsonerror.InternalServerError()
-	}
-	ev := res.StateEvents[tuple]
-	if ev == nil {
-		return util.JSONResponse{
-			Code: http.StatusForbidden,
-			JSON: jsonerror.Forbidden("User not in this room"),
```
oliverpool, line 66: The message will now be `user does not belong to room` (see `checkMemberInRoom`)

clientapi/routing/membership.go
```diff
@@ -388,15 +388,20 @@ func checkMemberInRoom(ctx context.Context, stateAPI currentstateAPI.CurrentStat
...
+	if ev == nil {
 		return &util.JSONResponse{
 			Code: http.StatusForbidden,
 			JSON: jsonerror.Forbidden("user does not belong to room"),
 		}
 	}
 	membership, err := ev.Membership()
-	if err != nil || membership != "join" {
+	if err != nil {
+		util.GetLogger(ctx).WithError(err).Error("Member event isn't valid")
```
neilalexander, line 400: More accurately, `Membership()` returns an error when it's *either* an invalid membership state *or* not a membership event.

federationapi/routing/join.go
```diff
@@ -253,11 +253,12 @@ func SendJoin(
 	// there isn't much point in sending another join event into the room.
 	alreadyJoined := false
 	for _, se := range stateAndAuthChainResponse.StateEvents {
+		if !se.StateKeyEquals(*event.StateKey()) {
```
neilalexander, line 256: I suspect this might panic if `event.StateKey()` returns `nil`?

clientapi/routing/membership.go
oliverpool: Yes, but since we requested events of the type `gomatrixserverlib.MRoomMember` and of the current user, this should only error on wrong content.

federationapi/routing/join.go
oliverpool: This is checked already above (line 168)

neilalexander: So it is!

roomserver/internal/perform_join.go
```diff
@@ -204,11 +204,12 @@ func (r *RoomserverInternalAPI) performJoinRoomByID(
 		// a member of the room.
 		alreadyJoined := false
 		for _, se := range buildRes.StateEvents {
+			if !se.StateKeyEquals(*event.StateKey()) {
```
neilalexander, line 207: We probably want to preserve `se.StateKey() != nil` here for safety. I don't *think* there's a reason that the query API would return something strange to the builder but it's not worth panicking if it does.

oliverpool: This nil-check is already implemented inside `StateKeyEquals`

neilalexander: It is for `se`, but not for `*event` which has to dereference into a `string` type or it'll likely panic.

oliverpool: I can simply replace  `*event.StateKey()` with `userID`

neilalexander: Perhaps the sensible thing to do is `event.StateKeyEquals(userID)` after the event is built. Then you should be able to be confident about `*event.StateKey()` being OK on line 207.

oliverpool: I feel it would be like testing that the implementation of the builder is right.

Maybe I could add a simple unit-test for that (calling the builder just like in this function and checking the `StateKeyEquals`).

Then I can use either `se.StateKeyEquals(userID)` or `se.StateKeyEquals(*event.StateKey())` without having to think about it :-)

federationapi/routing/join.go
Kegsay: That's not ideal, relying on a check roughly 100 lines above. This function is far too long :/

Kegsay (approved):

neilalexander (approved):

PR #1340, opened 2020-08-25
Enforce history visibility etc for /rooms/{roomID}/state

### LINKED ISSUE #665: GET /state... as banned should get the room’s state as it was when banned

neilalexander:
I did this at some point in the past for `/rooms/{roomID}/state/{type}/{statekey}` but not for `/rooms/{roomID}/state`. This fixes that.

This technically fixes #665, but there's an possibly an issue in Sytest at matrix-org/sytest#940 around this.

### DIFF
--- a/clientapi/routing/routing.go
+++ b/clientapi/routing/routing.go
@@ -201,7 +201,7 @@ func Setup(
 		if err != nil {
 			return util.ErrorResponse(err)
 		}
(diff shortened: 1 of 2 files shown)


Kegsay (commented):
There's a lot of copypasta from the `/state/{type}/{key}` form, can we factor this out?

Kegsay (approved):

PR #1352, opened 2020-08-27
Remove device DB from clientapi

Kegsay:


userapi/userapi_test.go
```diff
@@ -15,35 +15,36 @@ import (
...
 		t.Fatalf("failed to create account DB: %s", err)
 	}
-	deviceDB, err := devices.NewDatabase(&config.DatabaseOptions{
-		ConnectionString: "file::memory:",
-	}, serverName)
-	if err != nil {
-		t.Fatalf("failed to create device DB: %s", err)
+	cfg := &config.UserAPI{
+		DeviceDatabase: config.DatabaseOptions{
+			ConnectionString: "file::memory:",
```
neilalexander, line 34: To ensure this doesn't become flakey, do we need to set the maximum number of connections to 1?

neilalexander (commented):
Overall fine, but I think there are opportunities to prevent locks using writers.

neilalexander (approved):

Issue #1348, opened 2020-08-27
Possible transaction in transaction error


neilalexander:
Sytest in Buildkite [kicked this up recently](https://buildkite.com/matrix-dot-org/dendrite/builds/2884#e5b23005-9c98-45ae-bbb7-9918d319f098) in the SQLite Full HTTP tests, but I can't reproduce it and another run didn't yield the same result:
```
Test 149 Ephemeral messages received from clients are correctly expired... [server]: time="2020-08-27T09:40:13.110089246Z" level=panic msg="roomserver output log: write event failure" func="onNewRoomEvent\n\t" file=" [/src/syncapi/consumers/roomserver.go:161]" add="[$vFF1x4gOKURkf37dwcsqRUcJFUzT5L-HqMWpOoy02WY]" del="[]" error="sqlutil.WithTransaction.Begin: cannot start a transaction within a transaction" event="{\"auth_events\":[],\"content\":{\"creator\":\"@anon-20200827_093918-79:localhost:8800\",\"room_version\":\"5\"},\"depth\":1,\"hashes\":{\"sha256\":\"WgtujbAQjY8pgV6jqevB3v5U7HKwPc/rvXfj23ycB00\"},\"origin\":\"localhost:8800\",\"origin_server_ts\":1598521213081,\"prev_events\":[],\"prev_state\":[],\"room_id\":\"!CWa3zy3kpPsO898F:localhost:8800\",\"sender\":\"@anon-20200827_093918-79:localhost:8800\",\"signatures\":{\"localhost:8800\":{\"ed25519:ns5n\":\"eRDyl01g6PD/xY/VHEZcdDj2TlA+K24xdEQ0JBQ6e4jZNYrImpBzK4UOrepGoo9+YGZR/iBaBTlGogfmtCgFDA\"}},\"state_key\":\"\",\"type\":\"m.room.create\"}"
--
  | [server]: panic: (*logrus.Entry) (0x10432c0,0xc0002a4770)
  | [server]:
  | [server]: goroutine 109 [running]:
  | [server]: github.com/sirupsen/logrus.Entry.log(0xc0000ea230, 0xc00095e450, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
  | [server]: 	/gopath/pkg/mod/github.com/sirupsen/logrus@v1.6.0/entry.go:259 +0x339
[...]

Issue #1159, opened 2020-06-24
sqlite/http: /sync 500s forever


Kegsay:
Flavour: sqlite with HTTP APIS
```
time="2020-06-24T13:38:46.988060755Z" level=error msg="rp.currentSyncForUser failed" func="github.com/matrix-org/dendrite/syncapi/sync.(*RequestPool).OnIncomingSyncRequest" file="/src/syncapi/sync/requestpool.go:73" device_id=wjCq_DGv error="context cance
led" limit=20 req.id=TLp3yLoyiiTY req.method=GET req.path=/_matrix/client/r0/sync since=s0_0 timeout=0s user_id="@anon-20200624_133759-77:localhost:8800"
time="2020-06-24T13:38:46.988155247Z" level=info msg="Responding (55 bytes)" func=github.com/matrix-org/util.respond file="/gopath/pkg/mod/github.com/matrix-org/util@v0.0.0-20190711121626-527ce5ddefc7/json.go:170" code=500 req.id=TLp3yLoyiiTY req.method=G
ET req.path=/_matrix/client/r0/sync
...
time="2020-06-24T13:38:47.247178933Z" level=error msg="rp.currentSyncForUser failed" func="github.com/matrix-org/dendrite/syncapi/sync.(*RequestPool).OnIncomingSyncRequest" file="/src/syncapi/sync/requestpool.go:73" device_id=rcN-qsSO error="cannot start 
a transaction within a transaction" limit=20 req.id=mlxSWkVI0BLN req.method=GET req.path=/_matrix/client/r0/sync since=s0_0 timeout=0s user_id="@anon-20200624_133759-80:localhost:8800"
time="2020-06-24T13:38:47.247337996Z" level=info msg="Responding (55 bytes)" func=github.com/matrix-org/util.respond file="/gopath/pkg/mod/github.com/matrix-org/util@v0.0.0-20190711121626-527ce5ddefc7/json.go:170" code=500 req.id=mlxSWkVI0BLN req.method=GET req.path=/_matrix/client/r0/sync
```

Looks like if the context is cancelled we don't release a txn causing "cannot start a transaction within a transaction". Because this is the HTTP API form failing, the problem code is likely one of the paths where syncapi talks to an internal API (eg userapi, roomserver). https://buildkite.com/matrix-dot-org/dendrite/builds/2198#03883788-c943-4789-82cc-a1a31602c0a7

PR #1363, opened 2020-08-28
Add some error wrapping to sync API

neilalexander:
This is in the vein hope that it might yield something more useful the next time #1159 or #1348 happen, but I can't seem to force the issue to happen.


Kegsay (approved):
This closes #1348 and #1159 - we knew it was double txn and related to contexts, so it feels like the upstream sqlite3 issue really is the cause here. We can re-open if we encounter it again (otherwise when will we ever close them, how long do we wait?!)

PR #1370, opened 2020-08-30
Peeking via MSC2753

ara4n:
This is a draft PR to gather feedback on the vague shape (and to help inform review of MSC2753), and to get something committed.

Full scope (although i'd like to merge this before it bitrots and handle the other checkboxes in other PRs):
 * [x] get feedback on whether [the design](https://github.com/matrix-org/dendrite/blob/matthew/peeking/docs/peeking.md) is remotely correct
 * [x] make it build
 * [x] make it work
 * [x] actually add `peek` block into the sync response
 * [x] respect history visibility
 * [x] add postgres
 * [ ] add `/unpeek`
 * [ ] show unpeeked rooms briefly in the `leave` block
 * [ ] support EDUs in peeked rooms
 * [ ] figure out whether to stop trying to wake up devices which are no longer syncing
 * [ ] stop peeking if history visibility changes
 * [ ] clean up peeks on since-less /sync
 * [ ] sort out unstable prefixes
 * [x] tracing
 * [x] complement / sytests - sytest in progress at https://github.com/matrix-org/sytest/pull/944

See also https://github.com/matrix-org/matrix-doc/pull/2753

This consciously doesn't implement [peeking over federation](https://github.com/matrix-org/matrix-doc/pull/2444), which should be a separate PR, but is the main reason for doing this.


docs/peeking.md
```diff
@@ -0,0 +1,19 @@
...
+Peeking is implemented as per [MSC2753](https://github.com/matrix-org/matrix-doc/pull/2753).
+
+Implementationwise, this means:
+ * Users call `/peek` and `/unpeek` on the clientapi from a given device.
+ * The clientapi delegates these via HTTP to the roomserver, which coordinates peeking in general for a given room
+ * The roomserver writes an NewPeek event into the kafka log headed to the syncserver
+ * The syncserver tracks the existence of the local peek in its DB, and then starts waking up the peeking devices for the room in question, putting it in the `peek` section of the /sync response.
+
+Questions (given this is [my](https://github.com/ara4n) first time hacking on Dendrite):
+ * The whole clientapi -> roomserver -> syncapi flow to initiate a peek seems very indirect.  Is there a reason not to just let syncapi itself host the implementation of `/peek`?
```
Kegsay, line 12: Requests to peek from the client and requests to peek from federation are the same/very similar, hence putting it in the roomserver makes more sense (clientapi -> roomserver and fedapi -> roomserver).

The one counterpoint is that this is truly only read-only: the room state is unmodified and the state stored is specific to syncapi only. We do make exceptions - filtering is a good example which exists in syncapi because it is specific to syncing only. If peeking didn't have the federation component I could argue that it should be in syncapi as it's just a filter (akin to the previous filter proposal), but the addition of federation complicates this.

docs/peeking.md
```diff
@@ -0,0 +1,19 @@
...
+
+In future, peeking over federation will be added as per [MSC2444](https://github.com/matrix-org/matrix-doc/pull/2444).
+ * The `roomserver` will kick the `federationsender` much as it does for a federated `/join` in order to trigger a federated `/peek`
+ * The `federationsender` tracks the existence of the remote peek in question
+ * The `federationsender` regularly renews the remote peek as long as there are still peeking devices syncing for it.
```
Kegsay, line 17: I don't feel the `federationsender` should know or care about peeking devices. Keeping it a dumb queue is preferable as it's already got a lot of stuff to track wrt backoff/blacklisting. I would propose to keep peek renewal a syncapi responsibility as it knows about which devices are syncing. How this interacts with roomserver yet I don't know, we might need to talk about this more.

roomserver/api/output.go
```diff
@@ -195,3 +200,11 @@ type OutputRedactedEvent struct {
 	// The value of `unsigned.redacted_because` - the redaction event itself
 	RedactedBecause gomatrixserverlib.HeaderedEvent
 }
+
+// An OutputNewPeek is written whenever a user starts peeking into a room
+// using a given device.
+type OutputNewPeek struct {
```
Kegsay, line 206: Random Q: Is it desirable for members in the room to know who is peeking?

syncapi/storage/shared/syncserver.go
```diff
@@ -961,6 +1024,42 @@ func (d *Database) getStateDeltas(
 		return nil, nil, err
 	}
 
+	// find out which rooms this user is peeking, if any.
+	// We do this before joins so joins overwrite peeks
+	peeks, err := d.Peeks.SelectPeeks(ctx, userID, device.ID)
```
Kegsay, line 1029: This will be racey with `d.Peeks.MarkPeeksAsOld(ctx, txn, userID, device.ID)` - use the same `txn` please.

syncapi/sync/notifier.go
```diff
@@ -48,11 +50,11 @@ type Notifier struct {
 // the joined users within each of them by calling Notifier.Load(*storage.SyncServerDatabase).
 func NewNotifier(pos types.StreamingToken) *Notifier {
 	return &Notifier{
-		currPos:             pos,
-		roomIDToJoinedUsers: make(map[string]userIDSet),
-		userDeviceStreams:   make(map[string]map[string]*UserDeviceStream),
-		streamLock:          &sync.Mutex{},
-		lastCleanUpTime:     time.Now(),
```
Kegsay, line 55: Looks like we lost this?

docs/peeking.md
Kegsay: fedapi -> roomserver -> syncapi for the record will matter if the answer to https://github.com/matrix-org/dendrite/pull/1370#discussion_r481071577 is yes, as syncapi will need to inform users in the room of the members/counts of people peeking.

roomserver/api/output.go
ara4n: good question. in msc1776 you got this for free; in msc2753 it's kept private. in the end, privacy is probably best, otherwise it's effectively a tracking pixel.

syncapi/sync/notifier.go
ara4n: mm, which this?

Kegsay: `lastCleanUpTime:     time.Now(),`

ara4n:
This is passing its sytests (when they don't wedge with db locked errors), and good enough for a proper review.

N.B. that it pulls in the goid txn debugging branch, which should be ignored for the purposes of review.

docs/peeking.md
ara4n: agreed.

docs/peeking.md
ara4n: yeah, having it in roomserver is making sense in the end.

clientapi/routing/routing.go
```diff
@@ -101,6 +101,17 @@ func Setup(
 			)
 		}),
 	).Methods(http.MethodPost, http.MethodOptions)
+	r0mux.Handle("/peek/{roomIDOrAlias}",
+		httputil.MakeAuthAPI(gomatrixserverlib.Join, userAPI, func(req *http.Request, device *userapi.Device) util.JSONResponse {
```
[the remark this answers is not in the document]
ara4n: this was a typo

syncapi/storage/sqlite3/peeks_table.go
```diff
@@ -0,0 +1,185 @@
...
+	"SELECT room_id, user_id, device_id FROM syncapi_peeks"
+
+const markPeeksAsOldSQL = "" +
+	"UPDATE syncapi_peeks SET new=false WHERE user_id = $1 and device_id = $2"
+
+type peekStatements struct {
+	db                            *sql.DB
+	streamIDStatements            *streamIDStatements
+	insertPeekStmt        		  *sql.Stmt
+	deletePeekStmt         		  *sql.Stmt
```
neilalexander, line 67: Something went wildly wrong with the indentation here :D Better to use spaces rather than tabs.

syncapi/storage/sqlite3/syncserver.go
```diff
@@ -106,3 +119,172 @@ func (d *SyncServerDatasource) prepare() (err error) {
 	}
 	return nil
 }
+
+func (d *SyncServerDatasource) Events(ctx context.Context, eventIDs []string) ([]gomatrixserverlib.HeaderedEvent, error) {
```
neilalexander, line 123: I'm confused by all of these functions here that are already in `shared`. Is this to do with the GOID tracing? If so, can we revert those commits?

Kegsay: Yes it's for goid tracing with dbctx. Should be removed.

neilalexander: I've backed this out now.

neilalexander:
I've massaged this PR quite a bit this afternoon— 

* Cleared some merge conflicts with my earlier sync API work
* Fixed a boatload of lint issues 
* Buffed out some of the bitrot by fast-forwarding `master` on top
* Backed out the GOID merge so that's all been cleaned up now
* Fixed a bug and a missing function in the full-HTTP `PerformPeek` implementation

syncapi/storage/shared/syncserver.go
```diff
@@ -950,6 +1059,18 @@ func (d *Database) getStateDeltas(
 			//       the timeline.
 			if membership := getMembershipFromEvent(&ev.Event, userID); membership != "" {
 				if membership == gomatrixserverlib.Join {
+					if peeking[roomID] {
+						// we automatically cancel our peeks when we join a room
+						err = d.Writer.Do(d.DB, txn, func(txn *sql.Tx) error {
```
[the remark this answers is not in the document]
ara4n: yeah, i dithered back and forth between doing it here and in the consumer. i forget why it landed here... will rethink.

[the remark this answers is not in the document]
ara4n: fixed in 034ff320

ara4n:
assuming CI passes, i think this is ready to merge.

syncapi/storage/postgres/peeks_table.go
```diff
@@ -0,0 +1,183 @@
...
+
+	"github.com/matrix-org/dendrite/internal"
+	"github.com/matrix-org/dendrite/internal/sqlutil"
+	"github.com/matrix-org/dendrite/syncapi/storage/tables"
+	"github.com/matrix-org/dendrite/syncapi/types"
+)
+
+const peeksSchema = `
+CREATE TABLE IF NOT EXISTS syncapi_peeks (
+	id BIGINT PRIMARY KEY DEFAULT nextval('syncapi_stream_id'),
```
Kegsay, line 30: By having this be the primary key means we effectively don't ever clobber entries in this table. If we had it being a unique key on room/user/device then we could just toggle the id/deleted status and bound growth of this table.

ara4n: hum, this is also going to go badly with DeletePeeks, which tries to set the IDs of the deleted rows to be the same streampos, i think.

I'm missing how removing the pkey would change how the table grows though...

Kegsay: > which tries to set the IDs of the deleted rows to be the same streampos

Postgres is fine because the query uses `nextval('syncapi_stream_id')` on each UPDATE. This will fail on SQLite though because as you say we do use the same stream ID.

> I'm missing how removing the pkey would change how the table grows though

It doesn't, it's just I expected to see it clobber on user/device/room instead which is what made me realise that the current form is unbounded growth.

ara4n: fixed in b1d1ccdb

ara4n:
sytest is still failing for me locally on postgres with the peeked blocks having no timeline entries, so i am very confused on how this is passing CI.

syncapi/storage/sqlite3/peeks_table.go
```diff
@@ -27,13 +27,14 @@ import (
...
-	id INTEGER PRIMARY KEY,
+	id INTEGER,
 	room_id TEXT NOT NULL,
 	user_id TEXT NOT NULL,
 	device_id TEXT NOT NULL,
 	deleted BOOL NOT NULL DEFAULT false,
     -- When the peek was created in UNIX epoch ms.
-    creation_ts INTEGER NOT NULL
+    creation_ts INTEGER NOT NULL,
+    UNIQUE(room_id, user_id, device_id)
```
Kegsay, line 37: This alone will still cause problems if you peek/unpeek/peek into the same room I think. The 2nd peek will fail to insert.

Kegsay (approved):
Otherwise LGTM!

neilalexander (approved):

syncapi/storage/sqlite3/peeks_table.go
ara4n: fixed by https://github.com/matrix-org/dendrite/pull/1370/commits/956184377276dd91e82e0103ffe24c9c8ddc0fd2

Issue #1297, opened 2020-08-24
Implement SSO for logins


Kegsay:
Sytests:
```
    × Interactive authentication types include SSO
    × Can perform interactive authentication with SSO
    × The user must be consistent through an interactive authentication session with SSO
    × The operation must be consistent through an interactive authentication session
    × Can login with 3pid and password using m.login.password
    × login types include SSO
    × /login/cas/redirect redirects if the old m.login.cas login type is listed
    × Can login with new user via CAS
```

anandv96:
I'm working on this issue with @rohitmohan96 . Can you assign it to us?

duritong:
Will CAS the only supported SSO mechanism or will you also add SAML? How will one be able todo usermapping?

olanod:
PR was closed, nobody that could work on this? 😞

tommie:
I'm forking @anandv96's PR and continuing. Since it included a solution to #403, and we need that first, I'll start by splitting and trying to get hat closed first.

thegergo02:
Is this still being worked on? This is the only thing holding me back from switching to Dandrite.

tommie:
No. I'm waiting for a review of PR #2014. So far, no response.

ynerant:
Hi!
I am also very interested in this feature, this is the main reason I still using Synapse for a personal project.
Do you know if 1) the PR #2014 is going to be merged soon, and 2) if that means that other login types will be supported once the PR got merged?
Thanks!

kegsay:
I am aiming to get #2014 merged soon yes, we're basically happy with it at this point. That is a pre-requisite for allowing SSO flows.

compgeniuses:
I am curious About one thing, Why is this still in Opn state, if it has been done and merged, or was it only components of its implementation that were implemented.

If an issue was implemented, and merged, or Done, should it not be marked as closed state, so that other people could know its done.

Or maybe Add new tags like.

Partially-Done
Being-implemented
etcetra

tommie:
It's not done. This was only a part of it. Using tags sounds like a good idea.

The next step is rebasing and creating the next PR from https://github.com/tommie/dendrite/commits/loginsso. @kegsay also asked me for some post-merge cleanups in #2014, so that also needs to get done.

compgeniuses:
ok sure

scatterd:
[...]

PR #1374, opened 2020-09-01
[WIP] Implement m.login.token and SSO login initial changes

anandv96:
Made some changes to implement CAS SSO support

- made config changes for cas 
- GET /login returns the `m.login.sso` flow if the config is enabled
  - `login types include SSO` test passes if changes are made to the sytest dendrite config
- redirect user to the CAS server upon hitting `/login/sso/redirect`
- validate token from CAS server
- create user if required
- (pending) redirect client to the `redirectUrl` with a login token to complete the login
- (pending) change sytest whitelist

Clarifications needed
- The spec mentions that a client should hit a User interactive flow endpoint to get a list of `flows` and `stages` (https://matrix.org/docs/spec/client_server/latest#id186). 
  - What endpoint needs to be hit? Not clear in the spec.
  - A few sytests are dependent on this

- To actually authenticate the user, the spec says that the client should use an `m.login.token`.
  - Has this been implemented already? Couldn't find many references in the codebase
  - What kind of token is used here? How can I create this token?

Signed-off-by: Anand Vasudevan <anandv96@disroot.org>
### Pull Request Checklist

### DIFF
--- a/clientapi/auth/login_token.go
+++ b/clientapi/auth/login_token.go
@@ -1,0 +1,209 @@
+// Copyright 2020 The Matrix.org Foundation C.I.C.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package auth
+
+import (
+	"context"
+	"crypto/rand"
+	"encoding/base64"
+	"errors"
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/matrix-org/dendrite/clientapi/jsonerror"
+	"github.com/matrix-org/dendrite/clientapi/userutil"
+	"github.com/matrix-org/dendrite/internal/config"
+	"github.com/matrix-org/dendrite/userapi/api"
+	"github.com/matrix-org/gomatrixserverlib"
+	"github.com/matrix-org/util"
+)
+
+// This file handles all the m.login.token logic
+
+// GetAccountByLocalpart function implemented by the appropriate database type
+type GetAccountByLocalpart func(ctx context.Context, localpart string) (*api.Account, error)
+
+// LoginTokenRequest struct to hold the possible parameters from an m.login.token http request
+type LoginTokenRequest struct {
+	Login
+	Token string `json:"token"`
+	TxnID string `json:"txn_id"`
+}
+
+// LoginTypeToken holds the configs and the appropriate GetAccountByLocalpart function for the database
+type LoginTypeToken struct {
+	GetAccountByLocalpart GetAccountByLocalpart
+	Config                *config.ClientAPI
+}
+
+// Name returns the expected type of "m.login.token"
+func (t *LoginTypeToken) Name() string {
+	return "m.login.token"
+}
+
+// Request returns a struct of type LoginTokenRequest
+func (t *LoginTypeToken) Request() interface{} {
+	return &LoginTokenRequest{}
+}
+
+// Type of the LoginToken
+type loginToken struct {
+	UserID       string
+	CreationTime int64
+	RandomPart   string
+}
+
+// Login completes the whole token validation, user verification for m.login.token
+// returns a struct of type *auth.Login which has the users details
+func (t *LoginTypeToken) Login(ctx context.Context, req interface{}) (*Login, *util.JSONResponse) {
+	r := req.(*LoginTokenRequest)
+	userID, err := validateLoginToken(r.Token, r.TxnID, &t.Config.Matrix.ServerName)
+	if err != nil {
+		return nil, &util.JSONResponse{
+			Code: http.StatusUnauthorized,
+			JSON: jsonerror.InvalidArgumentValue(err.Error()),
+		}
+	}
+	r.Login.Identifier.User = userID
+	r.Login.Identifier.Type = "m.id.user"
+
+	return &r.Login, nil
+}
+
+// Decodes and validates a LoginToken
+// Accepts the base64 encoded token string as param
+// Checks the time expiry, userID (only the format, doesn't check to see if the user exists)
+// Also checks the DB to see if the token exists
+// Returns the localpart if successful
+func validateLoginToken(tokenStr string, txnID string, serverName *gomatrixserverlib.ServerName) (string, error) {
+	token, err := decodeLoginToken(tokenStr)
+	if err != nil {
+		return "", err
+	}
+
+	// check whether the token has a valid time.
+	// TODO: should this 5 second window be configurable?
+	if time.Now().Unix()-token.CreationTime > 5 {
+		return "", errors.New("Token has expired")
+	}
+
+	// check whether the UserID is malformed
+	if !strings.Contains(token.UserID, "@") {
+		// TODO: should we reveal details about the error with the token or give vague responses instead?
+		return "", errors.New("Invalid UserID")
+	}
+	if _, err := userutil.ParseUsernameParam(token.UserID, serverName); err != nil {
+		return "", err
+	}
+
+	// check in the database
+	if err := checkDBToken(tokenStr, txnID); err != nil {
+		return "", err
+	}
+
+	return token.UserID, nil
+}
+
+// GenerateLoginToken creates a login token which is a base64 encoded string of (userID+time+random)
+// returns an error if it cannot create a random string
+func GenerateLoginToken(userID string) (string, error) {
+	// the time of token creation
+	timePart := []byte(strconv.FormatInt(time.Now().Unix(), 10))
+
+	// the random part of the token
+	randPart := make([]byte, 10)
+	if _, err := rand.Read(randPart); err != nil {
+		return "", err
+	}
+
+	// url-safe no padding
+	return base64.RawURLEncoding.EncodeToString([]byte(userID)) + "." + base64.RawURLEncoding.EncodeToString(timePart) + "." + base64.RawURLEncoding.EncodeToString(randPart), nil
+}
+
+// Decodes the given tokenStr into a LoginToken struct
+func decodeLoginToken(tokenStr string) (*loginToken, error) {
+	// split the string into it's constituent parts
+	strParts := strings.Split(tokenStr, ".")
+	if len(strParts) != 3 {
+		return nil, errors.New("Malformed token string")
+	}
+
+	var token loginToken
+	// decode each of the strParts
+	userBytes, err := base64.RawURLEncoding.DecodeString(strParts[0])
+	if err != nil {
+		return nil, errors.New("Invalid user ID")
+	}
+	token.UserID = string(userBytes)
+
+	// first decode the time to a string
+	timeBytes, err := base64.RawURLEncoding.DecodeString(strParts[1])
+	if err != nil {
+		return nil, errors.New("Invalid creation time")
+	}
+	// now convert the string to an integer
+	creationTime, err := strconv.ParseInt(string(timeBytes), 10, 64)
+	if err != nil {
+		return nil, errors.New("Invalid creation time")
+	}
+	token.CreationTime = creationTime
+
+	randomBytes, err := base64.RawURLEncoding.DecodeString(strParts[2])
+	if err != nil {
+		return nil, errors.New("Invalid random part")
+	}
+	token.UserID = string(randomBytes)
+
+	token = loginToken{
+		UserID:       string(userBytes),
+		CreationTime: creationTime,
+		RandomPart:   string(randomBytes),
+	}
+	return &token, nil
+}
+
+// Checks whether the token exists in the DB and whether the token is assigned to the current transaction ID
+// Does not validate the userID or the creation time expiry
+// Returns nil if successful
+func checkDBToken(tokenStr string, txnID string) error {
+	// if the client has provided a transaction id, try to lock the token to that ID
+	if txnID != "" {
+		if err := LinkToken(tokenStr, txnID); err != nil {
+			// TODO: should we abort the login attempt or something else?
+		}
+	}
+	return nil
+}
+
+// StoreLoginToken stores the login token in the database
+// Returns nil if successful
+func StoreLoginToken(tokenStr string) error {
+	return nil
+}
+
+// DeleteLoginToken Deletes a token from the DB
+// used to delete a token that has already been used
+// Returns nil if successful
+func DeleteLoginToken(tokenStr string) error {
+	return nil
+}
+
+// LinkToken Links a token to a transaction ID so no other client can try to login using that token
+// as specified in https://matrix.org/docs/spec/client_server/r0.6.1#token-based
+func LinkToken(tokenStr string, txnID string) error {
+	return nil
+}
--- a/clientapi/routing/login.go
+++ b/clientapi/routing/login.go
@@ -84,10 +84,9 @@ func Login(
 		// TODO: is the the right way to read the body and re-add it?
 		body, err := ioutil.ReadAll(req.Body)
 		if err != nil {
-			// TODO: is this appropriate?
 			return util.JSONResponse{
-				Code: http.StatusMethodNotAllowed,
-				JSON: jsonerror.NotFound("Bad method"),
+				Code: http.StatusBadRequest,
+				JSON: jsonerror.BadJSON("Bad JSON"),
 			}
 		}
 		// add the body back to the request because ioutil.ReadAll consumes the body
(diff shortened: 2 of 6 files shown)


Kegsay:
> The spec mentions that a client should hit a User interactive flow endpoint to get a list of flows and stages - What endpoint needs to be hit?

In this case the `/register` endpoint. The term "user interactive auth" is used to describe the set of JSON responses returned from endpoints which implement it for authentication, so anything which requires special auth (registration, login, deleting devices, changing passwords) can use "user interactive auth" to provide a standardised way to authenticate these operations.


> To actually authenticate the user, the spec says that the client should use an m.login.token.

We do not have token auth yet - see https://github.com/matrix-org/dendrite/issues/403

anandv96:
> > To actually authenticate the user, the spec says that the client should use an m.login.token.
> 
> We do not have token auth yet - see #403

#429 for `m.login.token` was closed without without merging, I'm not sure why though. 

Should I work on #403 now before finishing #1297 ? Can #429 be modified and used?

Kegsay:
Yes work on #403 first I think. The problem with #429 is that it used macaroons rather than just a random string like we do for many other things. You may be able to take bits of it and use it successfully though.

anandv96:
> The problem with #429 is that it used macaroons rather than just a random string like we do for many other things

Looks like the spec recommends that we use macaroons. https://matrix.org/docs/spec/client_server/latest#id575

They also say that we can implement access tokens in any way we want. https://matrix.org/docs/spec/client_server/latest#id180

Should I go ahead and use random string tokens for the login token?

neilalexander:
Yes, I would use random strings for now.

clientapi/routing/login.go
```diff
@@ -53,40 +56,120 @@ func passwordLogin() flows {
...
 	} else if req.Method == http.MethodPost {
-		typePassword := auth.LoginTypePassword{
-			GetAccountByPassword: accountDB.GetAccountByPassword,
-			Config:               cfg,
+		// TODO: is the the right way to read the body and re-add it?
+		body, err := ioutil.ReadAll(req.Body)
+		if err != nil {
+			// TODO: is this appropriate?
+			return util.JSONResponse{
+				Code: http.StatusMethodNotAllowed,
```
Kegsay, line 89: The method is fine, it's just they didn't give an HTTP body.

clientapi/routing/login.go
```diff
@@ -53,40 +56,120 @@ func passwordLogin() flows {
...
-		r := typePassword.Request()
-		resErr := httputil.UnmarshalJSONRequest(req, r)
-		if resErr != nil {
-			return *resErr
+		// add the body back to the request because ioutil.ReadAll consumes the body
+		req.Body = ioutil.NopCloser(bytes.NewBuffer(body))
+
+		// marshall the body into an unstructured json map
+		var jsonBody map[string]interface{}
+		if err := json.Unmarshal([]byte(body), &jsonBody); err != nil {
```
Kegsay, line 98: Use `json.NewDecoder(req.Body).Decode(&jsonBody)` instead of going to `body`.

clientapi/routing/login.go
```diff
@@ -53,40 +56,120 @@ func passwordLogin() flows {
...
+			return util.JSONResponse{
+				Code: http.StatusMethodNotAllowed,
+				JSON: jsonerror.NotFound("Bad method"),
+			}
 		}
-		login, authErr := typePassword.Login(req.Context(), r)
-		if authErr != nil {
-			return *authErr
+
+		loginType := jsonBody["type"].(string)
```
Kegsay, line 105: Won't this panic if the request body has no `type` key?

clientapi/routing/login.go
```diff
@@ -53,40 +56,120 @@ func passwordLogin() flows {
...
+		return *authErr
+	}
+
+	// make a device/access token
+	authResult := completeAuth(req.Context(), cfg.Matrix.ServerName, userAPI, login)
+
+	// the login is successful, delete the login token before returning the access token to the client
+	if authResult.Code == http.StatusOK {
+		if err := auth.DeleteLoginToken(r.(*auth.LoginTokenRequest).Token); err != nil {
+			// TODO: what to do here?
```
Kegsay, line 167: Just log it and continue. This is just to cleanup.

clientapi/routing/sso.go
```diff
@@ -0,0 +1,218 @@
...
+func SSORedirect(
+	req *http.Request,
+	accountDB accounts.Database,
+	cfg *config.ClientAPI,
+) util.JSONResponse {
+	// If dendrite is not configured to use SSO by the admin return bad method
+	if !cfg.CAS.Enabled || cfg.CAS.Server == "" {
+		return util.JSONResponse{
+			Code: http.StatusMethodNotAllowed,
+			JSON: jsonerror.NotFound("Bad method"),
```
Kegsay, line 53: There's nothing wrong with the HTTP method here, the code is wrong. Give a 501 Not Implemented or something.

clientapi/routing/sso.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+	}
+
+	// Try to parse the SSO URL configured to a url.URL type
+	ssoURL, err := url.Parse(cfg.CAS.Server)
```
Kegsay, line 58: You could do this on startup rather than at request time.

clientapi/routing/sso.go
```diff
@@ -0,0 +1,218 @@
...
+		return util.JSONResponse{
+			Code: http.StatusBadRequest,
+			JSON: jsonerror.MissingArgument("redirectUrl parameter missing"),
+		}
+	}
+	// Check if the redirect url is a valid URL
+	redirectURL, err := url.Parse(redirectURLStr)
+	if err != nil {
+		return util.JSONResponse{
+			Code: http.StatusInternalServerError,
```
Kegsay, line 78: This isn't a server error if we fail to parse the URL, it's the client who supplies the URL. Perhaps a 400 Bad Request would be more appropriate.

clientapi/routing/sso.go
```diff
@@ -0,0 +1,218 @@
...
+// - calls validateTicket to validate the ticket
+// - calls completeSSOAuth
+func ssoTicket(
+	req *http.Request,
+	redirectURL *url.URL,
+	accountDB accounts.Database,
+	cfg *config.ClientAPI,
+) util.JSONResponse {
+	// form the ticket validation URL from the config
+	ssoURL, err := url.Parse(cfg.CAS.Server + cfg.CAS.ValidateEndpoint)
```
Kegsay, line 109: This can be done at startup.

clientapi/routing/sso.go
```diff
@@ -0,0 +1,218 @@
...
+	redirectURL *url.URL,
+	accountDB accounts.Database,
+) util.JSONResponse {
+	// try to create an account with that username
+	// if the user exists, then we pick that user, else we create a new user
+	account, err := accountDB.CreateAccount(req.Context(), username, "", "")
+	if err != nil {
+		// some error
+		if err != sqlutil.ErrUserExists {
+			return util.JSONResponse{
```
Kegsay, line 187: Please log `err` at the very least!

Kegsay (commented):
Looking roughly correct, would be good to get sytests passing.

clientapi/routing/sso.go
```diff
@@ -0,0 +1,218 @@
+// Copyright 2017 Vector Creations Ltd
```
Half-Shot, line 1: Copyright notice is wrong here.

clientapi/routing/login.go
anandv96: I tried this, but then I had to do `req.Body = ioutil.NopCloser(bytes.NewBuffer(body))` twice because the body is consumed when reading from it. Is that fine or is there something else that I'm missing?

clientapi/routing/login.go
anandv96: Yes it did. I hadn't checked for that earlier

clientapi/routing/sso.go
anandv96: You're right. I changed this and a few other http code returns as well to something I thought were more appropriate

anandv96:
Made another commit with some of the changes from the comments.

I need some help with making database changes in `clientapi/auth/login_token.go` L179 onwards
https://github.com/matrix-org/dendrite/pull/1374/files#diff-1f4ba2edfc578ed53786cf30d3148eedR179

anandv96:
I changed `lib/SyTest/Homeserver/Dendrite.pm` in sytest for the test `login types include SSO` to pass. Should I wait until all the other tests pass before I make the change in the sytest repo?

clientapi/routing/sso.go
```diff
@@ -0,0 +1,203 @@
...
+	// If the request has a ticket param, validate the ticket instead of redirecting to SSO server
+	if ticket := req.FormValue("ticket"); ticket != "" {
+		return ssoTicket(req, redirectURL, accountDB, cfg)
+	}
+
+	// Adding the params to the sso url
+	ssoURL := cfg.CAS.URL
+	ssoQueries := make(url.Values)
+	// the service url that we send to CAS is homeserver.com/_matrix/client/r0/login/sso/redirect?redirectUrl=xyz
+	ssoQueries.Set("service", req.RequestURI)
```
clokep, line 83: I suspect it doesn't matter too much, but it could be clearer to redirect to a different endpoint instead of using `/login/sso/redirect` as both the redirect endpoint and the endpoint that resolves the ticket. I'm not sure if this is common in CAS implementations or not. Synapse uses a separate endpoint, but it is also under the `/_matrix` prefix, which it should *not* be.

anandv96: Yeah, I thought about this too and in the older version of the spec ([r0.4.0](https://matrix.org/docs/spec/client_server/r0.4.0#cas-based-client-login)) there are 2 endpoints specified for CAS auth (`/login/cas/redirect` and `/login/cas/ticket`). But the newer spec ([r0.6.1](https://matrix.org/docs/spec/client_server/r0.6.1#sso-client-login)) doesn't mention the separate `/ticket` endpoint, so I thought I should just go with the single endpoint. 

Splitting it up into 2 endpoints can be easily done. 

About it not being in the `/_matrix` prefix, I don't know enough about common practices in other CAS implementations to comment. The spec does say `/_matrix/client/r0/login/sso/redirect` though.

clokep: > Yeah, I thought about this too and in the older version of the spec ([r0.4.0](https://matrix.org/docs/spec/client_server/r0.4.0#cas-based-client-login)) there are 2 endpoints specified for CAS auth (`/login/cas/redirect` and `/login/cas/ticket`). But the newer spec ([r0.6.1](https://matrix.org/docs/spec/client_server/r0.6.1#sso-client-login)) doesn't mention the separate `/ticket` endpoint, so I thought I should just go with the single endpoint.

I believe the second endpoint is meant to be implementation specific!

> Splitting it up into 2 endpoints can be easily done.

I'd let @Kegsay or @neilalexander decide whether this should be done or not. 😄 

> About it not being in the `/_matrix` prefix, I don't know enough about common practices in other CAS implementations to comment. The spec does say `/_matrix/client/r0/login/sso/redirect` though.

Sorry for not being clear, I meant the ticket endpoint. The redirect endpoint must be  `/_matrix/client/r0/login/sso/redirect` as you noted.

anandv96: > I'd let @Kegsay or @neilalexander decide whether this should be done or not

Yeah I agree :sweat_smile: 

> I meant the ticket endpoint

The ticket endpoint is `/_matrix/client/r0/login/cas/ticket` according to the [old spec](https://matrix.org/docs/spec/client_server/r0.4.0#get-matrix-client-r0-login-cas-ticket), that could explain why synapse has it under `/_matrix` as well.

clientapi/auth/login_token.go
```diff
@@ -0,0 +1,209 @@
...
+// Returns the localpart if successful
+func validateLoginToken(tokenStr string, txnID string, serverName *gomatrixserverlib.ServerName) (string, error) {
+	token, err := decodeLoginToken(tokenStr)
+	if err != nil {
+		return "", err
+	}
+
+	// check whether the token has a valid time.
+	// TODO: should this 5 second window be configurable?
+	if time.Now().Unix()-token.CreationTime > 5 {
```
Kegsay, line 100: 5 seconds seems very short, perhaps 1 minute?

clientapi/auth/login_token.go
```diff
@@ -0,0 +1,209 @@
...
+		return "", errors.New("Token has expired")
+	}
+
+	// check whether the UserID is malformed
+	if !strings.Contains(token.UserID, "@") {
+		// TODO: should we reveal details about the error with the token or give vague responses instead?
```
Kegsay, line 106: Revealing details is fine for now.

clientapi/auth/login_token.go
```diff
@@ -0,0 +1,209 @@
...
+		return "", errors.New("Invalid UserID")
+	}
+	if _, err := userutil.ParseUsernameParam(token.UserID, serverName); err != nil {
```
Kegsay, line 109: Please add some prose here to explain why this check alone is insufficient (it doesn't check there is an `@`).

clientapi/routing/login.go
```diff
@@ -53,40 +56,127 @@ func passwordLogin() flows {
...
 		return util.JSONResponse{
 			Code: http.StatusOK,
-			JSON: passwordLogin(),
+			JSON: flows,
 		}
 	} else if req.Method == http.MethodPost {
-		typePassword := auth.LoginTypePassword{
-			GetAccountByPassword: accountDB.GetAccountByPassword,
-			Config:               cfg,
+		// TODO: is the the right way to read the body and re-add it?
```
Kegsay, line 84: Yes

clientapi/routing/sso.go
Kegsay: 2 endpoints please, both under `/_matrix` will do. If we don't put it under `/_matrix` then we need to mess around with routers more, and the only reason why we don't want it there is "just in case" the spec suddenly uses that endpoint for something, which feels unlikely.

clientapi/routing/sso.go
```diff
@@ -0,0 +1,203 @@
...
+}
+
+// validateTicket sends the ticket to the sso server to get it validated
+// the CAS server responds with an xml which contains the username
+// validateTicket returns the SSO User
+func validateTicket(
+	ssoURL string,
+) (string, error) {
+	// make the call to the sso server to validate
+	response, err := http.Get(ssoURL)
```
Kegsay, line 137: Please use an HTTP client here, as the default `http.Get` has [no timeout](https://medium.com/@nate510/don-t-use-go-s-default-http-client-4804cb19f779).

clientapi/routing/sso.go
```diff
@@ -0,0 +1,203 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	// extract the response from the sso server
+	data, err := ioutil.ReadAll(response.Body)
```
Kegsay, line 143: Don't forget to `defer response.Body.Close()`

clientapi/routing/sso.go
```diff
@@ -0,0 +1,203 @@
...
+				return util.JSONResponse{
+					Code: http.StatusInternalServerError,
+					JSON: jsonerror.Unknown("Could not query user"),
+				}
+			}
+		}
+	}
+	token, err := auth.GenerateLoginToken(account.UserID)
+	if err != nil || token == "" {
+		return util.JSONResponse{
```
Kegsay, line 189: Please do not drop `err` to the floor, log it.

Kegsay (changes requested):
There's a bunch of sytests here https://github.com/matrix-org/dendrite/issues/1297 - some of them should be passing by now right? Please update `sytest-whitelist`.

clientapi/auth/login_token.go
anandv96: I only choose 5 seconds because the spec suggested it: https://matrix.org/docs/spec/client_server/r0.6.1#handling-the-authentication-endpoint

Kegsay: Oh wow I see. In which case please just link to this on this line to justify why 5s.

clokep: 5 seconds sounds pretty short to me as well, if you have a connection or DNS failure that could easily time out. It seems that Synapse uses 2 minutes, although I was unable to find any reference to why that was chosen: https://github.com/matrix-org/synapse/blob/78d5f91de1a9baf4dbb0a794cb49a799f29f7a38/synapse/handlers/auth.py#L1323-L1325

anandv96: In that case, should I make this time configurable by the admin (with a default time of 5 seconds)?

Kegsay:
This PR will be closed on 23/12/2020 without further input.

anandv96:
> This PR will be closed on 23/12/2020 without further input.

Sorry. I got busy with other work, I will start working on this again this week (have time off at the end of the year)

Kegsay:
Closing this for now. Please feel free to re-open/ping when you get back to this.

tommie:
@anandv96 Can I take your code and fork this PR?

anandv96:
> @anandv96 Can I take your code and fork this PR?

Yes of course! There isn't **too** much left to do IIRC

tommie:
> Yes of course! There isn't too much left to do IIRC

Thanks! Yeah, and the fact that you've already had review rounds is a strong motivator to not start over. :)

PR #1376, opened 2020-09-01
Fix duplicate writers

neilalexander:
This fixes a couple of places where we had duplicate `ExclusiveWriter`s, causing `database is locked` errors.


Kegsay (approved):
🚢

PR #1379, opened 2020-09-02
Use federation sender for backfill/getting missing events

neilalexander:
This is an attempt to use the federation sender for backfill/getting missing events rather than using a federation client directly.


Kegsay (approved):
Looks fab, thanks!

PR #1382, opened 2020-09-02
Add Queryer and Inputer and factor out more RSAPI stuff

Kegsay:
This neatly splits up the RS API based on the functionality it provides,
whilst providing a useful place for code sharing via the `helpers` package.

This reduces the overall internal surface of RSAPI considerably. 

What remains to be done:
 - There's a lot of "helper" functions which aren't exactly optimised, and grown organically. Helpers are used in at least two distinct RSAPI areas (e.g query and perform), and they tend to be wrappers around storage API calls to map between different types.

### DIFF
--- a/roomserver/internal/api.go
+++ b/roomserver/internal/api.go
@@ -10,12 +10,19 @@ import (
 	"github.com/matrix-org/dendrite/internal/config"
 	"github.com/matrix-org/dendrite/roomserver/api"
 	"github.com/matrix-org/dendrite/roomserver/internal/perform"
+	"github.com/matrix-org/dendrite/roomserver/internal/query"
 	"github.com/matrix-org/dendrite/roomserver/storage"
 	"github.com/matrix-org/gomatrixserverlib"
 )
 
 // RoomserverInternalAPI is an implementation of api.RoomserverInternalAPI
 type RoomserverInternalAPI struct {
+	*query.Queryer
+	*perform.Inviter
+	*perform.Joiner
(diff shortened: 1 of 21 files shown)


roomserver/internal/api.go
```diff
@@ -2,20 +2,28 @@ package internal
...
 	"github.com/matrix-org/dendrite/roomserver/internal/perform"
+	"github.com/matrix-org/dendrite/roomserver/internal/query"
 	"github.com/matrix-org/dendrite/roomserver/storage"
 	"github.com/matrix-org/gomatrixserverlib"
 )
 
 // RoomserverInternalAPI is an implementation of api.RoomserverInternalAPI
 type RoomserverInternalAPI struct {
+	*input.Inputer
+	*query.Queryer
```
neilalexander, line 21: The names sit a bit funny with me at the moment, not least because of `Inputer` and `Queryer`. Maybe we can just stick with `Inputs`, `Queries`, `Invites`, `Joins` etc.

neilalexander (commented):
I like the direction generally, just worried about naming.

roomserver/internal/api.go
Kegsay: I want to interface them actually so the -er suffix makes more sense then, which would allow us to test parts of the RSAPI more easily. Happy to have the concrete impl be `Inputs` `Queries` etc.

Kegsay: Merging for now, will resolve later down the line.

PR #1386, opened 2020-09-03
FIFO ordering of input events

neilalexander:
This guarantees that input events into the roomserver (per room) are FIFO'd properly, by adding per-room workers. 

This also removes `EventID` from `api.InputRoomEventsResponse` since it is practically useless.

### DIFF
--- a/roomserver/internal/input/input.go
+++ b/roomserver/internal/input/input.go
@@ -40,7 +40,7 @@ type Inputer struct {
 
 type inputTask struct {
 	ctx   context.Context
-	event api.InputRoomEvent
+	event *api.InputRoomEvent
 	wg    *sync.WaitGroup
 	err   error // written back by worker, only safe to read when all tasks are done
 }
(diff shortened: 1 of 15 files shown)


Kegsay (approved):
Reads nicely!

roomserver/internal/input/input.go
```diff
@@ -74,18 +105,51 @@ func (r *Inputer) InputRoomEvents(
...
+		// the wait group, so that the worker can notify us when this specific
+		// task has been finished.
+		tasks[i] = &inputTask{
+			ctx:   ctx,
+			event: e,
+			wg:    wg,
+		}
+
+		// Send the task to the worker.
+		go func(task *inputTask) { worker.input <- task }(tasks[i])
```
Kegsay, line 140: You lose input ordering by doing this in a goroutine. There's no guarantee that the goroutine will pass this task into `worker.input` before the next loop

PR #1387, opened 2020-09-03
Move currentstateserver API to roomserver

Kegsay:
And implement some of the current state server storage interface in roomserver.

Currently nothing calls the new `Query` functions on RSAPI so tests etc still pass. Subsequent PRs will:
 - Implement remaining storage functions in roomserver.
 - Point everything which calls CSS to call RS instead.
 - Remove CSS.

### DIFF
--- a/roomserver/storage/shared/storage.go
+++ b/roomserver/storage/shared/storage.go
@@ -748,5 +748,5 @@ func (d *Database) GetKnownUsers(ctx context.Context, userID, searchString strin
 
 // GetKnownRooms returns a list of all rooms we know about.
 func (d *Database) GetKnownRooms(ctx context.Context) ([]string, error) {
-	return nil, fmt.Errorf("not implemented yet")
+	return nil, nil
 }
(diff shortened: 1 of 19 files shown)


roomserver/storage/shared/storage.go
```diff
@@ -711,3 +712,82 @@ func (d *Database) loadEvent(ctx context.Context, eventID string) *types.Event {
 	}
 	return &evs[0]
 }
+
+// GetStateEvent returns the current state event of a given type for a given room with a given state key
+// If no event could be found, returns nil
+// If there was an issue during the retrieval, returns an error
+func (d *Database) GetStateEvent(ctx context.Context, roomID, evType, stateKey string) (*gomatrixserverlib.HeaderedEvent, error) {
+	/*
```
neilalexander, line 720: Is this meant to be commented out?

neilalexander (approved):

roomserver/storage/shared/storage.go
Kegsay: Yes

PR #1390, opened 2020-09-03
Remove ServerACLs from the current state server

Kegsay:
Functionality moved to roomserver


roomserver/acls/acls.go
```diff
@@ -87,6 +88,7 @@ func compileACLRegex(orig string) (*regexp.Regexp, error) {
 }
 
 func (s *ServerACLs) OnServerACLUpdate(state *gomatrixserverlib.Event) {
+	util.GetLogger(context.TODO()).Infof("banana OnServerACLUpdate :D %s", string(state.Content()))
```
neilalexander, line 91: Should we remove these loglines? :D

neilalexander (approved):

roomserver/acls/acls.go
Kegsay: But but they're so happy... :P

PR #1391, opened 2020-09-04
Peeking over federation via MSC2444

ara4n:
First wave of peeking over federation a la MSC2444 to see how it feels.

I think this is now good to be merged as an initial wave of work, to stop it bitrotting, even though all the housekeeping is missing.

* [x] make outbound /peek over SS API
* [x] track outbound peeks in the federationsender db
* [x] handle inbound /peek requests over SS API
* [x] track inbound peeks in the federationsender db
* [x] make it build
* [x] make it work
* [x] support postgres
* [ ] validate the state received by /peek correctly
* [ ] fix the race between /peek and /send (which is causing the tests to be flakey)
* [x] don't re-peek over federation if we're already peeking
* [ ] handle initial history visibility
* [ ] handle history visibility if it transitions mid-peek
* [ ] handle inbound peek renewals
* [ ] expire stale inbound peeks
* [ ] handle the transition from peeking into a remote room to joining a remote room nicely
* [ ] housekeeping to renew outbound peeks automatically if there is a client syncing them
* [ ] remove remote peeks if all the local clients /unpeek
* [ ] decide if/how to unify the 4 near-identical `{in,out}bound-peeks-table`s
* [x] sytest - https://github.com/matrix-org/sytest/pull/953

Implements https://github.com/matrix-org/matrix-doc/pull/2444
Builds on https://github.com/matrix-org/dendrite/pull/1370
Requires https://github.com/matrix-org/gomatrixserverlib/pull/220


ara4n:
@neilalexander can I please get a quick review at this point before I flesh out all the edge cases?

federationsender/internal/perform.go
```diff
@@ -210,6 +211,161 @@ func (r *FederationSenderInternalAPI) performJoinUsingServer(
...
+	} else {
+		if err = r.db.AddOutboundPeek(ctx, serverName, roomID, peekID, respPeek.RenewalInterval); err != nil {
+			return err
+		}
+	}
+
+	respState := respPeek.ToRespState()
+	// logrus.Warnf("converted respPeek %#v to respState %#v", respPeek, respState)
+	// Send the newly returned state to the roomserver to update our local view.
+	if err = roomserverAPI.SendEventWithState(
```
neilalexander, line 357: I think we probably want to use `SendEventWithRewrite` here - that way the state snapshots should be generated properly for the state events too. You should be able to pass in `nil` for the `haveEventIDs` parameter and it should still do the right thing.

federationsender/storage/interface.go
```diff
@@ -50,7 +50,18 @@ type Database interface {
 	GetPendingPDUServerNames(ctx context.Context) ([]gomatrixserverlib.ServerName, error)
 	GetPendingEDUServerNames(ctx context.Context) ([]gomatrixserverlib.ServerName, error)
 
+	// XXX: why don't these have contexts passed in?
```
neilalexander, line 53: Largely because we want these things to happen regardless of what happens to the request context, so they instead generate their own contexts.

federationsender/storage/sqlite3/inbound_peeks_table.go
```diff
@@ -0,0 +1,176 @@
...
+	if s.deleteInboundPeekStmt, err = db.Prepare(deleteInboundPeekSQL); err != nil {
+		return
+	}
+	return
+}
+
+func (s *inboundPeeksStatements) InsertInboundPeek(
+	ctx context.Context, txn *sql.Tx, serverName gomatrixserverlib.ServerName, roomID, peekID string, renewalInterval int64,
+) (err error) {
+	nowMilli := time.Now().UnixNano() / int64(time.Millisecond)
```
neilalexander, line 101: Not that it changes much functionally but the idiomatic way to do this would be `gomatrixserverlib.AsTimestamp(time.Now())`, since `gomatrixserverlib.Timestamp`s are always expected to be millisecond-precision.

go.mod
```diff
@@ -1,5 +1,7 @@
 module github.com/matrix-org/dendrite
 
+replace github.com/matrix-org/gomatrixserverlib => ../gomatrixserverlib
```
neilalexander, line 3: This will break CI :-)

roomserver/api/wrapper.go
```diff
@@ -42,10 +42,11 @@ func SendEvents(
 
 // SendEventWithState writes an event with KindNew to the roomserver
 // with the state at the event as KindOutlier before it. Will not send any event that is
-// marked as `true` in haveEventIDs
+// marked as `true` in haveEventIDs.
 func SendEventWithState(
 	ctx context.Context, rsAPI RoomserverInternalAPI, state *gomatrixserverlib.RespState,
 	event gomatrixserverlib.HeaderedEvent, haveEventIDs map[string]bool,
+	roomVersion gomatrixserverlib.RoomVersion,
```
neilalexander, line 49: What's the reason for this change? The `event` (as a `HeaderedEvent`) should contain the room version already, and by being in the same room, all of the outliers are going to be the same room version as the title event.

roomserver/internal/perform/perform_inbound_peek.go
```diff
@@ -0,0 +1,122 @@
...
+		r.DB.LatestEventIDs(ctx, info.RoomNID)
+	if err != nil {
+		return err
+	}
+	// XXX: is this actually the latest of the latest events?
+	latestEvents, err := r.DB.EventsFromIDs(ctx, []string{latestEventRefs[0].EventID})
+	if err != nil {
+		return err
+	}
+	response.LatestEvent = latestEvents[0].Headered(info.RoomVersion)
```
neilalexander, line 72: You could potentially `gomatrixserverlib.ReverseTopologicalOrdering(latestEvents, gomatrixserverlib.TopologicalOrderByPrevEvents)` to try and deterministically find what looks to be the "newest" event.

roomserver/internal/perform/perform_peek.go
```diff
@@ -182,8 +198,6 @@ func (r *Peeker) performPeekRoomByID(
 		}
 	}
 
-	// TODO: handle federated peeks
```
neilalexander, line 185: I think probably the important thing here is to make sure that we don't attempt a federated peek if someone on the HS is already joined to the room, otherwise we risk rewriting the state unnecessarily. We should be able to trust our own local view of the room if that's the case.

neilalexander (commented):
I think the pattern overall looks OK—made some suggestions/comments below.

roomserver/api/wrapper.go
ara4n: oops, this was an accidental leftover from something

federationsender/storage/sqlite3/inbound_peeks_table.go
ara4n: hm, the UnixNano pattern pops up a /lot/ elsewhere in the codebase...

roomserver/internal/perform/perform_inbound_peek.go
ara4n: done

ara4n:
tests won't pass until https://github.com/matrix-org/gomatrixserverlib/pull/220 is mergéd

federationapi/routing/peek.go
```diff
@@ -0,0 +1,102 @@
...
+			RenewalInterval: renewalInterval,
+		},
+		&response,
+	)
+	if err != nil {
+		resErr := util.ErrorResponse(err)
+		return resErr
+	}
+
+	if !response.RoomExists {
```
Kegsay, line 86: We would've failed earlier as `QueryRoomVersionForRoom` would return a failure.

federationsender/api/api.go
```diff
@@ -111,6 +117,16 @@ type PerformJoinResponse struct {
...
 }
 
+type PerformOutboundPeekRequest struct {
+	RoomID string `json:"room_id"`
+	// The sorted list of servers to try. Servers will be tried sequentially, after de-duplication.
+	ServerNames types.ServerNames `json:"server_names"`
+}
+
+type PerformOutboundPeekResponse struct {
+	LastError *gomatrix.HTTPError
```
Kegsay, line 127: Shouldn't we be getting `gomatrixserverlib.RespPeek` somewhere?

roomserver/internal/perform/perform_inbound_peek.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+type InboundPeeker struct {
+	DB      storage.Database
+	Inputer *input.Inputer
+}
+
+// PerformInboundPeek handles peeking into matrix rooms, including over
+// federation by talking to the federationsender. called when a remote server
+// initiates a /peek over federation.
```
Kegsay, line 38: Feels bad that we don't call `PerformInboundPeek` when local clients call `/peek` like we do with all the other Perform stuff.

federationsender/internal/perform.go
```diff
@@ -266,6 +266,172 @@ func (r *FederationSenderInternalAPI) performJoinUsingServer(
...
+	ctx context.Context,
+	roomID string,
+	serverName gomatrixserverlib.ServerName,
+	supportedVersions []gomatrixserverlib.RoomVersion,
+) error {
+	// create a unique ID for this peek.
+	// for now we just use the room ID again. In future, if we ever
+	// support concurrent peeks to the same room with different filters
+	// then we would need to disambiguate further.
+	peekID := roomID
```
Kegsay, line 360: Annoyingly the MSC forbids this because:

> should be a string consisting of the characters [0-9a-zA-Z.=_-]. Its length must not exceed 8 characters and it should not be empty.

federationsender/types/types.go
```diff
@@ -49,3 +49,23 @@ func (e EventIDMismatchError) Error() string {
...
 	)
 }
+
+// tracks peeks we're performing on another server over federation
+type OutboundPeek struct {
+	PeekID            string
+	RoomID            string
+	ServerName        gomatrixserverlib.ServerName
+	CreationTimestamp int64
+	RenewedTimestamp  int64
```
Kegsay, line 59: As N points out, it would be nicer for this to be `gomatrixserverlib.Timestamp` types (which are just ints)

roomserver/internal/perform/perform_peek.go
```diff
@@ -151,11 +151,28 @@ func (r *Peeker) performPeekRoomByID(
 		}
 	}
 
-	// If the server name in the room ID isn't ours then it's a
-	// possible candidate for finding the room via federation. Add
-	// it to the list of servers to try.
+	// handle federated peeks
+	// FIXME: don't create an outbound peek if we already have one going.
 	if domain != r.Cfg.Matrix.ServerName {
```
Kegsay, line 156: This won't work in cases where you create the room, other servers join then you leave the room then try to peek into it, as this will resolve to `false`. We shouldn't ever be switching on the domain of the room ID like this.

docs/peeking.md
```diff
@@ -1,19 +1,26 @@
...
+   * If not, it hits the PerformInboundPeek on the roomserver to ask it for the current state of the room.
+   * The roomserver atomically (in theory) adds a NewInboundPeek to its kafka stream to tell the federationserver to start peeking.
+   * The federationsender receives the event, tracks the inbound peek in the federationsender_inbound_peeks table, and starts sending events to the peeking server.
+   * The federationsender evicts stale inbound peeks which haven't been renewed.
+
+For peeking into other server's rooms ("outbound peeks"):
+ * The `roomserver` will kick the `federationsender` much as it does for a federated `/join` in order to trigger a federated outbound `/peek`
+ * The `federationsender` tracks the existence of the outbound peek in in its federationsender_outbound_peeks table.
  * The `federationsender` regularly renews the remote peek as long as there are still peeking devices syncing for it.
   * TBD: how do we tell if there are no devices currently syncing for a given peeked room?  The syncserver needs to tell the roomserver
```
Kegsay, line 25: Why do we care? We time out the peek anyway, and really just because there are no devices _currently_ syncing doesn't mean we should stop peeking. Network interruptions are a reasonable reason, we should still be honouring the renewal period.

PR #1394, opened 2020-09-04
Sync bug fixes

neilalexander:
This fixes two things:

1. The state deltas could incorrectly overwrite a room that was `"join"` with a different membership state from the past.
2. Retired invites could cause a room to appear in the `"leave"` section even though we are joined to it.

### DIFF
--- a/syncapi/storage/shared/syncserver.go
+++ b/syncapi/storage/shared/syncserver.go
@@ -743,8 +743,10 @@ func (d *Database) addInvitesToResponse(
 		res.Rooms.Invite[roomID] = *ir
 	}
 	for roomID := range retiredInvites {
-		lr := types.NewLeaveResponse()
-		res.Rooms.Leave[roomID] = *lr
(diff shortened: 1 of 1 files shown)


Kegsay (approved):
Assuming CI is happy. Please please please add complement tests for this, or at least tell me what HTTP requests I need to do and I'll do it.

neilalexander:
Writing this here for whoever gets to writing the test first:

The first bug happens when you:
1. User A `/join` public room
2. User A `/leave` the room
3. User A `/join` the room again
4. User A does an *incremental sync* with `"since"` set to before the first join
5. User A should not see the room ID appear in `/sync` `"leave"` section

The second bug happens when you:
1. User A `/join` public room
2. User B `/invite` user B
3. User B `/leave` to reject the invite from user A
4. User B does a *full sync*
5. User B should see the room ID appear in `/sync` `"leave"` section
6. User B `/join` the room
7. User B does a *full sync*
8. User B should not see the room ID appear in `/sync` `"leave"` section

Issue #407, opened 2018-03-03
Account: Implement Password Change API Endpoint


anoadragon453:
**Spec URL:** https://matrix.org/docs/spec/client_server/r0.3.0.html#post-matrix-client-r0-account-password

**Phase:** 1

**Difficulty:** Medium

This endpoint allows users to reset their passwords. Learn how to authenticate with endpoints [here](https://github.com/anoadragon453/dendrite/wiki/How-to-Test-with-Authentication-to-the-Client-Server-API)!

This issue has been rated Medium due to having to authenticate via the [User-Interactive Authentication API](https://matrix.org/docs/spec/client_server/r0.3.0.html#user-interactive-authentication-api). Theoretically, as we already handle such requests in register.go: https://github.com/matrix-org/dendrite/blob/139cb7a01e1c391d0828c2d78182c39059b3a0fa/src/github.com/matrix-org/dendrite/clientapi/routing/register.go#L404-L482

One should be able to just pass on requests to the `handleRegistrationFlow` method. However, this may end up requiring some refactoring, as that method also handles registering a new user, which is not necessary in this case.

Details on how to test/reproduce this endpoint are below.

## Request

Use a REST API Client such as [RESTClient](https://addons.mozilla.org/en-US/firefox/addon/restclient/) or [Postman](https://chrome.google.com/webstore/detail/postman/fhbjgbiflinjbdggehcddcbncdddomop).

```
# POST _matrix/client/r0/account/password
Header: N/A
Body: {}
```

See the above **Spec URL** for how the server should properly respond. A `200 OK` with an empty body means the password was changed successfully.

**Summary:**

- [ ] Create a new endpoint in [routing.go](https://github.com/matrix-org/dendrite/blob/master/src/github.com/matrix-org/dendrite/clientapi/routing/routing.go) for `/account/password`.
- [ ] Implement `/account/password` in a new `password.go` file in [clientapi/routing](https://github.com/matrix-org/dendrite/tree/master/src/github.com/matrix-org/dendrite/clientapi/routing).
- [ ] Allow ability to authenticate through [User Interactive Auth API](https://matrix.org/docs/spec/client_server/r0.3.0.html#user-interactive-authentication-api) or existing token.
- [ ] Create integration test(s) for creating a user and changing their password.

MohitKS5:
can I take this?
My plan
[...]

PR #1397, opened 2020-09-04
Password changes

neilalexander:
This PR adds support for users being able to change their own password.

### DIFF
--- a/clientapi/auth/authtypes/logintypes.go
+++ b/clientapi/auth/authtypes/logintypes.go
@@ -5,6 +5,7 @@ type LoginType string
 
 // The relevant login types implemented in Dendrite
 const (
+	LoginTypePassword           = "m.login.password"
 	LoginTypeDummy              = "m.login.dummy"
 	LoginTypeSharedSecret       = "org.matrix.login.shared_secret"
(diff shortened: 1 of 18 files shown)


Kegsay (approved):
Closes https://github.com/matrix-org/dendrite/issues/407

Kegsay:
But `    × After changing password, a different session no longer works by default` doesn't pass?

neilalexander:
I don't really understand that `After changing password, a different session no longer works by default` test.

The test doesn't specify `logout_devices: true` so why would other access tokens be invalidated? The spec says:

> The homeserver SHOULD NOT revoke the access token provided in the request. Whether other access tokens for the user are revoked depends on the request parameters.

Must be a Synapse-ism.

neilalexander:
Synapse appears to default to `true` for the `logout_devices` field if it isn't specified. Dendrite doesn't.

Kegsay:
Which is correct?

neilalexander:
Looks like the spec does say about this after all:

> Whether the user's other access tokens, and their associated devices, should be revoked if the request succeeds. Defaults to true.

I've fixed Dendrite to default to `true` also.

PR #1399, opened 2020-09-04
Update GetStateEvent behaviour

neilalexander:
This updates `GetStateEvent` so that it actually matches the commented behaviour of returning `nil` if no matching state event was found but no errors happened otherwise, rather than returning an error.

This fixes the multivarious `Failed to get server ACLs` messages when Dendrite starts up.


Kegsay (approved):
Whoops..

Kegsay:
https://github.com/matrix-org/dendrite/issues/1400 for CI failure

PR #1402, opened 2020-09-05
Don't use more than 999 variables in SQLite querys.

### LINKED ISSUE #1369: Joining #matrix:matrix.org fails with "to many SQL variables"
Time="2020-08-27T18:53:06.018059362Z" level=error msg="failed to join user "@henrik:solver.nu" to room "!OGEhHVWSdvArJzumhm:matrix.org" through 1 server(s): last error joinCtx.CheckSendJoinResponse: respSendJoin: too many SQL variables" func="github.com/matrix-org/dendrite/federationsender/internal.(*FederationSenderInternalAPI).PerformJoin" file="github.com/matrix-org/dendrite@/federationsender/internal/perform.go:105"

The problem comes from.
func (s *serverKeyStatements) bulkSelectServerKeys() in serverkeyapi/storage/sqlite3/server_key_table.go

In this particular case there are 1230 id's in the request.

Similar to #1223

HenrikSolver:
Solve this problem in a more general and reusable way.
Also fix #1369

Signed-off-by: Henrik Sölver <henrik.solver@gmail.com>

### Pull Request Checklist

### DIFF
--- a/internal/sqlutil/sql.go
+++ b/internal/sqlutil/sql.go
@@ -15,10 +15,14 @@
 package sqlutil
 
 import (
+	"context"
 	"database/sql"
 	"errors"
 	"fmt"
 	"runtime"
+	"strings"
+
+	"github.com/matrix-org/util"
 )
 
 // ErrUserExists is returned if a username already exists in the database.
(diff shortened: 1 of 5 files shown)


internal/sqlutil/sql.go
```diff
@@ -107,3 +111,40 @@ func SQLiteDriverName() string {
...
+	return b
+}
+
+type QueryProvider interface {
+	QueryContext(ctx context.Context, query string, args ...interface{}) (*sql.Rows, error)
+}
+
+const SQLite3MaxVariables = 999
+
+//RunLimitedVariablesQuery split up a query with more varaibles than the used databse can handle in multiple queries.
```
Kegsay, line 128:
```suggestion
//RunLimitedVariablesQuery split up a query with more varaibles than the used database can handle in multiple queries.
```

internal/sqlutil/sql.go
```diff
@@ -107,3 +111,40 @@ func SQLiteDriverName() string {
...
+	var start int
+	for start < len(variables) {
+		n := minOfInts(len(variables)-start, SQLite3MaxVariables)
+		query := strings.Replace(query, "($1)", QueryVariadic(n), 1)
+		rows, err := qp.QueryContext(ctx, query, variables[start:start+n]...)
+		if err != nil {
+			return err
+		}
+		err = rowHandler(rows)
+		if err := rows.Close(); err != nil {
```
Kegsay, line 139: This can't be called `err` else it will replace the `err` from `rowHandler` without checking it.

internal/sqlutil/sql.go
```diff
@@ -107,3 +111,40 @@ func SQLiteDriverName() string {
...
+
+const SQLite3MaxVariables = 999
+
+//RunLimitedVariablesQuery split up a query with more varaibles than the used databse can handle in multiple queries.
+func RunLimitedVariablesQuery(ctx context.Context, query string, qp QueryProvider, rowHandler func(*sql.Rows) error, variables []interface{}) error {
+	var start int
+	for start < len(variables) {
+		n := minOfInts(len(variables)-start, SQLite3MaxVariables)
+		query := strings.Replace(query, "($1)", QueryVariadic(n), 1)
+		rows, err := qp.QueryContext(ctx, query, variables[start:start+n]...)
```
Kegsay, line 134: Given we have a nice interface for this, please can you add unit tests to check this function does what you think it should be doing and we don't have off-by-one errors somewhere.

Kegsay (changes requested):
Looks good, just a few things to make sure this is robust in the event of failure.

internal/sqlutil/sql.go
HenrikSolver: Actually, I don't think that is the case. The err from rows.Close() is in a new scope inside the if (https://play.golang.org/p/tnQq8KHKVwn), but for clarity it might be better to use another name.

internal/sqlutil/sql.go
HenrikSolver: Will do. And also the varaibles thingy :-)

internal/sqlutil/sql.go
HenrikSolver: Will do.

HenrikSolver:
Unit tests added.

Kegsay:
Added stuff to this PR in https://github.com/matrix-org/dendrite/pull/1425 and landed it, thanks!

PR #1403, opened 2020-09-07
Use background context when processing event with missing state

neilalexander:
This is another one of those "we reuse a context and it has unintended side effects" things.

In this case, it seems that if someone pushes us an event with `/send` that we need to fetch prev events for, that it can take a long time for us to do that and the original request can time out, expiring the context in the process. That means we give up and end up with things like:

`time="2020-09-04T10:40:43.939557621Z" level=error msg="Failed to lookup state after prev_event: $palcj7ro574oGLsiVwnxS6pa5LRHXXaQwq5vz4An5oM" func="github.com/matrix-org/dendrite/federationapi/routing.(*txnReq).processEventWithMissingState" error="Get \"matrix://dendrite.neilalexander.dev/_matrix/federation/v1/state/%21yomrOFwgFXzmeMAbzX:matrix.org?event_id=%24palcj7ro574oGLsiVwnxS6pa5LRHXXaQwq5vz4An5oM\": context canceled" req.id=WIibzUs43Hkx req.method=PUT req.path=/_matrix/federation/v1/send/1599208104074-2`

It's probably better for us to just wait it out and to keep trying. Then when the remote side retries, there'll be a better chance of us having fetched the missing events.

### DIFF
--- a/federationapi/routing/send.go
+++ b/federationapi/routing/send.go
@@ -369,6 +369,7 @@ func (t *txnReq) processEvent(e gomatrixserverlib.Event, isInboundTxn bool) erro
 	}
 
 	if !stateResp.PrevEventsExist {
+		t.context = context.Background()
 		return t.processEventWithMissingState(e, stateResp.RoomVersion, isInboundTxn)
 	}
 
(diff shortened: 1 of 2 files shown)


Kegsay (approved):
With a timeout please e.g 5min

Issue #1357, opened 2020-08-27
currentstateserver: Add GoDoc to internal API to say "don't use me in a consumer"


Kegsay:
Context: https://github.com/matrix-org/dendrite/issues/1272

Kegsay:
#1405

Issue #1272, opened 2020-08-14
flakey: Local device key changes get to remote servers


Kegsay:
Can fail on postgres, both `Local device key changes get to remote servers` and `Local device key changes get to remote servers with correct prev_id`. Cause: `destinations` slice is empty.

```go
	var queryRes stateapi.QueryRoomsForUserResponse
	err = t.stateAPI.QueryRoomsForUser(context.Background(), &stateapi.QueryRoomsForUserRequest{
		UserID:         m.UserID,
		WantMembership: "join",
	}, &queryRes)
	if err != nil {
		logger.WithError(err).Error("failed to calculate joined rooms for user")
		return nil
	}
	// send this key change to all servers who share rooms with this user.
	destinations, err := t.db.GetJoinedHostsForRooms(context.Background(), queryRes.RoomIDs)
	if err != nil {
		logger.WithError(err).Error("failed to calculate joined hosts for rooms user is in")
		return nil
	}
```
Possible repro:
 - roomserver produces join event. `/join` 200 OKs.
 - fedsender gets it first and stores it.
 - client does `/keys/upload`.
 - keyserver produces key change event.
 - fedsender gets key change event, runs code above.
 - currentstateserver still doesn't have the join so returns an incomplete set of room IDs.
 - querying them in `GetJoinedHostsForRooms` returns wrong `destinations`.

The same result can occur if currentstateserver got the join event but fedsender did not (as it won't be in the joined_hosts table). We need to have some guarantees about which components have seen which things when we rely on making lots of outbound hits like this.

https://buildkite.com/matrix-dot-org/dendrite/builds/2716#8d456180-6fd7-4bd6-8e11-1fe07b0ae6e5

Kegsay:
> I think as a rough rule of thumb: code in consumers should only ever make internal HTTP hits to the producer component as other components may not have that event yet
which for key changes implies that we need to store the joining user IDs in the fedsender so we don't need to query currentstateserver

Kegsay:
Specifically:
 - syncapi has `OutputKeyChangeEventConsumer` which calls `currentStateAPI.QuerySharedUsers`
 - fedsender has `KeyChangeConsumer` which calls `currentStateAPI.QueryRoomsForUser`.

Kegsay:
Obsolete when CSS dies.

Kegsay:
#1405

PR #1405, opened 2020-09-07
Remove current state server

### LINKED ISSUE #1365: Remove currentstateserver
### Context

The specification makes no mention of what guarantees are in place when a request to a Homeserver returns a 200 OK.
For example, if you leave a room, what guarantees are in place when `/leave` returns 200 OK? Does it:
- guarantee that any clients blocking on `/sync` have _already_ been notified of the leave via `/sync`, or merely
that they _will be_ notified? Currently most implementations assume the latter.
- guarantee that `GET /joined_rooms` omits the room that was left? Sytest assumes this is immediate.
- guarantee that any _new_ `/sync` requests after the 200 OK will omit the room? Some Sytests assume this.
- guarantee that device key updates informing users to no longer track the leaving user's keys will be sent to
clients in the room that was left?
- guarantee that all/any/some (eg a quorum of) federated servers have been notified of the leave?
- guarantee that requests to `GET /state/m.room.member/@left-user` will return `membership: leave`?

It's easy to say that we should guarantee nothing but that makes client implementations difficult. Most clients
will assume that 'snapshot' requests are using the most up-to-date state but that `/sync` may lag behind a bit.

### Why remove current state server?

It encourages a bad pattern of raciness whereby the roomserver knows the latest state but currentstateserver hasn't caught up yet. This can lead to sytest flakiness such as:

`/joined_rooms` works by:
- `clientapi` receives the request.
- Asks `currentstateserver` via `QueryRoomsForUser`.
- Returns the response.

The `/leave` will 200 OK when the `roomserver` has processed the event and produced it to Kafka logs. It's possible
that `/joined_rooms` will be requested before `currentstateserver` has consumed the Kafka log entry for the leave, producing a stale result and failing sytest.

This will fix/obviate #1272 #1357

Kegsay:
Closes #1365 #1272 #1357


neilalexander (commented):
Can you please also update `dendrite-config.yaml` to remove the dead section and the various places in `build/docker/` that we refer to the CSS?

PR #1411, opened 2020-09-08
Optional transactional isolation in Sync API

neilalexander:
I suspect that the transaction that we open in the sync API may be causing `database is locked` errors. This disables it for SQLite.


neilalexander:
Closing because a) it's out of date and b) may not do what we want anyway.

PR #1412, opened 2020-09-08
Exclude deleted keys from selectBatchDeviceKeysSQL

neilalexander:
This seems to stop us from returning useless entries when someone else over federation requests our keys and stops quite so much spinning when logging in to a new session.


neilalexander:
I haven't added an optional flag to this because it'll just add complexity when I can't see any gain for it — unless I'm missing something, I can't see where we'd want to bulk select including deleted keys in this way.

Kegsay (approved):
LGTM. There's various places in code where we check for empty `KeyJSON` which can be removed now.

PR #1415, opened 2020-09-09
Handle state with input event as new events

neilalexander:
This has some side-effects:

- The roomserver will generate state snapshots for each of the state events when joining a room over federation (previously they did not have one)
- The latest events updater will run for each of these events, resetting the forward extremities when joining a room over federation
- Output events from the roomserver are generated, notifying the sync API and other components about these events inline

However, this should make it possible to start implementing soft-failed events without breaking federated joins.

### DIFF
--- a/roomserver/api/wrapper.go
+++ b/roomserver/api/wrapper.go
@@ -40,9 +40,9 @@ func SendEvents(
 	return SendInputRoomEvents(ctx, rsAPI, ires)
 }
 
-// SendEventWithState writes an event with KindNew to the roomserver
-// with the state at the event as KindOutlier before it. Will not send any event that is
-// marked as `true` in haveEventIDs
+// SendEventWithState writes an event with KindNew to the roomserver along
+// with a number of rewrite and outlier events for state and auth events
+// respectively.
 func SendEventWithState(
 	ctx context.Context, rsAPI RoomserverInternalAPI, state *gomatrixserverlib.RespState,
 	event gomatrixserverlib.HeaderedEvent, haveEventIDs map[string]bool,
@@ -60,6 +60,14 @@ func SendEventWithState(
 	var ires []InputRoomEvent
 	var stateIDs []string
 
+	// This function generates three things:
+	// A - A set of "rewrite" events, which will form the newly rewritten
+	//     state before the event, which includes every rewrite event that
+	//     came before it in its state
+	// B - A set of "outlier" events, which are auth events but not part
+	//     of the rewritten state
+	// C - A "new" event, which include all of the rewrite events in its
+	//     state
 	for _, authOrStateEvent := range authAndStateEvents {
 		if authOrStateEvent.StateKey() == nil {
 			continue
@@ -68,16 +76,21 @@ func SendEventWithState(
 			continue
 		}
 
-		// Work out whether we should store the event as an outlier
-		// or if we should send it as a rewrite event.
+		// We will handle an event as if it's an outlier if one of the
+		// following conditions is true:
 		storeAsOutlier := false
 		if authOrStateEvent.Type() == event.Type() && *authOrStateEvent.StateKey() == *event.StateKey() {
-			// The event is the same as the event that we are looking
-			// to send as a new event.
+			// The event is a state event but the input event is going to
+			// replace it, therefore it can't be added to the state or we'll
+			// get duplicate state keys in the state block. We'll send it
+			// as an outlier because we don't know if something will be
+			// referring to it as an auth event, but need it to be stored
+			// just in case.
 			storeAsOutlier = true
 		} else if _, ok := isCurrentState[authOrStateEvent.EventID()]; !ok {
-			// The event doesn't belong to the current state but is
-			// actually an auth event.
+			// The event is an auth event and isn't a part of the state set.
+			// We'll send it as an outlier because we need it to be stored
+			// in case something is referring to it as an auth event.
 			storeAsOutlier = true
 		}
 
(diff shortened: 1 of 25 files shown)


roomserver/types/types.go
```diff
@@ -72,6 +74,24 @@ func (a StateEntry) LessThan(b StateEntry) bool {
...
+func DeduplicateStateEntries(a []StateEntry) []StateEntry {
+	result := a
+	if len(a) < 2 {
+		return a
+	}
+	sort.SliceStable(a, func(i, j int) bool {
+		return a[i].LessThan(a[j])
+	})
+	for i := 0; i < len(result)-1; i++ {
+		if result[i].StateKeyTuple == result[i+1].StateKeyTuple {
```
Kegsay, line 87: I don't understand the signifance of using `result` vs `a`. Needs comments.

roomserver/types/types_test.go
```diff
@@ -0,0 +1,26 @@
...
+		{StateKeyTuple{1, 1}, 1},
+		{StateKeyTuple{1, 1}, 2},
+		{StateKeyTuple{1, 1}, 3},
+		{StateKeyTuple{2, 2}, 4},
+		{StateKeyTuple{2, 3}, 5},
+		{StateKeyTuple{3, 3}, 6},
+	}
+	expected := []EventNID{3, 4, 5, 6}
+	entries = DeduplicateStateEntries(entries)
+	if len(entries) != 4 {
```
Kegsay, line 18: Needs commentary either here or on the function signature itself to explain what it's doing - de-duplicating based on type/skey and preferring higher event NIDs. Current comment isn't enough.

syncapi/storage/sqlite3/current_room_state_table.go
```diff
@@ -51,7 +51,7 @@ CREATE UNIQUE INDEX IF NOT EXISTS syncapi_event_id_idx ON syncapi_current_room_s
 const upsertRoomStateSQL = "" +
 	"INSERT INTO syncapi_current_room_state (room_id, event_id, type, sender, contains_url, state_key, headered_event_json, membership, added_at)" +
 	" VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9)" +
-	" ON CONFLICT (event_id, room_id, type, sender, contains_url)" +
+	" ON CONFLICT (room_id, type, state_key)" +
```
Kegsay, line 54: Why was this changed?

neilalexander: Because it's wrong and looks like it was copy-pasta'd from somewhere when we did the first round of SQLite-ing. The Postgres equivalent is `CONSTRAINT syncapi_room_state_unique UNIQUE (room_id, type, state_key)`.

roomserver/internal/input/input_latest_events.go
```diff
@@ -303,8 +307,14 @@ func (u *latestEventsUpdater) makeOutputNewRoomEvent() (*api.OutputEvent, error)
 		latestEventIDs[i] = u.latest[i].EventID
 	}
 
+	var outputType api.OutputRoomEventType
+	if u.isHistorical {
+		outputType = api.OutputRoomState
```
Kegsay, line 312: `isHistorical` feels like the wrong naming here - presumably you could get historical message events which then confusingly get marked as `OutputRoomState`?

roomserver/api/wrapper.go
```diff
@@ -47,34 +47,66 @@ func SendEventWithState(
...
-		})
-	}
+		if haveEventIDs[authOrStateEvent.EventID()] {
+			continue
+		}
+
+		// Work out whether we should store the event as an outlier
+		// or if we should send it as a rewrite event.
+		storeAsOutlier := false
+		if authOrStateEvent.Type() == event.Type() && *authOrStateEvent.StateKey() == *event.StateKey() {
```
Kegsay, line 74: Won't this panic if `event` isn't a state event?

roomserver/api/wrapper.go
```diff
@@ -47,34 +47,66 @@ func SendEventWithState(
...
+			// The event is the same as the event that we are looking
+			// to send as a new event.
+			storeAsOutlier = true
```
Kegsay, line 77: I don't grok why we therefore store this as an outlier.

roomserver/api/wrapper.go
neilalexander: There's a guard before it:
```
if authOrStateEvent.StateKey() == nil {
    continue
}
```

roomserver/internal/perform/perform_join.go
```diff
@@ -183,31 +183,34 @@ func (r *Joiner) performJoinRoomByID(
...
-		// Check that the domain isn't ours. If it's local then we don't
-		// need to do anything as our own copy of the room state will be
-		// up-to-date.
-		if inviterDomain != r.Cfg.Matrix.ServerName {
-			// Add the server of the person who invited us to the server list,
-			// as they should be a fairly good bet.
-			req.ServerNames = append(req.ServerNames, inviterDomain)
+	if !serverInRoom || (err == nil && isInvitePending) {
+		if err == nil && inviteSender != "" {
+			// Check if there's an invite pending.
```
Kegsay, line 196: I thought we just did that on `:193`?

roomserver/internal/perform/perform_join.go
```diff
@@ -183,31 +183,34 @@ func (r *Joiner) performJoinRoomByID(
...
+			if ierr != nil {
+				return "", fmt.Errorf("gomatrixserverlib.SplitID: %w", err)
+			}
 
-			// Perform a federated room join.
-			return req.RoomIDOrAlias, r.performFederatedJoinRoomByID(ctx, req)
+			// Check that the domain isn't ours. If it's local then we don't
+			// need to do anything as our own copy of the room state will be
+			// up-to-date.
+			if inviterDomain != r.Cfg.Matrix.ServerName {
```
Kegsay, line 205: This will be a problem though if they are the same (local invite) as we will fall through and call `performFederatedJoinRoomByID`.

roomserver/api/wrapper.go
Kegsay: There's no guard for `event` though which is also being de-referenced.

roomserver/api/wrapper.go
neilalexander: Well, there's three things that are given here:
- The event that we're sending (A)
- The state events (B)
- The auth events (C)

`authOrStateEvents` is B+C, so it's quite likely that A will replace something that's in B+C. If we don't store it as an outlier then we can't refer to it as an auth event.

roomserver/api/wrapper.go
```diff
@@ -47,34 +47,66 @@ func SendEventWithState(
...
+			})
+		} else {
+			ires = append(ires, InputRoomEvent{
+				Kind:          KindRewrite,
+				Event:         authOrStateEvent.Headered(event.RoomVersion),
+				AuthEventIDs:  authOrStateEvent.AuthEventIDs(),
+				HasState:      true,
+				StateEventIDs: stateIDs,
+			})
+			stateIDs = append(stateIDs, authOrStateEvent.EventID())
```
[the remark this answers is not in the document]
neilalexander: The goal here is to generate:
1. A set of rewrite events, which will form the new state
2. Possibly some outliers, which are auth events but don't form part of the new room state
3. A new event, which is the input event we're dealing with

Each time we process a rewrite event, we want it to have a state snapshot that refers to all of the rewrite events that came before it. Then it gets added to the list so that it appears in the state snapshot of the next rewrite event.

Then by the time we get to the new event at the end, we have a full state comprising all of the rewrite events.

roomserver/internal/perform/perform_join.go
neilalexander: I think you're right - there's an unintended fallthrough here.

roomserver/api/wrapper.go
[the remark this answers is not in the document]
Kegsay: This definitely needs clarification in the code then as I thought this was a bug!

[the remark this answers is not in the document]
neilalexander: I'll comment it out better :-)

roomserver/api/wrapper.go
Kegsay: This is still an issue.

PR #1416, opened 2020-09-09
Add support for database migrations

### LINKED ISSUE #1246: Database schema upgrades
We need to support schema upgrades. Use an off-the-shelf one?

Kegsay:
Closes #1246

This PR does NOT add any migrations as an example. I have
manually tested that the library works with SQL and Go based
upgrades correctly. Documentation should be sufficient for
devs to add migrations.

We probably want a bash script which builds goose and calls `up` with
the right CLI args, but for now this is a start.


go.mod
```diff
@@ -29,6 +29,7 @@ require (
 	github.com/ngrok/sqlmw v0.0.0-20200129213757-d5c93a81bec6
 	github.com/opentracing/opentracing-go v1.2.0
 	github.com/pkg/errors v0.9.1
+	github.com/pressly/goose v2.7.0-rc5+incompatible
```
neilalexander, line 32: I wonder if we should pin this to a non-release candidate tag.

neilalexander (approved):
I've only poked it lightly but it looks neat!

go.mod
Kegsay: There hasn't been a non RC build in a long time hence why I did this. Some of the examples only work on 2.7

PR #1417, opened 2020-09-09
fix lots of places where we didn't rollback txns upon go errs

ara4n:


Kegsay (approved):
Please wait on merging until https://buildkite.com/matrix-dot-org/dendrite/builds/3101#46998ab9-7309-4b95-b4dc-c1defe046ceb passes or fails otherwise it'll cancel it.

PR #1423, opened 2020-09-12
Update all usages of tx.Stmt to sqlutil.TxStmt

### LINKED ISSUE #1392: Replace all instances of txn.Stmt with sqlutil.TxStmt
This prevents us from panicking when we change transactions around and pass `nil` through instead, since `sqlutil.TxStmt` will handle a `nil` transaction safely.

samcday:
@neilalexander roping you into this one since you raised the original issue :)

I wanted to also add some kind of static check to forbid further use of `tx.Stmt` in the codebase, but there doesn't to appear to be any lint tools that can do that currently. I started looking at tools/analysis to write my own but figured it might be better to just focus on contributing to dendrite first xD

The thing I'm unsure about is why there was a handful of usages of `tx.StmtContext` instead of the bare `tx.Stmt`. The docs state the context is only used during statement preparation in this case. But shouldn't we be just threading the context through anywhere it's supported, whenever we've got one?

closes #1392

### Pull Request Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -3,4 +3,4 @@
 <!-- Please read CONTRIBUTING.md before submitting your pull request -->
 
 * [ ] I have added any new tests that need to pass to `testfile` as specified in [docs/sytest.md](https://github.com/matrix-org/dendrite/blob/master/docs/sytest.md)
-* [ ] Pull request includes a [sign off](https://github.com/matrix-org/dendrite/blob/master/CONTRIBUTING.md#sign-off)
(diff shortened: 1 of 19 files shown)


neilalexander:
> The thing I'm unsure about is why there was a handful of usages of tx.StmtContext instead of the bare tx.Stmt. The docs state the context is only used during statement preparation in this case. But shouldn't we be just threading the context through anywhere it's supported, whenever we've got one?

I think there's probably a difference in some places between where we prepare statements at startup vs preparing them before we need them (like in the SQLite variadic query cases).

neilalexander (approved):
Thanks for this!

Issue #610, opened 2019-01-08
Implement account deactivation


babolivier:
- [ ] Implement [POST /_matrix/client/r0/account/deactivate](https://matrix.org/docs/spec/client_server/r0.4.0.html#post-matrix-client-r0-account-deactivate)

Kegsay:
Sytests:
```
    × Can deactivate account
    × Can't deactivate account with wrong password
    × After deactivating account, can't log in with password
```

neilalexander:
Yes, by #1455.

PR #1455, opened 2020-09-30
Implement account deactivation

Lesterpig:
In this PR, "deactivation" only prevents users from login, with no data deletion.
I believe there is a small problem with sytest, and suggest the following modification, since I don't see why the `completed` field is expected in the deactivation response with invalid password.

*(I can provide a PR for sytest if you are OK with it)*

```diff
diff --git a/tests/14account/02deactivate.pl b/tests/14account/02deactivate.pl
index 20b03d6b..834246cc 100644
--- a/tests/14account/02deactivate.pl
+++ b/tests/14account/02deactivate.pl
@@ -43,7 +43,7 @@ test "Can't deactivate account with wrong password",
 
          my $body = decode_json $resp->content;
 
-         assert_json_keys( $body, qw( error errcode params completed flows ));
+         assert_json_keys( $body, qw( error errcode params flows ));
 
          my $errcode = $body->{errcode};
```

This PR also attempts to add a simple database migration (add a column to account_accounts).

See #610
Signed-off-by: Loïck Bonniot <git@lesterpig.com>

### Pull Request Checklist


userapi/storage/accounts/sqlite3/accounts_table.go
```diff
@@ -37,7 +37,9 @@ CREATE TABLE IF NOT EXISTS account_accounts (
     -- The password hash for this account. Can be NULL if this is a passwordless account.
     password_hash TEXT,
     -- Identifies which application service this account belongs to, if any.
-    appservice_id TEXT
+    appservice_id TEXT,
+    -- If the account is currently active
+    is_active BOOLEAN DEFAULT 1
```
neilalexander, line 42: It's one of those personal preference things I'm sure, but I can't help but feel like this would make more sense flipped, e.g. `is_deactivated BOOLEAN DEFAULT 0`, in case we end up ever adding more fields like this.

userapi/storage/accounts/sqlite3/deltas/20200929203058_is_active.sql
```diff
@@ -0,0 +1,38 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE account_accounts RENAME TO account_accounts_tmp;
```
neilalexander, line 3: I'm curious about the choice to recreate the table in this way - can we not add columns in SQLite using `ALTER TABLE ADD`?

neilalexander (commented):
Looks like a really promising start, just a couple notes!

userapi/storage/accounts/sqlite3/deltas/20200929203058_is_active.sql
Lesterpig: I don't think sqlite supports "IF NOT EXISTS" as postgres does.

The rationale here is that if someone attempts to run this migration, it will work even if the database is already up to date (but goose migration table does not exist). However, all accounts will be reactivated in the process.

Also, I've not tried, but for some column types it is not possible to add a default values with ALTER.

Kegsay (approved):
Excellent PR, thank you! I'm impressed you also did DB migrations and handled the faff with SQLite! Please send a sytest PR for the `completed` stuff and we'll get that merged as well.

Lesterpig:
Thanks for the review @Kegsay !
After some discussions in the #dendrite-dev channel, it looks like the `completed` field is indeed required.

So I see two options here:
1) Remove the field from sytest, but this means dendrite will act slightly differently than synapse on invalid password
2) Update the interactive auth workflow to add `completed` on failure - I can also propose a new dendrite PR for that

I plan to also rename `is_active` to `is_deactivated` as suggested in @neilalexander 's review :+1:

Kegsay:
> Update the interactive auth workflow to add completed on failure - I can also propose a new dendrite PR for that

Sounds good.

neilalexander (approved):
LGTM assuming Sytest passes!

PR #1458, opened 2020-10-01
Add nginx sample reverse proxy configs

neilalexander:
This adds two sample config files for using nginx as a reverse proxy in front of Dendrite.


docs/nginx/monolith-sample.conf
```diff
@@ -0,0 +1,24 @@
+server {
+    listen 443 ssl;
+    server_name my.hostname.com;
+
+    ssl_certificate /path/to/fullchain.pem;
+    ssl_certificate_key /path/to/privkey.pem;
+    ssl_dhparam /path/to/ssl-dhparams.pem;
+
+    proxy_set_header Host      $host;
+    proxy_set_header X-Real-IP $remote_addr;
```
Kegsay, line 10: Do we use `X-Real-IP` anywhere?

docs/nginx/monolith-sample.conf
```diff
@@ -0,0 +1,24 @@
...
+    proxy_read_timeout          600;
```
Kegsay: This is in seconds, so 10min?

docs/nginx/monolith-sample.conf
neilalexander: Don't think so, but figured it might as well be included in case we ever want to add IP filtering.

docs/nginx/monolith-sample.conf
neilalexander: Yep.

docs/nginx/monolith-sample.conf
Kegsay: I think the general header is `X-Forwarded-For` - at least it is on Apache.

Kegsay (approved):

PR #1463, opened 2020-10-01
Update gomatrixserverlib, sytest-whitelist

neilalexander:
This updates gomatrixserverlib to matrix-org/gomatrixserverlib#226.


Kegsay (approved):

neilalexander:
This is no longer needed as we're already pulling in a newer version of gmsl than this.

Issue #617, opened 2019-01-08
Implement history visibility


babolivier:
- [ ] Implement special case for [m.room.history_visibility](https://matrix.org/docs/spec/client_server/r0.4.0.html#m-room-history-visibility) state event

Requires adding restrictions and checks to every endpoint/feature that returns events to make sure the user requesting them is allowed to see them.

It can get tricky when we face a situation in which the user is only allowed access to a specific range of the history (if they've joined the room then left, if the history visibility state event allowed the user to see events at some point then got turned back to a setting meaning the user can't access it anymore, etc.).

prateek2211:
I would be taking this

Kegsay:
I'd advise against it @prateek2211 - I've partially implemented history visibility over federation for the purposes of backfilling, but there's a lot of complexity in this issue.

Kegsay:
Sytests:
```
    × m.room.history_visibility == "shared" allows/forbids appropriately for Real users
    × m.room.history_visibility == "invited" allows/forbids appropriately for Real users
    × m.room.history_visibility == "joined" allows/forbids appropriately for Real users
    × m.room.history_visibility == "default" allows/forbids appropriately for Real users
    × Real users can sync from shared guest_access rooms if joined
    × Real users can sync from invited guest_access rooms if joined
    × Real users can sync from joined guest_access rooms if joined
    × Only see history_visibility changes on boundaries
```

Kegsay:
Put on hold until rejected events are implemented, because they are needed to determine history visibility because:

> "Outbound federation rejects backfill containing invalid JSON for events in room version 6" fails. and that test backfills a "bad event" with a float for a field, which Dendrite just discards and drops to the floor, but that then breaks the DAG meaning when I go to check "what was the state of the room at this event" the check fails because the DAG is disjoint, so I can't progress with history vis without making rejected events a thing

ShadowJonathan:
I encountered this issue on three complement tests i was creating;

```go
// Tries to fetch an event before join, and succeeds.
// history_visibility: shared
...

// Tries to fetch an event between being invited and joined, and succeeds.
// history_visibility: invited
...

// Tries to fetch an event without having joined, and succeeds.
// history_visibility: world_readable
...
```

bronger:
For me, in rooms many quotations are replaced with “You are not allowed to see this, or the item is unavailable” (back-translated from German). May this issue here be the cause of this?

[...]

PR #1470, opened 2020-10-02
Add history visibility guards

Kegsay:
Default to 'joined' visibility to avoid leaking events, until we get
around to implementing history visibility completely. Related #617


neilalexander (approved):
Looks OK to me. Let's see what Sytest says.

PR #1479, opened 2020-10-05
Resolve state after event against current room state when determining latest state changes

neilalexander:
We were working out which state events were added/removed based on the difference between the room state `A` and the state after the event `B`. The problem is that `B` is worked out after the `prev_events` including any state changes, but there's nothing to say that `B` actually came after `A` in real terms.

This meant that we could handle events with old `prev_events` which would cause us to revert newer state changes, such as the bug I experienced joining Watercooler, because comparing `A` and `B` made it look like `B` deleted some state from `A`.

This changes the comparison now to compare the difference between `A` and `resolved(A+B)`, such that if the room state contains newer state events, they will be deconflicted by the state resolution algorithm and kept.

### DIFF
--- a/roomserver/internal/input/input_latest_events.go
+++ b/roomserver/internal/input/input_latest_events.go
@@ -215,10 +215,27 @@ func (u *latestEventsUpdater) latestState() error {
 	var err error
 	roomState := state.NewStateResolution(u.api.DB, *u.roomInfo)
 
-	// Get a list of the current latest events.
-	latestStateAtEvents := make([]types.StateAtEvent, len(u.latest))
+	// Get a list of the current room state events if available.
+	var currentState []types.StateEntry
+	if u.roomInfo.StateSnapshotNID != 0 {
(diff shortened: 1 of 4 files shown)


roomserver/internal/input/input_latest_events.go
```diff
@@ -215,10 +215,27 @@ func (u *latestEventsUpdater) latestState() error {
...
+	for i := range currentState {
+		latestStateAtEvents[i] = types.StateAtEvent{
+			BeforeStateSnapshotNID: u.roomInfo.StateSnapshotNID,
+			StateEntry:             currentState[i],
+		}
+		offset++
+	}
 	for i := range u.latest {
-		latestStateAtEvents[i] = u.latest[i].StateAtEvent
+		latestStateAtEvents[offset+i] = u.latest[i].StateAtEvent
```
Kegsay, line 238: Won't this have many duplicates?

neilalexander: Possibly, and that's okay, because the state resolution algorithm will work out which the truthful events are if any state key tuples are duplicated.

PR #1481, opened 2020-10-06
Remove KindRewrite

neilalexander:
`KindRewrite` made sense when I was first working on the soft-fail stuff because it meant that everything had a state snapshot, even if that was just a cumulative one provided by the given state IDs.

It turns out that actually this doesn't make sense, and it can make joining larger rooms *really* slow as it tries to build ever-growing state snapshots for the outliers.


Kegsay:
Why does it not make sense? Surely knowing _some_ part of the room state at an outlier is better than nothing at all?

neilalexander:
Well, philosophically it might make sense, but given the cost of doing it I am not so sure.

We were only treating current state events as `KindRewrite`, not things that came in the auth chain, because the state events were the only ones we could build up state for cumulatively. Therefore it's still quite likely we'd end up with a bunch of auth events that had no state before them.

The other thing is that, for the purposes of soft-fail, we never ended up using that state.

Kegsay (approved):

PR #1482, opened 2020-10-06
Add furl

neilalexander:
It's like `curl`, but for federation requests.


Kegsay (approved):
Love it, though it would be good if it support POST too ;)

neilalexander:
Now it does!

ara4n:
oh, i've wanted this FOREVER :D

PR #1483, opened 2020-10-06
Don't give up if we fail to fetch a key

neilalexander:
If we fail to fetch specific keys, we shouldn't just give up - gomatrixserverlib may well be happy to carry on ignoring specific auth/state events if they aren't important to anything else.


Kegsay (approved):
Linting is failing.

PR #1484, opened 2020-10-06
Update monolith -api behaviour

neilalexander:
This makes some changes to the monolith `-api` behaviour:

- The monolith internal listener is now on `localhost:18008` by default when `-api` is enabled
- Adds `-api-bind-address` for overriding the above
- The setup package now prefers setting up the external listener over the internal one, which ultimately makes more sense because that's where the majority of useful HTTP routing goes

### DIFF
--- a/internal/setup/base.go
+++ b/internal/setup/base.go
@@ -300,35 +300,35 @@ func (b *BaseDendrite) SetupAndServeHTTP(
 	externalRouter.PathPrefix(httputil.PublicFederationPathPrefix).Handler(b.PublicFederationAPIMux)
 	externalRouter.PathPrefix(httputil.PublicMediaPathPrefix).Handler(b.PublicMediaAPIMux)
 
-	go func() {
-		logrus.Infof("Starting %s listener on %s", b.componentName, internalServ.Addr)
-		if certFile != nil && keyFile != nil {
-			if err := internalServ.ListenAndServeTLS(*certFile, *keyFile); err != nil {
-				logrus.WithError(err).Fatal("failed to serve HTTPS")
-			}
-		} else {
-			if err := internalServ.ListenAndServe(); err != nil {
-				logrus.WithError(err).Fatal("failed to serve HTTP")
-			}
-		}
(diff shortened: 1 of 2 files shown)


Kegsay:
Why do we do this though?

neilalexander:
So there are "external" and "internal" listeners. The external listeners are where `/_matrix` APIs go. The internal ones are where we put `/api`, `/metrics`. etc. Depending on whether we run in monolith mode or polylith mode, the behaviour is a bit different.

In monolith mode, we were previously defaulting to both listeners being the same, so `/api`, `/metrics` and other sensitive endpoints could end up on the public listener. 

In polylith mode, whether a component starts both external+internal listeners or just the internal listener depends on whether it has public-facing APIs (like the client API, sync API, federation API, media API) or not.

This PR changes things around so that the internal listener now listens on `localhost` by default, ensuring that the internal routing now stays on the internal listener, unless explicitly overridden.

internal/setup/base.go
```diff
@@ -80,6 +80,7 @@ type BaseDendrite struct {
 const HTTPServerTimeout = time.Minute * 5
 const HTTPClientTimeout = time.Second * 30
 
+const NoInternalListener = ""
 const NoExternalListener = ""
```
Kegsay, line 84: What's going on here? You need some actual values here else `NoInternalListener` is the same as `NoExternalListener` - https://play.golang.org/p/bIAL8JumJtK

Kegsay (approved):

internal/setup/base.go
neilalexander: They're just designed to be human-readable rather than leaving `""` littered about the place, e.g. 
https://github.com/matrix-org/dendrite/blob/759bab6a0aa59c92122f330db25784eaad43967e/cmd/dendrite-monolith-server/main.go#L164-L168
https://github.com/matrix-org/dendrite/blob/759bab6a0aa59c92122f330db25784eaad43967e/cmd/dendrite-room-server/main.go#L35-L39

Kegsay: Maybe `setup.NoListener` instead?

PR #1491, opened 2020-10-07
Deep checking of forward extremities

neilalexander:
Beforehand, the forward extremities were calculated using a simple delta, but this could mean that gaps bigger than one reference could mess up the calculation and sometimes put old events into the forward extremities.

Now we actually consult the `roomserver_previous_events` table properly, which is much better at finding referenced events.

This PR also removes `KindBackfill` as we don't use it anywhere, and also adds `previous_events` references in `StoreEvent` so that we don't have any code paths where that doesn't happen.

### DIFF
--- a/roomserver/internal/input/input_latest_events.go
+++ b/roomserver/internal/input/input_latest_events.go
@@ -28,6 +28,7 @@ import (
 	"github.com/matrix-org/dendrite/roomserver/types"
 	"github.com/matrix-org/gomatrixserverlib"
 	"github.com/matrix-org/util"
+	"github.com/sirupsen/logrus"
 )
 
 // updateLatestEvents updates the list of latest events for this room in the database and writes the
@@ -146,24 +147,12 @@ func (u *latestEventsUpdater) doUpdateLatestEvents() error {
 		return fmt.Errorf("u.updater.StorePreviousEvents: %w", err)
 	}
 
-	// Get the event reference for our new event. This will be used when
-	// determining if the event is referenced by an existing event.
-	eventReference := u.event.EventReference()
-
-	// Check if our new event is already referenced by an existing event
-	// in the room. If it is then it isn't a latest event.
-	alreadyReferenced, err := u.updater.IsReferenced(eventReference)
-	if err != nil {
-		return fmt.Errorf("u.updater.IsReferenced: %w", err)
-	}
-
-	// Work out what the latest events are.
-	u.latest = calculateLatest(
+	// Work out what the latest events are. This will include the new
+	// event if it is not already referenced.
+	u.calculateLatest(
 		oldLatest,
-		alreadyReferenced,
-		prevEvents,
 		types.StateAtEventAndReference{
-			EventReference: eventReference,
+			EventReference: u.event.EventReference(),
 			StateAtEvent:   u.stateAtEvent,
 		},
 	)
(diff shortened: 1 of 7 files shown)


roomserver/storage/shared/storage.go
```diff
@@ -468,6 +468,11 @@ func (d *Database) StoreEvent(
 		if !isRejected { // ignore rejected redaction events
 			redactionEvent, redactedEventID, err = d.handleRedactions(ctx, txn, eventNID, event)
 		}
+		for _, prev := range event.PrevEvents() {
+			if err = d.PrevEventsTable.InsertPreviousEvent(ctx, txn, prev.EventID, prev.EventSHA256, eventNID); err != nil {
```
Kegsay, line 472: It's probably not safe to do this because of the comment:
```
// Insert an entry into the previous_events table.
// If there is already an entry indicating that an event references that previous event then
// add the event NID to the list to indicate that this event references that previous event as well.
// This should only be modified while holding a "FOR UPDATE" lock on the row in the rooms table for this room.
// The lock is necessary to avoid data races when checking whether an event is already referenced by another event.
```

roomserver/internal/input/input_latest_events.go
```diff
@@ -279,42 +254,46 @@ func (u *latestEventsUpdater) latestState() error {
...
-	if !alreadyReferenced && !alreadyInLatest {
-		// This event is not referenced by any of the events in the room
-		// and the event is not already in the latest events.
-		// Add it to the latest events
+	// Then check and see if our new event is already included in that set.
+	for _, l := range newLatest {
+		if l.EventReference.EventID == newEvent.EventReference.EventID && bytes.Equal(l.EventReference.EventSHA256, newEvent.EventReference.EventSHA256) {
+			// We've already referenced this event so we can just return
+			// the newly completed extremities at this point.
+			u.latest = newLatest
```
Kegsay, line 281: Just because _one_ forward extremity matches doesn't mean they all do surely?

roomserver/internal/input/input_latest_events.go
```diff
@@ -279,42 +254,46 @@ func (u *latestEventsUpdater) latestState() error {
...
+			return
+		}
+	}
+
+	// If the new event isn't already in the set then we'll check if it
+	// really should be.
+	referenced, err := u.updater.IsReferenced(newEvent.EventReference)
```
Kegsay, line 288: Didn't we just do this `IsReferenced` check? It's not clear why we need to do it again.

roomserver/internal/input/input_latest_events.go
```diff
@@ -279,42 +254,46 @@ func (u *latestEventsUpdater) latestState() error {
...
+	if err != nil {
+		logrus.WithError(err).Errorf("Failed to retrieve event reference for %q", newEvent.EventReference.EventID)
+	}
```
Kegsay, line 291: Fallthrough unexpectedly?

Kegsay (changes requested):

roomserver/internal/input/input_latest_events.go
neilalexander: This is covering the edge-case of the *new* event already being in the *old* forward extremities for some reason (e.g. we've seen it before).

roomserver/internal/input/input_latest_events.go
neilalexander: The previous `IfReferenced` checks are on the *old* events that were marked as extremities. This one is on the *new* event.

roomserver/storage/shared/storage.go
neilalexander: Hmm, good point, I can change this to use the latest rooms updater although that may still not do the right thing on SQLite anyway.

roomserver/internal/input/input_latest_events.go
neilalexander: There should have been an `else` here, fixed.

roomserver/internal/input/input_latest_events.go
Kegsay: This needs to be made clearer.

neilalexander: I updated the comments again.

Kegsay (approved):

roomserver/internal/input/input_latest_events.go
```diff
@@ -279,42 +254,50 @@ func (u *latestEventsUpdater) latestState() error {
...
-		}
-		if l.EventNID == newEvent.EventNID {
-			alreadyInLatest = true
+		referenced, err := u.updater.IsReferenced(l.EventReference)
+		if err != nil {
+			logrus.WithError(err).Errorf("Failed to retrieve event reference for %q", l.EventID)
 		}
-		if keep {
-			// Keep the event in the latest events.
+		if !referenced {
```
Kegsay, line 271: This is presumably still going to fire if `err != nil`

neilalexander: I've fixed that so it no longer does. I guess it's better to omit them than include them incorrectly.

PR #1492, opened 2020-10-07
Rename serverkeyapi to signingkeyserver

Kegsay:
We use "api" for public facing stuff and "server" for internal stuff.
As the server key API is internal only, we call it 'signing key server',
which also clarifies the type of key (as opposed to TLS keys, E2E keys, etc)


cmd/dendrite-demo-yggdrasil/main.go
```diff
@@ -78,7 +78,7 @@ func main() {
 	cfg.MediaAPI.Database.ConnectionString = config.DataSource(fmt.Sprintf("file:%s-mediaapi.db", *instanceName))
 	cfg.SyncAPI.Database.ConnectionString = config.DataSource(fmt.Sprintf("file:%s-syncapi.db", *instanceName))
 	cfg.RoomServer.Database.ConnectionString = config.DataSource(fmt.Sprintf("file:%s-roomserver.db", *instanceName))
-	cfg.ServerKeyAPI.Database.ConnectionString = config.DataSource(fmt.Sprintf("file:%s-serverkey.db", *instanceName))
+	cfg.SigningKeyServer.Database.ConnectionString = config.DataSource(fmt.Sprintf("file:%s-serverkey.db", *instanceName))
```
neilalexander, line 81: We probably want to update this filename default too.

cmd/dendritejs/main.go
```diff
@@ -168,7 +168,7 @@ func main() {
 	cfg.FederationSender.Database.ConnectionString = "file:/idb/dendritejs_fedsender.db"
 	cfg.MediaAPI.Database.ConnectionString = "file:/idb/dendritejs_mediaapi.db"
 	cfg.RoomServer.Database.ConnectionString = "file:/idb/dendritejs_roomserver.db"
-	cfg.ServerKeyAPI.Database.ConnectionString = "file:/idb/dendritejs_serverkey.db"
+	cfg.SigningKeyServer.Database.ConnectionString = "file:/idb/dendritejs_serverkey.db"
```
neilalexander, line 171: And here.

neilalexander (commented):
Please also update `dendrite-config.yaml` and `build/docker/config/dendrite-config.yaml` too :-)

PR #1493, opened 2020-10-07
Return 200 on join before time out

neilalexander:
The join will continue in the background and the room will eventually appear in `/sync`, but the client will otherwise just time out and probably present a worse message to the user.

I don't know if 20 seconds is optimal here but it feels reasonable-ish for now.

### DIFF
--- a/clientapi/routing/joinroom.go
+++ b/clientapi/routing/joinroom.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/matrix-org/dendrite/clientapi/auth/authtypes"
 	"github.com/matrix-org/dendrite/clientapi/httputil"
+	"github.com/matrix-org/dendrite/clientapi/jsonerror"
 	roomserverAPI "github.com/matrix-org/dendrite/roomserver/api"
 	"github.com/matrix-org/dendrite/userapi/api"
 	"github.com/matrix-org/dendrite/userapi/storage/accounts"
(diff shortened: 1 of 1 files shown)


clientapi/routing/joinroom.go
```diff
@@ -73,17 +74,34 @@ func JoinRoomByIDOrAlias(
...
-
-	return util.JSONResponse{
+	// This is the default response that we'll return, assuming nothing
+	// goes wrong within the timeframe.
+	ok := util.JSONResponse{
 		Code: http.StatusOK,
 		// TODO: Put the response struct somewhere internal.
 		JSON: struct {
 			RoomID string `json:"room_id"`
 		}{joinRes.RoomID},
```
Kegsay, line 84: This is unset at this point in time.

clientapi/routing/joinroom.go
```diff
@@ -73,17 +74,34 @@ func JoinRoomByIDOrAlias(
...
+			done <- joinRes.Error.JSONResponse()
+		} else {
+			done <- ok
+		}
+	}()
+
+	// Wait either for the join to finish, or for us to hit a reasonable
+	// timeout, at which point we'll just return a 200 to placate clients.
+	select {
+	case <-time.After(time.Second * 15):
```
Kegsay, line 102: Feels a bit low, but okay.

clientapi/routing/joinroom.go
neilalexander: doh, of course it is.

Kegsay (approved):

PR #1496, opened 2020-10-08
Remove notifs about key changes in syncapi

Kegsay:
The join/leave events themselves will wake up the right people so we
needn't do it twice.


neilalexander (approved):
LGTM if Sytest is happy.

PR #1498, opened 2020-10-08
Return a non-fatal error to the federation API on a state regression

neilalexander:
This stops us from returning a 500 and killing the entire transaction.


Kegsay (approved):
Assuming CI passes

PR #1502, opened 2020-10-09
use jsoniter instead of encoding/json

bn4t:
### Pull Request Checklist

Uses jsoniter instead of encoding/json wherever possible. With some structs from gomatrixserverlib it's currently not possible to use jsoniter.

Some sytest tests seem to fail with this change. I suspect this is because jsoniter sorts map keys differently than encoding/json.
If this is the case jsoniter could be configured to sort map keys like encoding/json.

Signed-off-by: `Benjamin Nater <me@bn4t.me>`

### DIFF
--- a/appservice/consumers/roomserver.go
+++ b/appservice/consumers/roomserver.go
@@ -16,7 +16,7 @@ package consumers
 
 import (
 	"context"
-	"encoding/json"
+	"github.com/json-iterator/go"
 
 	"github.com/matrix-org/dendrite/appservice/storage"
 	"github.com/matrix-org/dendrite/appservice/types"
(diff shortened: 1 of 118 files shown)


neilalexander (commented):
It might be easier to use named imports here, e.g. `import json "github.com/json-iterator/go`, that way we don't need to change every place we deal with JSON.

bn4t:
@neilalexander Okay, will change that.

bn4t:
So, it seems like the reason sytest tests failed was because gomatrixserverlib still used encoding/json. With a modified version of gomatrixserverlib that uses jsoniter the failures no longer seem to happen.

However now the `TestUpdateNoPrevID` unit test fails. Might be because jsoniter sorts map keys differently.

bn4t:
Okay, this is odd. For some reason the `TestUpdateNoPrevID` unit test works on my machine.

bn4t:
Sytest still fails despite using `ConfigCompatibleWithStandardLibrary`. Any idea why this could be @neilalexander ? I'm kinda stuck here.

bn4t:
Closing this because of lack of time from my side.

Issue #1441, opened 2020-09-24
Refactor cross-boundary error handling


neilalexander:
Errors in Dendrite are a mess.

The client API has `jsonerror.MatrixError`, which is nice because those are the things that are returned to clients, although it is maybe misplaced since we also might want to use those in the federation API.

[gomatrix](https://github.com/matrix-org/gomatrix) also has `*gomatrix.HTTPError`, and then in some places we have `roomserverAPI.PerformError`, and in other places we just throw regular opaque errors around. In some API functions we put `PerformError`s into the response struct, in others we just return an `error` from the function itself.

This makes it really difficult to return meaningful clients and other federated servers to errors consistently in any place that an API query is performed.

We should do the work to ensure that if we provide a `jsonerror.MatrixError`, that it is correctly marshalled and unmarshalled across API boundaries and that they make it all the way back to clients/federation endpoints, and tidy up the rest of the mess so that it's consistent and usable.

ara4n:
I agree this is super desirable, but i'm not sure it should technically block entering beta.

Kegsay:
> that it is correctly marshalled and unmarshalled across API boundaries and that they make it all the way back to clients/federation endpoints

iirc this broadly does happen, and there's sytests to prove it. The internal APIs are a mixture of error return / error struct though which should definitely be fixed. Agreed it shouldn't block beta.

PR #1503, opened 2020-10-09
Only return 500 on /send if a database error occurs

neilalexander:
`/send` can return errors in one of two ways - we can either record the error in the JSON response for specific PDUs, or we can return a HTTP status code and fail the entire transaction.

We seemed to be returning 500s readily for any but a small list of errors, but that also caught all sorts of things like failing to find state events. As a result we were missing lots of other events by killing the entire transaction and then other servers would back off because of the 500 error.

This updates it *temporarily* to just fail if it looks like a database error. We really need to do something about #1441 to solve this properly, since right now it's difficult to tell when an error occurs *from* an API call, or if an error occured with *making* the API call itself.

### DIFF
--- a/federationapi/routing/send.go
+++ b/federationapi/routing/send.go
@@ -16,6 +16,7 @@ package routing
 
 import (
 	"context"
+	"database/sql"
 	"encoding/json"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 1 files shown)


federationapi/routing/send.go
```diff
@@ -234,17 +235,10 @@ func (t *txnReq) processTransaction(ctx context.Context) (*gomatrixserverlib.Res
...
-	default:
-		switch err {
-		case context.Canceled:
-		case context.DeadlineExceeded:
-		default:
-			return true
-		}
+	switch err {
+	case sql.ErrConnDone:
+	case sql.ErrTxDone:
```
Kegsay, line 240: You need to add commentary on why we return 500 in these cases, and why we default to non-fatal errors.

Kegsay (approved):

federationapi/routing/send.go
neilalexander: Commentary is already at:

https://github.com/matrix-org/dendrite/blob/2bd0449c5b2b5cbefffbe145f6c4c183e4ff0552/federationapi/routing/send.go#L186-L200

PR #1514, opened 2020-10-12
Get missing event tweaks

neilalexander:
This makes the following tweaks to the codepath for getting missing events in `/send`:

- Send the backward extremity with state *before* sending the rest of the backfilled events without state
- No longer recurse `processEvent` (we'll just send input events instead)
- `QueryMissingAuthPrevEvents` now includes events that have no state in the list of missing `prev_events`, so we'll go and `/state_ids` them

### DIFF
--- a/federationapi/routing/send.go
+++ b/federationapi/routing/send.go
@@ -468,26 +468,18 @@ func (t *txnReq) processEventWithMissingState(ctx context.Context, e gomatrixser
 	// - fill in the gap completely then process event `e` returning no backwards extremity
 	// - fail to fill in the gap and tell us to terminate the transaction err=not nil
 	// - fail to fill in the gap and tell us to fetch state at the new backwards extremity, and to not terminate the transaction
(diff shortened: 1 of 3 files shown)


federationapi/routing/send.go
```diff
@@ -464,39 +468,87 @@ func (t *txnReq) processEventWithMissingState(ctx context.Context, e gomatrixser
...
+
+	// Then send all of the newer backfilled events, of which will all be newer
+	// than the backward extremity, into the roomserver without state. This way
+	// they will automatically fast-forward based on the room state at the
+	// extremity in the last step.
+	headeredNewEvents := make([]gomatrixserverlib.HeaderedEvent, len(newEvents))
+	for i, newEvent := range newEvents {
+		headeredNewEvents[i] = newEvent.Headered(roomVersion)
+	}
+	fmt.Println("Headered events:", len(headeredNewEvents))
```
Kegsay, line 539: Debug logging

Kegsay (approved):
Love it, thank you for removing the recursion!

PR #1516, opened 2020-10-12
Try to retrieve missing auth events from multiple servers

neilalexander:
This tries to get missing auth events from multiple servers, and is a bit more tolerant of missing auth/state events on `/send`.

### DIFF
--- a/federationapi/routing/send.go
+++ b/federationapi/routing/send.go
@@ -112,6 +112,8 @@ type txnReq struct {
 	haveEvents map[string]*gomatrixserverlib.HeaderedEvent
 	// new events which the roomserver does not know about
 	newEvents map[string]bool
+	// servers which we should fetch missing events from
+	servers []gomatrixserverlib.ServerName
 }
 
 // A subset of FederationClient functionality that txn requires. Useful for testing.
@@ -350,7 +352,7 @@ func (t *txnReq) processEvent(ctx context.Context, e gomatrixserverlib.Event, is
 	}
 	var stateResp api.QueryMissingAuthPrevEventsResponse
 	if err := t.rsAPI.QueryMissingAuthPrevEvents(ctx, &stateReq, &stateResp); err != nil {
-		return err
+		return fmt.Errorf("t.rsAPI.QueryMissingAuthPrevEvents: %w", err)
 	}
 
 	if !stateResp.RoomExists {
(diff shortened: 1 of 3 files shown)


federationapi/routing/send.go
```diff
@@ -367,53 +365,31 @@ func (t *txnReq) processEvent(ctx context.Context, e gomatrixserverlib.Event) er
 		return roomNotFoundError{e.RoomID()}
 	}
 
-	if len(stateResp.MissingAuthEventIDs) > 0 {
-		logger.Infof("Event refers to %d unknown auth_events", len(stateResp.MissingAuthEventIDs))
-
-		servers := []gomatrixserverlib.ServerName{t.Origin}
+	// We will need to know this when fetching missing auth or prev events.
+	getServers := func() {
+		t.servers = []gomatrixserverlib.ServerName{}
```
Kegsay, line 370: Feels like a peculiar form, why have this as a nested function which side effects by setting `t.servers`? Why not just have `t.getServers() []ServerName`?

federationapi/routing/send.go
```diff
@@ -431,6 +407,49 @@ func (t *txnReq) processEvent(ctx context.Context, e gomatrixserverlib.Event) er
...
+				logger.WithError(err).Warnf("Failed to retrieve auth event %q", missingAuthEventID)
+				continue withNextServer
+			}
+			ev, err := gomatrixserverlib.NewEventFromUntrustedJSON(tx.PDUs[0], stateResp.RoomVersion)
+			if err != nil {
+				logger.WithError(err).Warnf("Failed to unmarshal auth event %q", missingAuthEventID)
+				continue withNextServer
+			}
+			if err = t.processEvent(ctx, ev); err != nil {
+				return fmt.Errorf("recursive t.processEvent: %w", err)
```
Kegsay, line 440: Do we really need to recurse? We don't care about state snapshots and events having prev_events for auth events right?

Kegsay (approved):

federationapi/routing/send.go
neilalexander: Nah, I just hadn't changed it to `api.SendInputRoomEvents` yet. I'll do that.

federationapi/routing/send.go
neilalexander: I've refactored it to be a bit less strange!

Issue #1508, opened 2020-10-09
Issue fetching old signing key


aditsachde:
### Background information

- **Dendrite version or git SHA**: 0.1.0
- **Monolith or Polylith?**: Monolith
- **SQLite3 or Postgres?**: Postgres
- **Running in Docker?**: Yes
- **`go version`**: Default version in docker container

### Description

I had a synapse server at matrix.sachde.xyz which used the signing key seen [here](https://matrix.org/_matrix/key/v2/query/aditsachde.com/ed25519:a_AMlB). I switched to a dendrite server at dendrite.sachde.xyz, and when joining a room that I previously participated in, the join fails, because signature validation fails, as dendrite fails to retrieve the key. 

I have the old key, but synapse used a different format. Dendrite tries to fetch the key directly but fails. It should fall back to the perspective key server, as the key does exist on matrix.org

### Steps to reproduce

- Join and leave room on a synapse server.
- Deploy a dendrite server on the same domain.
- Try to join the room again, without specifiying the old key in the dendrite config.

<!--
Describe how what happens differs from what you expected.

If you can identify any relevant log snippets from server logs, please include
those (please be careful to remove any personal or private data). Please surround them with
``` (three backticks, on a line on their own), so that they are formatted legibly.

Alternatively, please send logs to @kegan:matrix.org or @neilalexander:matrix.org
with a link to the respective Github issue, thanks!
-->
```
time="2020-10-09T22:44:19.854957496Z" level=info msg="Checking event signatures for 1352 events of room state" func="Check\n\t" file=" [github.com/matrix-org/gomatrixserverlib@v0.0.0-20201006143701-222e7423a5e3/federationtypes.go:409]" context=missing
time="2020-10-09T22:44:19.994178749Z" level=info msg="Fetching 85 key(s)" func="handleFetcherKeys\n\t" file=" [github.com/matrix-org/dendrite@/signingkeyserver/internal/api.go:208]" fetcher_name="perspective server matrix.org"
time="2020-10-09T22:44:20.426154563Z" level=info msg="Fetching 1 key(s)" func="handleFetcherKeys\n\t" file=" [github.com/matrix-org/dendrite@/signingkeyserver/internal/api.go:208]" fetcher_name=DirectKeyFetcher
[...]

PR #1520, opened 2020-10-14
Document dendrite server key format

aditsachde:
This is information I'd figured out and originally wrote [here](https://github.com/matrix-org/dendrite/issues/1508#issuecomment-706659730), but hopefully with less ambiguity. 

### Pull Request Checklist

* [X] I have added any new tests that need to pass to `sytest-whitelist` as specified in [docs/sytest.md](https://github.com/matrix-org/dendrite/blob/master/docs/sytest.md)
* [X] Pull request includes a [sign off](https://github.com/matrix-org/dendrite/blob/master/docs/CONTRIBUTING.md#sign-off)

Signed-off-by: `Adit Sachde github@aditsachde.com`


jakehemmerle:
nit; might be worth putting "It is important to include the equals sign" in bold for emphasis for those who don't read the documentation and just copy the PEM template visually. They might miss the padding

Kegsay (approved):
Thanks!

docs/serverkeyformat.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Converting Synapse Keys
+
+If you have signing keys from a previous synapse server, you should ideally configure them as `old_private_keys` in your Dendrite config file. Synapse stores signing keys in the following format.
+
+```
+ed25519 <Key Handle> <Base64 Encoded Key Data>
+```
+
+To convert this key to Dendrite's PEM format, use the following template. **It is important to include the equals sign, as the key data needs to be padded to 32 bytes.**
```
neilalexander, line 21: Is this true in all cases, or is it dependent on the key?

aditsachde: From experimental testing, Dendrite complains about an early EOF if the key part of the PEM file is less than 32 bytes, probably because the PEM parser expects 32 bytes. Any key less than this length needs to be padded.

The length of the decoded base64 string generated by synapse seems to be 30 bytes, so keys generated by Synapse definetly have to be padded.

PR #1521, opened 2020-10-14
Improved state handling in /send

neilalexander:
This fixes some bugs in the federation API `/send` endpoint where we'd only request state tuples for events needed for auth, but then try to send on that state as if it were the full state before an event.

This instead requests the entire room state, and updates `QueryStateAfterEvents` so that it performs state resolution.

### DIFF
--- a/roomserver/internal/input/input_latest_events.go
+++ b/roomserver/internal/input/input_latest_events.go
@@ -133,8 +133,7 @@ func (u *latestEventsUpdater) doUpdateLatestEvents() error {
 
 	// If the event has already been written to the output log then we
 	// don't need to do anything, as we've handled it already.
-	hasBeenSent, err := u.updater.HasEventBeenSent(u.stateAtEvent.EventNID)
-	if err != nil {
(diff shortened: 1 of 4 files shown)


roomserver/internal/input/input_latest_events.go
```diff
@@ -296,11 +298,13 @@ func (u *latestEventsUpdater) calculateLatest(
 	referenced, err := u.updater.IsReferenced(newEvent.EventReference)
 	if err != nil {
 		logrus.WithError(err).Errorf("Failed to retrieve event reference for %q", newEvent.EventReference.EventID)
+		return fmt.Errorf("u.updater.IsReferenced (new): %w", err)
```
Kegsay, line 301: Why are we now returning from here but didn't before?

Kegsay (approved):
Very good find! LGTM

roomserver/internal/input/input_latest_events.go
neilalexander: I put this in because it feels like a genuine error condition, e.g. we failed to talk to the database or something. It doesn't feel right to just fall through and possibly end up with a wrong state snapshot.

PR #1527, opened 2020-10-15
Start Kafka connections for each component that needs them

neilalexander:
Kafka has a limit on only being able to consume a topic once per connection. That's how #1506 happened, since we only created a single producer and a single consumer connection in the `setup` package and then tried to consume on the key-change topic from more than one place.

This refactors the setup so that components now have to request Kafka setup themselves, which might result in multiple connections to the Kafka broker, but it will at least mean that we can consume the same topic in multiple different components in monolith mode.

Note that this does *not* change how Naffka behaves, where we continue to use only a single connection.

### DIFF
--- a/internal/setup/base.go
+++ b/internal/setup/base.go
@@ -26,13 +26,9 @@ import (
 	"github.com/matrix-org/gomatrixserverlib"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 
-	"github.com/matrix-org/naffka"
-	naffkaStorage "github.com/matrix-org/naffka/storage"
-
 	"github.com/matrix-org/dendrite/internal"
 	"github.com/matrix-org/dendrite/userapi/storage/accounts"
 
-	"github.com/Shopify/sarama"
 	"github.com/gorilla/mux"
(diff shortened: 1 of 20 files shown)


internal/setup/base.go
```diff
@@ -335,30 +325,46 @@ func (b *BaseDendrite) SetupAndServeHTTP(
...
+func setupNaffka(cfg *config.Kafka) (sarama.Consumer, sarama.SyncProducer) {
+	if naffkaDatabase != nil {
+		return naffkaDatabase, naffkaDatabase
+	}
+	naffkaDB, err := naffkaStorage.NewDatabase(string(cfg.Database.ConnectionString))
 	if err != nil {
 		logrus.WithError(err).Panic("Failed to setup naffka database")
 	}
-	naff, err := naffka.New(naffkaDB)
+	naffkaDatabase, err = naffka.New(naffkaDB)
```
Kegsay, line 365: Misleading name, as `naffkaDB` is the database, this is just the naffka instance.

Kegsay (approved):

Issue #608, opened 2019-01-08
Implement read receipts


babolivier:
- [ ] Implement [m.receipt](https://matrix.org/docs/spec/client_server/r0.4.0.html#m-receipt)
- [ ] Implement [POST /_matrix/client/r0/rooms/{roomId}/receipt/{receiptType}/{eventId}](https://matrix.org/docs/spec/client_server/r0.4.0.html#post-matrix-client-r0-rooms-roomid-receipt-receipttype-eventid)
- [ ] Implement read receipt in [/sync](https://matrix.org/docs/spec/client_server/r0.4.0.html#get-matrix-client-r0-sync)

I'd suggest giving https://matrix.org/docs/spec/client_server/r0.4.0.html#id99 a good read before starting working on this one.

Kegsay:
Sytests:
```
    × POST /rooms/:room_id/receipt can create receipts
    × Read receipts are sent as events
    × Receipts must be m.read

    × Outbound federation sends receipts
    × Inbound federation rejects receipts from wrong remote

    × Read receipts appear in initial v2 /sync
    × New read receipts appear in incremental v2 /sync
```

S7evinK:
I'd like to give this a try, if there's nobody else already working on it.

S7evinK:
```
× Read receipts are sent as events
```

Isn't necessary, since its a test on a [deprecated](https://matrix.org/docs/spec/client_server/r0.6.1#deprecated-get-matrix-client-r0-events) endpoint

PR #1528, opened 2020-10-15
Implement read receipts

S7evinK:
Second draft of implementing receipts. (Will hopefully solve #608)
Feedback welcome.

### Pull Request Checklist


syncapi/storage/interface.go
```diff
@@ -147,4 +149,8 @@ type Database interface {
 	PutFilter(ctx context.Context, localpart string, filter *gomatrixserverlib.Filter) (string, error)
 	// RedactEvent wipes an event in the database and sets the unsigned.redacted_because key to the redaction event
 	RedactEvent(ctx context.Context, redactedEventID string, redactedBecause *gomatrixserverlib.HeaderedEvent) error
+	// StoreReceipt stores new receipt events ofr
```
Kegsay, line 152:
```suggestion
	// StoreReceipt stores new receipt events
```

syncapi/storage/shared/syncserver.go
```diff
@@ -563,21 +567,69 @@ func (d *Database) addTypingDeltaToResponse(
...
+					},
+				},
+			}
+
+			ev.Content, err = json.Marshal(content)
+			if err != nil {
+				return err
+			}
+			jr.Ephemeral.Events = append(jr.Ephemeral.Events, ev)
+			res.Rooms.Join[roomID] = jr
```
Kegsay, line 606: What happens if there is already a `JoinResponse` in `res`? This will replace it which doesn't sound right?

syncapi/storage/shared/syncserver.go
```diff
@@ -563,21 +567,69 @@ func (d *Database) addTypingDeltaToResponse(
...
 }
 
+// addReceiptDeltaToResponse adds all receipt information to a sync response
+// since the specified position
+func (d *Database) addReceiptDeltaToResponse(
+	since types.StreamingToken,
+	joinedRoomIDs []string,
+	res *types.Response,
+) error {
+	var jr types.JoinResponse
```
Kegsay, line 577: By declaring this here won't this pollute read receipts across mutliple rooms?

syncapi/storage/tables/interface.go
```diff
@@ -156,3 +157,8 @@ type Filter interface {
 	SelectFilter(ctx context.Context, localpart string, filterID string) (*gomatrixserverlib.Filter, error)
 	InsertFilter(ctx context.Context, filter *gomatrixserverlib.Filter, localpart string) (filterID string, err error)
 }
+
+type Receipts interface {
+	UpsertReceipt(ctx context.Context, txn *sql.Tx, roomId, receiptType, userId, eventId string, timestamp gomatrixserverlib.Timestamp) (pos types.StreamPosition, err error)
+	SelectRoomReceiptsAfter(ctx context.Context, roomId string, streamPos types.StreamPosition) ([]eduAPI.InputReceiptEvent, error)
```
Kegsay, line 163: It feels wrong for the storage format here to be `eduAPI.InputReceiptEvent` - I could under it being the *output* event as that is what we consume, but to have it as the input feels wrong.

syncapi/storage/postgres/receipt_table.go
```diff
@@ -0,0 +1,106 @@
...
+	for rows.Next() {
+		r := api.InputReceiptEvent{}
+		err = rows.Scan(&r.RoomID, &r.Type, &r.UserID, &r.EventID, &r.Timestamp)
+		if err != nil {
+			return res, errors.Wrap(err, "unable to scan row to api.Receipts")
+		}
+		res = append(res, r)
+	}
+	if rows.Err() != nil || rows.Close() != nil {
+		return res, errors.Wrap(err, "error while scanning rows or error closing rows")
```
Kegsay, line 103: `err` will be nill in these cases. You'll need to unpack this e.g `if err = rows.Err(); err != nil {`

syncapi/storage/postgres/receipt_table.go
```diff
@@ -0,0 +1,106 @@
...
+	rows, err := r.selectRoomReceipts.QueryContext(ctx, roomId, streamPos)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to query room receipts")
+	}
+	var res []api.InputReceiptEvent
+	for rows.Next() {
+		r := api.InputReceiptEvent{}
+		err = rows.Scan(&r.RoomID, &r.Type, &r.UserID, &r.EventID, &r.Timestamp)
+		if err != nil {
+			return res, errors.Wrap(err, "unable to scan row to api.Receipts")
```
Kegsay, line 98: We fail to close `rows` in this case, prefer `defer rows.Close() // nolint: errcheck` on `:93`.

Kegsay: Also, please use native error wrapping in Go 1.13 as previously discussed.

syncapi/storage/postgres/receipt_table.go
```diff
@@ -0,0 +1,106 @@
...
+	return r, nil
+}
+
+func (r *receiptStatements) UpsertReceipt(ctx context.Context, txn *sql.Tx, roomId, receiptType, userId, eventId string, timestamp gomatrixserverlib.Timestamp) (pos types.StreamPosition, err error) {
+	stmt := sqlutil.TxStmt(txn, r.upsertReceipt)
+	err = stmt.QueryRowContext(ctx, roomId, receiptType, userId, eventId, timestamp).Scan(&pos)
+	return
+}
+
+func (r *receiptStatements) SelectRoomReceiptsAfter(ctx context.Context, roomId string, streamPos types.StreamPosition) ([]api.InputReceiptEvent, error) {
```
Kegsay, line 88: Why is this being done per room? This will scale pretty badly when the user joins lots of rooms, especially since we _always_ call this receipt function in every sync response. In other cases we use `WHERE room_id IN $1` or equivalent to batch up the requests.

syncapi/storage/sqlite3/receipt_table.go
```diff
@@ -0,0 +1,114 @@
...
+);
+CREATE INDEX IF NOT EXISTS syncapi_receipts_room_id_idx ON syncapi_receipts(room_id);
+`
+
+const upsertReceipt = "" +
+	"INSERT INTO syncapi_receipts" +
+	" (id, room_id, receipt_type, user_id, event_id, receipt_ts)" +
+	" VALUES ($1, $2, $3, $4, $5, $6)" +
+	" ON CONFLICT (room_id, receipt_type, user_id)" +
+	" DO UPDATE SET id = $1, event_id = $5, receipt_ts = $6"
```
Kegsay, line 52: This probably won't work as you expect on SQLite. You'll probably need to use `$7 $8 $9`

syncapi/storage/sqlite3/receipt_table.go
```diff
@@ -0,0 +1,114 @@
...
+	rows, err := r.selectRoomReceipts.QueryContext(ctx, roomId, streamPos)
+	if err != nil {
+		return []api.InputReceiptEvent{}, errors.Wrap(err, "unable to query room receipts")
+	}
+	var res []api.InputReceiptEvent
+	for rows.Next() {
+		r := api.InputReceiptEvent{}
+		err = rows.Scan(&r.RoomID, &r.Type, &r.UserID, &r.EventID, &r.Timestamp)
+		if err != nil {
+			return res, errors.Wrap(err, "unable to scan row to types.Receipts")
```
Kegsay, line 106: Missing `rows.Close()` if this is hit.

Kegsay (changes requested):
By and large this is an excellent PR, thank you! There's a few notable things to resolve:
 - Please use Go 1.13 error wrapping `fmt.Errorf("foo %w", err)` rather than importing a new errors package.
 - Please batch up requests when selecting receipts after a stream position in the `syncapi`. Currently this will be inefficient as it'll result in 100 select statements for every `/sync` request every time a user who happens to be in 100 rooms syncs.

syncapi/storage/shared/syncserver.go
S7evinK: Not yet sure if the solution to this will work properly.

S7evinK:
Somehow don't see the issue with the insert/select for postgres. :(
Looks fine to me, even after some hours of checking that issue.

syncapi/storage/postgres/deltas/20201017140347_create_receipt_table.go
```diff
@@ -0,0 +1,46 @@
+package deltas
```
Kegsay, line 1: You don't need to write a delta for brand new tables. The code will CREATE TABLE IF NOT EXISTS.

syncapi/storage/postgres/syncserver.go
```diff
@@ -43,44 +47,73 @@ func NewDatabase(dbProperties *config.DatabaseOptions) (*SyncServerDatasource, e
...
 	}
+
+	// Create tables before executing migrations so we don't fail if the table is missing,
+	// and THEN prepare statements so we don't fail due to referencing new columns
+	r := receiptStatements{}
+	if err = r.execSchema(d.db); err != nil {
+		return nil, err
+	}
+	m := sqlutil.NewMigrations()
+	deltas.LoadCreateReceiptTable(m)
```
Kegsay, line 63: No need to do deltas for new tables.

syncapi/storage/shared/syncserver.go
```diff
@@ -554,21 +558,84 @@ func (d *Database) addTypingDeltaToResponse(
...
-) (err error) {
-
+) error {
 	if fromPos.EDUPosition() != toPos.EDUPosition() {
-		err = d.addTypingDeltaToResponse(
-			fromPos, joinedRoomIDs, res,
-		)
+		// add typing deltas
+		if err := d.addTypingDeltaToResponse(fromPos, joinedRoomIDs, res); err != nil {
+			return errors.Wrap(err, "unable to apply typing delta to response")
```
Kegsay, line 626: Please do not use `pkg/errors`, we use `%w` in Go 1.13+

Kegsay (approved):
Few minor things, this is looking great! Thanks so much!

syncapi/storage/postgres/deltas/20201017140347_create_receipt_table.go
Kegsay: So just remove these delta files.

Kegsay (approved):
Assuming CI is happy, SHIPIT

PR #1531, opened 2020-10-16
KindOld

neilalexander:
This adds `KindOld` to the roomserver input API so that we don't try to send backfilled or missing events through the latest events updater.

### DIFF
--- a/roomserver/api/output.go
+++ b/roomserver/api/output.go
@@ -183,8 +183,14 @@ func (ore *OutputNewRoomEvent) AddsState() []gomatrixserverlib.HeaderedEvent {
 }
 
 // An OutputOldRoomEvent is written when the roomserver receives an old event.
+// This will typically happen as a result of getting either missing events
+// or backfilling. Downstream components may wish to send these events to
+// clients when it is advantageous to do so, but with the consideration that
+// the event is likely a historic event.
+//
 // Old events do not update forward extremities or the current room state,
-// therefore they must not be treated as if they do.
+// therefore they must not be treated as if they do. Downstream components
+// should build their current room state up from OutputNewRoomEvents only.
 type OutputOldRoomEvent struct {
 	// The Event.
 	Event gomatrixserverlib.HeaderedEvent `json:"event"`
--- a/roomserver/api/output.go
+++ b/roomserver/api/output.go
@@ -91,9 +91,6 @@ const (
 type OutputNewRoomEvent struct {
 	// The Event.
 	Event gomatrixserverlib.HeaderedEvent `json:"event"`
-	// Is the event a forward extremity in the room at the time of the output event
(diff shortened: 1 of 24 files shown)


roomserver/api/wrapper.go
```diff
@@ -58,7 +60,7 @@ func SendEventWithState(
 			continue
 		}
 		ires = append(ires, InputRoomEvent{
-			Kind:         KindOutlier,
+			Kind:         kind,
```
Kegsay, line 63: Why would we ever want this to NOT be KindOutlier? Also docs for this function are now lying

Kegsay: I'd expect to see this on `:75` not here.

roomserver/internal/input/input_events.go
```diff
@@ -119,7 +119,7 @@ func (r *Inputer) processRoomEvent(
 		// We haven't calculated a state for this event yet.
 		// Lets calculate one.
 		err = r.calculateAndSetState(ctx, input, *roomInfo, &stateAtEvent, event, isRejected)
-		if err != nil {
+		if err != nil && input.Kind != api.KindOld {
```
Kegsay, line 122: If err != nil but the kind is kind old we'll fall through which doesn't sound right?

Kegsay: Especially as we'll fallthrough without logging `err`

roomserver/api/wrapper.go
Kegsay: Moving this to `:75` makes `Outbound federation can backfill events... OK` but `Backfilled events whose prev_events are in a different room do not allow cross-room back-pagination... FAIL` still

neilalexander: You're right, I managed to edit the wrong line there.

roomserver/internal/input/input_events.go
neilalexander: Well, this is something I need to think about, because it's entirely possible that we'll fail to work out the state before a `KindOld` event but that doesn't mean we shouldn't then go on to handle redactions etc.

roomserver/api/output.go
```diff
@@ -178,6 +182,14 @@ func (ore *OutputNewRoomEvent) AddsState() []gomatrixserverlib.HeaderedEvent {
 	return append(ore.AddStateEvents, ore.Event)
 }
 
+// An OutputOldRoomEvent is written when the roomserver receives an old event.
+// Old events do not update forward extremities or the current room state,
+// therefore they must not be treated as if they do.
+type OutputOldRoomEvent struct {
```
Kegsay, line 188: Explain about whether they should notify clients please.

Kegsay (approved):

PR #1532, opened 2020-10-16
Take write lock for rate limit map

neilalexander:
This fixes a problem where we can write to the rate limit map whilst only having a read lock.

It also ensures that the cleaner goroutine isn't operating on the map while we try to do any of this.

### DIFF
--- a/clientapi/routing/rate_limiting.go
+++ b/clientapi/routing/rate_limiting.go
@@ -73,8 +73,13 @@ func (l *rateLimits) rateLimit(req *http.Request) *util.JSONResponse {
 	// let's create one.
 	rateLimit, ok := l.limits[caller]
 	if !ok {
-		l.limits[caller] = make(chan struct{}, l.requestThreshold)
-		rateLimit = l.limits[caller]
+		rateLimit = make(chan struct{}, l.requestThreshold)
+
+		l.limitsMutex.RUnlock()
+		l.limitsMutex.Lock()
+		l.limits[caller] = rateLimit
+		l.limitsMutex.Unlock()
+		l.limitsMutex.RLock()
 	}
 
 	// Check if the user has got free resource slots for this request.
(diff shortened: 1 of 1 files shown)


clientapi/routing/rate_limiting.go
```diff
@@ -38,6 +39,7 @@ func (l *rateLimits) clean() {
 		// empty. If they are then we will close and delete them,
 		// freeing up memory.
 		time.Sleep(time.Second * 30)
+		l.cleanMutex.Lock()
```
Kegsay, line 42: Why have we sprouted another mutex? Double locking makes me scared of deadlocks.

neilalexander: The alternative is that we release the read lock, take a write lock, make the write, release the write lock, retake the read lock, and in that time, the cleaner goroutine might have taken the lock instead.

This just stops the cleaner goroutine from running while requests are in flight.

Kegsay: The cleaner was taking a write lock to clean, so I don't see any problem?

neilalexander: The cleaner closes the channel if it thinks it's no longer needed, so the alternative was something less ideal, like:
```
l.limits[caller] = make(chan struct{}, l.requestThreshold)

l.limitsMutex.RUnlock()
l.limitsMutex.Lock()
rateLimit = l.limits[caller]
l.limitsMutex.Unlock()

// cleaner could win the lock and run here, closing our channel, and we haven't used it yet

l.limitsMutex.RLock()
```

Kegsay: Can we avoid this by assigning it into `l.limits` after we've done our work with it?

neilalexander: Another HTTP request could come in in the meantime, which would result in more than one channel for the caller.

Kegsay: OK then, SHIPIT

PR #1537, opened 2020-10-17
Retrieve remote file size without the Content-Size header

### LINKED ISSUE #1525: Downloading pictures from federated Synapse instance fails
### Background information

- **Dendrite version or git SHA**: 286dd408ae623b218f9c5a97d53f174c0ea19cc2
- **Running in Docker?**:  No
- **`go version`**: go1.15.2 linux/arm

### Description

When sending pictures from a federated synapse instance onto a public room of my dendrite server, the picture doesn't show up and when clicking on it we get the following JSON error :
```json
{"errcode":"M_NOT_FOUND","error":"Failed to download: error querying the database.: invalid response from remote server: strconv.ParseInt: parsing \"\": invalid syntax"}
```

### Some clues

Problem seems to be located at the following line : https://github.com/matrix-org/dendrite/blob/e3c2b081c7c197cad931d6525931dc9e960b93c3/mediaapi/routing/download.go#L696 

The header `Content-Length` seems to be missing from the HTTP response provided by the Synapse instance. Note that the problem doesn't appear with several other instances behind reverse proxies, which is not the case of the one I'm mentioning. 

Since the problem did not appear between instances of Synapse, I looked at Synapse's way of downloading pictures from federated instances and it effectively doesn't appear to use the `Content-Length` header to deduce file size (see https://github.com/matrix-org/synapse/blob/v1.21.0/synapse/rest/media/v1/media_repository.py#L332 and https://github.com/matrix-org/synapse/blob/1c262431f9bf768d106bf79a568479fa5a0784a1/synapse/http/matrixfederationclient.py#L939)

Should I report this problem in the Synapse repo as well?
I can possibly come up with a PR, just need some advice on what should be done.

Thanks for your help :)

darkgallium:
Fixes #1525.
This patch gets the size of the HTTP body by reading the size of the body buffer of the HTTP response object rather than relying on the Content-Size header that can be missing.

### Pull Request Checklist

Signed-off-by: `Florian Le Minoux <darkgallium+git@deuxfleurs.fr>`


neilalexander (commented):
It may be worth using the `Content-Length` header *if it is present*, and falling back to this as a last resort?

darkgallium:
> It may be worth using the `Content-Length` header _if it is present_, and falling back to this as a last resort?
Agreed and done.

mediaapi/routing/download.go
```diff
@@ -673,6 +675,36 @@ func (r *downloadRequest) fetchRemoteFileAndStoreMetadata(
...
+		if parseErr != nil {
+			r.Logger.WithError(parseErr).Warn("Failed to parse content length")
+			return 0, nil, errors.Wrap(parseErr, "invalid response from remote server")
+		}
+
+		reader = *body
+		contentLength = parsedLength
+	} else {
+		// Content-Length header is missing, we need to read the whole body to get its length
+		bodyBytes, readAllErr := ioutil.ReadAll(*body)
```
Kegsay, line 694: This is going to be a problem if the file is huge as it'll all be put into memory. Why do you need to do this? I _think_ the previous solution didn't do this but because you force pushed I can't cross-reference with what was previously done.

Lesterpig: Suggestion: return 0 by default. Final size is set using `bytesWritten` from `hash, bytesWritten, tmpDir, err := fileutils.WriteTempFile(ctx, reader, maxFileSizeBytes, absBasePath)` anyway.

mediaapi/routing/download.go
```diff
@@ -673,6 +674,43 @@ func (r *downloadRequest) fetchRemoteFileAndStoreMetadata(
...
+func (r *downloadRequest) GetContentLengthAndReader(contentLengthHeader string, body *io.ReadCloser, maxFileSizeBytes config.FileSizeBytes) (int64, io.Reader, error) {
+	reader := *body
+	var contentLength int64
+
+	if contentLengthHeader != "" {
+		// A Content-Length header is provided. Let's try to parse it.
+		parsedLength, parseErr := strconv.ParseInt(contentLengthHeader, 10, 64)
+		if parseErr != nil {
+			r.Logger.WithError(parseErr).Warn("Failed to parse content length")
+			return 0, nil, errors.Wrap(parseErr, "invalid response from remote server")
```
Kegsay, line 686: Can we please use `%w` - we shouldn't need to use `pkg/errors`.

Kegsay (approved):

ferologics:
@neilalexander @darkgallium could this change be extended to read content length of an upload request? We are getting an [error for missing Content-Length header](https://github.com/matrix-org/dendrite/blob/802f1c96f804f7a146e4e12e25b20c980a6af870/mediaapi/routing/upload.go#L227) when using the [Matrix iOS SDK](https://github.com/matrix-org/matrix-ios-sdk).

Kegsay:
Please open an issue for this @ferologics

ferologics:
@Kegsay done.

mediaapi/routing/download.go
```diff
@@ -673,6 +674,43 @@ func (r *downloadRequest) fetchRemoteFileAndStoreMetadata(
...
+	var contentLength int64
+
+	if contentLengthHeader != "" {
+		// A Content-Length header is provided. Let's try to parse it.
+		parsedLength, parseErr := strconv.ParseInt(contentLengthHeader, 10, 64)
+		if parseErr != nil {
+			r.Logger.WithError(parseErr).Warn("Failed to parse content length")
+			return 0, nil, fmt.Errorf("strconv.ParseInt: %w", parseErr)
+		}
+		if parsedLength > int64(maxFileSizeBytes) {
```
database64128, line 688: This has caused a regression where `media_api.max_file_size_bytes = 0` leads to failed media downloads.

PR #1538, opened 2020-10-18
Added systemd polylith service install script and fixed/appended INSTALL.md

dymattic:
Added a bash script for creating systemd unit files and enabling them:
- can be used for automated install
- lets user choose which servers to install services for

Changes to INSTALL.md:
- fixed some errors in INSTALL.md (at least what I noticed during install) 
- added section for polylith service setup

Signed-off-by: `Dmitrij Pastian dymattic@naise.io`


docs/INSTALL.md
```diff
@@ -109,7 +112,7 @@ Assuming that Postgres 9.5 (or later) is installed:
 * Create the component databases:
 
   ```bash
-  for i in account device mediaapi syncapi roomserver signingkeyserver federationsender appservice e2ekey naffka; do
+  for i in account device mediaapi syncapi roomserver signingkeyserver federationsender appservice e2ekey naffka userapi_accounts userapi_devices keyserver ; do
```
neilalexander, line 115: We already have `account`, `device` and `signingkeyserver` here. Is what you are saying with this change that there’s a sample config somewhere that disagrees with these instructions?

dymattic: Interesting, I didn't notice! 
It seems like the default config is looking for 
dendrite_userapi_accounts, dendrite_userapi_devices and dendrite_keyserver 
instead of 
dendrite_account, dendrite_device, and dendrite_ signingkeyserver.

dymattic: I think it isn't mentioned in particular ( at least I didn't find docs for switching to postgresql ).

I basically used the naming scheme of the sqlite-databases:

```
# Configuration for the Key Server (for end-to-end encryption).
key_server:
  internal_api:
    listen: http://localhost:7779
    connect: http://localhost:7779
  database:
    connection_string: file:keyserver.db
    max_open_conns: 100
    max_idle_conns: 2
    conn_max_lifetime: -1
```
to
```
# Configuration for the Key Server (for end-to-end encryption).
key_server:
  internal_api:
    listen: http://localhost:7779
    connect: http://localhost:7779
  database:
    connection_string: postgresql://dendrite:<password>@localhost/dendrite_keyserver
    max_open_conns: 100
    max_idle_conns: 2
    conn_max_lifetime: -1
```
___

```
# Configuration for the User API.
user_api:
  internal_api:
    listen: http://localhost:7781
    connect: http://localhost:7781
  account_database:
    connection_string: file:userapi_accounts.db
    max_open_conns: 100
    max_idle_conns: 2
    conn_max_lifetime: -1
  device_database:
    connection_string: file:userapi_devices.db
    max_open_conns: 100
    max_idle_conns: 2
    conn_max_lifetime: -1
```

to

```
# Configuration for the User API.
user_api:
  internal_api:
    listen: http://localhost:7781
    connect: http://localhost:7781
  account_database:
    connection_string: postgresql://dendrite:<password>@localhost/dendrite_userapi_accounts
    max_open_conns: 100
    max_idle_conns: 2
    conn_max_lifetime: -1
  device_database:
    connection_string: postgresql://dendrite:<password>@localhost/dendrite_userapi_devices
    max_open_conns: 100
    max_idle_conns: 2
    conn_max_lifetime: -1
```
___

```
# Configuration for the Signing Key Server (for server signing keys).
signing_key_server:
  internal_api:
    listen: http://localhost:7780
    connect: http://localhost:7780
  database:
    connection_string: file:signingkeyserver.db
    max_open_conns: 100
    max_idle_conns: 2
    conn_max_lifetime: -1
```
to

```
# Configuration for the Signing Key Server (for server signing keys).
signing_key_server:
  internal_api:
    listen: http://localhost:7780
    connect: http://localhost:7780
  database:
    connection_string: postgresql://dendrite:<password>@localhost/dendrite_signingkeyserver
    max_open_conns: 100
    max_idle_conns: 2
    conn_max_lifetime: -1
```

S7evinK:
Couldn't we use a systemd template file for this?
That would allow us to control a service by
`systemctl restart dendrite@mediaapi`
(where `mediaapi` may be replaced by any other component)
Unfortunately.. That's dependent on the systemd version.

dymattic:
I think this actually might be a great idea.
I will include an option in the script that lets the user decide which method to use.

Do you know if there is a way to check whether the systemd version supports templated services?

S7evinK:
Not sure, i just happen to remember that pre systemd 230 (i think) you couldn't use `%i` for the instance name directly in ExecStart as the binary (`ExecStart=/usr/bin/dendrite-%i` wouldn't work; had to be worked around by using something like `ExecStart=/bin/bash -c /usr/bin/dendrite-%i`)

neilalexander:
Afraid that v0.2.0 has probably made a bit of a mess of some of this PR, as we no longer have the separate polylith binaries, but just one, e.g.
```
./dendrite-polylith-multi roomserver
./dendrite-polylith-multi clientapi
./dendrite-polylith-multi syncapi
```
... etc.

dymattic:
@neilalexander Actually the changes you made with this release have saved me quite some headaches :D 
should be ready now

neilalexander (commented):
Overall this is neat, but I think it would be much more useful for people if the template service file and the installation script were separated out. Bit easier to audit and use for people who don’t like running scripts.

docs/systemd/polylith-example@.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+	Description= Dendrite Polylith Multi - %I
+	PartOf=polyDendrite.service
```
S7evinK, line 3: I'm not a fan of camelCase in unit files. Maybe change that to `PartOf=dendrite-polylith.service` (do we even need that?)
```suggestion
	PartOf=dendrite-polylith.service
```

docs/systemd/polylith-example@.service
```diff
@@ -0,0 +1,14 @@
...
+	After=network.target
+
+	[Service]
+	User=dendrite
+	WorkingDirectory=/opt/dendrite
+	Type=forking
```
S7evinK, line 9: Should probably be `Type=simple`, since Dendrite doesn't fork a new process.
```suggestion
	Type=simple
```

docs/systemd/polylith-example@.service
```diff
@@ -0,0 +1,14 @@
...
+	ExecStart=/opt/dendrite/bin/dendrite-polylith-multi --config=dendrite.yaml %i
+	Restart=on-failure
```
S7evinK, line 11:
```suggestion
	Restart=on-failure
	RestartSec=5s
```

Simply to let Kafka, Postgres or other components recover and not "DOS" them back to a failure state. :P

install-polylith-systemd-units.sh
```diff
@@ -0,0 +1,186 @@
...
+
+	[Service]
+	User=$USER
+	WorkingDirectory=$DENDRITEDIR
+	Type=forking
+	ExecStart=$DENDRITEDIR/bin/dendrite-polylith-multi --config=dendrite.yaml %i
+	Restart=on-failure
+
+	[Install]
+	WantedBy=multi-user.target
```
S7evinK, line 67: See suggestions above for the unit file.

S7evinK (commented):
Mostly checked the service, so no opinions on the install script.

Kegsay:
We've thought about this a lot and have decided that we do not want to have install scripts like this at this time. Entry scripts are the main interface between end-users and this project and as such have many design/API decisions around them. We're not yet at the stage where we feel we can add scripts and maintain them. Thank you for your contribution.

saerdnaer:
@Kegsay Could you merge the other parts of this PR besides the install script? At least me personally, it would have saved one hour or so...

Kegsay:
Which other bits exactly were helpful to you?

saerdnaer:
I can create PR if that helps...?

Issue #1329, opened 2020-08-24
Fix failing room state endpoints (setting state twice)


kegsay:
Sytests:
```
    × Setting state twice is idempotent
```

MadLittleMods:
Looks related to https://github.com/matrix-org/dendrite/issues/1311

piyushsingariya:
Is this issue still open? Can I have a look over this?

MadLittleMods:
Go for it @piyushsingariya!

piyushsingariya:
@MadLittleMods Can you explain to me a bit about this issue? as I am new and having a hard time understanding Dendrite's issues and its codebase. It would be helpful if you could guide me.
Thanks in advance :)

keremgocen:
hey @piyushsingariya were you able to progress on this one? I can try to help if you need any or I'm also happy to take a look myself, looking for a #good-first-issue to get myself familiarised with this dendrite.

piyushsingariya:
yeah go ahead @keremgocen , I'm not focusing on this.

keremgocen:
I'm looking at this and left a comment [here](https://github.com/matrix-org/dendrite/pull/1539/files#r667537781). It's not obvious to me what should be the fix;
- should I look at https://github.com/matrix-org/dendrite/blob/master/clientapi/routing/joinroom.go and check if the user is already joined somehow
- or it's about the event being returned and the new event returned must have a different id here https://github.com/matrix-org/dendrite/blob/master/clientapi/routing/sendevent.go

Not sure how the 2nd event ends up having the same id, when I look at the failing test code here: https://github.com/matrix-org/sytest/blob/master/tests/30rooms/01state.pl#L287

Any guidance would be helpful as this is my first issue, thanks.

Tert0:
I get
```
Test 137 Setting state twice is idempotent... EXPECTED FAIL
```
as output for SyTests.

Does that mean that it works?

kegsay:
No, you need it to be `... OK`

ParthGala2k:
Hello, I plan to apply to matrix for the dendrite project as part of GSoC this year and was looking to familiarise myself with the codebase.
If no one is looking at this issue, is it okay if I have a go at it ?

kegsay:
Still failing as of today: `2022-05-20T11:37:09.7789547Z     Test 594 outliers whose auth_events are in a different room are correctly rejected... EXPECTED FAIL`

kegsay:
[...]

PR #1539, opened 2020-10-18
Fix state endpoints

djohnson1865:
Should fix underlying issues in #1329 

Doesn't fix the tests associated because it looks like those invoke `/rooms/{roomID}/initialSync`which returns a forbidden status. 
Modifying the return status  in clientapi/routing/routing.go will cause the tests to pass, however wasn't sure of any underlying client bugs that might cause. 

### Pull Request Checklist

Signed-off-by: `Devon Johnson <djohnson1865@gmail.com>`

### DIFF
--- a/clientapi/routing/sendevent.go
+++ b/clientapi/routing/sendevent.go
@@ -15,6 +15,7 @@
 package routing
 
 import (
+	"bytes"
 	"net/http"
 	"sync"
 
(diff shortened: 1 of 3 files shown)


djohnson1865:
Can't seem to replicate the failing test on my local build of Dendrite running the latest sytest docker image.

clientapi/routing/sendevent.go
```diff
@@ -80,6 +81,29 @@ func SendEvent(
...
+	stateRes := api.QueryLatestEventsAndStateResponse{}
+	if err := rsAPI.QueryLatestEventsAndState(req.Context(), &api.QueryLatestEventsAndStateRequest{
+		RoomID:       roomID,
+		StateToFetch: []gomatrixserverlib.StateKeyTuple{},
+	}, &stateRes); err != nil {
+		util.GetLogger(req.Context()).WithError(err).Error("queryAPI.QueryLatestEventsAndState failed")
+		return jsonerror.InternalServerError()
+	}
+
+	// Check that currnet event doesn't match previous events
```
neilalexander, line 93: I'm not really sure why we would want to do this? It doesn't really seem related to the fixes at hand.

Not to mention that the `QueryLatestEventsAndState` operation is an *expensive* thing to be calling on every sent event.

clientapi/routing/state.go
```diff
@@ -141,7 +141,7 @@ func OnIncomingStateRequest(ctx context.Context, device *userapi.Device, rsAPI a
 			util.GetLogger(ctx).WithError(err).Error("Failed to QueryMembershipForUser")
 			return jsonerror.InternalServerError()
 		}
-		for _, ev := range stateRes.StateEvents {
+		for _, ev := range stateAfterRes.StateEvents {
```
neilalexander, line 144: This is a good catch!

neilalexander (changes requested):
Thanks for catching the `state.go` bug there—just don't really understand the changes in `sendevent.go`.

neilalexander:
We've accepted commit `fdbf47b` from your branch into `master`, for addressing other tests it might be wise to open separate PRs.

Thanks for the contribution!

clientapi/routing/sendevent.go
djohnson1865: This could easily not be the most optimal way to check for previous events, first blush at hacking on Dendrite. It was meant to be an attempt at making setting state idempotent by sending the previous identical event rather than sending a newly created event. 

Not sure if there is a better place in the code base to check for this, or a less expensive way to do so.

keremgocen: I'm looking at #1329, what would be the recommended fix for this?

PR #1546, opened 2020-10-19
fix create-account

ThatNerdyPikachu:
### Pull Request Checklist

Signed-off-by: `Pika <pika@lasagna.dev>`

This PR fixes `create-account` by removing the need to create a device, so it only uses the singular user database. In the future, this should be changed to read values from `dendrite.yaml`, rather than requiring manual input for the database and server name fields.


Kegsay:
Seems reasonable, as people can just `curl localhost` to login and then make a device. Thoughts @neilalexander ?

neilalexander (approved):
Yep, I'm happy to remove the device creation - it seems largely pointless.

PR #1548, opened 2020-10-20
Exclude old state events from sync when consumed from roomserver

### LINKED ISSUE #1547: Events shown in wrong order
### Background information

- **Dendrite version or git SHA**: 0.2rc
- **Running in Docker?**: no
- **`go version`**: 1.15

got a room on dendrite federated to matrix.org. events are shown in the order of Thursday, Saturday and then Friday seen from matrix.org 🤔

neilalexander:
When consuming old room events from the roomserver, the sync API was not excluding them from sync, which I think is responsible for some older events appearing incorrectly in the timeline. This should therefore fix #1547.


neilalexander:
As far as I can tell, the sytests that this has caused to fail probably shouldn't have been passing before.

The sync API doesn't currently have the ability to insert events into the history of the timeline - they can only get added as if they had just happened - and I get the feeling this has caused us to pass a couple of tests by fluke.

Half-Shot:
>The sync API doesn't currently have the ability to insert events into the history of the timeline - they can only get added as if they had just happened

Concerns from the peanut gallery: If I send dendrite an extremely delayed transaction from my HS, containing events from ages ago, will the sync API treat those as new?

Kegsay (approved):
Sounds plausible and a reasonable compromise for now!

PR #1550, opened 2020-10-20
Remove invalid state delta check

neilalexander:
This should no longer be needed since:

- the reverse topological ordering is fixed
- backfilled events are no longer treated as new and therefore not candidates to become forward extremities


Kegsay (approved):
Assuming CI is happy, for some definition of happy..

PR #1552, opened 2020-10-20
Multi-personality polylith binary

neilalexander:
This gets rid of all the various polylith components in `./cmd` and replaces them with a single `./cmd/dendrite-polylith-multi`, which takes the name of the component as the first argument, i.e.
```
./bin/dendrite-polylith-multi clientapi
./bin/dendrite-polylith-multi --config=dendrite.yaml roomserver
```

This also updates the Docker stuff so we now have exactly two images: `matrixdotorg/dendrite-monolith` and `matrixdotorg/dendrite-polylith`, and have updated all of the Docker samples to match.

This also updates the documentation to state that the minimum PostgreSQL version is 9.6.


go.mod
```diff
@@ -42,6 +42,8 @@ require (
 	golang.org/x/crypto v0.0.0-20200820211705-5c72a883971a
 	gopkg.in/h2non/bimg.v1 v1.1.4
 	gopkg.in/yaml.v2 v2.3.0
+	honnef.co/go/tools v0.0.1-2019.2.3
+	upspin.io v0.0.0-20200124224713-5ddde7b8e6ff
```
Kegsay, line 46: How did we sprout these deps?

Kegsay (approved):

go.mod
neilalexander: No idea. Upstream dependency maybe?

PR #1556, opened 2020-10-21
Refactor forward extremities

neilalexander:
This is another refactor of forward extremities so that we don't check for references from outliers etc, which we were seemingly getting wrong even before the deep-checking PR.

It also adds a `resolve-state` utility for probing the resolved state for one or more state snapshots from the database, which is a useful debugging tool when trying to figure out where state changed/disappeared.

### DIFF
--- a/cmd/resolve-state/main.go
+++ b/cmd/resolve-state/main.go
@@ -1,0 +1,129 @@
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"os"
+	"strconv"
+
+	"github.com/matrix-org/dendrite/internal/caching"
+	"github.com/matrix-org/dendrite/internal/setup"
+	"github.com/matrix-org/dendrite/roomserver/state"
+	"github.com/matrix-org/dendrite/roomserver/storage"
+	"github.com/matrix-org/dendrite/roomserver/types"
+	"github.com/matrix-org/gomatrixserverlib"
+)
+
+// This is a utility for inspecting state snapshots and running state resolution
+// against real snapshots in an actual database.
+// It takes one or more state snapshot NIDs as arguments, along with a room version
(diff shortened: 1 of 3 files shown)


cmd/resolve-state/main.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/matrix-org/dendrite/roomserver/state"
+	"github.com/matrix-org/dendrite/roomserver/storage"
+	"github.com/matrix-org/dendrite/roomserver/types"
+	"github.com/matrix-org/gomatrixserverlib"
+)
+
+// This is a utility for inspecting state snapshots and running state resolution
+// against real snapshots in an actual database.
+// It takes one or more state snapshot NIDs as arguments, along with a room version
+// to use for unmarshalling events, and will produce resolved output.
```
Kegsay, line 21: Including actual usage would be nice e.g `$ resolve-state -roomversion 5 what_goes_here`

Kegsay (approved):
Can you please summarise in a PR comment for posterity how the previous algorithm was wrong and how this PR fixes it. I grok your actions but I don't know how this is more correct than before.

neilalexander:
There are two things that this PR fixes.

The first is that, beforehand, we were checking if any event known to the roomserver was referencing a forward extremity rather than just known extremities. This included outliers and other events that weren't woven into the room topology yet, so it could cause us to accidentally get rid of forward extremities when we didn't really mean to (and any state after them in the process).

The PR now ensures that we're only checking references *between extremities*, so we might accumulate more extremities as a result, but we won't get rid of any until they've actually been superseded on the graph.

The second is that, because of the fact that we could potentially invalidate all of our extremities by outliers, that meant that we could potentially fall back on just using the new event and all of it's associated state, and it would never be resolved against potentially other good extremities.

PR #1557, opened 2020-10-21
Fix RewritesState bug

neilalexander:
Apparently we were overwriting `RewritesState` when generating new output events and didn't bother to check `u.rewritesState` the second time. This meant that we were probably sending rewrite output events far more often than we intended to, confusing downstream component state.

This also updates the sync API to only discard state, rather than the entire room history, since that can cause problems with events going missing from the timeline, and also fixes a bug that can result in a panic if the federation sender gets the same joined hosts more than once.

### DIFF
--- a/roomserver/internal/input/input_latest_events.go
+++ b/roomserver/internal/input/input_latest_events.go
@@ -114,7 +114,6 @@ type latestEventsUpdater struct {
 	newStateNID types.StateSnapshotNID
 }
 
-// nolint:gocyclo
 func (u *latestEventsUpdater) doUpdateLatestEvents() error {
 	u.lastEventIDSent = u.updater.LastEventIDSent()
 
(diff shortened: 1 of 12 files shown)


roomserver/internal/input/input_latest_events.go
```diff
@@ -125,7 +124,8 @@ func (u *latestEventsUpdater) doUpdateLatestEvents() error {
 	// then start with an empty set - none of the forward extremities
 	// that we knew about before matter anymore.
 	oldLatest := []types.StateAtEventAndReference{}
-	if !u.stateAtEvent.Overwrite {
+	if !u.rewritesState {
+		u.oldStateNID = u.updater.CurrentStateSnapshotNID()
```
Kegsay, line 128: Whilst this makes sense, why did you change it so it's only set if we don't rewrite state? I think `updater.CurrentStateSnapshotNID()` just returns an int - the `New...` function did the DB query so it's not even an optimisation. Maybe to act as a guard against accidentally fiddling with it and calculating deltas against things we shouldn't?

Kegsay (approved):
Reads well!

roomserver/internal/input/input_latest_events.go
neilalexander: The idea is that, if we are rewriting state, we *want* the state snapshot NID to be zero because that means that the deltas produced later on will be against an empty snapshot (effectively a complete rewrite, where `AddsStateEventIDs` will contain the entire new state).

If we aren't rewriting state, we want to know what the current state is so that we produce a real delta.

PR #1558, opened 2020-10-22
Version 0.2.1

neilalexander:
* Forward extremities are now calculated using only references from other extremities, rather than including outliers, which should fix cases where state can become corrupted
* Old state events will no longer be processed by the sync API as new, which should fix some cases where clients incorrectly believe they have joined or left rooms
* More SQLite database locking issues have been resolved in the latest events updater
* Internal HTTP API calls are now made using H2C (HTTP/2) in polylith mode, mitigating some potential head-of-line blocking issues
* Roomserver output events no longer incorrectly flag state rewrites
* Notification levels are now parsed correctly in power level events
* Invalid UTF-8 is now correctly rejected when making federation requests

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,8 +8,8 @@
 * Old state events will no longer be processed by the sync API as new, which should fix some cases where clients incorrectly believe they have joined or left rooms
 * Internal HTTP API calls are now made using H2C (HTTP/2) in polylith mode, mitigating some potential head-of-line blocking issues
 * Roomserver output events no longer incorrectly flag state rewrites
-* Notification levels are now parsed correctly in power level events
-* Invalid UTF-8 is now correctly rejected when making federation requests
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -1,5 +1,16 @@
...
+## Dendrite 0.2.1 (2020-10-22)
+
+### Fixes
+
+* Forward extremities are now calculated using only references from other extremities, rather than including outliers, which should fix some state resets
+* Old state events will no longer be processed by the sync API as new, which should fix some cases where clients incorrectly believe they have joined or left rooms
+* Internal HTTP API calls are now made using H2C (HTTP/2) in polylith mode, mitigating some potential head-of-line blocking issues
+* Roomserver output events no longer incorrectly flag state rewrites
+* Notification levels are now parsed correctly in power level events
+* Invalid UTF-8 is now correctly rejected when making federation requests
```
Kegsay, line 12: Please credit respective people.

Kegsay (approved):

PR #1563, opened 2020-10-24
Add possibilty to configure MaxMessageBytes for sarama

S7evinK:
This adds the possibility to configure the `MaxMessageBytes` used in sarama.
The default is to low, if joining large rooms. (User count?)
```go
c.Producer.MaxMessageBytes = 1000000
```

Also adds some more logging when sending messages to Kafka/Naffka in `roomserver`

### Pull Request Checklist

### DIFF
--- a/dendrite-config.yaml
+++ b/dendrite-config.yaml
@@ -74,8 +74,11 @@ global:
     # Kafka.
     use_naffka: true
 
-    # The maximal size of messages passed between Kafka producers and consumers
-    max_message_bytes: 1048576
+    # The max size a Kafka message is allowed to use.
+    # You only need to change this value, if you encounter issues with too large messages.
+    # Must be less than/equal to "max.message.bytes" configured in Kafka.
+    # Defaults to 8388608 bytes.
+    # max_message_bytes: 8388608
 
     # Naffka database options. Not required when using Kafka.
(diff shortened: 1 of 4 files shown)


Kegsay:
It feels strange to have this as a config variable. Admins will have no idea what a sensible value is, nor the trade-off between having it low vs high. In addition, the term "message" has a chance to be confused with a Matrix event which this is not (it'll be the entire send join response triggering the error which has the entire room state contained within the join event). I think just setting this to a sensible high value is preferable to offloading this decision to the server admin.

S7evinK:
Right, my thinking was, that if the admin can configure Kafka to allow a specific message size, he might as well set that at the application level. 
As the default seems to work in most cases, there's only the need to change it if there's an issue. (My case was, that there were 6k events to be added, resulting in a ~5MB message)

Should i comment the value in the config and leave the default value at 8MB (8388608 bytes)?

S7evinK:
Besides changing the `MaxMessageSize`, we could also think about enabling compression in sarama.
Possible compression codecs: GZIP, Snappy, LZ4, ZSTD
As with everything related to compression, comes with the trade off of more allocations etc.

Kegsay (approved):

Issue #1306, opened 2020-08-24
Implement whois


Kegsay:
Spec: https://matrix.org/docs/spec/client_server/r0.6.1#get-matrix-client-r0-admin-whois-userid
Sytests:
```
    × /whois
```

DavidSpenler:
I can work on this

MadLittleMods:
It looks like https://github.com/matrix-org/dendrite/pull/1573 is merged now so closing this issue.

PR #1573, opened 2020-10-29
Implement whois

DavidSpenler:
### Pull Request Checklist

Signed-off-by: `David Spenler <dsspenle@uwaterloo.ca>`

### DIFF
--- a/clientapi/routing/admin_whois.go
+++ b/clientapi/routing/admin_whois.go
@@ -1,0 +1,88 @@
+// Copyright 2020 David Spenler
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package routing
+
+import (
+	"net/http"
+
+	"github.com/matrix-org/dendrite/clientapi/jsonerror"
+	"github.com/matrix-org/dendrite/userapi/api"
+
+	"github.com/matrix-org/util"
+)
+
+type adminWhoisResponse struct {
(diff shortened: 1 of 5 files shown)


tulir:
What's the point of this without admin accounts? (I don't know why the endpoint is in the spec at all)

DavidSpenler:
This PR addresses #1306. It gets us closer to spec compliance regardless of how useful it is at present.

clientapi/routing/admin_whois.go
```diff
@@ -0,0 +1,93 @@
...
+			"": deviceInfo{
+				Sessions: []sessionInfo{
+					sessionInfo{
+						Connections: connections,
+					},
+				},
+			},
+		},
+	}
+
```
S7evinK, line 88: As per the spec, "Each key is an identifier for one of the user's devices."
Synapse just uses an empty string, not sure if we should do the same. (Well, if the `device.ID` is empty, we'll get the same result)

As i had started on this endpoint, when i extended the table with `LastSeenIP` etc.:
```suggestion
	res := make(map[string]deviceInfo)
	for _, device := range devicesResp.Devices {
		connInfo := connectionInfo{
			IP:        device.LastSeenIP,
			LastSeen:  device.LastSeenTS,
			UserAgent: device.UserAgent,
		}
		dev, ok := res[device.ID]
		if !ok {
			dev.Sessions = []sessionInfo{{}}
		}
		dev.Sessions[0].Connections = append(dev.Sessions[0].Connections, connInfo)
		res[device.ID] = dev
	}
```

DavidSpenler:
We've made the changes you suggested.

neilalexander (commented):
Do you have a plan for protecting this endpoint? We don't currently have a concept of admin accounts and this looks like it could expose user IP addresses at current.

DavidSpenler:
The endpoint in its current state does not allow the user to query anyone other than themselves. We assumed that the additional permission of admins to query anyone would have to wait until the admin functionality itself is implemented.

DavidSpenler:
Can I get an update on this?

neilalexander (approved):
We had a bit of discussion about this since we don't have any concept of admin accounts and/or plans for other unspecced admin endpoints, but since this one is in the spec and seems to do what the spec says, we will merge it.

Other unspecced admin endpoints are going to be excluded via matrix-org/sytest#981 so they won't affect our test compliance.

PR #1579, opened 2020-11-12
Fix failing federation tests

S7evinK:
This solves #1327 (notifications test is already passing)

I'm a bit unhappy with getting the power levels, is there a better way?

### Pull Request Checklist

### DIFF
--- a/clientapi/routing/sendevent.go
+++ b/clientapi/routing/sendevent.go
@@ -15,6 +15,7 @@
 package routing
 
 import (
+	"context"
 	"net/http"
 	"sync"
 
@@ -80,6 +81,12 @@ func SendEvent(
 		return *resErr
 	}
 
+	if e.Type() == gomatrixserverlib.MRoomPowerLevels {
+		if ok, plErr := checkPowerLevels(req.Context(), roomID, rsAPI, e); !ok {
+			return *plErr
+		}
+	}
+
 	var txnAndSessionID *api.TransactionID
 	if txnID != nil {
 		txnAndSessionID = &api.TransactionID{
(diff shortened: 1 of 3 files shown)


clientapi/routing/sendevent.go
```diff
@@ -120,6 +127,46 @@ func SendEvent(
 	return res
 }
 
+func checkPowerLevels(
+	ctx context.Context,
+	roomID string,
+	rsAPI api.RoomserverInternalAPI,
+	event *gomatrixserverlib.Event,
+) (ok bool, response *util.JSONResponse) {
+	req := api.QueryLatestEventsAndStateRequest{RoomID: roomID}
```
neilalexander, line 136: We should probably be setting `StateToFetch` in `req` so that we aren't pulling out lots of state events that we don't need - ultimately we only care about the create event and the existing power level event.

roomserver/internal/input/input_events.go
```diff
@@ -210,6 +219,34 @@ func (r *Inputer) processRoomEvent(
...
+		return true, err
+	}
+	var creator string
+	for _, v := range resp.StateEvents {
+		if v.Type() == gomatrixserverlib.MRoomCreate {
+			creator = v.Sender()
+			break
+		}
+	}
+	if event.Sender() != creator && r.ServerName != event.Origin() {
```
neilalexander, line 238: I don't think this check is doing the right thing - ultimately what we want to do is check if the event is sent by the same creator *only if* there isn't a power level event already in the room.

If there is an existing power level event then we don't need to do this, as the soft-fail check will capture it if the current state events don't allow the new event.

clientapi/routing/sendevent.go
```diff
@@ -120,6 +127,46 @@ func SendEvent(
...
+		pl, _ := v.PowerLevels() // we only care about the existence of power levels, so ignore any errors
+		if pl != nil {
+			return true, nil
+		}
+		if v.Type() == gomatrixserverlib.MRoomCreate {
+			creator = v.Sender()
+		}
+	}
+
+	if creator != "" && event.Sender() != creator {
```
neilalexander, line 156: I don't know if we need a check in the client API for this? The roomserver should reject an event that it isn't happy with, and the client API won't be involved in remote servers pushing state to us.

neilalexander (changes requested):

Kegsay:
This PR has been marked as stale and will be closed in 1 week if there are no further commits.

S7evinK:
Closed, for being stale and waiting for https://github.com/matrix-org/dendrite/issues/1441.

kegsay:
This was fixed in https://github.com/matrix-org/dendrite/pull/1920

PR #1583, opened 2020-11-16
Pass pointers to events — reloaded

neilalexander:
Depends on matrix-org/gomatrixserverlib#240 and replaces #950.


Kegsay (approved):
> Significantly lower memory usage when joining rooms, too
Joined two rooms, one with 600 members, the other with 160-ish, and it didn't break over 100MB
That would have spiked up to several hundred MB before

PR #1589, opened 2020-11-18
MSC2836:  Threading - part one

Kegsay:
This is a huge PR which:
 - Adds the concepts of hooks (monolith only)
 - Adds an `mscs` package with MSC2836.
 - Add custom config fields for enabling MSC2836.
 - Adds custom `/unstable` endpoints to both CS API and SS API.
 - Adds some tests for MSC2836.

Currently this only works *locally*.

Best grab a cup of coffee for this one @neilalexander :D

### DIFF
--- a/internal/hooks/hooks.go
+++ b/internal/hooks/hooks.go
@@ -19,20 +19,20 @@ package hooks
 import "sync"
 
 const (
-	// KindNewEvent is a hook which is called with *gomatrixserverlib.HeaderedEvent
+	// KindNewEventPersisted is a hook which is called with *gomatrixserverlib.HeaderedEvent
 	// It is run when a new event is persisted in the roomserver.
 	// Usage:
-	//   hooks.Attach(hooks.KindNewEvent, func(headeredEvent interface{}) { ... })
-	KindNewEvent = "new_event"
-	// KindModifyNewEvent is a hook which is called with *gomatrixserverlib.HeaderedEvent
+	//   hooks.Attach(hooks.KindNewEventPersisted, func(headeredEvent interface{}) { ... })
+	KindNewEventPersisted = "new_event_persisted"
+	// KindNewEventReceived is a hook which is called with *gomatrixserverlib.HeaderedEvent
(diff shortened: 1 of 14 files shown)


internal/hooks/hooks.go
```diff
@@ -0,0 +1,74 @@
...
+package hooks
+
+import "sync"
+
+const (
+	// KindNewEvent is a hook which is called with *gomatrixserverlib.HeaderedEvent
+	// It is run when a new event is persisted in the roomserver.
+	// Usage:
+	//   hooks.Attach(hooks.KindNewEvent, func(headeredEvent interface{}) { ... })
+	KindNewEvent = "new_event"
```
neilalexander, line 26: Might be good to name these a bit more accurately, e.g. `KindNewEventPersisted` and `KindNewEventReceived` or something.

internal/mscs/msc2836/msc2836.go
```diff
@@ -0,0 +1,549 @@
...
+	constRoomIDKey   = "relationship_room_id"
+	constRoomServers = "relationship_servers"
+)
+
+type EventRelationshipRequest struct {
+	EventID         string `json:"event_id"`
+	MaxDepth        int    `json:"max_depth"`
+	MaxBreadth      int    `json:"max_breadth"`
+	Limit           int    `json:"limit"`
+	DepthFirst      *bool  `json:"depth_first"`
```
neilalexander, line 49: I don't really love the pointerisation of the `bool`s and `applyDefaults` here. I think it would feel better just to set the defaults after instantiating the empty object but before calling `json.Decode`/`json.Unmarshal`. That way there's no need to deference anywhere else and any omitted values won't override the defaults.

neilalexander (approved):
The hooking API is a nice touch and overall there isn't too much jumping out at me, just a couple minor stylistic things.

internal/mscs/msc2836/msc2836.go
Kegsay: Good point, thanks!

PR #1590, opened 2020-11-18
Allow = in user identifiers

bodqhrohro:
While I was breaking through all the TDD bureaucracy to finally push this fix,
it turned out that it already got fixed in #1578. Still I push the rest of the
changes (basically, everything except of actually fixing the bug ;)).

`=` is crucial for matrix-bifrost which turns XMPP @'s into =40.

### Pull Request Checklist

Signed-off-by: Bohdan Horbeshko <bodqhrohro@gmail.com>


bodqhrohro:
> It's curious that these sytests didn't show as passed-but-not-whitelisted before, but thanks for adding them!

That's because I PR'ed those tests right before, and they are not even merged yet. There were only tests for restricted characters, but not for allowed characters; that's why this issue wasn't spotted so long.

PR #1593, opened 2020-11-20
Update last seen on sync requests

neilalexander:
This PR sends a request to the user API on an incoming sync to update the last seen IP address and timestamp for the device.

### DIFF
--- a/dendrite-config.yaml
+++ b/dendrite-config.yaml
@@ -300,6 +300,11 @@ sync_api:
     max_idle_conns: 2
     conn_max_lifetime: -1
 
+  # This option controls which HTTP header to inspect to find the real remote IP
+  # address of the client. This is likely required if Dendrite is running behind
+  # a reverse proxy server.
+  # real_ip_header: X-Real-IP
+
 # Configuration for the User API.
 user_api:
   internal_api:
--- a/internal/httputil/httpapi.go
+++ b/internal/httputil/httpapi.go
@@ -17,6 +17,7 @@ package httputil
 import (
 	"context"
 	"io"
+	"net"
 	"net/http"
 	"net/http/httptest"
 	"net/http/httputil"
@@ -59,6 +60,20 @@ func MakeAuthAPI(
 		logger := util.GetLogger((req.Context()))
 		logger = logger.WithField("user_id", device.UserID)
 		req = req.WithContext(util.ContextWithLogger(req.Context(), logger))
+
+		// check if the forwarding proxy, if there is one, has provided a real address.
+		if realIP := req.Header.Get("X-Real-IP"); realIP != "" {
+			if ip := net.ParseIP(realIP); ip != nil {
+				req.RemoteAddr = realIP
+			}
+		} else if forwardedFor := req.Header.Get("X-Forwarded-For"); forwardedFor != "" {
+			addresses := strings.Split(forwardedFor, ",")
+			if len(addresses) > 0 {
+				if ip := net.ParseIP(addresses[0]); ip != nil {
+					req.RemoteAddr = addresses[0]
+				}
+			}
+		}
 
 		return f(req, device)
 	}
--- a/internal/httputil/httpapi.go
+++ b/internal/httputil/httpapi.go
@@ -17,7 +17,6 @@ package httputil
 import (
 	"context"
 	"io"
-	"net"
 	"net/http"
 	"net/http/httptest"
 	"net/http/httputil"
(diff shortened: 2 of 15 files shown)


internal/httputil/httpapi.go
```diff
@@ -60,6 +61,20 @@ func MakeAuthAPI(
 		logger = logger.WithField("user_id", device.UserID)
 		req = req.WithContext(util.ContextWithLogger(req.Context(), logger))
 
+		// check if the forwarding proxy, if there is one, has provided a real address.
```
Kegsay, line 64: This _has_ to be gated behind a config option else malicious clients can spoof their IP by literally sending `X-Real-IP: 1.2.3.4` - something like:
```
clientapi.ip_header: X-Forwarded-For
```
might do the trick.

syncapi/sync/requestpool.go
```diff
@@ -74,6 +74,14 @@ func (rp *RequestPool) OnIncomingSyncRequest(req *http.Request, device *userapi.
 		"limit":     syncReq.limit,
 	})
 
+	lsreq := &userapi.PerformLastSeenUpdateRequest{
+		UserID:     device.UserID,
+		DeviceID:   device.ID,
+		RemoteAddr: req.RemoteAddr,
+	}
+	lsres := &userapi.PerformLastSeenUpdateResponse{}
+	go rp.userAPI.PerformLastSeenUpdate(req.Context(), lsreq, lsres) // nolint:errcheck
```
Kegsay, line 83: We probably want to add some sort of rate-limiting here, we don't need millisecond accuracy on the last seen timestamp, heck even just by the minute would do. I would be tempted to:
 - Have a sync map of deltas (e.g `[user_id + device_id] => []remoteAddrs`)
 - Have a background worker which periodically (every 1 minute?) purges the map.

The danger of this current solution is that I have no confidence that we don't tight loop under some cases, which will cause a _lot_ of goroutines to be made, resources consumed, etc.

Kegsay (changes requested):
Few things, but overall yay!

dendrite-config.yaml
```diff
@@ -300,6 +300,11 @@ sync_api:
     max_idle_conns: 2
     conn_max_lifetime: -1
 
+  # This option controls which HTTP header to inspect to find the real remote IP
+  # address of the client. This is likely required if Dendrite is running behind
+  # a reverse proxy server.
+  # real_ip_header: X-Real-IP
```
Kegsay, line 306: We presumably want the same config option on `client_api`?

neilalexander: Nothing in the client API right now actually updates the last seen, only sync.

syncapi/sync/requestpool.go
```diff
@@ -37,18 +40,63 @@ import (
...
+}
+
+func (rp *RequestPool) updateLastSeen(req *http.Request, device *userapi.Device) {
+	value, _ := rp.lastseen.LoadOrStore(device.UserID+device.ID, time.Now().Add(-time.Minute))
+	lastseen := value.(time.Time)
+	if time.Since(lastseen) < time.Minute {
+		return
+	}
+
+	remoteAddr := req.RemoteAddr
```
Kegsay, line 82: Doesn't this mean if you hit the server with the same device but different IPs we drop all but 1? Is that an issue?

syncapi/sync/requestpool.go
```diff
@@ -37,18 +40,63 @@ import (
...
+	go rp.cleanLastSeen()
+	return rp
+}
+
+func (rp *RequestPool) cleanLastSeen() {
+	for {
+		rp.lastseen.Range(func(key interface{}, value interface{}) bool {
+			lastseen := value.(time.Time)
+			if time.Since(lastseen) > time.Minute*2 {
+				rp.lastseen.Delete(key)
```
Kegsay, line 67: I was expecting a slightly simpler method I guess:
 - When a new request comes in, just blindly store (key=`device.UserID+device.ID`, value=`req.RemoteAddr`) all the time.
 - This cleaner goroutine ranges every minute and deletes _everything_, converting each item into a goroutine request to the user API.

This does mean that some updates will be done quicker than others, but no update should take longer than the refresh interval (1min).

syncapi/sync/requestpool.go
neilalexander: It will, but I don't know what the alternative is. If we check if the IP address changed OR the time is over a minute, then two or more devices syncing at the same time will just result in the value flapping and hitting the user API a lot more often.

syncapi/sync/requestpool.go
neilalexander: I can certainly do that.

dendrite-config.yaml
neilalexander: We could put the option into the global section and then other components could pull it in if needed?

Kegsay: I think there's only 3 client-facing components: media/sync/client API. Probably shouldn't pollute global.

Kegsay (approved):

PR #1596, opened 2020-11-24
MSC2836 threading: part 2

Kegsay:
- Add `QueryAuthChain` to `roomserver`.
- Add `MSC2836EventRelationships` to `fedsender`.
- Update GMSL to have a version which has `MSC2836EventRelationships` on `FederationClient`.
- Implement `direction: up` in `/event_relationships`.
- Do all sorts of shenanigans in `msc2836.go` to get an initial cut of federated threading working:
    * Add federation endpoint for `/event_relationships`.
    * Tweak `reqCtx` to process requests slightly differently for fed/client (fed requests will not trigger more fed requests to avoid infinite loops).
    * When looking up events over fed, use `/event_relationships`.
    * Add helpers to convert event IDs into servers to query, which is done by looking up the event in the roomserver then for that room ID querying fedsender via `QueryJoinedHostServerNamesInRoom`
    * Add `children` and `children_hash` to `unsigned` for both client/federated responses.
    * Store an `explored` flag for each event in a thread, which is `false` initially if there are unexplored children. This gets reset to `false` every time we disagree on the number of children or the hash, and is set to `true` when we do a federated `/event_relationships` request with that event ID (to stop us endlessly hitting federation).

Tested via https://github.com/matrix-org/complement/pull/45


federationsender/inthttp/client.go
```diff
@@ -416,3 +417,35 @@ func (h *httpFederationSenderInternalAPI) LookupServerKeys(
 	}
 	return response.ServerKeys, nil
 }
+
+type eventRelationships struct {
```
neilalexander, line 421: Wouldn't we usually define `EventRelationshipsRequest` and `EventRelationshipsResponse` in the `api` package?

neilalexander (approved):
Looks good to me!

PR #1602, opened 2020-12-02
Optimise CheckServerAllowedToSeeEvent

neilalexander:
This PR tries to optimise `CheckServerAllowedToSeeEvent` so that we only unmarshal state events where the state keys actually end in the server name in question. This is because unmarshalling those events is very expensive in big rooms.

Screenshot 2020-12-02 at 11 32 19


Kegsay (approved):
Elegant solution, I like it!

PR #1610, opened 2020-12-03
Don't recalculate event IDs so often

neilalexander:
Uses `NewEventFromTrustedJSONWithEventID` in the roomserver (see matrix-org/gomatrixserverlib#243) so that we don't burn so much CPU time calculating the event ID over and over again when we already know it.

### DIFF
--- a/roomserver/storage/shared/storage.go
+++ b/roomserver/storage/shared/storage.go
@@ -332,7 +332,7 @@ func (d *Database) Events(
 				return nil, err
 			}
 		}
-		result.Event, err = gomatrixserverlib.NewEventFromStoredJSON(
+		result.Event, err = gomatrixserverlib.NewEventFromTrustedJSONWithEventID(
 			eventID, eventJSON.EventJSON, false, roomVersion,
 		)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


roomserver/storage/shared/storage.go
```diff
@@ -322,14 +322,18 @@ func (d *Database) Events(
 		if roomID, ok := d.Cache.GetRoomServerRoomID(roomNID); ok {
 			roomVersion, _ = d.Cache.GetRoomVersion(roomID)
 		}
+		eventID := ""
+		if id, ierr := d.EventsTable.SelectEventID(ctx, nil, eventJSON.EventNID); ierr == nil {
```
Kegsay, line 326: Can we not do this in bulk before entering the for loop? This will be kinda slow otherwise. E.g `map[EventNID]string`

roomserver/storage/shared/storage.go
```diff
@@ -929,7 +937,11 @@ func (d *Database) GetBulkStateContent(ctx context.Context, roomIDs []string, tu
 	result := make([]tables.StrippedEvent, len(events))
 	for i := range events {
 		roomVer := eventNIDToVer[events[i].EventNID]
-		ev, err := gomatrixserverlib.NewEventFromTrustedJSON(events[i].EventJSON, false, roomVer)
+		eventID := ""
+		if id, err := d.EventsTable.SelectEventID(ctx, nil, events[i].EventNID); err == nil {
```
Kegsay, line 941: Same deal, can we do this outside the for loop?

Kegsay (approved):

PR #1612, opened 2020-12-03
Fix broken image in CONTRIBUTING.md

TR-SLimey:
The relative path seemed to return a 404 so I replaced it with the full URL

Signed-off-by: `tr_slimey@protonmail.com | @tr_slimey:an-atom-in.space`


neilalexander:
Thanks for this!

PR #1613, opened 2020-12-03
GitHub action for automatic multiarch Docker build

TR-SLimey:
Not sure if this is something you'd want to do or not, but I figured I'd open a PR as a proof of concept anyway.

I've just created a GitHub action which builds the Dendrite docker image on release using buildx, and then pushes the image to docker hub while tagging it with the release version also ( saves me building unofficial images every release :P ).

By default it produces `linux/amd64`, `linux/arm/v7` and `linux/arm64` variants to match Synapse, but obviously that can be tweaked easily.

Note: If you do merge this, don't forget to create the `DOCKER_TOKEN` secret.

Signed-off-by: `tr_slimey@protonmail.com | @tr_slimey:an-atom-in.space`


Kegsay:
We probably only want this to execute for tags if I'm honest. We push to master frequently.

TR-SLimey:
Yeah, I actually thought about that after opening the PR but didn't have time to change it. I'll look into it though when I have a minute.

TR-SLimey:
That should do it :D

Kegsay:
Passing this to @anoadragon453 who has more GH Actions knowledge than I.

.github/workflows/docker-build-and-push.yml
```diff
@@ -0,0 +1,68 @@
...
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      -
+        name: Login to DockerHub
+        uses: docker/login-action@v1 
+        with:
+          username: ${{ env.DOCKER_HUB_USER }}
+          password: ${{ secrets.DOCKER_TOKEN }}
+      -
+        name: Build temporary (builder) image
```
anoadragon453, line 35: Why are we calling this a temporary builder image?

anoadragon453 (commented):
Heads up secrets are added to a repository via https://github.com/matrix-org/dendrite/settings/secrets/actions.

Looks mostly sane. It'd be nice to test this in a separate branch (or pushing to a temporary user?) before pushing to production with it.

I'll need to check internally whether we're happy pushing to docker hub via github actions, but I'm not against it personally.

Just a small q below.

.github/workflows/docker-build-and-push.yml
TR-SLimey: From what I've gathered, that's the image which contains the go toolchain and which builds Dendrite, and only the binaries get copied to the actual monolith and polylith images (to save space I believe - Neil mentioned this somewhere), while this image is just discarded, hence it is temporary and a builder. That's mostly just from a quick glance at the `Dockerfile`s though so I may be wrong.

anoadragon453: Cool OK, and thanks for the info. Do you want to try that? Otherwise I can do so sometime tomorrow possibly.

TR-SLimey: I've done those tests already on my fork ( I'm nowhere near confident enough to open a PR without testing :P ), and it seemed to work pretty well, but I think it might be good for you to test also, to avoid "works on my end" kinds of bugs :P

neilalexander:
Testing against `neilalexander/dendrite*` — here goes!

PR #1614, opened 2020-12-04
Federation sender event cache

neilalexander:
Right now, each of the destination queues will try to pull PDUs out of the federation sender database. This can result in a lot of duplicate database hits for the same events in rooms with a lot of servers. This adds a federation event cache to keep them in memory for a while so that we can let up on the database significantly.

I still don't really feel 100% happy with the way that the federation sender handles these queues but this just makes things less bad for now.

### DIFF
--- a/federationsender/storage/shared/storage_edus.go
+++ b/federationsender/storage/shared/storage_edus.go
@@ -69,7 +69,16 @@ func (d *Database) GetNextTransactionEDUs(
 			nids: nids,
 		}
 
-		blobs, err := d.FederationSenderQueueJSON.SelectQueueJSON(ctx, txn, nids)
+		retrieve := make([]int64, 0, len(nids))
+		for _, nid := range nids {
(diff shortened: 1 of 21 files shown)


federationsender/storage/shared/storage_edus.go
```diff
@@ -111,6 +120,7 @@ func (d *Database) CleanEDUs(
 			}
 			if count == 0 {
 				deleteNIDs = append(deleteNIDs, nid)
+				d.Cache.EvictFederationSenderQueuedEDU(nid)
```
Kegsay, line 123: So this gets called when _all_ servers have been sent this EDU?

Kegsay (approved):

federationsender/storage/shared/storage_edus.go
neilalexander: Yep, exactly. It's just a way of cleaning up the cache a bit rather than leaving it there until it's evicted by something else.

PR #1618, opened 2020-12-06
[WIP] Implement archived rooms

DavidSpenler:
I've added logic to check for the `include_leave` parameter in the filter, to retrieve room events for 'left' rooms and to populate the 'leave' field of the response with them (currently, the field is left blank regardless of the filter parameters) for complete sync requests. However, the test still does not pass.

It seems that while synapse performs the final sync request as an incremental sync, dendrite for some reason does a complete sync, causing it to return extraneous state events (I added the leave-population functionality initially to the complete sync portion of the code because I assumed dendrite was correct in doing this).

As the title says, work in progress PR. Please add comments or concerns where necessary.

### Pull Request Checklist

Signed-off-by: `David Spenler <davidspenler@gmail.com>`


Kegsay:
This PR has been marked as stale and will be closed in 1 week if there are no further commits.

PR #1623, opened 2020-12-09
Adjust latest events updater

neilalexander:
There were a couple of cases where `u.latest` could not be set, and then we'd go and consume it in `makeOutputNewRoomEvent` anyway, so that's fixed now.

This also updates one of the checks to be simpler.


Kegsay (approved):
Does this explain the missing fwd extremities?

neilalexander:
I really hope so!

PR #1624, opened 2020-12-09
Don't recalculate event ID so often in sync

neilalexander:
This is similar to #1610 but is for the sync API rather than the roomserver. It reduces CPU usage quite dramatically on large syncs.

This also temporarily comments out a return error when we're missing some state events, as otherwise we return a 500 and Element thinks the server is dead. It's not a permanent fix, but we saw this on `dendrite.matrix.org` happening quite often and some sync is better than no sync in that case.


Kegsay (approved):

Kegsay:
Assuming CI passes LGTM

PR #1628, opened 2020-12-10
Refactor sync tokens

### LINKED ISSUE #1627: Same read receipts sent down /sync constantly
### Background information

- dendrite.matrix.org
- **Dendrite version or git SHA**: 0.3.3+bad81c02

### Description

In a DM I just created, I keep receiving the same read receipts, for the exact same events, with the exact same event IDs, every few syncs.

### Steps to reproduce

- Create a room.
- Send some messages.
- Check the ephemeral events the client receives.

The read receipts should not be resent multiple times, but they are.

neilalexander:
This PR overhauls the streaming and topological tokens, as they had a number of issues.

The `StreamingToken` and `TopologyToken` structs are now more clearly defined, and separate out the different positions that we care about.

Beforehand, this was mixing unrelated EDU types together (send-to-device messages, read receipts, typing notifications) which was actually causing things to be sent down `/sync` multiple times incorrectly because the typing notification cache is reset back to zero when Dendrite restarts, so this should fix #1627.

It also seems to improve backward pagination in my testing on `dendrite.neilalexander.dev`, but I don't know if that's a fluke.

### DIFF
--- a/syncapi/consumers/eduserver_sendtodevice.go
+++ b/syncapi/consumers/eduserver_sendtodevice.go
@@ -107,7 +107,7 @@ func (s *OutputSendToDeviceEventConsumer) onMessage(msg *sarama.ConsumerMessage)
 	s.notifier.OnNewSendToDevice(
 		output.UserID,
 		[]string{output.DeviceID},
-		types.NewStreamToken(0, streamPos, nil),
+		types.NewStreamToken(0, 0, 0, streamPos, nil),
 	)
 
 	return nil
--- a/syncapi/consumers/eduserver_sendtodevice.go
+++ b/syncapi/consumers/eduserver_sendtodevice.go
@@ -107,7 +107,7 @@ func (s *OutputSendToDeviceEventConsumer) onMessage(msg *sarama.ConsumerMessage)
 	s.notifier.OnNewSendToDevice(
 		output.UserID,
 		[]string{output.DeviceID},
-		types.NewStreamToken(0, 0, 0, streamPos, nil),
+		types.StreamingToken{SendToDevicePosition: streamPos},
 	)
 
 	return nil
--- a/syncapi/types/types.go
+++ b/syncapi/types/types.go
@@ -16,7 +16,6 @@ package types
 
 import (
 	"encoding/json"
-	"errors"
 	"fmt"
 	"sort"
 	"strconv"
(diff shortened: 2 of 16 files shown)


syncapi/consumers/eduserver_sendtodevice.go
```diff
@@ -107,7 +107,7 @@ func (s *OutputSendToDeviceEventConsumer) onMessage(msg *sarama.ConsumerMessage)
 	s.notifier.OnNewSendToDevice(
 		output.UserID,
 		[]string{output.DeviceID},
-		types.NewStreamToken(0, streamPos, nil),
+		types.NewStreamToken(0, 0, 0, streamPos, nil),
```
Kegsay, line 110: I think it would be better if NewStreamingToken accepted a struct at this point so we can just omit 0 values and get more intuitive labels on what stream we're using rather than relying on the position.

neilalexander: If that's the case, we had might as well just remove `NewStreamToken` and `NewTopologyToken` altogether and instantiate the structs directly :-)

Kegsay: Maybe we should then?

neilalexander: I'm happy to go through and do that tomorrow!

syncapi/types/types.go
```diff
@@ -223,151 +231,121 @@ func (t *TopologyToken) Decrement() {
...
-			var offset int64
-			offset, err = strconv.ParseInt(segments[2], 10, 64)
-			if err != nil {
-				return
-			}
-			logs[segments[0]] = &LogPosition{
-				Partition: int32(partition),
-				Offset:    offset,
-			}
+	categories := strings.Split(tok[1:], ".")
```
Kegsay, line 304: This will panic if `len(tok) == 1`?

Kegsay (approved):
Clearer!

syncapi/types/types.go
neilalexander: It shouldn't do, as it will just result in a zero-length string slice. The tests seem OK with it.

syncapi/consumers/eduserver_sendtodevice.go
neilalexander: Did this now, as it didn't take long.

PR #1631, opened 2020-12-11
Give receipts their own stream ID in the database

neilalexander:
It's incorrect for the read receipts to be sharing the PDU stream ID, given that it now has its own position in the `StreamingToken`.

This PR gives it a stream ID of its own.


syncapi/storage/postgres/syncserver.go
```diff
@@ -86,6 +88,11 @@ func NewDatabase(dbProperties *config.DatabaseOptions) (*SyncServerDatasource, e
 	if err != nil {
 		return nil, err
 	}
+	m := sqlutil.NewMigrations()
+	deltas.LoadFixSequences(m)
+	if err = m.RunDeltas(d.db, dbProperties); err != nil {
```
Kegsay, line 93: It's a bit more complicated - https://github.com/matrix-org/dendrite/blob/master/roomserver/storage/postgres/storage.go#L45

Kegsay: Oh this doesn't add a new column?

neilalexander: Nope, which is mostly why I thought I could probably get away with it as it is.

Kegsay (approved):

PR #1632, opened 2020-12-11
Fix device list position in streaming token

### LINKED ISSUE #1625: Device keys get marked out of date every second sync
### Background information

- **Dendrite version or git SHA**: 0.3.3+bad81c02
- **SQLite3 or Postgres?**: Whatever runs dendrite.matrix.org
- **Running in Docker?**: Whatever runs dendrite.matrix.org
- **`go version`**: Whatever runs dendrite.matrix.org

### Description

After logging in with Nheko on dendrite.matrix.org, I get the following update in every second sync, causing my client to mark those users keys as out of date:

```
    "device_lists": {
        "changed": [
            "@jonah:aragon.sh",
            "@neilalexander:dendrite.matrix.org",
            "@florian:wolkenplanet.de",
            "@ninchuka:matrix.org",
            "@kim:sosnowkadub.de",
            "@midov:midov.pl",
            "@mathijs:dendrite.matrix.org",
            "@kegan:dendrite.kegsay.com",
            "@alexander:eiselecloud.de",
            "@deepbluev7:dendrite.matrix.org",
            "@neilalexander:dendrite.neilalexander.dev",
            "@office:ddt-consult.de",
            "@jaywink:federator.dev",
            "@mewmew:blob.cat",
            "@midek:construct.midov.moe",
            "@schwukas:lkas.cc",
            "@ruff_wizard:pixie.town",
            "@travis:dendrite.matrix.org",
            "@tom:lant.uk",
            "@android_ex.ai:midov.pl",
            "@xethos:xethos.net"
        ]
    },
```

This also includes my own user, @deepbluev7:dendrite.matrix.org, who only has this one device.

I also get 2 times the same sync token, only differing in the last number: s69732_6464.dl-0-9131 and s69737_6464.dl-0-0.

The last number seems to be the same in every second sync. It looks like Nheko is running 2 parallel /sync streams here for some reason, but the sync tokens actually flip flop the last number after each sync, which then causes dendrite to send device_lists for one of those /sync's?

### Steps to reproduce

- register to dendrite.matrix.org
- login with Nheko
- join a room
- watch your device list

<!--

``` (three backticks, on a line on their own), so that they are formatted legibly.

-->

### Expected behaviour

Device lists should not change, if they don't actually change.

neilalexander:
`SelectKeyChanges` was only setting `latestOffset` if there was more one or more keys, but not if there were zero keys changed. `latestOffset` was put into the `NextBatch` token, which is why the offset kept alternating and being set back to zero.

This sets the `latestOffset` to the `fromOffset` initially, so that if no keys are returned, we keep the same offset in the `NextBatch`.

This *should* fix #1625.

### DIFF
--- a/keyserver/storage/postgres/key_changes_table.go
+++ b/keyserver/storage/postgres/key_changes_table.go
@@ -82,6 +82,7 @@ func (s *keyChangesStatements) SelectKeyChanges(
 	if toOffset == sarama.OffsetNewest {
 		toOffset = math.MaxInt64
 	}
+	latestOffset = fromOffset
 	rows, err := s.selectKeyChangesStmt.QueryContext(ctx, partition, fromOffset, toOffset)
 	if err != nil {
 		return nil, 0, err
--- a/keyserver/storage/sqlite3/key_changes_table.go
+++ b/keyserver/storage/sqlite3/key_changes_table.go
@@ -83,6 +83,7 @@ func (s *keyChangesStatements) SelectKeyChanges(
 	if toOffset == sarama.OffsetNewest {
 		toOffset = math.MaxInt64
 	}
+	latestOffset = fromOffset
 	rows, err := s.selectKeyChangesStmt.QueryContext(ctx, partition, fromOffset, toOffset)
 	if err != nil {
 		return nil, 0, err


Kegsay (approved):

neilalexander:
This fix is included in #1641 so closing.

PR #1641, opened 2020-12-14
More sane next batch handling, typing notification tweaks, give invites their own stream position, device list fix

neilalexander:
This PR does a few things:

- Replaces a few places where we'd bounce backwards and forwards between `string` and `types.StreamPosition` with a single `NextBatch` stream position which gets marshalled to/from text as needed by the JSON marshaller/unmarshaller
- Tweaks the typing code so it's a bit cleaner and so that the server restarting doesn't break typing notifications until the cache is cleared
- Invites now have their own stream position instead of abusing the PDU one
- Adds the device list fix from #1632 for good measure

### DIFF
--- a/syncapi/storage/shared/syncserver.go
+++ b/syncapi/storage/shared/syncserver.go
@@ -629,7 +629,6 @@ func (d *Database) addReceiptDeltaToResponse(
 			}
 			read.User[receipt.UserID] = eduAPI.ReceiptTS{TS: receipt.Timestamp}
 			content[receipt.EventID] = read
-			res.NextBatch.ReceiptPosition++
 		}
 		ev.Content, err = json.Marshal(content)
 		if err != nil {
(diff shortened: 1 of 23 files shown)


syncapi/storage/shared/syncserver.go
```diff
@@ -628,6 +629,7 @@ func (d *Database) addReceiptDeltaToResponse(
 			}
 			read.User[receipt.UserID] = eduAPI.ReceiptTS{TS: receipt.Timestamp}
 			content[receipt.EventID] = read
+			res.NextBatch.ReceiptPosition++
```
Kegsay, line 632: We never used to do this before, why do this now?

syncapi/storage/shared/syncserver.go
```diff
@@ -582,6 +582,7 @@ func (d *Database) addTypingDeltaToResponse(
 			jr.Ephemeral.Events = append(jr.Ephemeral.Events, ev)
 			res.Rooms.Join[roomID] = jr
 		}
+		res.NextBatch.TypingPosition = types.StreamPosition(d.EDUCache.GetLatestSyncPosition())
```
Kegsay, line 585: Does this not have the chance to race (a call to `GetTypingUsersIfUpdatedAfter` vs a call to `GetLatestSyncPosition`)?

Kegsay (approved):

Kegsay (approved):

PR #1643, opened 2020-12-15
Hit the database far less in Events to find room NIDs and room versions

neilalexander:
This replaces `SelectRoomNIDForEventNID` with `SelectRoomNIDsForEventNIDs`, so that when we're pulling lots of event entries out of the database, we're only making one statement call like this instead of potentially tens or hundreds of them.

This also replaces `SelectRoomVersionForRoomNID` with `SelectRoomVersionsForRoomNIDs`, just as above.

This currently lights up on the flame graph pretty badly in a lot of places:

Screenshot 2020-12-15 at 16 11 39

### DIFF
--- a/roomserver/storage/postgres/events_table.go
+++ b/roomserver/storage/postgres/events_table.go
@@ -439,6 +439,7 @@ func (s *eventStatements) SelectRoomNIDsForEventNIDs(
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


Kegsay (approved):
LGTM - though under what circumstances do we hammer `Events()` with lots of different events in different rooms?

neilalexander:
I don't think we do, but the API shape doesn't forbid it, so I thought it was better to do the right thing in that case.

PR #1646, opened 2020-12-16
Add RoomInfo cache, remove RoomServerRoomNIDsCache

neilalexander:
This adds a cache for `types.RoomInfo` in the roomserver, and removes the `RoomServerRoomNIDsCache` since `RoomInfoCache` covers the same use-case.

I've deliberately not done anything with the `RoomVersionCache` in this PR because it needs a bit more thought - the main place that the `RoomVersionCache` helps us is in polylith mode, hence why it's mostly handled at the API boundaries and not internally within the roomserver.

### DIFF
--- a/roomserver/storage/shared/storage.go
+++ b/roomserver/storage/shared/storage.go
@@ -330,13 +330,23 @@ func (d *Database) Events(
 	for _, n := range roomNIDs {
 		uniqueRoomNIDs[n] = struct{}{}
 	}
-	roomNIDList := make([]types.RoomNID, 0, len(uniqueRoomNIDs))
+	roomVersions := make(map[types.RoomNID]gomatrixserverlib.RoomVersion)
+	fetchNIDList := make([]types.RoomNID, 0, len(uniqueRoomNIDs))
 	for n := range uniqueRoomNIDs {
-		roomNIDList = append(roomNIDList, n)
+		if roomID, ok := d.Cache.GetRoomServerRoomID(n); ok {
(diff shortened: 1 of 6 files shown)


roomserver/storage/shared/storage.go
```diff
@@ -124,7 +124,15 @@ func (d *Database) StateEntriesForTuples(
 }
 
 func (d *Database) RoomInfo(ctx context.Context, roomID string) (*types.RoomInfo, error) {
-	return d.RoomsTable.SelectRoomInfo(ctx, roomID)
+	if roomInfo, ok := d.Cache.GetRoomInfo(roomID); ok {
```
Kegsay, line 127: Is this safe? The room nid and room version may be immutable but the state snapshot nid is not.

Kegsay (approved):

roomserver/storage/shared/storage.go
neilalexander: If the roominfo is present in the cache, then the latest events updater will keep it up-to-date: https://github.com/matrix-org/dendrite/pull/1646/files#diff-e863b7b084310a691d13d9747108c96f6fac397eb7c09411c770aa7c83b4b0ffR108-R114

It has to be a mutable cache for that reason (that and the room can suddenly become not-a-stub when we start updating the latest events).

neilalexander: Also this is safe because this cache is strictly used only in the roomserver and not cross-component.

neilalexander: Adding comments to this effect.

Issue #1518, opened 2020-10-13
Registration restrictions and app service namespace restrictions are also applied to app services


phol:
### Background information

- **Dendrite version or git SHA**: v0.1.0
- **Monolith or Polylith?**: Monolith
- **SQLite3 or Postgres?**: Postgres
- **Running in Docker?**: No
- **`go version`**: go version go1.15.2 linux/amd64

### Description

I am experimenting with Dendrite and was trying out to set up the [mautrix-whatsapp](https://github.com/tulir/mautrix-whatsapp) bridge.

It requires an exclusive namespace, but when `exclusive: true` is enabled in its registration file, it errors with

```log
[ERROR] Failed to connect to homeserver: failed to ensure registered: failed to POST /_matrix/client/r0/register: M_EXCLUSIVE (HTTP 400): This username is reserved by an application service.. Retrying in 10 seconds...
```

Setting `exclusive: false` remediates that issue, however, disabling registration in dendrite also causes errors:

```log
[ERROR] Failed to connect to homeserver: failed to ensure registered: failed to POST /_matrix/client/r0/register: 403 403 Forbidden
```

After setting `exclusive: false` and `registration_disabled: false` and dropping and recreating the bridge, it does register as can be viewed in the Dendrite logs:

```log
Oct 13 11:29:02 dendrite dendrite-monolith-server[23628]: time="2020-10-13T09:29:02.818448029Z" level=info msg="Processing registration request" func="Register\n\t" file=" [/root/dendrite/clientapi/routing/register.go:507]" auth.type="" req.id=eAmml6MbH2Mm req.method=POST req.path=/_matrix/client/r0/register session_id="" username=whatsappbot
```

I can also invite it into rooms, but it doesn't respond to commands. I've not figured out what the cause of that last issue is yet but I'm suspecting this might have to do with notifications not being implemented yet, making it irrelevant for this specific issue.

### Steps to reproduce

- Install Dendrite
- Install Mautrix WhatsApp and try to register it in the app services section in dendrite.yaml.
- Restart Dendrite, start Mautrix Whatsapp
- Watch the logs.

<!--
Describe how what happens differs from what you expected.

If you can identify any relevant log snippets from server logs, please include
those (please be careful to remove any personal or private data). Please surround them with
``` (three backticks, on a line on their own), so that they are formatted legibly.

Alternatively, please send logs to @kegan:matrix.org or @neilalexander:matrix.org
with a link to the respective Github issue, thanks!
-->

[...]

PR #1652, opened 2020-12-17
Do not check if a username is exclusive if the request is for an appservice in /register

Half-Shot:
Fixes https://github.com/matrix-org/dendrite/issues/1518 #1656 

I'm not quite sure how I'd go about writing tests for this one, pointers would be useful :)

### DIFF
--- a/clientapi/routing/register.go
+++ b/clientapi/routing/register.go
@@ -627,6 +627,10 @@ func handleRegistrationFlow(
 		// Add Dummy to the list of completed registration stages
 		AddCompletedSessionStage(sessionID, authtypes.LoginTypeDummy)
 
+	case "":
+		// An empty auth type means that we want to fetch the available
+		// flows. It can also mean that we want to register as an appservice
+		// but that is handed above.
 	default:
 		return util.JSONResponse{
 			Code: http.StatusNotImplemented,
--- a/clientapi/routing/register.go
+++ b/clientapi/routing/register.go
@@ -589,6 +589,9 @@ func handleRegistrationFlow(
 
 	// TODO: email / msisdn auth types.
 	accessToken, accessTokenErr := auth.ExtractAccessToken(req)
+	if accessTokenErr != nil {
+		return util.MessageResponse(http.StatusForbidden, "Access token error: "+accessTokenErr.Error())
+	}
 
 	// Appservices are special and are not affected by disabled
 	// registration or user exclusivity.
--- a/clientapi/routing/register.go
+++ b/clientapi/routing/register.go
@@ -329,9 +329,13 @@ func UserIDIsWithinApplicationServiceNamespace(
 	appservice *config.ApplicationService,
 ) bool {
 
-	var local, _, err = gomatrixserverlib.SplitID('@', userID)
+	var local, domain, err = gomatrixserverlib.SplitID('@', userID)
 	if err != nil {
 		// Not a valid userID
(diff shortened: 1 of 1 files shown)


clientapi/routing/register.go
```diff
@@ -488,10 +488,16 @@ func Register(
 		return *resErr
 	}
 
+	// Extract access token here
+	accessToken, _ := auth.ExtractAccessToken(req)
+
 	// Make sure normal user isn't registering under an exclusive application
 	// service namespace. Skip this check if no app services are registered.
+	// If an access token is provided, ignore this check this is an appservice
+	// request and we will validate in validateApplicationService
```
tulir, line 496: Does the check always happen when there's an access token? i.e. can you now bypass username reservation checks by providing a bogus access token and using dummy login?

Half-Shot: Further down if we see an access token in the request, we'll treat it like an appservice request. There isn't a way to pass in a token without going through the appservice handler.

Half-Shot: Unless of course, you pass in a different login type. Hmph.

Half-Shot: Have made appservice handling entirely special.

TR-SLimey:
Here's some pointers for you:

```
0xf4e13a8b
0xaff35b2e
0xbf35e3e7
```
:P

Half-Shot:
Somehow I knew someone was going to make that joke @TR-SLimey :p

Half-Shot:
I've also included a fix for #1656

neilalexander (approved):

clientapi/routing/register.go
```diff
@@ -328,7 +328,18 @@ func UserIDIsWithinApplicationServiceNamespace(
...
 ) bool {
+
+	var local, _, err = gomatrixserverlib.SplitID('@', userID)
+	if err != nil {
+		// Not a valid userID
+		return false
+	}
+
 	if appservice != nil {
+		if appservice.SenderLocalpart == local {
```
Half-Shot, line 339: Thought: We should check that the domain matches the current server. Is there an easy way to get the value of domain of the server?

neilalexander: Added a check.

clientapi/routing/register.go
```diff
@@ -581,11 +588,33 @@ func handleRegistrationFlow(
 	// TODO: Handle mapping registrationRequest parameters into session parameters
 
 	// TODO: email / msisdn auth types.
+	accessToken, accessTokenErr := auth.ExtractAccessToken(req)
```
Kegsay, line 591: Can we deal with `accessTokenErr` ASAP please rather than rely on a complicated if statement on `:595` and a subsequent function call? Seems confusing.

Kegsay (approved):
It's important that we handle errors quickly, I don't like how we pass `err` into `handleApplicationServiceRegistration`. The code looks fine though.

clientapi/routing/register.go
```diff
@@ -581,11 +588,39 @@ func handleRegistrationFlow(
...
+	if accessTokenErr != nil {
```
Half-Shot, line 592: Hold on a sec. A missing access token will raise an error, except that a missing access token is *only* an error for appservices. /register should work without an access token for non-as based requests

(it's at times like these I wish we had a dedicated endpoint for appservices)

Issue #1429, opened 2020-09-20
Add smart /health endpoint that checks if dependent services available


S7evinK:
**Is your feature request related to a problem? Please describe.**
Adding a `/health` endpoint allows orchestration tools to see if a component is healthy and "restart" the process, if needed.

**Describe the solution you'd like**
Ping the database connection and/or other components required to run a specific component.

Kegsay:
You can use `/_matrix/client/versions` for this for now. This is still relevant in polylith deployments though which don't have this. I _think_ if metrics are enabled you can hit prometheus but obviously it doesn't actually mean that the service is ready for connections yet.

MadLittleMods:
We're tracking a dumb `/health` endpoint that returns 200 OK in https://github.com/matrix-org/dendrite/issues/1668. We can use this as a follow-up to make something smarter.

For reference with Gitter, we have a dumb `/health_check` and another smart `/health_check/full` like this issue is after, [`server/api/index.js#L36-50`](https://gitlab.com/gitterHQ/webapp/-/blob/5abd9cdd6cf2ca7da62f4fd320f67aee1d38b94c/server/api/index.js#L36-50)

whatlimits:
[](https://help.twitter.com/en/rules-and-policies/twitter-username-squatting)

S7evinK:
https://github.com/alexliesenfeld/health

Might be worth a look for this.

matslarson:
👍 for this, even if it's just a dumb `200 OK`. I have a Helm chart for a polylith deployment in k8s almost ready to go, but the hangup is that there's no way for the ingress loadbalancer to healthcheck the components. Using `/metrics` as a hack doesn't really work either, not only is it not really indicative of whether the service is ready, but it requires auth that can't be passed in a path and listens on the internal port while the loadbalancer would be checking the external port for service availability.

MadLittleMods:
This was added in https://github.com/matrix-org/dendrite/pull/2303 and available in Dendrite 0.7.0 (2022-03-25).

> Healthcheck endpoints are now available at:
>  - `/_dendrite/monitor/up`, which will return 200 when Dendrite is ready to accept requests
>  - `/_dendrite/monitor/health`, which will return 200 if healthy and 503 if degraded for some reason
>
> *-- https://github.com/matrix-org/dendrite/blob/main/CHANGES.md#dendrite-070-2022-03-25*

PR #1657, opened 2020-12-18
Add health check endpoint

MadLittleMods:
Add health check (`/health`) endpoint, http://localhost:8008/health

Fix https://github.com/matrix-org/dendrite/issues/1668

## Dev notes

 - [`setup/base.go`](https://github.com/matrix-org/dendrite/blob/4a0461378ad0d0af2ba701fb07dea505be469c92/setup/base.go#L279-L282)
 - [`/_matrix/client/versions`](https://github.com/matrix-org/dendrite/blob/4a0461378ad0d0af2ba701fb07dea505be469c92/clientapi/routing/routing.go#L65-L82)

---

```
./build.sh && ./bin/dendrite-demo-libp2p --port 8085
```

 - http://localhost:8085/api/health
 - http://localhost:8085/_matrix/client/health

```
./build.sh && sudo ./bin/dendrite-monolith-server --tls-cert server.crt --tls-key server.key --config dendrite.yaml -http-bind-address :8009
```

 - http://localhost:8009/api/health
 - http://localhost:8009/_matrix/client/health

---

In [`setup/base.go` -> `SetupAndServeHTTP`](https://github.com/matrix-org/dendrite/blob/4a0461378ad0d0af2ba701fb07dea505be469c92/setup/base.go#L279-L282). But `SetupAndServeHTTP` function never gets called in `./bin/dendrite-demo-libp2p` so the health checks aren't available. But this does work for normal monolith (`/api/health`, `/_matrix/client/health`).
```go
b.InternalAPIMux.HandleFunc("/health", func(w http.ResponseWriter, req *http.Request) {
		fmt.Println("health hit1")
		w.WriteHeader(200)
	}).Methods(http.MethodGet)

	b.PublicClientAPIMux.HandleFunc("/health", func(w http.ResponseWriter, req *http.Request) {
		fmt.Println("health hit2")
		w.WriteHeader(200)
	}).Methods(http.MethodGet)

externalRouter.Handle("/health", httputil.MakeInternalAPI("Health", func(req *http.Request) util.JSONResponse {
		return util.JSONResponse{
			Code: http.StatusOK,
			JSON: struct {
				Versions bool `json:"success"`
			}{true},
		}
	}),
	)
```

---

Next to [`/_matrix/client/versions`](https://github.com/matrix-org/dendrite/blob/4a0461378ad0d0af2ba701fb07dea505be469c92/clientapi/routing/routing.go#L65-L82). Adds to `/_matrix/client/health`
```go
publicAPIMux.Handle("/health",
		httputil.MakeExternalAPI("health", func(req *http.Request) util.JSONResponse {
			return util.JSONResponse{
				Code: http.StatusOK,
				JSON: struct {
					Versions bool `json:"success"`
				}{true},
			}
		}),
	).Methods(http.MethodGet)
```

### Pull Request Checklist

* [ ] Pull request includes a [sign off](https://github.com/matrix-org/dendrite/blob/master/docs/CONTRIBUTING.md#sign-off)

### DIFF
--- a/setup/base.go
+++ b/setup/base.go
@@ -26,6 +26,7 @@ import (
 	"github.com/matrix-org/dendrite/internal/caching"
 	"github.com/matrix-org/dendrite/internal/httputil"
 	"github.com/matrix-org/gomatrixserverlib"
+	"github.com/matrix-org/util"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"golang.org/x/net/http2"
 	"golang.org/x/net/http2/h2c"
@@ -310,6 +311,17 @@ func (b *BaseDendrite) SetupAndServeHTTP(
 			Handler: h2c.NewHandler(internalRouter, internalH2S),
 		}
 	}
+
+	externalRouter.Handle("/health",
+		httputil.MakeExternalAPI("health", func(req *http.Request) util.JSONResponse {
+			return util.JSONResponse{
+				Code: http.StatusOK,
+				JSON: struct {
+					Versions bool `json:"success"`
+				}{true},
+			}
+		}),
+	).Methods(http.MethodGet)
 
 	internalRouter.PathPrefix(httputil.InternalPathPrefix).Handler(b.InternalAPIMux)
 	if b.Cfg.Global.Metrics.Enabled {
--- a/setup/base.go
+++ b/setup/base.go
@@ -317,7 +317,7 @@ func (b *BaseDendrite) SetupAndServeHTTP(
 			return util.JSONResponse{
 				Code: http.StatusOK,
 				JSON: struct {
-					Versions bool `json:"success"`
+					Success bool `json:"success"`
 				}{true},
 			}
 		}),


setup/base.go
```diff
@@ -311,6 +312,18 @@ func (b *BaseDendrite) SetupAndServeHTTP(
 		}
 	}
 
+	b.InternalAPIMux.HandleFunc("/health", func(w http.ResponseWriter, req *http.Request) {
+		w.WriteHeader(200)
+	})
+
+	externalRouter.Handle("/health", httputil.MakeInternalAPI("Health", func(req *http.Request) util.JSONResponse {
```
MadLittleMods, line 319: I just have these 2 attempts at handling `/health` to see what works (neither currently)

What's the distinction/difference between `internalRouter` and `externalRouter`?

Looking further, it seems like `b.InternalAPIMux` feeds into `internalRouter`.

---

Which addresses go where?

```
$  ./build.sh && ./bin/dendrite-demo-libp2p --port 8085
...
Our addresses: [/ip4/192.168.1.118/tcp/57741 /ip4/127.0.0.1/tcp/57741 /ip6/::1/tcp/57742]
```

 - http://localhost:8085/health -> 404 page not found
 - http://localhost:57741/health ->ERR_INVALID_HTTP_RESPONSE (same for below)
 - http://192.168.1.118:57741/health
 - http://127.0.0.1:57741/health
 - http://localhost:57742/health
 - http://192.168.1.118:57742/health
 - http://127.0.0.1:57742/health

neilalexander: In monolith mode, the internal and external routers can end up being the same thing, but in full API mode or polylith mode, internal APIs go on the internal router and world-facing Matrix APIs (like the CS API and SS API) go on the external router.

None of those libp2p addresses will accept HTTP - those are for libp2p peering - you'll need to check port 8085, if that's what you specified with `--port`.

S7evinK:
IMHO this doesn't fix/close #1429 .
Simply returning http 200 in polylith mode just tells us that the service is up and running, not that it can access other required services/components. (like database, kafka)
See #1476 for my try/idea. :)

MadLittleMods:
@S7evinK That PR was closed off with a note to just go the simple route which this PR aims to do.

> I would prefer it if, for now, we just had a simple HTTP handler which responded with 200 OK in response to `/health`
>
> https://github.com/matrix-org/dendrite/pull/1476#pullrequestreview-521561001

S7evinK:
I know, i just wanted to say that the issue should not be closed when merging this PR. :)

setup/base.go
MadLittleMods: Thanks @neilalexander!

I think my main confusion is that `setup/base.go` -> `SetupAndServeHTTP` isn't called in `cmd/dendrite-demo-libp2p/main.go` like it is for the monolith and polylith binaries so my health check wasn't being configured.

MadLittleMods:
@S7evinK 👍 I created https://github.com/matrix-org/dendrite/issues/1668 to track this and we can use https://github.com/matrix-org/dendrite/issues/1429 as a follow-up to improve it.

setup/base.go
```diff
@@ -311,6 +312,17 @@ func (b *BaseDendrite) SetupAndServeHTTP(
...
+	externalRouter.Handle("/health",
+		httputil.MakeExternalAPI("health", func(req *http.Request) util.JSONResponse {
+			return util.JSONResponse{
+				Code: http.StatusOK,
+				JSON: struct {
+					Versions bool `json:"success"`
+				}{true},
+			}
+		}),
+	).Methods(http.MethodGet)
```
MadLittleMods, line 324: One problem I am seeing is that `setup/base.go` -> `SetupAndServeHTTP` isn't called in [`cmd/dendrite-demo-libp2p/main.go`](https://github.com/matrix-org/dendrite/blob/4a0461378ad0d0af2ba701fb07dea505be469c92/cmd/dendrite-demo-libp2p/main.go) like it is for the [monolith](https://github.com/matrix-org/dendrite/blob/4a0461378ad0d0af2ba701fb07dea505be469c92/cmd/dendrite-monolith-server/main.go#L161-L170) and [polylith](https://github.com/matrix-org/dendrite/blob/4a0461378ad0d0af2ba701fb07dea505be469c92/cmd/dendrite-polylith-multi/personalities/clientapi.go#L40-L44) binaries so the health check isn't being configured in the `libp2p` demo. This means that when you run `./build.sh && ./bin/dendrite-demo-libp2p`, http://localhost:8080/health isn't available.

To be clear, the current code works fine in the monolith/polylith. Just have a problem with the `libp2p` demo

@neilalexander Can we use the standard `SetupAndServeHTTP` function in `cmd/dendrite-demo-libp2p/main.go` instead of using our own custom [`http.ListenAndServe(httpBindAddr, httpRouter)`](https://github.com/matrix-org/dendrite/blob/4a0461378ad0d0af2ba701fb07dea505be469c92/cmd/dendrite-demo-libp2p/main.go#L216) call (adapt whatever is necessary)? It looks like we do some [`embed.Embed(...)` stuff](https://github.com/matrix-org/dendrite/blob/4a0461378ad0d0af2ba701fb07dea505be469c92/cmd/dendrite-demo-libp2p/main.go#L205) to the router which makes it different from using `SetupAndServeHTTP`

ex.
```go
httpAddr := config.HTTPAddress("http://" + httpBindAddr)
httpAPIAddr := httpAddr

base.Base.SetupAndServeHTTP(
	httpAPIAddr, // internal API
	httpAddr,    // external API
	nil, nil,    // TLS settings
)
```

setup/base.go
```diff
@@ -311,6 +312,17 @@ func (b *BaseDendrite) SetupAndServeHTTP(
 		}
 	}
 
+	externalRouter.Handle("/health",
```
MadLittleMods, line 315: Where is best to add a test? I don't see any tests for `/_matrix/client/versions` for example.

MadLittleMods:
@neilalexander Friendly ping for review 🙂

neilalexander:
Will take a look shortly!

MadLittleMods:
@neilalexander 🦡🦡🦡 The badgers are itching to ship 😁. Poking for review again

setup/base.go
```diff
@@ -311,6 +312,17 @@ func (b *BaseDendrite) SetupAndServeHTTP(
...
+		httputil.MakeExternalAPI("health", func(req *http.Request) util.JSONResponse {
+			return util.JSONResponse{
+				Code: http.StatusOK,
+				JSON: struct {
+					Versions bool `json:"success"`
```
S7evinK, line 320:
```suggestion
 					Success bool `json:"success"`
```

S7evinK (changes requested):
Besides that little nitpick, looks good to me.

setup/base.go
MadLittleMods: Thanks @S7evinK! Updated 😀

S7evinK (approved):

setup/base.go
neilalexander, line 315: My feeling is that this almost certainly wants to be on `internalRouter`, not `externalRouter`. Otherwise it'll be exposed on the public mux in monolith mode, and it won't be in the right place with `-api` set.

setup/base.go
neilalexander: To be honest I'm not terribly worried about the P2P demos here, as long as this does the right thing for monolith and polylith.

setup/base.go
MadLittleMods: @neilalexander and I discussed this a bit in a sync call but it's not a simple problem because we have to contend with monolith and polylith mode and making it available in each mode.

We probably want this on `internalRouter` because in the future, it will share more information about the state of databases, etc (see https://github.com/matrix-org/dendrite/issues/1429)

But we might need to put metrics and health together and add some new logic to make it work.

Kegsay:
This PR has been marked as stale and will be closed in 1 week without further input.

PR #1660, opened 2020-12-18
Re-add the appservice alias query on unknown alias

### LINKED ISSUE #1655: Appservice alias query support is broken
This was broken in https://github.com/matrix-org/dendrite/pull/994. Dendrite does not query appservices when a user requests a unknown alias. 

Spec: https://matrix.org/docs/spec/application_service/r0.1.2#get-matrix-app-v1-rooms-roomalias

If dendrite encounters a request to join (either over federation or locally) an alias, and the alias is local but unknown to the homeserver then it should try to query the appservice using the above endpoint. The appservice should then handle the request by either calling createRoom with the given alias, making the alias exist, or send a failed status code where dendrite should then drop the request it was handling with a not found error.

Dendrite will know which appservice to query by the namespaces given in the registration data of the appservice.

neilalexander:
At the time I commented this out during a roomserver API refactor because it wasn't obvious what it did and it created an import cycle. I've now restored it to close #1655.

### DIFF
--- a/appservice/api/query.go
+++ b/appservice/api/query.go
@@ -20,9 +20,9 @@ package api
 import (
 	"context"
 	"database/sql"
+	"errors"
 
 	"github.com/matrix-org/dendrite/clientapi/auth/authtypes"
-	"github.com/matrix-org/dendrite/internal/eventutil"
 	"github.com/matrix-org/dendrite/userapi/storage/accounts"
 	"github.com/matrix-org/gomatrixserverlib"
 )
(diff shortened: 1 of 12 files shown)


appservice/api/query.go
```diff
@@ -109,7 +109,7 @@ func RetrieveUserProfile(
 
 	// If no user exists, return
 	if !userResp.UserIDExists {
-		return nil, eventutil.ErrProfileNoExists
+		return nil, errors.New("no known profile for given user ID")
```
neilalexander, line 112: I'm the root of an import cycle :-(

Kegsay (approved):

PR #1661, opened 2020-12-18
Ensure we wake for our own device list updates

neilalexander:
This guarantees that we will be woken up for our own device list updates.


Kegsay (approved):
Seems plausible. We did this already in https://github.com/matrix-org/dendrite/blob/fac71edc62508e517b188e00fb64651910761ff0/syncapi/internal/keychange.go#L240 but that's only going to work for catchups - if the user was already waiting on `/sync` then it would be skipped.

PR #1665, opened 2020-12-18
fix imports

6543:
fix imports

Signed-off-by: 6543 <6543@obermui.de>

### DIFF
--- a/userapi/storage/accounts/sqlite3/storage.go
+++ b/userapi/storage/accounts/sqlite3/storage.go
@@ -22,9 +22,6 @@ import (
 	"strconv"
 	"sync"
 
-	// Import the sqlite3 package
-	_ "github.com/mattn/go-sqlite3"
-
 	"github.com/matrix-org/dendrite/clientapi/auth/authtypes"
 	"github.com/matrix-org/dendrite/internal/sqlutil"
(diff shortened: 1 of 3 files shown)


userapi/storage/accounts/sqlite3/storage.go
```diff
@@ -22,14 +22,16 @@ import (
 	"strconv"
 	"sync"
 
+	// Import the sqlite3 package
+	_ "github.com/mattn/go-sqlite3"
```
Kegsay, line 26: We don't technically need this here as we are importing it in other places. It's a little misleading to put this here because we don't always use `mattn/go-sqlite3` - in WASM mode we use https://github.com/matrix-org/go-sqlite3-js

6543: then the best way is to just rm the comment as a whole

userapi/storage/accounts/sqlite3/storage.go
```diff
@@ -22,14 +22,16 @@ import (
...
+
```
6543, line 27:
```suggestion
```

6543:
@Kegsay :rocket:

Kegsay (approved):
Thanks!

PR #1666, opened 2020-12-18
Reduce database impact on getStateDeltas

neilalexander:
This performs a simpler query first to work out the smallest range to satisfy the state deltas, if there are indeed any at all.


Kegsay:
In principle LGTM, unsure how much this would actually help though?

Kegsay:
Also looks like SQLite is unhappy in CI: `storage_test.go:223: failed to do sync: d.addPDUDeltaToResponse: d.getStateDeltas: sql: Scan error on column index 1, name "MIN(id)": converting NULL to int64 is unsupported`

PR #1667, opened 2020-12-18
Limit sync properly for getStateDeltas

neilalexander:
Apparently we don't provide the limit to `getStateDeltas`.


Kegsay:
Hmm, I don't know when `limit` was added to state filters but I'm fairly sure we don't want to do this. Clients will not get the complete room state if we do this.

PR #1675, opened 2021-01-02
Add MSC's section to example config

jaywink:
The config section for MSC's was missing from the example config. Even though these are experimental features, it might be good to add the section into the example config for visibility.

Signed-off-by: Jason Robinson <mail@jasonrobinson.me>

### DIFF
--- a/dendrite-config.yaml
+++ b/dendrite-config.yaml
@@ -253,6 +253,18 @@ media_api:
     height: 480
     method: scale
 
+# Configuration for experimental MSC's
+mscs:
+  # A list of enabled MSC's
+  # Currently valid values are:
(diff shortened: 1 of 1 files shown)


dendrite-config.yaml
```diff
@@ -253,6 +253,18 @@ media_api:
...
 
+# Configuration for experimental MSC's
+mscs:
+  # A list of enabled MSC's
+  # Currently valid values are:
+  # - msc2836         (Threading, see https://github.com/matrix-org/matrix-doc/pull/2836)
+  mscs: []
+  database:
+    connection_string: file:mscs.db
+    max_open_conns: 10
```
neilalexander, line 264: Just as a note to look at, but we need to make sure that the various `max_open_conns` in the sample config don’t collectively exceed 100.

Kegsay (approved):

PR #1680, opened 2021-01-04
Send events to appservice based on room membership

### LINKED ISSUE #1679: Events should be forwarded to appservices who have members joined to the room
https://matrix.org/docs/spec/application_service/r0.1.2#registration

>Similarly, the application service is said to be interested in a given event if one of the application service's namespaced users is the target of the event, **or is a joined member of the room where the event occurred**.

Half-Shot:
Fixes #1679 

~~Note that this does not currently work and needs input from maintainers~~ Thanks Neil!


appservice/consumers/roomserver.go
```diff
@@ -159,5 +159,33 @@ func (s *OutputRoomEventConsumer) appserviceIsInterestedInEvent(ctx context.Cont
 		}).WithError(err).Errorf("Unable to get aliases for room")
 	}
 
+	// Check if any of the members in the room match the appservice
+	// TODO: appserviceIsInterestedInEvent should be wrapped in a cache
+	// as this part can involve a lot of DB lookups
+	membershipReq := api.QueryMembershipsForRoomRequest{
```
Half-Shot, line 165: This seems to be the wrong function to call. It expects a userId (presumably because this call is expected to handle `/members`). I'd like to just get membership of the room without the context of a sender user.

appservice/consumers/roomserver.go
```diff
@@ -159,5 +159,33 @@ func (s *OutputRoomEventConsumer) appserviceIsInterestedInEvent(ctx context.Cont
...
+		JoinedOnly: true,
+		RoomID:     event.RoomID(),
+	}
+	var membershipRes api.QueryMembershipsForRoomResponse
+
+	// XXX: This should be membership at the time of the event, not current membership. I'm not sure
```
Half-Shot, line 171: Need help on this one also.

appservice/consumers/roomserver.go
```diff
@@ -159,5 +159,33 @@ func (s *OutputRoomEventConsumer) appserviceIsInterestedInEvent(ctx context.Cont
...
+	// how to extract this though, help?!
+	// TODO: We also don't need to get all the members. Could we query the DB directly with a set of
```
Half-Shot, line 173: I leave this up to the maintainers on the urgency of this, but obviously each event going through dendrite will currently check the membership of a room. If there is an easy way to cache *this* function call and invalidate it when the membership (or aliases) changes, then let's do that.

appservice/consumers/roomserver.go
```diff
@@ -159,5 +159,33 @@ func (s *OutputRoomEventConsumer) appserviceIsInterestedInEvent(ctx context.Cont
 		}).WithError(err).Errorf("Unable to get aliases for room")
 	}
 
+	// Check if any of the members in the room match the appservice
```
Half-Shot, line 162: I leave this up to the maintainers on the urgency of this, but obviously each event going through dendrite will currently check the membership of a room. If there is an easy way to cache this function call and invalidate it when the membership (or aliases) changes, then let's do that.

appservice/consumers/roomserver.go
Half-Shot: We're using `QueryStateAfterEvents` which seems to do the right thing.

appservice/consumers/roomserver.go
Half-Shot: Opened #1682 to track, we're consciously deferring this work as it's difficult to have caches span several components at once.

appservice/consumers/roomserver.go
```diff
@@ -125,6 +125,8 @@ func (s *OutputRoomEventConsumer) filterRoomserverEvents(
 
 // appserviceIsInterestedInEvent returns a boolean depending on whether a given
 // event falls within one of a given application service's namespaces.
+//
+// TODO: This should be cached, see https://github.com/matrix-org/dendrite/issues/1682
 func (s *OutputRoomEventConsumer) appserviceIsInterestedInEvent(ctx context.Context, event *gomatrixserverlib.HeaderedEvent, appservice config.ApplicationService) bool {
```
Half-Shot, line 130: > appservice/consumers/roomserver.go:130:1: cyclomatic complexity 15 of func `(*OutputRoomEventConsumer).appserviceIsInterestedInEvent` is high (> 13) (gocyclo)

Can I do something about this?

Half-Shot: solved by splitting out to another function

appservice/consumers/roomserver.go
```diff
@@ -123,8 +123,39 @@ func (s *OutputRoomEventConsumer) filterRoomserverEvents(
 	return nil
 }
 
+// appserviceHasMembershipInRoom returns a boolean depending on whether a given
+// appservice has membership at the time a given event was created.
+func (s *OutputRoomEventConsumer) appserviceHasMembershipForEvent(ctx context.Context, event *gomatrixserverlib.HeaderedEvent, appservice config.ApplicationService) bool {
```
Kegsay, line 128: `appserviceHasMembershipForEvent` or `appserviceHasMembershipInRoom`? Comments disagree.

neilalexander (approved):
Based on our chat yesterday I think this should be fine for the short term.

appservice/consumers/roomserver.go
Kegsay: Also what does "has membership" mean? Joined? Invited? Left? They are all memberships...

appservice/consumers/roomserver.go
```diff
@@ -123,8 +123,39 @@ func (s *OutputRoomEventConsumer) filterRoomserverEvents(
...
+	// Check if any of the members in the room match the appservice
+	membershipReq := api.QueryStateAfterEventsRequest{
+		PrevEventIDs: []string{event.EventID()},
+		RoomID:       event.RoomID(),
+	}
+	var membershipRes api.QueryStateAfterEventsResponse
+
+	// XXX: This could potentially race if the state for the event is not known yet
+	// e.g. the event came over federation but we do not have the full state persisted.
+	if err := s.rsAPI.QueryStateAfterEvents(ctx, &membershipReq, &membershipRes); err == nil {
```
Kegsay, line 138: This is the wrong API to use because we're pulling out the entire room state for this call. We can't make use of `QueryStateAfterEventsRequest.StateToFetch` because we want to glob all room member events. Use `QueryMembershipsForRoomRequest` instead:
```go
// QueryMembershipsForRoomRequest is a request to QueryMembershipsForRoom
type QueryMembershipsForRoomRequest struct {
	// If true, only returns the membership events of "join" membership
	JoinedOnly bool `json:"joined_only"`
	// ID of the room to fetch memberships from
	// ID of the user sending the request
	Sender string `json:"sender"`
}

// QueryMembershipsForRoomResponse is a response to QueryMembershipsForRoom
type QueryMembershipsForRoomResponse struct {
	// The "m.room.member" events (of "join" membership) in the client format
	JoinEvents []gomatrixserverlib.ClientEvent `json:"join_events"`
	// True if the user has been in room before and has either stayed in it or
	// left it.
	HasBeenInRoom bool `json:"has_been_in_room"`
	// True if the user asked to forget this room.
	IsRoomForgotten bool `json:"is_room_forgotten"`
}
```

Kegsay (changes requested):
This is too heavy on memory at present.

appservice/consumers/roomserver.go
Half-Shot: `QueryMembershipsForRoomRequest` doesn't work here because it takes a `sender` param, and we don't have a sender here.

appservice/consumers/roomserver.go
Half-Shot: Changed function to reflect reality.

roomserver/internal/query/query.go
```diff
@@ -242,6 +242,24 @@ func (r *Queryer) QueryMembershipsForRoom(
 		return err
 	}
 
+	if request.Sender == "" {
```
Kegsay, line 245: Please comment what the behaviour is supposed to be with an absent sender.

Kegsay (commented):
It's still not great that we are pulling out a ton of `ClientEvent` structs for _every event_ we send to an AS, but this is marginally better than before.

roomserver/internal/query/query.go
neilalexander: Done, although I had already added a comment into the `api` package.

PR #1683, opened 2021-01-05
Disable gocyclo

neilalexander:
Proposing to disable the `gocyclo` linter. It's an annoyingly blunt tool, we had 43 `nolint:gocyclo` directives scattered around the codebase and at that point there's really not much use in it anyway.


Kegsay:
I'd rather increase the threshold than remove it entirely. Yes it's blunt but it serves a useful purpose.

neilalexander:
Will increase the limit but will do so in a new PR.

PR #1684, opened 2021-01-05
Fix appsevice alias queries part 2

### LINKED ISSUE #1655: Appservice alias query support is broken

Half-Shot:
Fixes #1655 

Depends on https://github.com/matrix-org/dendrite/pull/1680

The previous PR #1660 didn't work for a few reasons.
- It was not called if a user attempted to join an alias that didn't exist
- If the call to the DB to fetch the alias gave an error, the AS would not be called at all.

This PR addresses these problems. I've had to add a new field `ShouldHitAppservice` to GetRoomIDForAliasRequest to avoid a loop where the AS would hit the createRoom code to create a room, which would then hit the AS, which would then hit the createRoom code...etc.

I believe this works but I welcome words of experience on how to do this better.


Zauberstuhl:
With this patch I am able to use https://github.com/tulir/mautrix-telegram with dendrite.
I cannot really comment on the patch itself but I though it would be worth noticing that your patch adds value :)

Thanks !

Half-Shot:
Yeah, it also got my bridges going for me. I've just not had time to finish it off!

anoadragon453:
@Half-Shot Should this PR be taken out of draft state?

Zauberstuhl:
I upgrade from 0.3.10 to 0.3.11 and now I am running into following issue:

```
time="2021-03-02T21:13:46.868062675Z" level=panic msg="failed to generate bot account for appservice" func="NewInternalAPI\n\t" file=" [github.com/matrix-org/dendrite/appservice/appservice.go:75]" appservice=telegram error="Username already exists"
panic: (*logrus.Entry) 0xc000382d20

goroutine 1 [running]:
github.com/sirupsen/logrus.Entry.log(0xc0000dac40, 0xc00037f920, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        github.com/sirupsen/logrus@v1.7.0/entry.go:255 +0x325
github.com/sirupsen/logrus.(*Entry).Log(0xc000382cb0, 0x0, 0xc0000ef878, 0x1, 0x1)
        github.com/sirupsen/logrus@v1.7.0/entry.go:283 +0xf0
github.com/sirupsen/logrus.(*Entry).Logf(0xc000382cb0, 0xc000000000, 0x104e917, 0x2d, 0x0, 0x0, 0x0)
        github.com/sirupsen/logrus@v1.7.0/entry.go:329 +0xe5
github.com/sirupsen/logrus.(*Entry).Panicf(...)
        github.com/sirupsen/logrus@v1.7.0/entry.go:367
github.com/matrix-org/dendrite/appservice.NewInternalAPI(0xc0000e5780, 0x11b2ec0, 0xc00038ce40, 0x11bb078, 0xc0000f3100, 0xc0003828c0, 0x11ad500)
        github.com/matrix-org/dendrite/appservice/appservice.go:75 +0x8c8
main.main()
        github.com/matrix-org/dendrite/cmd/dendrite-monolith-server/main.go:124 +0x550
```

Is this related?

neilalexander (approved):

PR #1688, opened 2021-01-07
Sync refactor — Part 1

neilalexander:
This PR refactors the sync API somewhat to break out the streams into individual stream providers, and to remove a bunch of spaghetti code.

This should make the sync API significantly easier to reason about and work on.


syncapi/consumers/clientapi.go
```diff
@@ -92,7 +96,8 @@ func (s *OutputClientDataConsumer) onMessage(msg *sarama.ConsumerMessage) error
 		}).Panicf("could not save account data")
 	}
 
-	s.notifier.OnNewEvent(nil, "", []string{string(msg.Key)}, types.StreamingToken{PDUPosition: pduPos})
+	s.streams.AccountDataStreamProvider.Advance(streamPos)
```
Kegsay, line 99: It feels odd that writes to create new positions are done here in the consumer but recording this is in the provider. It feels poorly encapsulated, as I would expect a stream provider to be able to handle insertion too.

syncapi/consumers/roomserver.go
```diff
@@ -290,9 +299,12 @@ func (s *OutputRoomEventConsumer) onRetireInviteEvent(
 		}).Panicf("roomserver output log: remove invite failure")
 		return nil
 	}
+
 	// Notify any active sync requests that the invite has been retired.
 	// Invites share the same stream counter as PDUs
```
Kegsay, line 304: This comment presumably lies now?

syncapi/consumers/roomserver.go
```diff
@@ -327,12 +340,13 @@ func (s *OutputRoomEventConsumer) onRetirePeek(
 		}).Panicf("roomserver output log: write peek failure")
 		return nil
 	}
-	// tell the notifier about the new peek so it knows to wake up new devices
-	s.notifier.OnRetirePeek(msg.RoomID, msg.UserID, msg.DeviceID)
```
Kegsay, line 331: I didn't see where this got moved to?

syncapi/storage/interface.go
```diff
@@ -30,6 +28,26 @@ import (
 
 type Database interface {
 	internal.PartitionStorer
+
+	MaxStreamTokenForPDUs(ctx context.Context) (types.StreamPosition, error)
```
Kegsay, line 32: Token or Position? We should be consistent in our terminology and I think you mean Position.

syncapi/streams/stream_accountdata.go
```diff
@@ -0,0 +1,132 @@
...
+				gomatrixserverlib.ClientEvent{
+					Type:    datatype,
+					Content: gomatrixserverlib.RawJSON(databody),
+				},
+			)
+			req.Response.Rooms.Join[r] = j
+		}
+	}
+
+	return p.LatestPosition(ctx)
```
Kegsay, line 63: Won't this race? E.g `QueryAccountData` returns stale data but `LatestPosition` returns the latest position so we never get the latest info?

syncapi/streams/stream_pdu.go
```diff
@@ -0,0 +1,303 @@
...
+	ctx context.Context,
+	req *types.SyncRequest,
+) types.StreamPosition {
+	from := types.StreamPosition(0)
+	to := p.LatestPosition(ctx)
+
+	// Get the current sync position which we will base the sync response on.
+	r := types.Range{
+		From:      to,
+		To:        0,
```
Kegsay, line 38: This feels like it's the wrong way around?

syncapi/syncapi.go
```diff
@@ -50,57 +53,53 @@ func AddPublicRoutes(
...
+		logrus.WithError(err).Panicf("failed to load notifier")
 	}
 
-	requestPool := sync.NewRequestPool(syncDB, cfg, notifier, userAPI, keyAPI, rsAPI)
+	requestPool := sync.NewRequestPool(syncDB, cfg, userAPI, keyAPI, rsAPI, streams, notifier)
 
 	keyChangeConsumer := consumers.NewOutputKeyChangeEventConsumer(
 		cfg.Matrix.ServerName, string(cfg.Matrix.Kafka.TopicFor(config.TopicOutputKeyChangeEvent)),
-		consumer, notifier, keyAPI, rsAPI, syncDB,
+		consumer, keyAPI, rsAPI, syncDB, notifier, streams,
```
Kegsay, line 67: It would be nicer if we just passed in the stream required by `KeyChangeEventConsumer`, etc.

Kegsay (changes requested):
Definitely an improvement from before!

syncapi/consumers/clientapi.go
neilalexander: Indeed, but given that the `StreamProvider` and `PartitionedStreamProvider` interfaces have to be generic, I didn't really know how best to handle inserts and didn't give it a lot of thought. Happy to revisit this in the next PR.

syncapi/syncapi.go
neilalexander: Done for all but the roomserver consumer, which currently uses multiple streams.

syncapi/streams/stream_pdu.go
neilalexander: I've added a comment to this but it's right - for a complete sync, we want to show the most recent messages in the room, so we start at the current position and work backwards.

syncapi/streams/stream_accountdata.go
neilalexander: It probably will, but this is a bit of a downside to having the user API own this information and not send it out via kafka. We can fix this too in a future PR.

syncapi/storage/interface.go
neilalexander: Fixed - it's Position.

syncapi/consumers/roomserver.go
neilalexander: This was a mistake, have put it back, but added the position into the parameters too, since we still need to update the PDU position for peeks for now.

syncapi/consumers/roomserver.go
neilalexander: It does, and it's gone.

syncapi/syncapi.go
neilalexander: Now done it for the roomserver consumer too, just passed in more than one.

Kegsay (approved):

MadLittleMods:
Some follow-up fixes to this refactor in https://github.com/matrix-org/dendrite/pull/1703

PR #1690, opened 2021-01-08
Add madvdontneed=1 to decrease RAM consumption for systemd service

rany2:


phol:
This setting is no longer necessary with go versions >`go1.16beta1`, according to [this commit](https://github.com/golang/go/commit/05e6d28849293266028c0bc9e9b0f8d0da38a2e2).

rany2:
I think it's best to have this until Dendrite's minimum Go requirements no longer need `madvdontneed=1`

neilalexander:
Thanks for this!

PR #1695, opened 2021-01-11
Exclude old events from sync, don't wake streams

neilalexander:
"Old" events — that is, events that have been pulled in because they were missing `prev_events` or `auth_events` — shouldn't appear at the bottom of the sync stream for clients and shouldn't wake up notifiers.


Kegsay:
In theory, LGTM. Tests are unhappy though.

Kegsay:
I feel like this is stale at this point.

Kegsay:
We should be waking streams for /g_m_e traffic, therefore I'm closing this PR.

PR #1700, opened 2021-01-11
MSC2946: Spaces Summary

Kegsay:
This is an implementation of [MSC2946](https://github.com/matrix-org/matrix-doc/blob/kegan/spaces-summary/proposals/2946-spaces-summary.md). With some tests.

Doesn't yet paginate.

### DIFF
--- a/setup/mscs/msc2946/msc2946.go
+++ b/setup/mscs/msc2946/msc2946.go
@@ -288,16 +288,40 @@ func (w *walker) publicRoomsChunk(roomID string) *gomatrixserverlib.PublicRoom {
 
 // authorised returns true iff the user is joined this room or the room is world_readable
 func (w *walker) authorised(roomID string) bool {
-	var queryRes roomserver.QueryMembershipForUserResponse
-	err := w.rsAPI.QueryMembershipForUser(w.ctx, &roomserver.QueryMembershipForUserRequest{
+	hisVisTuple := gomatrixserverlib.StateKeyTuple{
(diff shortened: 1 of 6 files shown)


setup/mscs/msc2946/msc2946.go
```diff
@@ -0,0 +1,366 @@
...
+	base *setup.BaseDendrite, rsAPI roomserver.RoomserverInternalAPI, userAPI userapi.UserInternalAPI,
+) error {
+	db, err := NewDatabase(&base.Cfg.MSCs.Database)
+	if err != nil {
+		return fmt.Errorf("Cannot enable MSC2946: %w", err)
+	}
+	hooks.Enable()
+	hooks.Attach(hooks.KindNewEventPersisted, func(headeredEvent interface{}) {
+		he := headeredEvent.(*gomatrixserverlib.HeaderedEvent)
+		hookErr := db.StoreReference(context.Background(), he)
```
neilalexander, line 82: I realise now that filtering takes place in `StoreReference`, but that wasn't immediately obvious and left me wondering why we'd try to find references in all timeline events. Might make sense to perform the filtering here instead, or at the very least comment what's going on, as leaving the storage API to filter what we give it isn't really a pattern we use elsewhere.

neilalexander (approved):
Overall looks fine to me!

setup/mscs/msc2946/msc2946.go
Kegsay: Will tweak this in the next PR.

PR #1702, opened 2021-01-12
Simplify send-to-device messaging

neilalexander:
This PR refactors the send-to-device stream in the sync API so that it takes advantage of having its own stream positions, and simplifies the interface significantly. 

It should also fix some bugs with send-to-device message delivery, which may have been unreliable before.

### DIFF
--- a/syncapi/storage/sqlite3/deltas/20210112130000_sendtodevice_sentcolumn.go
+++ b/syncapi/storage/sqlite3/deltas/20210112130000_sendtodevice_sentcolumn.go
@@ -16,7 +16,6 @@ package deltas
 
 import (
 	"database/sql"
-	"fmt"
 
 	"github.com/matrix-org/dendrite/internal/sqlutil"
 )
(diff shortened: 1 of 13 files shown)


syncapi/storage/sqlite3/deltas/20210112130000_sendtodevice_sentcolumn.go
```diff
@@ -0,0 +1,67 @@
+// Copyright 2020 The Matrix.org Foundation C.I.C.
```
Kegsay, line 1: 2021

Kegsay (approved):

syncapi/storage/sqlite3/deltas/20210112130000_sendtodevice_sentcolumn.go
neilalexander: Oh yeah, that happened didn't it?

PR #1703, opened 2021-01-12
Sync refactor fixes

neilalexander:
It turns out that it makes sense for each of the consumers to have access to `*streams.Streams` after all, because then we can send the *absolutely latest* stream token for all streams, rather than just delta-ing based on the thing we just consumed from Kafka. This should hopefully fix the problem where some sync positions are advancing much slower than they should be.

We also now only hit the notifier if the stream *actually* caused a stream update.


neilalexander:
Testing this further, I don't think this is doing the right thing.

PR #1708, opened 2021-01-13
Update /messages pagination token behaviour

### LINKED ISSUE #1707: Infinite loading Back Pagination with Element Android
### Description

According to the spec for GET /_matrix/client/r0/rooms/{roomId}/messages
link :  https://matrix.org/docs/spec/client_server/r0.6.1#id267

The start field of the response should be equal to the start parm:
> The token the pagination starts from. If dir=b this will be the token supplied in from.

Actually Dendrite is not doing it, and this is 'breaking' the android back pagination reach end detection causing endless loading when back paginating.

As per logs

> [github.com/matrix-org/dendrite/syncapi/routing/messages.go:198] OnIncomingMessagesRequest
	 Responding    backwards=true from=t10_1128 limit=50 req.id=zU95gMfPHTWk req.method=GET req.path="/_matrix/client/r0/rooms/!VMuQwSnXtHTdHLBe:localhost/messages"

And response is
> return_end=t0_0 return_start=t10_74 to=t0_0 user_id="@alice:localhost"

returned start is t10_74 instead of t10_1128

neilalexander:
This should fix #1707 hopefully. We should be setting `start` to the `from=` value, so as to not confuse clients by accidentally introducing gaps. This also fixes a problem where the `start` and `end` could be the wrong way around.

It also fixes a minor thing where we incorrectly applied filtering to world-readable rooms.

### DIFF
--- a/syncapi/routing/messages.go
+++ b/syncapi/routing/messages.go
@@ -283,13 +283,6 @@ func (r *messagesReq) retrieveEvents() (
 
 	// Sort the events to ensure we send them in the right order.
 	if r.backwardOrdering {
-		// A stream/topological position is a cursor located between two events.
-		// While they are identified in the code by the event on their right (if
-		// we consider a left to right chronological order), tokens need to refer
-		// to them by the event on their left, therefore we need to decrement the
-		// end position we send in the response if we're going backward.
-		end.Decrement()
-
 		// This reverses the array from old->new to new->old
 		reversed := func(in []*gomatrixserverlib.HeaderedEvent) []*gomatrixserverlib.HeaderedEvent {
 			out := make([]*gomatrixserverlib.HeaderedEvent, len(in))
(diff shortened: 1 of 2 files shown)


syncapi/routing/messages.go
```diff
@@ -273,6 +273,14 @@ func (r *messagesReq) retrieveEvents() (
 		return []gomatrixserverlib.ClientEvent{}, *r.from, *r.to, nil
 	}
 
+	// Get the position of the first and the last event in the room's topology.
+	// This position is currently determined by the event's depth, so we could
+	// also use it instead of retrieving from the database. However, if we ever
+	// change the way topological positions are defined (as depth isn't the most
+	// reliable way to define it), it would be easier and less troublesome to
+	// only have to change it in one place, i.e. the database.
+	start, end, err = r.getStartEnd(events)
```
Kegsay, line 282: Why move this here? Will this be correct for backwards and forwards pagination?

syncapi/routing/messages.go
```diff
@@ -388,33 +388,27 @@ func (r *messagesReq) filterHistoryVisible(events []*gomatrixserverlib.HeaderedE
...
-		r.ctx, events[0].EventID(),
-	)
-	if err != nil {
-		err = fmt.Errorf("EventPositionInTopology: for start event %s: %w", events[0].EventID(), err)
-		return
-	}
-	if r.backwardOrdering && events[len(events)-1].Type() == gomatrixserverlib.MRoomCreate {
-		// We've hit the beginning of the room so there's really nowhere else
-		// to go. This seems to fix Riot iOS from looping on /messages endlessly.
-		end = types.TopologyToken{}
```
Kegsay, line 401: We've lost this?

Kegsay (approved):

syncapi/routing/messages.go
neilalexander: I don't know if we strictly need this anymore, but I'll add it back in and `NOTSPEC` it.

PR #1716, opened 2021-01-18
Use go build instead of go install

c-x-berger:
`go install` doesn't like to cross-compile things. (Try running build.sh with `GOARCH` set to something other than what it "should" be.)

With `go build`, it appears that cross-compilation is really, really straightforward. Simply install a compiler for your target platform and set `GOARCH` and `CC` accordingly.

I need to test the produced binaries on a real ARM chip, but this approach does seem to work. (Go is rather new to me, so I'm flying by the seat of my pants here.)

See also #1613, also in the vein of "making ARM easier"

### Pull Request Checklist

Signed-off-by: `Caleb Xavier Berger <caleb.x.berger@gmail.com>`

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -17,13 +17,6 @@ else
     export FLAGS=""
 fi
 
-for pkg in ./cmd/*; do
-    if [[ -d "$pkg" && ! -L "$pkg" ]]; then
-        # https://unix.stackexchange.com/a/94307/125869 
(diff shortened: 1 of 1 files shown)


c-x-berger:
Not sure why the test is failing - I checked against the `go install` line and got the same MD5s for executables generated with `go install` and `go build`.

neilalexander (approved):

PR #1721, opened 2021-01-19
Basic sync filtering

neilalexander:
This provides basic filtering based on `limit`, `sender`, `not_sender`, `types` and `not_types`. 

PostgreSQL is pretty simple as we can prepare sufficiently complex statements ahead of time, but in SQLite we don't have that luxury so we have to build the queries as we need them. 

This also fixes a couple of minor overlaps in SQLite where multiple streams would consume the same stream ID counters.

I'll deal with the lack of `contains_url` in some places separately.

### DIFF
--- a/syncapi/storage/postgres/output_room_events_table.go
+++ b/syncapi/storage/postgres/output_room_events_table.go
@@ -84,12 +84,20 @@ const selectEventsSQL = "" +
 const selectRecentEventsSQL = "" +
 	"SELECT event_id, id, headered_event_json, session_id, exclude_from_sync, transaction_id FROM syncapi_output_room_events" +
 	" WHERE room_id = $1 AND id > $2 AND id <= $3" +
-	" ORDER BY id DESC LIMIT $4"
+	" AND ( $4::text[] IS NULL OR     sender  = ANY($4)  )" +
+	" AND ( $5::text[] IS NULL OR NOT(sender  = ANY($5)) )" +
+	" AND ( $6::text[] IS NULL OR     type LIKE ANY($6)  )" +
+	" AND ( $7::text[] IS NULL OR NOT(type LIKE ANY($7)) )" +
+	" ORDER BY id DESC LIMIT $8"
 
 const selectRecentEventsForSyncSQL = "" +
 	"SELECT event_id, id, headered_event_json, session_id, exclude_from_sync, transaction_id FROM syncapi_output_room_events" +
 	" WHERE room_id = $1 AND id > $2 AND id <= $3 AND exclude_from_sync = FALSE" +
-	" ORDER BY id DESC LIMIT $4"
+	" AND ( $4::text[] IS NULL OR     sender  = ANY($4)  )" +
+	" AND ( $5::text[] IS NULL OR NOT(sender  = ANY($5)) )" +
+	" AND ( $6::text[] IS NULL OR     type LIKE ANY($6)  )" +
+	" AND ( $7::text[] IS NULL OR NOT(type LIKE ANY($7)) )" +
+	" ORDER BY id DESC LIMIT $8"
 
 const selectEarlyEventsSQL = "" +
 	"SELECT event_id, id, headered_event_json, session_id, exclude_from_sync, transaction_id FROM syncapi_output_room_events" +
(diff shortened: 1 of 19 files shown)


syncapi/sync/request.go
```diff
@@ -51,16 +44,14 @@ func newSyncRequest(req *http.Request, device userapi.Device, syncDB storage.Dat
 			return nil, err
 		}
 	}
-	timelineLimit := DefaultTimelineLimit
 	// TODO: read from stored filters too
+	filter := gomatrixserverlib.DefaultFilter()
 	filterQuery := req.URL.Query().Get("filter")
 	if filterQuery != "" {
 		if filterQuery[0] == '{' {
 			// attempt to parse the timeline limit at least
```
Kegsay, line 52: This comment isn't true anymore.

syncapi/sync/request.go
```diff
@@ -69,23 +60,20 @@ func newSyncRequest(req *http.Request, device userapi.Device, syncDB storage.Dat
 				util.GetLogger(req.Context()).WithError(err).Error("gomatrixserverlib.SplitID failed")
 				return nil, err
 			}
-			f, err := syncDB.GetFilter(req.Context(), localpart, filterQuery)
-			if err == nil {
-				timelineLimit = f.Room.Timeline.Limit
+			if f, err := syncDB.GetFilter(req.Context(), localpart, filterQuery); err == nil {
+				filter = *f
+			} else {
+				panic(err)
```
Kegsay, line 66: We shouldn't panic, just fail the request?

syncapi/storage/postgres/output_room_events_table.go
```diff
@@ -84,17 +84,29 @@ const selectEventsSQL = "" +
 const selectRecentEventsSQL = "" +
 	"SELECT event_id, id, headered_event_json, session_id, exclude_from_sync, transaction_id FROM syncapi_output_room_events" +
 	" WHERE room_id = $1 AND id > $2 AND id <= $3" +
-	" ORDER BY id DESC LIMIT $4"
+	" AND ( $4::text[] IS NULL OR     sender  = ANY($4)  )" +
+	" AND ( $5::text[] IS NULL OR NOT(sender  = ANY($5)) )" +
+	" AND ( $6::text[] IS NULL OR     type LIKE ANY($6)  )" +
+	" AND ( $7::text[] IS NULL OR NOT(type LIKE ANY($7)) )" +
```
Kegsay, line 90: Why `LIKE`? I thought they were direct matching like `sender`?

syncapi/storage/sqlite3/filtering.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+	}
+	if count := len(nottypes); count > 0 {
+		query += " AND type NOT IN " + sqlutil.QueryVariadicOffset(count, offset)
+		for _, v := range nottypes {
+			params, offset = append(params, v), offset+1
+		}
+	}
+	if order != "" {
+		query += " ORDER BY id " + order
```
Kegsay, line 49: I don't like literally using `order` here as it opens up a SQL injection if the caller isn't careful. We are careful to only pass in `ASC` or `DESC`, but nothing stops a caller passing in an attacker controller input. It would be better if `order` was an enum and we `switch`ed off that.

syncapi/storage/sqlite3/output_room_events_table.go
```diff
@@ -298,16 +269,18 @@ func (s *outputRoomEventsStatements) InsertEvent(
...
 
-	addStateJSON, err := json.Marshal(addState)
-	if err != nil {
-		return 0, err
+	var addStateJSON, removeStateJSON []byte
+	if len(addState) > 0 {
+		addStateJSON, err = json.Marshal(addState)
+	}
+	if len(removeState) > 0 {
+		removeStateJSON, err = json.Marshal(removeState)
```
Kegsay, line 277: If `addState` contains invalid JSON but `removeState` does not then we will miss the error because on `:274` we assign to err:
```
addStateJSON, err = json.Marshal(addState)
```
fail to check it, then clobber `err` here. Add a `if err != nil` check please.

Kegsay (approved):
Broadly speaking, excellent. LGTM

syncapi/sync/request.go
neilalexander: You're right, that was from testing. It's fixed now.

syncapi/storage/sqlite3/filtering.go
neilalexander: Done.

syncapi/storage/sqlite3/output_room_events_table.go
neilalexander: Done.

syncapi/storage/postgres/output_room_events_table.go
neilalexander: Apparently we have support (in Postgres at least) for wildcard event types. From the spec:

> types | [string] | A list of event types to include. If this list is absent then all event types are included. A '*' can be used as a wildcard to match any sequence of characters.

This won't work on SQLite though.

PR #1723, opened 2021-01-20
Build Docker images with cross compilation instead of QEMU

c-x-berger:
This PR utilizes Go's cross-compilation powers and some Docker features to build Docker images much, much faster than the current system. To the tune of 13-14 minutes faster.

Since Dendrite requires a C compiler ([`sqlite3`](https://github.com/mattn/go-sqlite3) has some C code), we can't just set `GOARCH` and call it a day. We need a toolchain for the target architecture. Alpine doesn't package cross compiling toolchains, so instead we can just grab a pre-built toolchain from [musl.cc](https://musl.cc).
(This is skipped when `GOARCH` is the same as `GOHOSTARCH`. Also, while it's not much of a practical concern today, musl.cc doesn't have a complete "N to N" matrix... yet.)

`build/docker/Dockerfile` replaces `Dockerfile.monolith` and `Dockerfile.polylith`. It can build both the monolith and polylith images with a build `ARG`. This is documented in `build/docker/README.md`.
Unfortunately, it looks like the changes needed for cross-compilation support (e.g. use of `BUILDPLATFORM`) make the Dockerfile require BuildKit. This isn't a problem for CI, and from what I can tell, BuildKit comes with Docker now, but it does need to be manually enabled. (If anyone can get the image to build without BuildKit using `--build-arg BUILDPLATFORM=thing`, I'd love to know *how*.)

A notable downside of this approach is that it does complicate adding C dependencies. If Dendrite were to (for example) depend on `libpng` in the future, we would need to come back and deal with fetching that library for the target architecture. Other C-linked-to-Go libraries are liable to use `pkg-config`, which is another (much easier to tackle) problem.
That said, Dendrite currently only has one C dependency beyond the standard library, so this may not be a big deal.

Changes required to merge:
- [ ] only build on tagged releases
- [ ] tag images with version number
- [ ] push to Docker Hub and/or GitHub container registry

### Pull Request Checklist

Signed-off-by: `Caleb Xavier Berger <caleb.x.berger@gmail.com>`


Kegsay:
I am a fan of using Go cross-compiling and Docker instead of QEMU, but it's depressingly complicated with the C deps as you've noticed. I'm probably not best placed to review this, but in principle I am happy with this.

neilalexander (commented):
Thanks for submitting this and apologies that it's taken so long for a review.

While innovative and the principle is good, I am worried that this introduces a number of moving parts which are much harder for us to manage if we either want to add new platforms or new cgo dependencies in the future for whatever reason.

The current xbuild/qemu approach might be slow but it's straight-forward to understand and maintain, so I'm going to go ahead and close this for now.

PR #1726, opened 2021-01-20
Add sync API memberships table

neilalexander:
This table tracks the most recent time that a user transitioned to each membership state. This will allow us to find the most recent time that the user was invited to the room, joined the room, left the room, was kicked from the room etc.

This will allow us to work out what the boundaries are for history visibility.

### DIFF
--- a/syncapi/storage/postgres/memberships_table.go
+++ b/syncapi/storage/postgres/memberships_table.go
@@ -1,5 +1,4 @@
-// Copyright 2017-2018 New Vector Ltd
-// Copyright 2019-2020 The Matrix.org Foundation C.I.C.
+// Copyright 2021 The Matrix.org Foundation C.I.C.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -94,7 +93,7 @@ func (s *membershipsStatements) UpsertMembership(
 	_, err = sqlutil.TxStmt(txn, s.upsertMembershipStmt).ExecContext(
 		ctx,
 		event.RoomID(),
-		event.EventID(),
+		*event.StateKey(),
 		membership,
 		event.EventID(),
 		streamPos,
--- a/syncapi/storage/postgres/memberships_table.go
+++ b/syncapi/storage/postgres/memberships_table.go
@@ -1,0 +1,95 @@
+// Copyright 2017-2018 New Vector Ltd
+// Copyright 2019-2020 The Matrix.org Foundation C.I.C.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


syncapi/storage/postgres/memberships_table.go
```diff
@@ -0,0 +1,112 @@
+// Copyright 2017-2018 New Vector Ltd
+// Copyright 2019-2020 The Matrix.org Foundation C.I.C.
```
Kegsay, line 2: Lies

syncapi/storage/postgres/output_room_events_topology_table.go
```diff
@@ -115,10 +116,10 @@ func NewPostgresTopologyTable(db *sql.DB) (tables.Topology, error) {
 // on the event's depth.
 func (s *outputRoomEventsTopologyStatements) InsertEventInTopology(
 	ctx context.Context, txn *sql.Tx, event *gomatrixserverlib.HeaderedEvent, pos types.StreamPosition,
-) (err error) {
-	_, err = s.insertEventInTopologyStmt.ExecContext(
+) (topoPos types.StreamPosition, err error) {
```
Kegsay, line 119: Ew, this means we rely on depth even more?

syncapi/storage/sqlite3/memberships_table.go
```diff
@@ -0,0 +1,120 @@
...
+	streamPos, topologicalPos types.StreamPosition,
+) error {
+	membership, err := event.Membership()
+	if err != nil {
+		return fmt.Errorf("event.Membership: %w", err)
+	}
+	_, err = sqlutil.TxStmt(txn, s.upsertMembershipStmt).ExecContext(
+		ctx,
+		event.RoomID(),
+		event.EventID(),
```
Kegsay, line 97: This should be the user ID?

syncapi/storage/postgres/memberships_table.go
```diff
@@ -0,0 +1,112 @@
...
+	streamPos, topologicalPos types.StreamPosition,
+) error {
+	membership, err := event.Membership()
+	if err != nil {
+		return fmt.Errorf("event.Membership: %w", err)
+	}
+	_, err = sqlutil.TxStmt(txn, s.upsertMembershipStmt).ExecContext(
+		ctx,
+		event.RoomID(),
+		event.EventID(),
```
Kegsay, line 97: likewise, this should be the state key.

Kegsay (changes requested):

syncapi/storage/postgres/memberships_table.go
neilalexander: How is it I *never* remember to fix these?

syncapi/storage/postgres/output_room_events_topology_table.go
neilalexander: Well, only in so far as that we're replicating it into the memberships table so that we can work out the topological bounds for `/messages`. If we stopped using depth for the topological position, we'd stop using it here too.

syncapi/storage/sqlite3/memberships_table.go
neilalexander: You're right - fixed.

Kegsay (approved):

PR #1728, opened 2021-01-22
DNS caching

neilalexander:
Depends on matrix-org/gomatrixserverlib#251. This is **not** enabled by default.

The DNS caching behaviour is wildly different across Linux distributions and different platforms, so for deployments where no local caching resolver is available, this will stop Dendrite from hitting DNS so often.

### DIFF
--- a/build/docker/config/dendrite-config.yaml
+++ b/build/docker/config/dendrite-config.yaml
@@ -100,7 +100,7 @@ global:
     # Maximum number of entries to hold in the DNS cache, and
     # for how long those items should be considered valid.
     cache_size: 256
-    cache_lifetime: 600
+    cache_lifetime: 300
 
 # Configuration for the Appservice API.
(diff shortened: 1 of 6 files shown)


build/docker/config/dendrite-config.yaml
```diff
@@ -91,6 +91,17 @@ global:
...
+  # DNS cache options. The DNS cache may reduce the load on DNS servers
+  # if there is no local caching resolver available for use.
+  dns_cache:
+    # Whether or not the DNS cache is enabled.
+    enabled: false
+
+    # Maximum number of entries to hold in the DNS cache, and
+    # for how long those items should be considered valid.
+    cache_size: 256
+    cache_lifetime: 300
```
Kegsay, line 103: Needs units. Minutes? Seconds?

Kegsay (approved):

PR #1734, opened 2021-01-25
Graceful shutdowns

neilalexander:
This adds `process.ProcessContext` for allowing the consumers and HTTP listeners to shut down gracefully.

### DIFF
--- a/appservice/appservice.go
+++ b/appservice/appservice.go
@@ -89,7 +89,7 @@ func NewInternalAPI(
 	// We can't add ASes at runtime so this is safe to do.
 	if len(workerStates) > 0 {
 		consumer := consumers.NewOutputRoomEventConsumer(
-			base.Cfg, consumer, appserviceDB,
+			base.ProcessContext, base.Cfg, consumer, appserviceDB,
 			rsAPI, workerStates,
 		)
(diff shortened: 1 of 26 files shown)


Kegsay:
I don't feel that adding graceful shutdowns is going to help us out much. There are still many cases of uncontrolled shutdowns (e.g pull the power cord) which could corrupt state. Ideally, we should be fault tolerant to shutdowns at any point in the code, which implies we do some form of transaction log (even if that's done via the database). I'd rather spend time on that and get that working correctly, rather than adding graceful shutdowns which would be obsolete if we did the former.

neilalexander:
At the moment I'm a bit worried that killing Dendrite effectively simulates a crash every time, but mostly I'm concerned about the `gobind`/iOS stuff—when the app goes into the background, we get some notice that the process is about to be killed (a few seconds) but after that it is killed à la `SIGKILL`. Since we can't edit the database there to recover, we need to shut down safely.

Kegsay (approved):
As far as this diff is concerned, LGTM. I'm hesistant about adding this since I feel it's spurious, but if it actually solves an iOS problem then sure let's go for it.

PR #1736, opened 2021-01-25
Increase roomserver parallelisation

neilalexander:
This PR changes the RS input API so that, rather than blocking the input for a room for the entire `processRoomEvent` function, more of it happens in parallel. We do mutex on the latest events updater, because this is the place where we can hit races, so we do still have some wait contention there.

I don't really know if this reduces backpressure a lot, but it should stop us from waiting so much for state-before-event calculations, database persistence etc.


neilalexander:
It should be noted that I don't know if this will introduce other problems, especially in cases where the database is down or we fail to generate output events or something.

Kegsay:
What's the status of this PR @neilalexander ?

neilalexander:
I'll close this for now. It *works* to a degree, but it also causes a number of tests to fail as the ordering then becomes indeterminate thanks to the goroutine scheduling.

PR #1740, opened 2021-01-29
Add m.room.create to invite stripped state

Kegsay:
MSC1772 needs this because the create event contains info on if
the room is a space or not. The create event itself isn't sensitive
so other people may find this useful too.


neilalexander (approved):
J'approve!

PR #1741, opened 2021-01-29
Complete sync performance

neilalexander:
This improves complete sync performance by:

* Adding a global pool of 256 workers into the PDU stream provider and handling rooms in parallel
* Not sending device list updates when not needed

### DIFF
--- a/syncapi/streams/stream_devicelist.go
+++ b/syncapi/streams/stream_devicelist.go
@@ -19,7 +19,7 @@ func (p *DeviceListStreamProvider) CompleteSync(
 	ctx context.Context,
 	req *types.SyncRequest,
 ) types.LogPosition {
-	return p.IncrementalSync(ctx, req, types.LogPosition{}, p.LatestPosition(ctx))
+	return p.LatestPosition(ctx)
 }
 
 func (p *DeviceListStreamProvider) IncrementalSync(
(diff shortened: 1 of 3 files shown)


syncapi/streams/stream_pdu.go
```diff
@@ -2,18 +2,48 @@ package streams
...
+	"time"
 
 	"github.com/matrix-org/dendrite/syncapi/types"
 	userapi "github.com/matrix-org/dendrite/userapi/api"
 	"github.com/matrix-org/gomatrixserverlib"
+	"go.uber.org/atomic"
 )
 
+const PDU_STREAM_QUEUESIZE = 2048
+const PDU_STREAM_WORKERS = 256
```
Kegsay, line 15: Please comment what these configure and why too low is bad and too high is bad e.g:
```go
// The max number of per-room goroutines to have running.
// Too high and this will consume lots of CPU, too low and sync responses will take longer to process.
const PDU_STREAM_WORKERS = 256
```

syncapi/streams/stream_devicelist.go
```diff
@@ -19,7 +19,7 @@ func (p *DeviceListStreamProvider) CompleteSync(
 	ctx context.Context,
 	req *types.SyncRequest,
 ) types.LogPosition {
-	return p.IncrementalSync(ctx, req, types.LogPosition{}, p.LatestPosition(ctx))
+	return p.LatestPosition(ctx)
```
Kegsay, line 22: Cute that this is a one-liner :)

Kegsay (approved):

PR #1753, opened 2021-02-05
Fix crash in membership updater

neilalexander:
While the behaviour of #1748 was definitely not right, it turns out it was actually preventing us from dereferencing a nil pointer which would cause a crash.

This behaviour is probably better — comments inline to explain why, since it’s especially fiddly.

### DIFF
--- a/roomserver/internal/input/input_membership.go
+++ b/roomserver/internal/input/input_membership.go
@@ -106,6 +106,23 @@ func (r *Inputer) updateMembership(
 		// immediately, unless it's a Join update (e.g. profile update).
 		return updates, nil
 	}
+	
+	// In an ideal world, we shouldn't ever have "add" be nil and "remove" be
+	// set, as this implies that we're deleting a state event without replacing
+	// it (a thing that ordinarily shouldn't happen in Matrix). However, state
+	// resets are sadly a thing occasionally and we have to account for that.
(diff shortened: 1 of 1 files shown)


f12systems:
ran this on my machine and got a different 'crash' that seems to indicate the earlier issue
left the dB in a locked state

Specifically 
        Signalling other goroutines waiting for this goroutine to generate the thumbnail.  Height=480 MediaID=QiYNbjnWfZTiNYOaXhQGRNDW Origin=matrix.org ResizeMethod=scale Width=640 req.id=04NKqGdcNXgV req.method=GET req.path=/_matrix/media/r0/thumbnail/matrix.org/QiYNbjnWfZTiNYOaXhQGRNDW
panic: the ContinualConsumer in "syncapi/clientapi" failed to SetPartitionOffset: database is locked

goroutine 72 [running]:
github.com/matrix-org/dendrite/internal.(*ContinualConsumer).consumePartition(0xc0003aa300, 0x558211175340, 0xc000372400)
        github.com/matrix-org/dendrite/internal/consumers.go:129 +0x2ed
created by github.com/matrix-org/dendrite/internal.(*ContinualConsumer).StartOffsets
        github.com/matrix-org/dendrite/internal/consumers.go:107 +0x545

neilalexander:
This has fixed crashes on d.m.org and d.n.dev so merging.

PR #1761, opened 2021-02-11
Implement /createRoom power_level_content_override

t3chguy:
Doesn't look like sytest covered this parameter
It is one used by spaces so would be nice to land swiftly, lmk if you need me to do more.

### DIFF
--- a/clientapi/routing/createroom.go
+++ b/clientapi/routing/createroom.go
@@ -261,7 +261,8 @@ func createRoom(
 	powerLevelContent := eventutil.InitialPowerLevelsContent(userID)
 	if r.PowerLevelContentOverride != nil {
 		// Merge powerLevelContentOverride fields by unmarshalling it atop the defaults
-		powerLevelContentOverrideBytes, err := json.Marshal(r.PowerLevelContentOverride)
+		var powerLevelContentOverrideBytes []byte
+		powerLevelContentOverrideBytes, err = json.Marshal(r.PowerLevelContentOverride)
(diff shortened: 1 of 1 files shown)


clientapi/routing/createroom.go
```diff
@@ -38,16 +38,17 @@ import (
...
+	Name                      string                        `json:"name"`
+	Visibility                string                        `json:"visibility"`
+	Topic                     string                        `json:"topic"`
+	Preset                    string                        `json:"preset"`
+	CreationContent           map[string]interface{}        `json:"creation_content"`
+	InitialState              []fledglingEvent              `json:"initial_state"`
+	RoomAliasName             string                        `json:"room_alias_name"`
+	GuestCanJoin              bool                          `json:"guest_can_join"`
+	RoomVersion               gomatrixserverlib.RoomVersion `json:"room_version"`
+	PowerLevelContentOverride map[string]interface{}        `json:"power_level_content_override"`
```
neilalexander, line 51: You can probably simplify this by just setting the type here to `json.RawMessage` to defer un-marshalling. Then you won’t need to marshal it again before un-marshalling it over the top of `&powerLevelContent`.

t3chguy: Great; much cleaner. Done.

neilalexander:
Sytest meltdown in CI seems to be related to matrix-org/sytest#1011.

neilalexander (approved):
LGTM pending Sytest (which is actually running now).

PR #1766, opened 2021-02-16
Don't query for servers so often in /send

neilalexander:
This reduces the amount that we try to fetch remote servers in `/send`, and also doesn't cap it at 5 servers anymore, since that might not be a good thing to do in a room where servers are likely to be offline.

It also doesn't hit the internal API for missing auth events when there are no missing auth events.

It also removes `RemoveConflictsAdhoc` from the roomserver, since that's unneeded — it does exactly the same as `gomatrixserverlib.ResolveConflicts` does.


federationapi/routing/send.go
```diff
@@ -482,14 +483,10 @@ func (t *txnReq) retrieveMissingAuthEvents(
 		missingAuthEvents[missingAuthEventID] = struct{}{}
 	}
 
-	servers := t.getServers(ctx, e.RoomID())
-	if len(servers) > 5 {
-		servers = servers[:5]
```
Kegsay, line 487: For rooms like Matrix HQ this could result in slow message processing..

neilalexander: Well, in the worst case when the first set of servers we try are not available, but conversely, in a P2P room, we can't assume servers will be up. It's probably better to just try more servers rather than give up and drop the event.

federationapi/routing/send.go
```diff
@@ -558,7 +555,7 @@ func (t *txnReq) processEventWithMissingState(ctx context.Context, e *gomatrixse
 	// event ids and then use /event to fetch the individual events.
 	// However not all version of synapse support /state_ids so you may
 	// need to fallback to /state.
-
+	t.servers = t.getServers(ctx, e.RoomID())
```
Kegsay, line 558: Is this really the first place where servers is unpopulated? I'm not convinced we don't need to call `getServers` in more places. I would just have the `t.servers` cache in `getServers` and if it's nil (which it shouldn't legit be) then do the lookup, rather than call `getServers` here and using `t.servers` elsewhere which I feel may result in uninitialised usage.

Kegsay (changes requested):
Overall looks good, but worried about `t.servers` not be inited.

federationapi/routing/send.go
neilalexander: Done.

Kegsay (approved):

PR #1768, opened 2021-02-17
Gradually evict oldest cache entries

neilalexander:
Right now, when a cache grows, it never ever shrinks again, so the baseline memory usage of the Dendrite process creeps up over time.

This PR just ticks over once per minute and evicts the oldest item from each cache, so we have a chance to reclaim some of that memory for things that haven't been hit in a while.

### DIFF
--- a/internal/caching/impl_inmemorylru.go
+++ b/internal/caching/impl_inmemorylru.go
@@ -93,7 +93,12 @@ func cacheCleaner(caches ...*InMemoryLRUCachePartition) {
 	for {
 		time.Sleep(time.Minute)
 		for _, cache := range caches {
-			cache.lru.RemoveOldest()
+			// Hold onto the last 10% of the cache entries, since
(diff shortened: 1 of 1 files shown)


internal/caching/impl_inmemorylru.go
```diff
@@ -83,6 +89,15 @@ func NewInMemoryLRUCache(enablePrometheus bool) (*Caches, error) {
 	}, nil
 }
 
+func cacheCleaner(caches ...*InMemoryLRUCachePartition) {
+	for {
+		time.Sleep(time.Minute)
+		for _, cache := range caches {
+			cache.lru.RemoveOldest()
```
Kegsay, line 96: It would be nice to keep a minimum number of items in the cache, else I fear a small quiet time may result in lots of CPU churn as we re-populate the cache.

Kegsay (approved):

PR #1774, opened 2021-02-26
Return the current OTK count on an empty upload request

### LINKED ISSUE #1773: Dendrite doesn't return the current number of unclaimed one-time keys on an empty keys upload request.
### Background information

- **Dendrite version or git SHA**: 0.3.10
- **SQLite3 or Postgres?**: SQLite
- **Running in Docker?**: Yes
- **`go version`**: go1.16 linux/amd64

### Description

The js-sdk and thus Element web use, under certain circumstances, an empty `/keys/upload` requests to check the current number of unclaimed one-time keys on the server. Dendrite returns this response for such an empty request:
```json
{
    "one_time_key_counts": {}
}
```

The js-sdk interprets this response as having 0 one-time keys uploaded due to the logic over [here](https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/crypto/index.js#L1953).

It then proceeds to upload 5 signed one-time keys and gets the correct response from Dendrite:
```json
{
    "one_time_key_counts": {
        "signed_curve25519": 55
    }
}
```

The js-sdk then resets their count over [here](https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/crypto/index.js#L1953), expecting a fresh count from the sync. The next sync from Dendrite doesn't contain the one-time key counts either, so [this](https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/sync.js#L1363) logic never gets executed and the js-sdk will perform another count check using an empty upload request.

This spirals into the js-sdk slowly but surely uploading an infinite amount of one-time keys, since libolm can hold only 100 one-time keys in its fixed size buffer, older one-time keys get dropped. Dropping one-time keys like that will result in invalid Olm sessions being created, and thus clients won't be able to share room keys.

### Steps to reproduce

- start a Dendrite server
- user Element web to log in to the Dendrite server
- observe either the console log containing these log lines:
> generating oneTimeKeys
> calling _uploadOneTimeKeys
- or alternatively use the network tab in the inspector or mitmproxy to observe the constant key upload calls

<!--

``` (three backticks, on a line on their own), so that they are formatted legibly.

-->

The [spec](https://matrix.org/docs/spec/client_server/latest#post-matrix-client-r0-keys-upload) contains this explanation for the response of the keys/upload request:

> Required. For each key algorithm, the number of unclaimed one-time keys of that type currently held on the server for this device. 

The spec suggests that a `/keys/upload` requests must always return the number of one-time keys the server has, and the js-sdk can depend on this.

When and if the sync response should contain updated one-time key counts is not so clear cut, the field in the sync response is optional and no rules or logic are given when the count should be added to the response. There are a couple of options:

1. Return it always - this is what Synapse currently does.
2. When the count decreases - this is probably the minimal required thing for the server to do, otherwise clients might never replenish their one-time keys.
3. When the count changes - this sounds sensible and would have avoided the bug in Element web as well, since the js-sdk discards the count from an actual upload response and expects the next sync to contain the counts.

### PR DESCRIPTION
**Pull request by neilalexander**
This should hopefully fix #1773 and improve E2EE reliability.

### CODE REVIEW

File: keyserver/internal/internal.go [hunk 1]
```diff
@@ -513,6 +513,18 @@ func (a *KeyInternalAPI) uploadLocalDeviceKeys(ctx context.Context, req *api.Per
 }
 
 func (a *KeyInternalAPI) uploadOneTimeKeys(ctx context.Context, req *api.PerformUploadKeysRequest, res *api.PerformUploadKeysResponse) {
+	if len(req.OneTimeKeys) == 0 {
```
Kegsay, line 516: Can we add an assertion to fail the request if `req.UserID` or `req.DeviceID` are empty?

Kegsay (approved):

PR #1775, opened 2021-02-26
WIP: Implement failing media API tests

### LINKED ISSUE #1303: Implement failing media API tests
Sytests:
```
    × Can fetch images in room
    × Test URL preview
    × Can read configuration endpoint
```

DelusionalOptimist:
Fixes #1303 

 × Can fetch images in room
 × Test URL preview
 × ~~Can read configuration endpoint~~

Previews _just work_ as of now, pretty tentative implementation
Currently storing the response body for the url to be previewed is heavily inspired by the `/upload` endpoint. Not sure but probably gotta make a `url_cache` in `media_store` (similar to synapse).

Would be grateful to receive suggestions and comments. :grin: 

### Pull Request Checklist

Signed-off-by: `Rudraksh Pareek <54525605+DelusionalOptimist@users.noreply.github.com>`

### DIFF
--- a/mediaapi/routing/preview_url.go
+++ b/mediaapi/routing/preview_url.go
@@ -1,0 +1,400 @@
+// Copyright 2017 Vector Creations Ltd
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package routing
+
+import (
+	"context"
+	"crypto/rand"
+	"encoding/hex"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"path"
+	"strconv"
+	"strings"
+	"time"
+
+	"golang.org/x/net/html"
+
+	"github.com/matrix-org/dendrite/clientapi/jsonerror"
+	"github.com/matrix-org/dendrite/mediaapi/fileutils"
+	"github.com/matrix-org/dendrite/mediaapi/storage"
+	"github.com/matrix-org/dendrite/mediaapi/types"
+	"github.com/matrix-org/dendrite/setup/config"
+	"github.com/matrix-org/util"
+	log "github.com/sirupsen/logrus"
+)
+
+// the parameters included in the incoming preview_url request
+type PreviewUrlRequest struct {
+	url url.URL
+	ts  types.UnixMs
+}
+
+// metadata of the url (media/html) to be previewed
+type mediaInfo struct {
+	MediaMetadata *types.MediaMetadata
+	Logger        *log.Entry
+}
+
+// PreviewUrlResponse defines the format of the JSON response
+// https://matrix.org/docs/spec/client_server/latest#get-matrix-media-r0-preview-url
+// TODO: add more fields to the response
+type PreviewUrlResponse struct {
+	MatrixImageSize int64  `json:"matrix:image:size"`
+	OgImage         string `json:"og:image"`
+	OgSiteName      string `json:"og:site_name"`
+	OgType          string `json:"og:type"`
+	OgTitle         string `json:"og:title"`
+	OgUrl           string `json:"og:url"`
+	OgDescription   string `json:"og:description"`
+}
+
+// PreviewUrl implements GET /preview_url.
+// Current implementation gets the url, parses the meta tags to obtain OG
+// data and returns a JSONResponse for the client to process.
+// TODO: avoid this endpoint in encrypted rooms
+func PreviewUrl(
+	req *http.Request, cfg *config.MediaAPI, db storage.Database) util.JSONResponse {
+
+	// parsing the request
+	r, err := parseRequest(req)
+	if err != nil {
+		return *err
+	}
+
+	// downloading metadata for the url
+	mediaInfo, err := r.downloadUrl(cfg, req.Context(), db)
+	if err != nil {
+		return *err
+	}
+
+	// preparing the response
+	response, err := mediaInfo.prepareResponse(cfg)
+	if err != nil {
+		return *err
+	}
+
+	return util.JSONResponse{
+		Code: http.StatusOK,
+		JSON: response,
+	}
+}
+
+// parseRequest parses the incoming preview request to extract the url and ts.
+// Returns either a parsed PreviewUrlRequest or error formatted as
+// util.JSONResponse
+func parseRequest(req *http.Request) (*PreviewUrlRequest, *util.JSONResponse) {
+
+	// get the url to be previewed from the request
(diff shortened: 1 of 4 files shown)


mediaapi/routing/preview_url.go
```diff
@@ -0,0 +1,400 @@
...
+				Code: http.StatusBadRequest,
+				JSON: jsonerror.InvalidArgumentValue(
+					"Couldn't parse 'ts' to a valid integer" + err.Error()),
+			}
+		}
+
+		request := &PreviewUrlRequest{
+			url: *parsedUrl,
+			// Convert timestamp to ms
+			ts: types.UnixMs(ts / 1000000),
```
kegsay, line 135: `tsStr` will be in milliseconds already looking at https://github.com/matrix-org/synapse/blob/4b965c862dc66c0da5d3240add70e9b5f0aa720b/synapse/rest/media/v1/preview_url_resource.py#L198 so it should just be fine to do `ts: types.UnixMs(ts)`.

mediaapi/routing/preview_url.go
```diff
@@ -0,0 +1,400 @@
...
+// formatted as util.JSONResponse.
+// Current implementation for saving response is heavily similar to /upload,
+// need to work on this.
+// TODO: better logging
+func (previewReq *PreviewUrlRequest) downloadUrl(cfg *config.MediaAPI, ctx context.Context, db storage.Database) (*mediaInfo, *util.JSONResponse) {
+
+	urlString := previewReq.url.String()
+
+	// Get the URL
+	response, err := http.Get(urlString)
```
kegsay, line 163: Don't use the default client as it has no timeout, so malicious preview URLs could just not return anything and consume a file descriptor on the server. Use a client with a sensible timeout please e.g `&http.Client{Timeout:30 * time.Second}`

mediaapi/routing/preview_url.go
```diff
@@ -0,0 +1,400 @@
...
+	urlToPreview := req.URL.Query().Get("url")
+	if len(urlToPreview) == 0 {
+		return nil, &util.JSONResponse{
+			Code: http.StatusBadRequest,
+			JSON: jsonerror.MissingArgument("Missing URL"),
+		}
+	}
+
+	// parse the url
+	parsedUrl, err := url.Parse(urlToPreview)
```
kegsay, line 112: We need to check this URL against a blacklist of URLs otherwise this is just an arbitrary proxy which is 😨 

See https://github.com/matrix-org/synapse/blob/4b965c862dc66c0da5d3240add70e9b5f0aa720b/synapse/rest/media/v1/preview_url_resource.py#L202 and https://github.com/matrix-org/synapse/blob/240b3ce253b2b36d79a0a6cc18aefd70ad066de3/synapse/config/repository.py#L323

mediaapi/routing/preview_url.go
```diff
@@ -0,0 +1,400 @@
...
+	if response.StatusCode == 404 {
+		return nil, &util.JSONResponse{
+			Code: http.StatusBadRequest,
+			JSON: jsonerror.NotFound("Given url returned 404"),
+		}
+	}
+
+	// Save the response body to the temporary dir
+	// TODO: should be a new directory like url_cache maybe and
+	hash, bytesWritten, tmpDir, err := fileutils.WriteTempFile(ctx, response.Body, cfg.AbsBasePath)
```
kegsay, line 181: We definitely want to set a cap on the number of bytes we're willing to write to file, otherwise malicious urls can return a 10GB file and consume lots of disk space. 1MB should be plenty.

mediaapi/routing/preview_url.go
```diff
@@ -0,0 +1,400 @@
...
+		}
+	}
+
+	// metadata of the media/webpage to preview
+	urlMetadata := &mediaInfo{
+		MediaMetadata: &types.MediaMetadata{
+			Origin:        cfg.Matrix.ServerName,
+			FileSizeBytes: bytesWritten,
+			Base64Hash:    hash,
+			ContentType:   types.ContentType(response.Header["Content-Type"][0]),
```
kegsay, line 195: Use `response.Header.Get("Content-Type")` instead please. This otherwise will panic if the response has no Content-Type header.

mediaapi/routing/preview_url.go
```diff
@@ -0,0 +1,400 @@
...
+	}).Info("File downloaded")
+
+	if resErr := urlMetadata.storeFileAndMetadata(ctx, tmpDir, cfg.AbsBasePath, db); resErr != nil {
+		return nil, resErr
+	}
+
+	return urlMetadata, nil
+}
+
+func (m *mediaInfo) generateMediaID(ctx context.Context, db storage.Database) (types.MediaID, error) {
```
kegsay, line 263: This function will race with multiple uploads. If 2 people upload a file at the same time it's possible that they will be allocated the same media ID which will then be checked in the database, see it's free and then overwrite each other later on. This is however unlikely given the 32 bytes of entropy used here so maybe that's okay.

mediaapi/routing/preview_url.go
```diff
@@ -0,0 +1,400 @@
...
+		}
+	}
+
+	file := strings.NewReader(string(fileString))
+
+	tree, err := html.Parse(file)
+	if err != nil {
+		return nil, &util.JSONResponse{
+			Code: http.StatusInternalServerError,
+			JSON: jsonerror.Unknown("Couldn't get the path to the stored file " + err.Error()),
```
kegsay, line 363: Incorrect error message.

mediaapi/routing/preview_url.go
```diff
@@ -0,0 +1,400 @@
...
+		}
+	}
+
+	// map for storing the content values
+	m := make(map[string]string)
+
+	// Iterating the *html.Node, looking for og data
+	var f func(*html.Node)
+	f = func(n *html.Node) {
```
kegsay, line 372: Why not just do `f := func(n *html.Node)` to avoid the declaration on `:371`?

mediaapi/routing/preview_url.go
```diff
@@ -0,0 +1,400 @@
...
+			//parse attributes of the tag
+			for _, a := range n.Attr {
+				if a.Key == "property" && strings.HasPrefix(a.Val, "og:") {
+					slice := strings.Split(a.Val, ":")
+					m[slice[1]] = parseContent(n)
+				}
+			}
+		}
+		for c := n.FirstChild; c != nil; c = c.NextSibling {
+			f(c)
```
kegsay, line 384: We probably want to limit how much we're willing to explore on the HTML to avoid pathological cases which consume lots of CPU.

mediaapi/routing/preview_url.go
```diff
@@ -0,0 +1,400 @@
+// Copyright 2017 Vector Creations Ltd
```
kegsay, line 1: Copyrights are wrong.

kegsay (changes requested):
URL previews aren't an easy thing to tackle due to the amount of paranoia you need to apply when downloading arbitrary URLs from clients. I didn't really think about this when tagging this issue as `good-first-issue` so that's my bad. At present, there is still a lot of work which needs to be done before this is suitable to be merged.

DelusionalOptimist:
Hey @kegsay! Thanks for taking a look at this.
I agree this is some terrible work lol.
I learnt a lot while working on this so thanks for that too but unfortunately I've also got other commitments now and I'm unable to carry this on.
Closing this : )

PR #1777, opened 2021-03-02
Fix SQLite registration bug

neilalexander:
This fixes two things:

* The error type returned by SQLite on a violated constraint wasn't triggering the upstream code to check for duplicate accounts, causing the registration flow to proceed unexpectedly
* The order of insertions on an account creation has been changed to try the account insertion before the account data/profile


Kegsay (approved):
Wait for CI but this LGTM. This does mean that duplication /register requests for the same user will now fail even if the login credentials are correct, but this is probably fine since the client should be using /login.

neilalexander:
The sytest failure `Local device key changes get to remote servers with correct prev_id` is about to be blacklisted in a separate PR, the other tests are green.

PR #1778, opened 2021-03-02
Use new path for MSC2946

Kegsay:
Relies on https://github.com/matrix-org/gomatrixserverlib/pull/255 for outbound requests.


neilalexander (approved):
I assume you'll need to merge matrix-org/gomatrixserverlib#255 and then `go get ...gomatrixserverlib@squashedcommitid`, `go mod tidy` etc before merging this, but otherwise LGTM.

PR #1781, opened 2021-03-03
Docker fix and tweak

stoically:
- Recent docker versions want a trailing slash on the target when COPYing multiple files
- Default docker dendrite-config.yaml points to /var/dendrite/media as media storage path, which isn't mounted and can easily result in data loss (happened to me)

### Pull Request Checklist

Signed-off-by: `stoically <stoically@protonmail.com>`


neilalexander (approved):
Good spot, thanks for this!

PR #1785, opened 2021-03-03
Tweak AS registration check and AS component HTTP clients

neilalexander:
This tweaks the check for AS registration vs normal registration, since it seems that AS users are incorrectly being subjected to normal username checks (e.g. not being allowed to have usernames prefixed with `_`).

This also makes the following changes to the appservice component:
* A new appservice client is used for appservice HTTP requests
* A new `DisableTLSValidation` option has been added to the appservice config section for appservice traffic only

### DIFF
--- a/appservice/query/query.go
+++ b/appservice/query/query.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"net/http"
 	"net/url"
-	"time"
 
 	"github.com/matrix-org/dendrite/appservice/api"
 	"github.com/matrix-org/dendrite/setup/config"
@@ -159,11 +158,4 @@ func (a *AppServiceQueryAPI) UserIDExists(
 
 	response.UserIDExists = false
 	return nil
-}
-
-// makeHTTPClient creates an HTTP client with certain options that will be used for all query requests to application services
-func makeHTTPClient() *http.Client {
-	return &http.Client{
-		Timeout: time.Second * 30,
-	}
 }
--- a/appservice/query/query.go
+++ b/appservice/query/query.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/matrix-org/dendrite/appservice/api"
 	"github.com/matrix-org/dendrite/setup/config"
+	"github.com/matrix-org/gomatrixserverlib"
 	opentracing "github.com/opentracing/opentracing-go"
 	log "github.com/sirupsen/logrus"
 )
@@ -33,7 +34,7 @@ const userIDExistsPath = "/users/"
 
 // AppServiceQueryAPI is an implementation of api.AppServiceQueryAPI
 type AppServiceQueryAPI struct {
-	HTTPClient *http.Client
+	HTTPClient *gomatrixserverlib.Client
 	Cfg        *config.Dendrite
 }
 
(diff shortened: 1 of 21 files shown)


setup/base.go
```diff
@@ -290,6 +290,21 @@ func (b *BaseDendrite) CreateClient() *gomatrixserverlib.Client {
 	return client
 }
 
+// CreateClient creates a new client (normally used for media fetch requests).
+// Should only be called once per component.
+func (b *BaseDendrite) CreateAppserviceClient() *gomatrixserverlib.Client {
```
Kegsay, line 295: Comments lie.

appservice/query/query.go
```diff
@@ -33,7 +34,7 @@ const userIDExistsPath = "/users/"
 
 // AppServiceQueryAPI is an implementation of api.AppServiceQueryAPI
 type AppServiceQueryAPI struct {
-	HTTPClient *http.Client
+	HTTPClient *gomatrixserverlib.Client
```
Kegsay, line 37: Why are we using this client instead of a normal one? GMSL.Client is for federation traffic only I thought? ASes do not get sent signed traffic.

clientapi/routing/register.go
```diff
@@ -496,11 +497,20 @@ func Register(
...
+	// Is this an appservice registration? It will be if the access
+	// token is supplied
+	accessToken, accessTokenErr := auth.ExtractAccessToken(req)
+
 	// Squash username to all lowercase letters
 	r.Username = strings.ToLower(r.Username)
-
-	if resErr = validateUsername(r.Username); resErr != nil {
-		return *resErr
+	if r.Auth.Type == authtypes.LoginTypeApplicationService || accessTokenErr == nil {
```
Kegsay, line 506: I'd prefer it if we only look at the access token if the login type is set to AS?

clientapi/routing/register.go
```diff
@@ -679,6 +691,8 @@ func handleApplicationServiceRegistration(
 	cfg *config.ClientAPI,
 	userAPI userapi.UserInternalAPI,
 ) util.JSONResponse {
+	logrus.Warnf("APPSERVICE Is appservice registration %q", r.Username)
```
Kegsay, line 694: Debug logging

Kegsay (changes requested):

appservice/query/query.go
neilalexander: `gomatrixserverlib.Client` is for normal requests (like what the media API uses to fetch media), `gomatrixserverlib.FederationClient` is for signed requests.

clientapi/routing/register.go
neilalexander: It turns out that this isn't always the case. It's another one of those "spec is vague" things but Synapse also treats all calls to `/register` with an `access_token` as if they are coming from an AS.

neilalexander: Decided with @Half-Shot to instead enforce `authtypes.LoginTypeApplicationService` only.

appservice/query/query.go
Kegsay: What's the difference between GMSL.Client and net/http.Client?

neilalexander: Not much ,mostly just that `gomatrixserverlib.Client` has all of the correct wiring already to give it a timeout and/or disable TLS validation without needing to reinvent the wheel again in the AS code.

Kegsay (approved):

neilalexander:
I need to fix the sytests again to send `m.login.application_service` as the auth type, so I will do that before merging.

Half-Shot:
> I need to fix the sytests again to send `m.login.application_service` as the auth type, so I will do that before merging.

https://github.com/matrix-org/sytest/issues/1018

PR #1788, opened 2021-03-05
Use default transport for AS traffic

neilalexander:
This might fix the problem where it tries to enforce HTTPS.

cc @Half-Shot


neilalexander:
Oh something is horribly wrong with this PR.

PR #1791, opened 2021-03-06
Add CORS to nginx config

Cogitri:
Without this entry, setups where users have the homeserver on the URL
matrix.myurl.com but want the servername to be myurl.com don't work by default
since clients like element.io can't connect to the homeserver

### DIFF
--- a/docs/nginx/monolith-sample.conf
+++ b/docs/nginx/monolith-sample.conf
@@ -16,6 +16,9 @@ server {
     }
 
     location /.well-known/matrix/client {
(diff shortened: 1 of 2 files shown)


neilalexander:
I wonder if there's any downside to us allowing this in all cases, rather than having it commented out?

Cogitri:
Hm, not sure about that either, to be perfectly honest

Kegsay (approved):
It's as comments, which seems reasonable as a prompt for end users. We can worry about making it on-by-default another time.

PR #1792, opened 2021-03-08
Return a more useful error on /register AS spec compliance violation

neilalexander:
If you supply an `access_token` to `/register` but don't supply the login type then we should return a more useful error, so that it's more obvious that it's an appservice bug and not a Dendrite one.


Half-Shot (approved):
Looks reasonable to me :+1:

PR #1795, opened 2021-03-08
Inhibit login for AS users

neilalexander:
As per matrix-org/matrix-doc#2917, so that we don't go creating access tokens for those users.


neilalexander:
Turns out this isn't really necessary, but we may want to revisit it if the spec requires it in the future.

Issue #1644, opened 2020-12-16
Log permission denied issue when trying to get started


MadLittleMods:
Running through the ["get started"](https://github.com/matrix-org/dendrite#get-started) section in the readme on a fresh macbook results in a permission error when it tries to write some logs.

```sh
$ ./bin/dendrite-monolith-server --tls-cert server.crt --tls-key server.key --config dendrite.yaml
FATA[2020-12-16T06:39:03.379411000Z] [github.com/matrix-org/dendrite/internal/log.go:155] setupFileHook
Couldn't create directory /var/log/dendrite: "mkdir /var/log/dendrite: permission denied"
```

It seems necessary to use `sudo` when launching Dendrite to get it to work out of the box.

### Can the log directory be configured?

Yes, this can be configured in your [`dendrite.yaml` config](https://github.com/matrix-org/dendrite/blob/38318b0f162a41ac9d1aa60dc3b7693eae033851/dendrite-config.yaml#L350) under `logging`

### Potential solutions

 1. Update the readme instructions to note about changing the config for `logging` in addition to the `server_name`
 1. Update the log directory to default somewhere with permissions like in the project itself.

Kegsay:
We really shouldn't be using `/var/log` at all as a log location imo. `./logs` would be better since we'll be more likely to be able to write to it.

PR #1799, opened 2021-03-17
Use log directory that we will be able to write to

MadLittleMods:
Use log directory that we will be able to write to.

Fix https://github.com/matrix-org/dendrite/issues/1644

### Pull Request Checklist

* [ ] ~~Pull request includes a [sign off](https://github.com/matrix-org/dendrite/blob/master/docs/CONTRIBUTING.md#sign-off)~~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@
 /_test
 /vendor/bin
 /docker/build
+/logs
 
 # Architecture specific extensions/prefixes
 *.[568vq]
--- a/dendrite-config.yaml
+++ b/dendrite-config.yaml
@@ -383,4 +383,4 @@ logging:
 - type: file
   level: info
   params:
-    path: /var/log/dendrite
+    path: ./logs


dendrite-config.yaml
```diff
@@ -383,4 +383,4 @@ logging:
 - type: file
   level: info
   params:
-    path: /var/log/dendrite
+    path: ./logs
```
MadLittleMods, line 386: Ran this locally a few time and it creates the `./logs` directory just fine.

.gitignore
```diff
@@ -19,6 +19,7 @@
 /_test
```
MadLittleMods, line 19: It's unclear to me why the Sytest job is failing and what it is complaining about, https://buildkite.com/matrix-dot-org/dendrite/builds/4671#46eca488-3a58-4493-8b43-9866ec5e9760

Kegsay (changes requested):
Thanks for this! We use this path in numerous other places though, I think I need @neilalexander to help on which need to be changed (the ones for docker are probably fine staying as `/var/log`:
```
/cmd/generate-config/main.go:103:				"path": "/var/log/dendrite",
/build/docker/config/dendrite-config.yaml:344:    path: /var/log/dendrite
```

.gitignore
Kegsay: The key change test is flakey annoyingly.

neilalexander:
... which makes me realise that we should probably update the Docker Compose sample files to map the media and log directories somewhere outside of the container.

MadLittleMods:
@Kegsay It looks like both of those places are related to the Docker images. `generate-config` is only used in Docker and isn't documented elsewhere. And the other instance `/build/docker/config/dendrite-config.yaml` is self-explanatory as Docker.

Anything else stopping this from merging? Seems like we want to iterate a bit on the Docker setup but we can handle in another PR.

Kegsay:
Agreed.

PR #1802, opened 2021-03-23
Add more metrics

Kegsay:
- Add metric in `federationapi` for timings of `txn.processEvent` including what operations were required.
- Add metric in `keyserver` for number of times we call update device lists.

### DIFF
--- a/federationapi/routing/send.go
+++ b/federationapi/routing/send.go
@@ -34,27 +34,55 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+const (
+	// Event was passed to the roomserver
+	MetricsOutcomeOK = "ok"
+	// Event failed to be processed
+	MetricsOutcomeFail = "fail"
+	// Event failed auth checks
+	MetricsOutcomeRejected = "rejected"
+	// Terminated the transaction
+	MetricsOutcomeFatal = "fatal"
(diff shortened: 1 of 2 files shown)


federationapi/routing/send.go
```diff
@@ -34,27 +34,55 @@ import (
...
 	)
 	eduCountTotal = prometheus.NewCounter(
 		prometheus.CounterOpts{
 			Namespace: "dendrite",
 			Subsystem: "federationapi",
 			Name:      "recv_edus",
+			Help:      "Number of incoming EDUs from remote servers",
+		},
+	)
+	processEventSummary = prometheus.NewSummaryVec(
```
neilalexander, line 72: Would this be better as a Histogram rather than a Summary? I think we have similar statistics in the roomserver so you should be able to look at what they are.

neilalexander (approved):
Overall looks good!

federationapi/routing/send.go
Kegsay: I'm copying what state res does currently, which uses Summary. Looking at the docs, Histogram is useful when you know the ranges (how long things will take to complete) but Summary is good when you don't know. I don't think we know the timings yet, so I'll keep it as Summary for now.

PR #1803, opened 2021-03-23
Add Sentry support

Kegsay:
Untested but seems legit?


neilalexander (approved):
Looks interesting, would be nice to test it somewhere.

Kegsay:
Tested and does appear to work now. Waiting for CI before merging.

Issue #587, opened 2018-11-07
Implement filtering


babolivier:
This issue is tracking the progress of the implementation of filtering on the CS API. AFAIK @CromFr has already been working on implementing that on `/sync`. Will also need to be implemented on `/messages` when that endpoint itself gets implemented.

CromFr:
Yay filters ! 🎉 

My current filter-related PRs:

- #436 implements the use of filter objects when storing / retrieving them in the database
- #438 and #513 implements the filtering logic inside the event query functions (but uses default filters inside IncrementalSync/CompleteSync so the /sync behaviour is unchanged), respectively for state and account data events
- #388 (closed) was my first attempt at implementing the filtering logic, all in one huge and messy PR.

I should have in my local repo some branches for other event types, and the final IncrementalSync/CompleteSync implementation, but requires some additional work to be sane & clean.

Kegsay:
Sytests:
```
    × Can pass a JSON filter as a query parameter
    × Can request federation format via the filter
    × A filtered timeline reaches its limit
    × Check creating invalid filters returns 4xx
```

Kegsay:
@neilalexander did you do this in https://github.com/matrix-org/dendrite/pull/1721 ?

neilalexander:
We have filtering in `/sync` but I don't think we have it in `/messages` yet.

spaetz:
So, was that implemented by https://github.com/matrix-org/dendrite/pull/2347 and can be closed then?

kegsay:
Still failing.

Azanul:
@kegsay The `TODO` is [here](https://github.com/matrix-org/dendrite/blob/2e1fe589375b650f9b2d9a09e1fcffb3ab6fe5b6/syncapi/routing/messages.go#L188)

By your comment on the closed PR, I reckon that it needs to be implemented [here](https://github.com/matrix-org/dendrite/blob/2e1fe589375b650f9b2d9a09e1fcffb3ab6fe5b6/syncapi/storage/shared/storage_sync.go#L176)

Please clarify

Azanul:
@kegsay I'm new to dendrite. Could you please help me with this?
To make the invalid filters test pass, do I need to make changes in `parseRoomEventFilter` or after parsing but before querying?

PR #1809, opened 2021-03-26
Implement filtering for /roomId/messages

hulthe:
Make [`GET /_matrix/client/r0/rooms/{roomId}/messages`](https://matrix.org/docs/spec/client_server/r0.6.1#get-matrix-client-r0-rooms-roomid-messages) accept a `filter` parameter in
accordance with the spec (hopefully).

Related issue: #587 

This is my first PR to a matrix-project, so there are a couple of thing I'm unsure about:

- [`Database.GetEventsInTopologicalRange`][topological] doesn't support filters, so if the client specifies a
  topological `from`-token filtering wont work. Not sure if this is intended.

- `limit` can be set both as a query parameter, and as a part of RoomEventFilter.
  The spec doesn't seem to specify which takes precedence if both are specified.
  For now I've assumed that the query parameter takes precedence.

- Curiously, the default `limit` value in [`gomatrixserverlib.RoomEventFilter`][REF] is 20. I'm not sure
  if this is something I should adhere to, or event if it conforms to the spec, since the default for
  the query parameter is 10.

- Unlike [`/sync`][sync], the spec doesn't seem to allow for the filter payload to be a `filter_id`
  instead of a JSON RoomEventFilter. Is this intended, or a spec oversight?

[topological]: https://github.com/matrix-org/dendrite/blob/6e44450cc9b541e5e79b7e601462fd3d3cf35fe5/syncapi/storage/interface.go#L110 "GetEventsInTopologicalRange"
[REF]: https://github.com/matrix-org/gomatrixserverlib/blob/18abed1923fd09d5302090c8e91fab384b6cefb9/filter.go#L141 "RoomEventFilter"
[sync]: https://matrix.org/docs/spec/client_server/r0.6.1#get-matrix-client-r0-sync "/sync spec"

### SyTest
Sadly, no new tests are passing (there doesn't seem to be many tests which handle filtering /messages),
although a couple of tests seem to be closer to passing than before:
- `Ephemeral messages received from clients are correctly expired`
- `Ephemeral messages received from servers are correctly expired`

These now fail for other reasons than filtering not working :)

### Pull Request Checklist

Signed-off-by: `Joakim Hulthe <joakim@hulthe.net>`


hulthe:
Not sure why SyTest is failing in the CI. It seems to work on my machine :confused:

syncapi/routing/messages.go
```diff
@@ -234,19 +249,17 @@ func (r *messagesReq) retrieveEvents() (
...
 	if r.fromStream != nil {
 		toStream := r.to.StreamToken()
 		streamEvents, err = r.db.GetEventsInStreamingRange(
-			r.ctx, r.fromStream, &toStream, r.roomID, &eventFilter, r.backwardOrdering,
+			r.ctx, r.fromStream, &toStream, r.roomID, &r.filter, r.backwardOrdering,
 		)
 	} else {
 		streamEvents, err = r.db.GetEventsInTopologicalRange(
-			r.ctx, r.from, r.to, r.roomID, r.limit, r.backwardOrdering,
+			r.ctx, r.from, r.to, r.roomID, r.filter.Limit, r.backwardOrdering,
```
Kegsay, line 262: We need the filter object passed through to `GetEventsInTopologicalRange` and for it to be implemented there unfortunately!

Kegsay (commented):
We need to implement the topological filtering bit.

> limit can be set both as a query parameter, and as a part of RoomEventFilter.
The spec doesn't seem to specify which takes precedence if both are specified.
For now I've assumed that the query parameter takes precedence.

Your assumption is the same as mine: `limit` takses precedence.

I would go for a 20 limit, not 10. 

> Unlike /sync, the spec doesn't seem to allow for the filter payload to be a filter_id
instead of a JSON RoomEventFilter. Is this intended, or a spec oversight?

Almost certainly a spec oversight. I didn't even know `?filter=` was A Thing on `/messages`!

kegsay:
Stale. Please re-open when this has been brough up-to-date and the outstanding issues resolved. Thanks!

Issue #599, opened 2019-01-07
Implement OpenID module


babolivier:
- [ ] Implement [POST /_matrix/client/r0/user/{userId}/openid/request_token](https://matrix.org/docs/spec/client_server/r0.4.0.html#post-matrix-client-r0-user-userid-openid-request-token)
- [ ] Implement [GET /_matrix/federation/v1/openid/userinfo](https://matrix.org/docs/spec/server_server/unstable.html#get-matrix-federation-v1-openid-userinfo)

Kegsay:
Sytests:
```
    × Can generate a openid access_token that can be exchanged for information about a user
    × Invalid openid access tokens are rejected
    × Requests to userinfo without access tokens are rejected
```

jakehemmerle:
Starting work on this!

BruceMacD:
I'll review Jake's previous work here and finish this off.

PR #1812, opened 2021-03-29
Implement OpenID module (#599)

BruceMacD:
#599 
This change implements:
POST /_matrix/client/r0/user/{userId}/openid/request_token
```
curl --location --request POST 'http://{dendrite_base_url}/_matrix/client/r0/user/{user_id}/openid/request_token' \
--header 'Authorization: Bearer {token}' \
--header 'Content-Type: application/json' \
--data-raw '{}'
```
GET /_matrix/federation/v1/openid/userinfo
```
curl --location --request GET 'http://{dendrite_base_url}/_matrix/federation/v1/openid/userinfo?access_token={token}'
```

- Unrelated: change Riot references to Element in client API routing

Signed-off-by: Bruce MacDonald <contact@bruce-macdonald.com>


clientapi/routing/routing.go
```diff
@@ -469,7 +469,7 @@ func Setup(
 		}),
 	).Methods(http.MethodPost, http.MethodOptions)
 
-	// Stub endpoints required by Riot
+	// Stub endpoints required by Element
```
BruceMacD, line 472: This change isn't related to this issue, but it seemed like some nice house keeping. Let me know if I should pull it out.

federationapi/routing/routing.go
```diff
@@ -460,4 +460,10 @@ func Setup(
 			return QueryDeviceKeys(httpReq, request, keyAPI, cfg.Matrix.ServerName)
 		},
 	)).Methods(http.MethodPost)
+
+	v1fedmux.Handle("/openid/userinfo",
+		httputil.MakeExternalAPI("federation_openid_userinfo", func(req *http.Request) util.JSONResponse {
```
BruceMacD, line 465: Is an external API correct here or should I be checking client authorization?

clientapi/routing/routing.go
Kegsay: Nope, it's fine to leave this in as it's pretty obvious. Thanks for doing it!

federationapi/routing/routing.go
Kegsay: Nope, external API sounds right as it's an unauthenticated endpoint.

clientapi/routing/routing.go
```diff
@@ -685,6 +685,19 @@ func Setup(
 		}),
 	).Methods(http.MethodGet)
 
+	r0mux.Handle("/user/{userId}/openid/request_token",
```
Kegsay, line 688:
```suggestion
	r0mux.Handle("/user/{userID}/openid/request_token",
```
Tiny nit: We tend to use Go naming conventions for the path params.

clientapi/routing/routing.go
```diff
@@ -685,6 +685,19 @@ func Setup(
...
+		httputil.MakeAuthAPI("openid_request_token", userAPI, func(req *http.Request, device *userapi.Device) util.JSONResponse {
+			if r := rateLimits.rateLimit(req); r != nil {
+				return *r
+			}
+			vars, err := httputil.URLDecodeMapValues(mux.Vars(req))
+			if err != nil {
+				return util.ErrorResponse(err)
+			}
+			return CreateOpenIDToken(req, userAPI, device, vars["userId"], cfg)
```
Kegsay, line 697:
```suggestion
			return CreateOpenIDToken(req, userAPI, device, vars["userID"], cfg)
```

federationapi/routing/openid.go
```diff
@@ -0,0 +1,65 @@
...
+
+	req := userapi.QueryOpenIDTokenRequest{
+		Token: token,
+	}
+
+	var openIDTokenAttrResponse userapi.QueryOpenIDTokenResponse
+	err := userAPI.QueryOpenIDToken(httpReq.Context(), &req, &openIDTokenAttrResponse)
+	if err != nil {
+		util.GetLogger(httpReq.Context()).WithError(err).Error("userAPI.QueryOpenIDToken failed")
+	}
```
Kegsay, line 51: Unclear if this fallthrough is intentional or not. Comment if it is intentional please. The code will probably still work as-is due to `openIDTokenAttrResponse.Sub == ""`.

userapi/api/api.go
```diff
@@ -226,6 +228,27 @@ type PerformAccountDeactivationResponse struct {
...
+
+// QueryOpenIDTokenRequest is the request for QueryOpenIDToken
+type QueryOpenIDTokenRequest struct {
+	Token string
+}
+
+// QueryOpenIDTokenResponse is the response for QueryOpenIDToken
+type QueryOpenIDTokenResponse struct {
+	Sub       string // The Matrix User ID that generated the token
+	ExpiresTS int64
```
Kegsay, line 249: When using ints to describe timestamps, please put units in the field name e.g `ExpiresAtMS` would probably do the trick. `Expires` on its own is ambiguous if its a timestamp or absolute duration, but `ExpiresAt` indicates a timestamp reasonably well, so we don't have to do things like `ExpiresTSMS` :D

userapi/internal/api.go
```diff
@@ -414,3 +417,42 @@ func (a *UserInternalAPI) PerformAccountDeactivation(ctx context.Context, req *a
 	res.AccountDeactivated = err == nil
 	return err
 }
+
+// PerformOpenIDTokenCreation creates a new token that a relying party uses to authenticate a user
+func (a *UserInternalAPI) PerformOpenIDTokenCreation(ctx context.Context, req *api.PerformOpenIDTokenCreationRequest, res *api.PerformOpenIDTokenCreationResponse) error {
+	token := util.RandomString(24)
+	tokenHash := getTokenHash(token)
+	nowMS := time.Now().UnixNano() / int64(time.Millisecond)
+	expiresMS := nowMS + (3600 * 1000) // 60 minutes
```
Kegsay, line 426: Given we don't use `nowMS` anymore, why not just do `time.Now().Add(60 * time.Minute).UnixNano() / int64(time.Millisecond)` as `expiresMS`?

Kegsay: Also, the spec doesn't seem to mention an expiration time, so this should probably be configurable in the YAML.

userapi/internal/api.go
```diff
@@ -414,3 +417,42 @@ func (a *UserInternalAPI) PerformAccountDeactivation(ctx context.Context, req *a
...
+
+	res.Sub = openIDTokenAttrs.UserID
+	res.ExpiresTS = openIDTokenAttrs.ExpiresTS
+
+	return nil
+}
+
+// used for getting the format in which the token is stored to prevent plaintext storage of sensitive info
+func getTokenHash(token string) hash.Hash32 {
+	tokenHash := fnv.New32a()
```
Kegsay, line 455: Whilst I appreciate the thought, we shouldn't need to do this. There's several reasons here.

Firstly, if the database is compromised then an attacker can just query information directly. Given this token is only used to get the user ID, there's nothing of interest being protected here data wise.

Secondly, the point of hashing passwords is to prevent the password getting leaked in the event of a database compromise. Leaked passwords are valuable because people tend to reuse passwords between services, so there is intrinsic value to them beyond just access to the compromised service.

Thirdly, if you _are_ hashing passwords, you definitely don't want to use something like FNV which is designed for speed. Hashing once (even with something better suited like SHA256) is trivial to break. For example, Google search `000c285457fc971f862a79b786476c78812c8897063c6fa9c045f579a3b2d63f` and tell me what my password is :) This is why libraries and algorithms exist for securely storing passwords, the most popular being bcrypt and scrypt. These algorithms hash many times, include salt (extra data to make sure 2 people with the same password don't get the same end value), etc. These algorithms are designed to be computationally (and in scrypt's case, memory) expensive to avoid brute force attacks.

So in conclusion:
 - The token is not guarding anything sensitive.
 - The token itself is not valuable.

So please just insert the token directly into the database.

userapi/storage/accounts/sqlite3/storage.go
```diff
@@ -41,13 +41,15 @@ type Database struct {
...
 	threepids    threepidStatements
+	openIDTokens tokenStatements
 	serverName   gomatrixserverlib.ServerName
 	bcryptCost   int
 
 	accountsMu     sync.Mutex
 	profilesMu     sync.Mutex
 	accountDatasMu sync.Mutex
 	threepidsMu    sync.Mutex
+	openIDsMu      sync.Mutex
```
Kegsay, line 52: You shouldn't need a mutex for this. I _think_ you only need mutexes for tables which are modified on registration/login.

Kegsay (changes requested):
Lots of comments but overall this is a very strong PR, thank you for your work on this! The majority of my comments are minor, and I appreciate you doing all the sytests for this feature!

federationapi/routing/openid.go
BruceMacD: That was intentional in this case, I just didn't want to return an internal server error in the case that there was no token in the database.

userapi/api/api.go
BruceMacD: That's much nicer with MS. I changed all the expiry names to match this convention.

userapi/internal/api.go
BruceMacD: Definitely no expiry in the spec, making this part of the config makes sense. Added to the user API config.

userapi/internal/api.go
BruceMacD: All good point, I just switched to plaintext storage.

userapi/storage/accounts/sqlite3/storage.go
BruceMacD: Yeah, this only ever gets written to. The mutex was not needed.

BruceMacD:
XSS failure is due to an unchanged file.

Kegsay (approved):

Kegsay:
Excellent thank you! Will merge once CI runs again.

PR #1831, opened 2021-04-13
Remove the Content-Length requirement on upload request

### LINKED ISSUE #1825: Allow media uploads to specify no Content-Length header
# Problem

When uploading a [video] file through the `_matrix/client/media/v1/upload` endpoint the server returns is failing because the dendrite server returns a `411` error: `{"errcode":"M_UNKNOWN","error":"HTTP Content-Length request header must be greater than zero."}`.

# Motivation

Seeing that the dendrite server is in active development, instead of potentially patching the issue on JS / iOS / Android SDKs the required change can be done on the backend.

# Suggested Design

Reuse the work done in #1537 and retrieve remote file size without the Content-Length request header. In similar canon, we can relax the requirement for Content-Length request header and fall back to the temporary file size.

# Alternatives

Set the `Content-Length` request header on mobile client iOS SDK in [here](https://github.com/matrix-org/matrix-ios-sdk/blob/develop/MatrixSDK/MXRestClient.m#L3640). 

I haven't checked client SDKs on other platforms (web / android) but I assume it can be done so there in a similar fashion.

# Additional Information
### Screenshots

<details>
<summary>Charles Proxy recorded request</summary>

2576456625-image

</details>

### cURL reproducing the error

```bash
curl -H 'Host: dendrite.globekeeper.com' -H 'Accept: */*' -H 'User-Agent: com.apple.dt.xctest.tool/17501 (iPhone; iOS 14.4; Scale/3.00)' -H 'Accept-Language: en;q=1' -H 'Authorization: Bearer eS2jdBtvTY-OxbukomkZOd0XOxBjlEvde8pI7M0Ioss' --data-binary "" --compressed 'https://dendrite.globekeeper.com/_matrix/media/v1/upload'
```

### Dendrite server version

`0.3.11`

ferologics:
The requirement for Content-Length header in upload requests can be relaxed because we can fallback to temporary file size to make sure it does not exceed the configured max content size.

Resolves #1825

Signed-off-by: `Frantisek Hetes f.hetes@gmail.com`

### DIFF
--- a/mediaapi/routing/upload.go
+++ b/mediaapi/routing/upload.go
@@ -221,12 +221,6 @@ func (r *uploadRequest) doUpload(
 
 // Validate validates the uploadRequest fields
 func (r *uploadRequest) Validate(maxFileSizeBytes config.FileSizeBytes) *util.JSONResponse {
-	if r.MediaMetadata.FileSizeBytes < 1 {
-		return &util.JSONResponse{
-			Code: http.StatusLengthRequired,
-			JSON: jsonerror.Unknown("HTTP Content-Length request header must be greater than zero."),
-		}
-	}
 	if maxFileSizeBytes > 0 && r.MediaMetadata.FileSizeBytes > types.FileSizeBytes(maxFileSizeBytes) {
 		return &util.JSONResponse{
 			Code: http.StatusRequestEntityTooLarge,
--- a/mediaapi/routing/upload.go
+++ b/mediaapi/routing/upload.go
@@ -158,8 +158,8 @@ func (r *uploadRequest) doUpload(
 		}
 	}
 
-	// If Content-Length header is not set, check if temp file size exceeds max file size configuration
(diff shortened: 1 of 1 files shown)


mediaapi/routing/upload.go
```diff
@@ -158,6 +158,12 @@ func (r *uploadRequest) doUpload(
 		}
 	}
 
+	// If Content-Length header is not set, check if temp file size exceeds max file size configuration
+	if r.MediaMetadata.FileSizeBytes < 1 && bytesWritten > types.FileSizeBytes(*cfg.MaxFileSizeBytes) {
```
Kegsay, line 162: Please universally do this check i.e:
```go
if bytesWritten > types.FileSizeBytes(*cfg.MaxFileSizeBytes) {
```
as we will want to assert this regardless of what the client puts as the Content-Length.

Kegsay (approved):
Otherwise LGTM! Thanks.

PR #1834, opened 2021-04-14
Remove the 'Content-Type' request header requirement

### LINKED ISSUE #1833: Allow media uploads to specify no Content-Type header
# Problem

When uploading a [video] file to the `_matrix/client/media/v1/upload` endpoint via iOS Matrix client the server returns a `411` error: `{"errcode":"M_UNKNOWN","error":"HTTP Content-Type request header must be set."}`.

# Motivation

The requirement seems arbitrary at the moment, and it is not used for anything instrumental for the upload request to succeed.

# Suggested Design

Remove the restriction on `Content-Length` header in [here](https://github.com/matrix-org/dendrite/blob/master/mediaapi/routing/upload.go#L241).

# Alternatives

Prevent sending the initial empty upload request on iOS SDK.

# Additional Information
### Screenshots

<details>
<summary>Charles Proxy recorded request</summary>

image

</details>

### Dendrite server version

`0.3.11`

ferologics:
### Pull Request Checklist

The requirement for `Content-Type` header in upload requests does not seem to be necessary for the request so succeed and it's removal allows the iOS Matrix client SDK to upload video content.

Resolves #1833 

Signed-off-by: `Frantisek Hetes f.hetes@gmail.com`

### DIFF
--- a/mediaapi/routing/upload.go
+++ b/mediaapi/routing/upload.go
@@ -233,13 +233,6 @@ func (r *uploadRequest) Validate(maxFileSizeBytes config.FileSizeBytes) *util.JS
 			JSON: jsonerror.Unknown(fmt.Sprintf("HTTP Content-Length is greater than the maximum allowed upload size (%v).", maxFileSizeBytes)),
 		}
 	}
-	// TODO: Check if the Content-Type is a valid type?
-	if r.MediaMetadata.ContentType == "" {
-		return &util.JSONResponse{
-			Code: http.StatusBadRequest,
-			JSON: jsonerror.Unknown("HTTP Content-Type request header must be set."),
-		}
-	}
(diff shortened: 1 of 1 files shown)


neilalexander (approved):
Overall this is OK, but I am wondering if there was a reason why this was there in the first place.

ferologics:
The root cause for this missing header is described in [another issue I opened for iOS SDK client](https://github.com/matrix-org/matrix-ios-sdk/issues/1076). As far as reasonable defaults go, I think it makes total sense to require the header, but I wasn't able to find anything in the specification that would explicitly require it.

Kegsay:
The content type is useful to infer when to do thumbnails - e.g on `image/*` only. Clients can always lie about that though so we do need to do fallbacks/handle if they are wrong, so it's reasonable to omit this information I think.

Kegsay (approved):

PR #1837, opened 2021-04-19
Implement registering with an email support

### LINKED ISSUE #1298: Implement registering with an email support
Spec: https://matrix.org/docs/spec/client_server/r0.6.1#email-based-identity-homeserver
Sytests:
```
    × Can register using an email address
```

PiotrKozimor:
Support for registration via email sent by **identity server**. Resolves #1298, but there is a catch.
Test case `Can register using an email address` validates registration when dendrite itself sends verification email. In this PR dendrite sends `/_matrix/identity/v2/validate/email/requestToken` request to identity server (which will trigger sending email). New test case must be implemented in sytest. Proposed name: `Can register using an email address via identity server`

Allowable authentication flows must be passed via config file. This PR setups following convention (at the same time putting step towards removing `Derived`):
```yaml
client_api:
  registration:
    flows:
      - stages:
        - m.login.email.identity
  login:
    flows:
      - stages:
        - m.login.password
``` 

What's left to do?
* [x] Authenticate requests to identity server - I leave it **with no authentication** using V1 Identity API
* [x] Verify identity server certificates - there is configuration to add custom CA to system certificate pool.
* [x] Implement test case in sytest - [PR](https://github.com/matrix-org/sytest/pull/1034)

### Pull Request Checklist

### DIFF
--- a/clientapi/routing/register.go
+++ b/clientapi/routing/register.go
@@ -823,7 +823,7 @@ func checkAndCompleteFlow(
 	cfg *config.ClientAPI,
 	userAPI userapi.UserInternalAPI,
 ) util.JSONResponse {
-	if checkFlowCompleted(flow, cfg.Derived.Registration.Flows) {
+	if checkFlowCompleted(flow, cfg.Registration.Flows) {
 		// This flow was completed, registration can continue
 		return completeRegistration(
 			req.Context(), userAPI, r.Username, r.Password, "", req.RemoteAddr, req.UserAgent(),
@@ -836,7 +836,7 @@ func checkAndCompleteFlow(
 	return util.JSONResponse{
 		Code: http.StatusUnauthorized,
 		JSON: newUserInteractiveResponse(sessionID,
-			cfg.Derived.Registration.Flows, cfg.Derived.Registration.Params),
+			cfg.Registration.Flows, cfg.Registration.Params),
 	}
 }
 
--- a/clientapi/routing/register.go
+++ b/clientapi/routing/register.go
@@ -356,7 +356,7 @@ func validateEmailIdentity(
 	cred *threepidCreds,
 ) *util.JSONResponse {
 	url := fmt.Sprintf(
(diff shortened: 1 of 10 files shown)


PiotrKozimor:
Regarding failing CodeQL checks - for both `Uncontrolled data used in network request` issue is reported while in fact identity server URI is checked against list of trusted ones.

clientapi/routing/register.go
```diff
@@ -310,6 +313,80 @@ func validateRecaptcha(
...
+) *util.JSONResponse {
+	if err := isTrusted(cred.IDServer, cfg); err != nil {
+		return &util.JSONResponse{
+			Code: http.StatusForbidden,
+			JSON: jsonerror.NotTrusted(cred.IDServer),
+		}
+	}
+	util.GetLogger(ctx).Infof("conecting to identity server: %s", cred.IDServer)
+	url := fmt.Sprintf(
+		"https://%s/_matrix/identity/api/v1/3pid/getValidated3pid",
```
richvdh, line 329: dendrite isn't my project, but:

this endpoint is deprecated by [MSC2713](https://github.com/matrix-org/matrix-doc/pull/2713) and it's likely that sydent will soon drop support for it.

In general, homeservers should not be delegating responsibility for email address validation to identity servers, since it allows a compromised ID server to be used to take over homeserver accounts.

PiotrKozimor: Thank you for comment. I think I have misunderstood concept of ID server. It makes sense that homeserver sends email on its own. So ID server would be user after registration to publish association, so that other may find user by email, right? I suppose also that auth stage for login should be also done without ID server. I am right?

richvdh: yes, that's right.

PiotrKozimor:
Closing this as long as we want HS to send validation emails - see [this issue](https://github.com/matrix-org/synapse/issues/1710).

PR #1843, opened 2021-04-23
Use LimitReader to prevent DoS risk

### LINKED ISSUE #1832: Apply limited writer to uploads
When reviewing https://github.com/matrix-org/dendrite/pull/1831 it became apparent that Dendrite will blindly dump files to disk _then_ check the size of the file. This is a DoS risk as it allows clients to upload arbitrarily sized files which will then be cleaned up later. We should instead use a `LimitedReader` when doing uploads, returning an error if the limit is reached.

S7evinK:
Should fix #1832 

### Pull Request Checklist


S7evinK:
Needs further investigation:
```
// Because Copy is defined to read from src until EOF, it does
// not treat an EOF from Read as an error to be reported.
```

S7evinK:
In the current state, files with **exactly** the same size as the `MaxFileSizeBytes` will be handled as if they exceeded the limit.

Kegsay:
Fab thanks for this. Please just modify `LimitReader` to do `MaxConfigSizeBytes + 1` to avoid the exact number incorrectly failing. This should simplify checks then as you can keep it as `if actual_size > max_size`

Kegsay (approved):

PR #1855, opened 2021-05-03
Split the select+update query for txn_id counter

### LINKED ISSUE #1852: txn_id is not incremented
### Background information

- **Dendrite version or git SHA**: d6e9b7b307ff0d7541046ec33890d49239c7a6ca
- **Running in Docker?**: no
- **`go version`**: go1.15.2 linux/amd64

### Description

Every new transaction for an appservice gets txnID=1, and thus gets ignored by an appservice (`matrix-appservice`-based, in my case).

A deeper insight shows that txn_id does not even get incremented in the `appservice_counters` table, despite it's supposed to.

### Steps to reproduce

- Configure an appservice.
- Run Dendrite and the appservice.
- Make some events that are sent to an appservice.
- Capture `/transactions/{txnId}` requests and find out they all have txnId=1 even if the appservice correctly responds with `200 OK`.

<!--

``` (three backticks, on a line on their own), so that they are formatted legibly.

-->

bodqhrohro:
The update part wasn't executed actually for SQLite, so it is moved to
a separate statement.

Fixes #1852.

### Pull Request Checklist

Signed-off-by: Bohdan Horbeshko <bodqhrohro@gmail.com>

### DIFF
--- a/appservice/storage/sqlite3/txn_id_counter_table.go
+++ b/appservice/storage/sqlite3/txn_id_counter_table.go
@@ -32,14 +32,18 @@ INSERT OR IGNORE INTO appservice_counters (name, last_id) VALUES('txn_id', 1);
 `
 
 const selectTxnIDSQL = `
-  SELECT last_id FROM appservice_counters WHERE name='txn_id';
-  UPDATE appservice_counters SET last_id=last_id+1 WHERE name='txn_id';
+  SELECT last_id FROM appservice_counters WHERE name='txn_id'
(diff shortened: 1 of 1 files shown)


S7evinK (approved):

Kegsay (approved):
Thanks for this! Ideally we should be using transactions for this but because this is SQLite we already serialise blocks using `s.writer.Do` so it's fine to leave the transaction out.

PR #1866, opened 2021-05-30
Add NATS JetStream support

S7evinK:
This adds experimental support for NATS JetStream by implementing the required interfaces for sarama.

Currently uses my experimental [NATS JetStream sarama implementation](https://github.com/S7evinK/saramajetstream), seems like there are still a few issues in there, as e.g. receipts/read markers don't seem to always come through.

On a side note, I was able to join #dendrite-dev and still receive messages in other rooms.

Kafka (with 3 nodes, 1 zookeeper) is using about ~4G of memory.. NATS on the other side (also 3 nodes, no manager needed) - 120M. (As reported by Prometheus) Kafka/NATS are both idle.

Some screenshots from NATS:
image

image

### Pull Request Checklist

* [ ] Pull request includes a [sign off](https://github.com/matrix-org/dendrite/blob/master/docs/CONTRIBUTING.md#sign-off)


Kegsay:
Fantastic work on this, thanks @S7evinK for all of your contributions thus far! I'll try to block out some time to look at this, but it'll take a while to migrate fully to NATS. We still need to sketch out and profile whether the swap would be net positive for large deployments. The main drivers for wanting to swap away from Kafka/Naffka are:
 - Unbounded log growth - yes Kafka has a retention period but it's become clearer as the project has developed that we primarily want tiny retention periods, particularly for embedded use cases.
 - Inability to send ephemeral events (generic pubsub to allow us to request actions on other components and register callbacks). Being able to do this will open up more ways to prevent race conditions between components.

S7evinK:
Even after some more testing and trying to fix things, still the same issue with receipts not coming through. Looks like, at least that's what i guess, that only `m.fully_read` make their way to Element.

After a few days in #dendrite:
```
# nats stream report
Obtaining Stream stats

+---------------------------------------------------------------------------------------------------------------+
|                                                 Stream Report                                                 |
+---------------------------------+---------+-----------+----------+---------+------+---------+-----------------+
| Stream                          | Storage | Consumers | Messages | Bytes   | Lost | Deleted | Replicas        |
+---------------------------------+---------+-----------+----------+---------+------+---------+-----------------+
| DendriteOutputSendToDeviceEvent | File    | 2         | 0        | 0 B     | 0    | 0       | nats-cluster-0* |
| DendriteOutputTypingEvent       | File    | 2         | 0        | 0 B     | 0    | 0       | nats-cluster-0* |
| DendriteOutputClientData        | File    | 1         | 57       | 9.7 KiB | 0    | 0       | nats-cluster-0* |
| DendriteOutputReceiptEvent      | File    | 2         | 535      | 164 KiB | 0    | 0       | nats-cluster-2* |
| DendriteOutputKeyChangeEvent    | File    | 2         | 798      | 711 KiB | 0    | 0       | nats-cluster-2* |
| DendriteOutputRoomEvent         | File    | 2         | 1,537    | 4.0 MiB | 0    | 0       | nats-cluster-1* |
+---------------------------------+---------+-----------+----------+---------+------+---------+-----------------+
```

kegsay:
@neilalexander has championed this and I think has a different (?) branch with support for in-process NATS? If so, I think this PR can be closed.

That being said, thank you very very very much @S7evinK for trying this out before myself and N! We're grateful to see some work on this from the community as it's a major change which will be needed before we exit beta.

neilalexander:
Yes, I've been working on this in the `neilalexander/nats` branch, which is just a continuation of this PR. We're ultimately blocked at the moment on nats-io/nats-server#2360, nats-io/nats.go#774 and nats-io/nats-server#2363 before we can progress though.

neilalexander:
I've merged my changes back into this PR.

S7evinK:
Everything seems to be OK so far, besides the `DendriteOutputKeyChangeEvent`s, those seem to pile up.
Receipts also work much better than before :)

```
# nats stream report
Obtaining Stream stats

+---------------------------------------------------------------------------------------------------------------+
|                                                 Stream Report                                                 |
+---------------------------------+---------+-----------+----------+---------+------+---------+-----------------+
| Stream                          | Storage | Consumers | Messages | Bytes   | Lost | Deleted | Replicas        |
+---------------------------------+---------+-----------+----------+---------+------+---------+-----------------+
| DendriteOutputClientData        | File    | 1         | 0        | 0 B     | 0    | 0       | nats-cluster-2* |
| DendriteOutputPresenceData      | File    | 0         | 0        | 0 B     | 0    | 0       | nats-cluster-2* |
| DendriteOutputPresenceData      | File    | 1         | 0        | 0 B     | 0    | 0       | nats-cluster-1* |
| DendriteOutputReceiptEvent      | File    | 2         | 0        | 0 B     | 0    | 0       | nats-cluster-2* |
| DendriteOutputRoomEvent         | File    | 2         | 0        | 0 B     | 0    | 0       | nats-cluster-0* |
| DendriteOutputSendToDeviceEvent | File    | 2         | 0        | 0 B     | 0    | 0       | nats-cluster-2* |
| DendriteOutputKeyChangeEvent    | File    | 2         | 766      | 681 KiB | 0    | 0       | nats-cluster-1* |
+---------------------------------+---------+-----------+----------+---------+------+---------+-----------------+
```

roomserver/internal/input/input.go
```diff
@@ -19,19 +19,19 @@ import (
 	"context"
 	"encoding/json"
 	"sync"
-	"time"
 
-	"github.com/Shopify/sarama"
+	"github.com/Arceliar/phony"
```
kegsay, line 23: I would much rather this was done in a different PR :(

roomserver/internal/input/input.go
```diff
@@ -42,105 +42,154 @@ var keyContentFields = map[string]string{
...
-type inputWorker struct {
-	r       *Inputer
-	running atomic.Bool
-	input   *fifoQueue
+// onMessage is called when a new event arrives in the roomserver input stream.
+func (r *Inputer) Start() error {
+	_, err := r.JetStream.Subscribe(
+		r.InputRoomEventTopic,
+		func(msg *nats.Msg) {
+			_ = msg.InProgress()
```
kegsay, line 58: Needs more documentation. Why do we do this?

roomserver/internal/input/input.go
```diff
@@ -42,105 +42,154 @@ var keyContentFields = map[string]string{
...
+			roomID := msg.Header.Get("room_id")
+			defer roomserverInputBackpressure.With(prometheus.Labels{"room_id": roomID}).Dec()
+			var inputRoomEvent api.InputRoomEvent
+			if err := json.Unmarshal(msg.Data, &inputRoomEvent); err != nil {
+				_ = msg.Term()
+				return
+			}
+			inbox, _ := r.workers.LoadOrStore(roomID, &phony.Inbox{})
+			inbox.(*phony.Inbox).Act(nil, func() {
+				_ = msg.InProgress()
```
kegsay, line 68: Why do we call `InProgress` twice for the same message?

roomserver/internal/input/input.go
```diff
@@ -42,105 +42,154 @@ var keyContentFields = map[string]string{
...
+func (r *Inputer) Start() error {
+	_, err := r.JetStream.Subscribe(
+		r.InputRoomEventTopic,
+		func(msg *nats.Msg) {
+			_ = msg.InProgress()
+			roomID := msg.Header.Get("room_id")
+			defer roomserverInputBackpressure.With(prometheus.Labels{"room_id": roomID}).Dec()
+			var inputRoomEvent api.InputRoomEvent
+			if err := json.Unmarshal(msg.Data, &inputRoomEvent); err != nil {
+				_ = msg.Term()
```
kegsay, line 63: Why do we do this here but in other cases we just `Ack()` or `Nack()`?

roomserver/internal/input/input.go
```diff
@@ -42,105 +42,154 @@ var keyContentFields = map[string]string{
...
+					sentry.CaptureException(err)
+					_ = msg.Respond([]byte(err.Error()))
+				} else {
+					hooks.Run(hooks.KindNewEventPersisted, inputRoomEvent.Event)
+				}
+				_ = msg.Ack()
+			})
+		},
+		nats.ManualAck(),
+		nats.MaxDeliver(0),
```
kegsay, line 79: Needs more documentation.

roomserver/internal/input/input.go
```diff
@@ -42,105 +42,154 @@ var keyContentFields = map[string]string{
...
+				return
+			case err := <-responses:
+				if err != nil {
+					response.ErrMsg = err.Error()
+					return
+				}
 			}
-			task.wg.Done()
-		case <-time.After(time.Second * 5):
-			return
```
kegsay, line 87: We don't seem to have a timeout anymore?

roomserver/internal/input/input.go
```diff
@@ -42,105 +42,154 @@ var keyContentFields = map[string]string{
...
-			task, ok := w.input.pop()
-			if !ok {
-				continue
+// InputRoomEvents implements api.RoomserverInternalAPI
+func (r *Inputer) InputRoomEvents(
+	ctx context.Context,
+	request *api.InputRoomEventsRequest,
+	response *api.InputRoomEventsResponse,
+) {
+	if request.Asynchronous {
```
kegsay, line 90: So if I understand correctly, `InputRoomEvents` is called by other components who want to input room events into a room server. If this flag is set then it does this by publishing to jetstream in an async fashion, fine. However, if this is sync then we seem to process the event on the other component? `r.processRoomEvent(context.TODO(), &inputRoomEvent)` on `:116`? I don't get it.

kegsay (changes requested):
Two main complaints:
 - Phony changes are an unwelcome addition to this already enormous PR. 
 - Scattered `Ack()` calls throughout consumer code makes me extremely nervous that we will forget to ACK in places and I don't know what happens then? I guess it will get stuck on that event which is obviously not ideal.

setup/jetstream/nats.go
```diff
@@ -0,0 +1,93 @@
+package jetstream
+
+import (
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/Shopify/sarama"
```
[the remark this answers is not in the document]
neilalexander: Key change consumers still use `github.com/S7evinK/saramajetstream` because they expect a partition number as well as an offset. They will need to be rewritten for NATS nativeness.

roomserver/internal/input/input.go
neilalexander: We did talk a bit about this at the time in the internal room — it pretty much boils down to the fact that buffered channels can fill up and block, unbuffered channels block anyway and `phony` was about the easiest way to ensure that we can continue to queue things up in memory without creating unexpected backpressure. The changes are limited to `input.go` (instead of managing goroutines ourselves).

roomserver/internal/api.go
```diff
@@ -63,13 +64,17 @@ func NewRoomserverAPI(
...
+			InputRoomEventTopic:  inputRoomEventTopic,
 			OutputRoomEventTopic: outputRoomEventTopic,
-			Producer:             producer,
+			JetStream:            consumer,
 			ServerName:           cfg.Matrix.ServerName,
 			ACLs:                 serverACLs,
 		},
 		// perform-er structs get initialised when we have a federation sender to use
 	}
+	if err := a.Inputer.Start(); err != nil {
```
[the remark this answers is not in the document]
neilalexander: Right — that's because in Naffka/Kafka world, we didn't have an "input stream" for the roomserver, we just sent things to the room server by calling the input API and blocking. In the NATS branch, the roomserver has a durable input queue, so the input API calls don't block.

roomserver/internal/input/input.go
neilalexander: Thinking about this, given that `MaxDeliver` is set to 0 for the consumer, we probably don't need these at all anymore.

roomserver/internal/input/input.go
neilalexander: This was because of the fact that the inbox might not process the task until later, so we reset the delivery timer when we start *working* on the event as well. As above though in the other comment, I don't think we actually need these at all, so I will remove them.

roomserver/internal/input/input.go
neilalexander: So this is something that I've been wanting to refactor for a while so that we're using request-reply semantics within NATS itself, but I figured that was probably deserving of it's own PR later. What happens if we're performing a synchronous request, we queue up the request straight onto the room inboxes directly rather than going via the NATS input stream and then wait for the responses. It's the exact same process that we follow when we consume a message from the input stream, just that we don't "block" there.

roomserver/internal/input/input.go
neilalexander: This was to do with trying to queue things up using request-reply, rather than putting things straight onto the inboxes themselves for synchronous requests. I never finished it and decided it would be better to do that in a separate PR later, but the `Term` here is pretty much equivalent here to an `Ack`.

dendrite-config.yaml
```diff
@@ -68,34 +68,27 @@ global:
...
+    in_memory: false
 
-    # The max size a Kafka message is allowed to use.
-    # You only need to change this value, if you encounter issues with too large messages.
-    # Must be less than/equal to "max.message.bytes" configured in Kafka.
-    # Defaults to 8388608 bytes.
-    # max_message_bytes: 8388608
+    # Persistent directory to store JetStream streams in. This directory 
+    # should be preserved across Dendrite restarts.
+    storage_path: ./
```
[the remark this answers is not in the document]
neilalexander: It's where the JetStream streams get persisted to disk, yes.

dendrite-config.yaml
```diff
@@ -68,34 +68,27 @@ global:
   # to other servers and the federation API will not be exposed.
   disable_federation: false
 
-  # Configuration for Kafka/Naffka.
-  kafka:
-    # List of Kafka broker addresses to connect to. This is not needed if using
-    # Naffka in monolith mode.
+  # Configuration for NATS JetStream
+  jetstream:
```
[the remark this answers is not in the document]
neilalexander: Good idea, I have done this.

roomserver/roomserver_test.go
```diff
@@ -1,407 +0,0 @@
-package roomserver
```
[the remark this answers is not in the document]
neilalexander: The tests in this file are all based on being able to satisfy the `sarama` producer interface with a dummy implementation, which we cannot do with NATS. We need to rewrite these tests from the ground up.

roomserver/internal/api.go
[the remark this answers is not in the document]
kegsay: Gotcha, thanks!

roomserver/roomserver_test.go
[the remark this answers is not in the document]
kegsay: SGTM

setup/jetstream/nats.go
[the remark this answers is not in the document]
kegsay: Okay we should replace this ASAP then - can we do this next?

kegsay (approved):
LGTM - Let's do it!

r3k2:
Awesome! how do we change our current server/setup to test this?

neilalexander:
Take a look at the `jetstream` section in the [version 2 sample config file](https://github.com/matrix-org/dendrite/blob/master/dendrite-config.yaml) and make sure you have a `store_dir` that is read-writable by the Dendrite process, and you *should* be able to just upgrade to it with any luck. Let me know if you have any problems.

PR #1867, opened 2021-06-02
Validate configured server name

neilalexander:
Apparently it's possible to configure Dendrite with nonsense server names which will break clients and/or federation. This PR enforces IDNA validation if there's a domain name present in the `server_name` configuration option, or sufficient IP literal (and optional port) validation if not.


Kegsay:
Please add some tests. Also, doesn't GMSL do something similar when checking event signatures?

neilalexander:
Closing this for now, since we need to spend some time working out the right thing to do here. It broke all sorts the last time I tested this on d.m.org.

PR #1869, opened 2021-06-07
ThreePid component

PiotrKozimor:
First step to implement #1298. 

`ThreePid` component is introduced which would handle validate session. It will have interface as defined in [api.go](https://github.com/globekeeper/dendrite/blob/piotr-kozimor/hs-sends-emails/threepid/api.go). 
Database implementation for storing sessions is defined by `storage` interface. 

### Pull Request Checklist

### DIFF
--- a/threepid/api.go
+++ b/threepid/api.go
@@ -1,0 +1,38 @@
+package threepid
+
+import "context"
+
+type API interface {
+	CreateSession(context.Context, *CreateSessionRequest, *Session) error
+	ValidateSession(context.Context, *ValidateSessionRequest, struct{}) error
+	GetThreePidForSession(context.Context, *SessionOwnership, *GetThreePidForSessionResponse) error
+	DeleteSession(context.Context, *SessionOwnership, struct{}) error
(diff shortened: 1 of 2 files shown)


Kegsay:
Please don't do this as a separate component. There's no need to make every feature a component. Components should be used when there is a real risk of high CPU/RAM consumption such that offloading that to another machine and/or sharding it is desirable. In this case, that is not a concern. There is a cost to adding components hence the desire to keep these as low as feasibly possible.

PiotrKozimor:
Ok, it makes sense. I guess I will merge the defined `API` interface with `UserInternalAPI`. Do you have any comments for interface that I have already defined?

kegsay:
Merging it with `UserInternalAPI` makes sense as it should scale with the number of users. It's hard to comment on the API when I can't see how it's being used unfortunately.

At a glance it seems plausible, just be careful not to introduce race conditions (`IsSessionValidated` followed by `ValidateSession` for example will race if a session is in the process of being validated when `IsSessionValidated` is called).

PiotrKozimor:
I am closing this for clarity, in favour of #1970

Issue #1467, opened 2020-10-02
Move to jsoniter


Kegsay:
We should move to use https://github.com/json-iterator/go because we spend a not insignificant amount of time doing JSON parsing. In terms of expected speedup:

 - std decode 35510 ns/op
 - jsoniter decode 5623 ns/op (~6x speedup)
 - std encode 2213 ns/op
 - jsoniter encode 837 ns/op (~3x speedup)

6543:
just an idear ... github.com/goccy/go-json is even faster sometimes, what do you think?

l0nax:
We have used goccy and jsoniter and from what I can tell goccy is the clear winner when it comes to performance.

PR #1870, opened 2021-06-10
Use "[]byte" instead of "json.RawMessage" if posible

6543:
one step towards #1467, can not remove it completely, since gomatrixserverlib struts use it too ...

### DIFF
--- a/clientapi/auth/user_interactive_test.go
+++ b/clientapi/auth/user_interactive_test.go
@@ -62,7 +62,7 @@ func TestUserInteractivePasswordLogin(t *testing.T) {
 		UserID:     fmt.Sprintf("@alice:%s", serverName),
 	}
 	// valid password requests
-	testCases := []json.RawMessage{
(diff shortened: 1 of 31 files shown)


neilalexander (commented):
I don't understand why you would want to do this?

`json.RawMessage` is a special type that the `json` library specifically uses to defer unmarshalling until a later time, which we do deliberately for a number of reasons. 

It is just a type alias of `[]byte` anyway, so you can still slice it if needed.

6543:
https://github.com/matrix-org/gomatrixserverlib/pull/260#issuecomment-858612394

kegsay:
Sorry, but we don't want to do this. We can still move to jsoniter without doing this AFAIK.

6543:
nothing against it ... sure - will come up with a huger diff wo use *renamed imports* to introduce **jsoniter** and one with **go-json** so we can compare compatibility & speed

PR #1872, opened 2021-06-11
Use NotFound instead of Forbidden for missing account data

adamgreig:
Currently Dendrite returns HTTP 404 but with an M_FORBIDDEN error type when a client requests account data of a non-existent type. Synapse [returns M_NOT_FOUND instead](https://github.com/matrix-org/synapse/blob/4b965c862dc66c0da5d3240add70e9b5f0aa720b/synapse/rest/client/v2_alpha/account_data.py#L62). The spec [doesn't mention behaviour when not found](https://matrix.org/docs/spec/client_server/r0.6.1#get-matrix-client-r0-user-userid-account-data-type).

### Pull Request Checklist

Signed-off-by: `Adam Greig <adam@adamgreig.com>`

### DIFF
--- a/clientapi/routing/account_data.go
+++ b/clientapi/routing/account_data.go
@@ -69,7 +69,7 @@ func GetAccountData(
 
 	return util.JSONResponse{
 		Code: http.StatusNotFound,
-		JSON: jsonerror.Forbidden("data not found"),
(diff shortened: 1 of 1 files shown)


S7evinK (approved):

neilalexander (approved):
Makes sense, thanks!

PR #1875, opened 2021-06-13
Set MaxFileSizeBytes <= 0 to "unlimited"

S7evinK:
Not sure if it's the right place, but this should allow an "unlimited" file size if it's set below 0.

From #dendrite:matrix.org
> In case it's helpful, the dendrite yaml was configured to allow unlimited max file size (since nginx can handle file size), by setting maxfilesize_bytes to zero.

### Pull Request Checklist

### DIFF
--- a/mediaapi/routing/upload.go
+++ b/mediaapi/routing/upload.go
@@ -147,6 +147,18 @@ func (r *uploadRequest) doUpload(
 	//   r.storeFileAndMetadata(ctx, tmpDir, ...)
 	// before you return from doUpload else we will leak a temp file. We could make this nicer with a `WithTransaction` style of
 	// nested function to guarantee either storage or cleanup.
+
+	// should not happen, but prevents any int overflows
+	if cfg.MaxFileSizeBytes != nil && *cfg.MaxFileSizeBytes+1 <= 0 {
+		r.Logger.WithFields(log.Fields{
+			"MaxFileSizeBytes": *cfg.MaxFileSizeBytes + 1,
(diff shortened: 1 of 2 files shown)


mediaapi/routing/upload.go
```diff
@@ -147,6 +147,18 @@ func (r *uploadRequest) doUpload(
 	//   r.storeFileAndMetadata(ctx, tmpDir, ...)
 	// before you return from doUpload else we will leak a temp file. We could make this nicer with a `WithTransaction` style of
 	// nested function to guarantee either storage or cleanup.
+
+	// should not happen, but prevents any int overflows
+	if cfg.MaxFileSizeBytes != nil && *cfg.MaxFileSizeBytes+1 <= 0 {
```
kegsay, line 152: The nil check is spurious given we will panic on `:162` if it is `nil`.

neilalexander (approved):

setup/config/config_mediaapi.go
```diff
@@ -57,6 +58,11 @@ func (c *MediaAPI) Verify(configErrs *ConfigErrors, isMonolith bool) {
 	checkNotEmpty(configErrs, "media_api.database.connection_string", string(c.Database.ConnectionString))
 
 	checkNotEmpty(configErrs, "media_api.base_path", string(c.BasePath))
+	// allow "unlimited" file size
+	if c.MaxFileSizeBytes != nil && *c.MaxFileSizeBytes <= 0 {
```
kegsay, line 62: It's not really unilmited though. I would just remove the `io.LimitReader` if `MaxFileSizeBytes <= 0`.

Issue #1813, opened 2021-03-29
LDAP


cyberb:
I am new to matrix and I would like to add dendrite to Syncloud store.
For this I need LDAP integration (login, user creation), which I can try adding if there is no working solution is available yet.

drequivalent:
YES PLEASE!

cyberb:
https://github.com/matrix-org/dendrite/pull/2986

cyberb:
So it was rejected in favour of the future OIDC (which is not available right now as I understand)
For those who are desperate like me to get LDAP support right now, feel free to use my branch which I will rebase from time to time.
https://github.com/cyberb/dendrite/commits/syncloud

cyberb:
it has this new config section:
```
 ldap:
    enabled: true
    uri: "ldap://localhost:389"
    base_dn: "dc=syncloud,dc=org"

    # admin bind modes uses a separate admin ldap account to perform user search
    admin_bind_enabled: true
    admin_bind_dn: "cn=admin,dc=syncloud,dc=org"
    admin_bind_password: "syncloud"
    search_base_dn: "ou=users,dc=syncloud,dc=org"
    search_filter: "(&(objectclass=inetOrgPerson)(cn={username}))"
    search_attribute: "cn"

    # direct user bind if admin bind is disabled
    user_bind_dn: "cn={username},ou=users,dc=syncloud,dc=org"

    # is user an admin or not
    admin_group_dn: "cn=syncloud,ou=groups,dc=syncloud,dc=org"
    admin_group_filter: "(memberUid={username})"
    admin_group_attribute: "memberUid"
```

Danieloni1:
Hey @cyberb,
Thanks for the contribution!
I am indeed desperate like you to have LDAP support.
Will it work for me out of the box right now?

cyberb:
It works for me in Syncloud project, use the fork and add the config, let me know here if it does not work I will help or fix the code.

cyberb:
Start with user_bind_dn mode

drequivalent:
> So it was rejected in favour of the future OIDC (which is not available right now as I understand)
![The_What ](https://user-images.githubusercontent.com/1463277/230247274-f78318d7-a95b-41f1-ba8d-ed4a403c3527.jpg)

LDAP is the established standard, you pretty much *have* to support it, in order to be usable in integrated environments.

Danieloni1:
@cyberb 
Thanks pal!
[...]

PR #1877, opened 2021-06-16
Ldap

phoenix-2021:
### Pull Request Checklist

Login in to dendrite with username and password from ldap.

Signed-off-by: `jiafeng zheng <jiafengzheng@workly.ai>`


DoM1niC:
#1813 will close

phoenix-2021:
> ### Pull Request Checklist
> Login in to dendrite with username and password from ldap.
> 
> Signed-off-by: `jiafeng zheng <jiafengzheng@workly.ai>`

phoenix-2021:
can someone review the code or guide me what to do next? I'd love to perfect this pr can be merged.

cmd/sytest/sytest.go
```diff
@@ -0,0 +1,114 @@
+package main
```
kegsay, line 1: This binary seems unrelated to this PR.

clientapi/auth/password.go
```diff
@@ -62,6 +70,101 @@ func (t *LoginTypePassword) Login(ctx context.Context, req interface{}) (*Login,
...
+		} else {
+			addr = "ldap://" + t.Config.LDAP.Host + ":" + t.Config.LDAP.Port
+		}
+
+		var conn *ldap.Conn
+		conn, err = ldap.DialURL(addr)
+		if err != nil {
+			return nil, &util.JSONResponse{
+				Code: http.StatusUnauthorized,
+				JSON: jsonerror.InvalidUsername(err.Error()),
```
kegsay, line 86: Unclear why this would be an invalid username. It's a badly configured LDAP address.

clientapi/auth/password.go
```diff
@@ -62,6 +70,101 @@ func (t *LoginTypePassword) Login(ctx context.Context, req interface{}) (*Login,
 			JSON: jsonerror.InvalidUsername(err.Error()),
 		}
 	}
+	if len(t.Config.LDAP.Host) > 0 {
+		addr := ""
+		if t.Config.LDAP.TLS {
+			addr = "ldaps://" + t.Config.LDAP.Host + ":" + t.Config.LDAP.Port
+		} else {
+			addr = "ldap://" + t.Config.LDAP.Host + ":" + t.Config.LDAP.Port
```
kegsay, line 78: It might be easier to just have `t.Config.LDAPAddr` which can then mux together TLS/host/port e.g `ldaps://localhost:8008`

clientapi/auth/password.go
```diff
@@ -62,6 +70,101 @@ func (t *LoginTypePassword) Login(ctx context.Context, req interface{}) (*Login,
...
+				JSON: jsonerror.InvalidUsername(err.Error()),
+			}
+		}
+		defer conn.Close()
+
+		e1 := conn.Bind(t.Config.LDAP.BindDN, t.Config.LDAP.BindPSWD)
+		if e1 != nil {
+			return nil, &util.JSONResponse{
+				Code: http.StatusUnauthorized,
+				JSON: jsonerror.InvalidUsername(err.Error()),
```
kegsay, line 95: Should be using `e1` not `err`

clientapi/auth/password.go
```diff
@@ -62,6 +70,101 @@ func (t *LoginTypePassword) Login(ctx context.Context, req interface{}) (*Login,
...
+			}
+		}
+		filter := fmt.Sprintf("(&%s(%s=%s))", t.Config.LDAP.Filter, "uid", localpart)
```
kegsay, line 98: `"uid"` can just be put into the `Sprintf`?

clientapi/auth/password.go
```diff
@@ -62,6 +70,101 @@ func (t *LoginTypePassword) Login(ctx context.Context, req interface{}) (*Login,
...
+		searchRequest := ldap.NewSearchRequest(t.Config.LDAP.BaseDN, ldap.ScopeWholeSubtree, ldap.NeverDerefAliases, 0, 0, false, filter, []string{"uid"}, nil)
```
kegsay, line 99: We probably want this to be customisable.

clientapi/auth/password.go
```diff
@@ -62,6 +70,101 @@ func (t *LoginTypePassword) Login(ctx context.Context, req interface{}) (*Login,
...
+		if len(sr.Entries) > 1 {
+			return nil, &util.JSONResponse{
+				Code: http.StatusUnauthorized,
+				JSON: jsonerror.BadJSON("'user' must be duplicated."),
+			}
+		}
+		if len(sr.Entries) == 0 {
+			_, err = t.GetAccountByPassword(ctx, localpart, r.Password)
+			if err != nil {
+				return nil, &util.JSONResponse{
```
kegsay, line 117: Rather than duplicating the code on `:168-176` it would be better to fork the LDAP code into a separate function and communicate success/failure/fatal failure that way.

kegsay (commented):
Aside from comments, this is a working opt-in LDAP solution. Dendrite will need an LDAP solution inevitably in the future. However, I don't know if we're at that point yet. There's numerous things to consider which this PR doesn't do yet:
 - Where in the LDAP directory to search: this PR searches everywhere the bind user has permission to.
 - Support for anonymous clients (that is, no bind user at all)
 - Any kind of SASL authentication (specifically ones which are more secure and do not send the plaintext password to the server, e.g [RFC2222](https://www.ietf.org/rfc/rfc2222.txt)

We would also need some sort of Complement tests for this realistically. Thoughts @neilalexander ?

For comparison, Synapse has LDAP support, with a config that looks like:
```yaml
password_providers:
#    # Example config for an LDAP auth provider
#    - module: "ldap_auth_provider.LdapAuthProvider"
#      config:
#        enabled: true
#        uri: "ldap://ldap.example.com:389"
#        start_tls: true
#        base: "ou=users,dc=example,dc=com"
#        attributes:
#           uid: "cn"
#           mail: "email"
#           name: "givenName"
#        #bind_dn:
#        #bind_password:
#        #filter: "(objectClass=posixAccount)"
```

clientapi/auth/password.go
```diff
@@ -62,6 +70,101 @@ func (t *LoginTypePassword) Login(ctx context.Context, req interface{}) (*Login,
...
+			}
+			return &r.Login, nil
+		}
+
+		userDN := sr.Entries[0].DN
+		err = conn.Bind(userDN, r.Password)
+		if err != nil {
+			return nil, &util.JSONResponse{
+				Code: http.StatusUnauthorized,
+				JSON: jsonerror.InvalidUsername(err.Error()),
```
neilalexander, line 130: I suspect this is the wrong return code for an invalid bind DN, since this error code will be returned to the client as if the user had entered a wrong account password?

clientapi/auth/password.go
```diff
@@ -62,6 +70,101 @@ func (t *LoginTypePassword) Login(ctx context.Context, req interface{}) (*Login,
 			JSON: jsonerror.InvalidUsername(err.Error()),
 		}
 	}
+	if len(t.Config.LDAP.Host) > 0 {
```
neilalexander, line 73: Does this mean that we can't have local accounts in addition to LDAP accounts?

clientapi/auth/password.go
```diff
@@ -62,6 +70,101 @@ func (t *LoginTypePassword) Login(ctx context.Context, req interface{}) (*Login,
...
+		}
+
+		_, err = t.GetAccountByLocalpart(ctx, localpart)
+		if err != nil {
+			if err == sql.ErrNoRows {
+				var accRes api.PerformAccountCreationResponse
+				err = t.UserAPI.PerformAccountCreation(ctx, &api.PerformAccountCreationRequest{
+					AppServiceID: "",
+					Localpart:    localpart,
+					Password:     r.Password,
```
neilalexander, line 141: This looks like we replicate the user's LDAP password from the point of their first login into the Dendrite account database, which I don't think we want to do since we should probably already be asking LDAP each time the user authenticates with their *current* password.

neilalexander (changes requested):
Thanks for the contribution! Ultimately this PR seems reasonable but I agree with @Kegsay that we need to support, at a minimum:
* restricting the search base to specific parts of the directory
* restricting authentication to certain LDAP group memberships
* avoiding plain-text auth

It's also important that we get the account management flow right here — for example, some questions to consider:
* what happens if you delete or disable an LDAP account after creation?
* can we convert an LDAP account to a local user or back?
* what happens if you have local users and then enable LDAP later?

cmd/sytest/sytest.go
phoenix-2021: Yes, it should be removed from here.

clientapi/auth/password.go
phoenix-2021: Good idea, it should be changed like this.

phoenix-2021:
> Thanks for the contribution! Ultimately this PR seems reasonable but I agree with @kegsay that we need to support, at a minimum:
> 
> * restricting the search base to specific parts of the directory
> * restricting authentication to certain LDAP group memberships
> * avoiding plain-text auth
> 
> It's also important that we get the account management flow right here — for example, some questions to consider:
> 
> * what happens if you delete or disable an LDAP account after creation?
> * can we convert an LDAP account to a local user or back?
> * what happens if you have local users and then enable LDAP later?

Thank you very much for your guidance!
Since I am not familiar with dendrite and matrix, I have no idea how to deal with the relationship between local accounts and ldap, but I will continue to understand.
I have a small question,  thank according to my understanding of LDAP, field LDAP.BaseDN in config can be used to specify the search scope, field LDAP.Filter can be used to filter groups, Configured by the administrator here.

neilalexander:
I tried to merge the latest `main` commits into this branch but it appears the PR doesn't allow maintainer edits. Please check the `Allow edits from maintainers` box.

neilalexander:
Closing this for now because it is stale. If you get some bandwidth for this then please let us know and we can reopen it.

Issue #602, opened 2019-01-07
Implement the presence module


babolivier:
- [ ] Implement the [m.presence](https://matrix.org/docs/spec/client_server/r0.4.0.html#m-presence) event 
- [ ] Implement [PUT /_matrix/client/r0/presence/{userId}/status](https://matrix.org/docs/spec/client_server/r0.4.0.html#put-matrix-client-r0-presence-userid-status)
- [ ] Implement [GET /_matrix/client/r0/presence/{userId}/status](https://matrix.org/docs/spec/client_server/r0.4.0.html#get-matrix-client-r0-presence-userid-status)
- [ ] Implement [POST /_matrix/client/r0/presence/list/{userId}](https://matrix.org/docs/spec/client_server/r0.4.0.html#post-matrix-client-r0-presence-list-userid)
- [ ] Implement [GET /_matrix/client/r0/presence/list/{userId}](https://matrix.org/docs/spec/client_server/r0.4.0.html#get-matrix-client-r0-presence-list-userid)
- [ ] Implement setting and getting presence through [/sync](https://matrix.org/docs/spec/client_server/r0.4.0.html#get-matrix-client-r0-sync)

Kegsay:
Related sytests:
```
    × GET /presence/:user_id/status fetches initial status
    × PUT /presence/:user_id/status updates my presence
    × Presence change reports an event to myself
    × Friends presence changes reports events
    × Presence changes are reported to local room members
    × Presence changes are also reported to remote room members
    × Presence changes to UNAVAILABLE are reported to local room members
    × Presence changes to UNAVAILABLE are reported to remote room members
    × New federated private chats get full presence information (SYN-115)
    × Left room members do not cause problems for presence
    Sync API:
    × Existing members see new member's presence
    × Get presence for newly joined members in incremental sync
    × User sees their own presence in a sync
    × User sees updates to presence from other users in the incremental sync.
```

alexkursell:
I would love to take a crack at implementing this! I assume that the implementation would be somewhat similar to #1528.

ptman:
Assuming this is the resource sink it is for synapse, there should be a toggle for turning presence off.

S7evinK:
> I would love to take a crack at implementing this! I assume that the implementation would be somewhat similar to #1528.

@alexkursell: Are you still on it or need help? Should be pretty similar to #1528.

neilalexander:
This landed in Dendrite 0.8.0.

PR #1879, opened 2021-06-18
[WIP] Add Presence module

S7evinK:
WIP of a presence implementation. (#602)

Todo:
- [x] Implement `GetPresence`
- [x] Add `m.presence` to gomatrixserverlib
- [x] Implement `/sync`
- [x] Implement federationapi & sender

Questions:
- Where to store the data? (memory is not possible, as we need to store e.g. `status_msg`)
- Add a new component? (would make sense for the future, scaling wise [IMHO])

### Pull Request Checklist


r3k2:
When will this be on master?

S7evinK:
@r3k2 There are still some missing pieces (e.g. database), which I didn't start yet.

S7evinK:
@kegsay @neilalexander 
Opinions on where to store the data?

`syncapi` (same as receipts) - would require an `inthttp` package for getting the presence from the `clientapi`
`eduserver` - would require a new database

The other idea would be, as mentioned, a new component with the known pros/cons.

kegsay:
Adding components is extremely costly and incurs IPC costs on any public API. We generally want to stick logic to orthogonal scaling points. Presence will scale based on the number of users independent of rooms, which is the same as the user API, so that feels like the most logical place to put this.

r3k2:
Good work with this @S7evinK I seen in the last 2 weeks there was a lot of things added, and you stuck on the checks... any ETA on this?

S7evinK:
Thing is, Sytest is using deprecated endpoints, so most of the [tests](https://github.com/matrix-org/dendrite/issues/602#issuecomment-679217922) listed will not pass at the moment. I've created an [issue](https://github.com/matrix-org/sytest/issues/1108) for that.

I've been running this for a week or so and only found one issue i need to fix. (unsure if the outbound federation is working correctly, but inbound looks good)

kegsay:
If you're only stuck on the sytests, I can convert them to be using `/sync`.

S7evinK:
I *think* the sytests are the last missing pieces. Thanks for having a look at them. :)
Tested the federation part earlier today with two dendrite monoliths, seems to work so far. Will have a look at optimizing the incoming path, as currently every presence event is directly send to the userapi, which should be better in bulk.

r3k2:
yay! @S7evinK  thanks for your work on this. hopefully @kegsay  etc will push to master soon.

S7evinK:
Unfortunately still stuck with the following tests. full-api mode is a bit worse, probably just error handling, didn't have a look at that yet.
```
Presence: 67% (10/15 tests)
    × Newly joined room includes presence in incremental sync
    × Get presence for newly joined members in incremental sync
    × User sees their own presence in a sync
    × User sees updates to presence from other users in the incremental sync.
    × New federated private chats get full presence information (SYN-115)
```

thinkwelltwd:
I'm doing some due diligence research about moving to a multi-homeserver Matrix deployment. For performance benefits, I'd be very glad to deploy Dendrite, but for the lack of presence. Is there any sort of ETA (ie 1-6 months) for this PR, or should I just plan to go with Synapse?

S7evinK:
As there have been many changes since I've started working on this, and I now know better (hopefully), I'll close this for now. But definitely will revisit this, unfortunately no ETA on this.

Issue #1308, opened 2020-08-24
Fix issues around /roomId/members


kegsay:
Spec: https://matrix.org/docs/spec/client_server/r0.6.1#get-matrix-client-r0-rooms-roomid-members
Sytests:
```
    × Can get rooms/{roomId}/members at a given point
    × Can filter rooms/{roomId}/members
```

Grotax:
So I investigated the first API endpoint.
I guessed that
https://github.com/matrix-org/dendrite/blob/master/clientapi/routing/memberships.go#L60
needs to be extended with an "At" and passing it though from the client request.
So I ended up at 
https://github.com/Grotax/dendrite/blob/master/roomserver/storage/interface.go#L129

But I guess then the At would need to be implemented for each DB implementation.

I think it makes sense to only query the data in the DB we actually need but I'm not a 100% sure.
So I'm just gonna leave this here for now.

akhil-dh1:
@Kegsay @Grotax I would like to work on this issue. I am new to Matrix and I am not familiar with its codebase. It would be great if anyone could provide me some pointers on how to begin fixing this issue.

alam0rt:
@kegsay  - been also taking a look at this, quite new to Matrix (codebase wise).

Looking at how `GetMembershipEventNIDsForRoom()` is implemented, I'm not really sure how to take the provided `at` and filter events using it.

Forgive the horror within https://github.com/matrix-org/dendrite/pull/2041, but any pointers would be great. Keen to implement.

kegsay:
Still failing as of today.

ShadowJonathan:
And as of today, while converting the test to complement

S7evinK:
Implemented as of https://github.com/matrix-org/dendrite/pull/2827

PR #1883, opened 2021-06-22
Make room members filterable

alexchan150:
For: https://github.com/matrix-org/dendrite/issues/1308

As mentioned in the issue, changes to rooms/{roomId}/members to include an "At" query param would require database changes to allow for picking out membership events at timestamps, so it will not be included in this PR. Instead, add filtering for type of membership (ie. join, invite, leave, ban).

### Pull Request Checklist

Signed-off-by: `Alex Chan <alexchan150@gmail.com>`


clientapi/routing/memberships.go
```diff
@@ -57,10 +58,45 @@ func GetMemberships(
 	_ *config.ClientAPI,
 	rsAPI api.RoomserverInternalAPI,
 ) util.JSONResponse {
+	membership := req.URL.Query().Get("membership")
+	notMembership := req.URL.Query().Get("not_membership")
+
+	regexpMembershipFilter, _ := regexp.Compile("join|invite|leave|ban")
```
kegsay, line 64: This is insufficient for validation. Strings like `joinaaa` will pass. You need to use `^` and `$` to ensure there are no other characters in the string. In addition, this can be done statically and you can use `regexp.MustCompile` to avoid building the regexp on every request.

roomserver/storage/shared/storage.go
```diff
@@ -343,14 +343,52 @@ func (d *Database) GetMembership(ctx context.Context, roomNID types.RoomNID, req
...
+				eventNIDs = append(eventNIDs, filteredEventNIDs...)
+			}
+			if filter == "invite" {
+				filteredEventNIDs, err = d.MembershipTable.SelectMembershipsFromRoomAndMembership(
+					ctx, roomNID, tables.MembershipStateInvite, localOnly,
+				)
+				if err != nil {
+					return eventNIDs, err
+				}
+				eventNIDs = append(eventNIDs, filteredEventNIDs...)
```
kegsay, line 376: This seems needlessly complex. Ultimately, you're mapping `filter` into a `tables.MembershipStateXXX`. Why can't you just do:
```go
state := tables.MembershipStateLeaveOrBan
switch (filter) {
    case "invite":
        state = tables.MembershipStateInvite
    case "join":
        state = tables.MembershipStateJoin
    case "leave":
        fallthrough
    case "ban":
        state = tables.MembershipStateLeaveOrBan
    default:
        return nil, fmt.Errorf("unknown membership filter: %v", filter)
}

filteredEventNIDs, err = d.MembershipTable.SelectMembershipsFromRoomAndMembership(
    ctx, roomNID, state, localOnly,
)
```

roomserver/storage/shared/storage.go
```diff
@@ -343,14 +343,52 @@ func (d *Database) GetMembership(ctx context.Context, roomNID types.RoomNID, req
...
+			}
+			if (filter == "ban" || filter == "leave") && !leaveOrBan {
+				filteredEventNIDs, err = d.MembershipTable.SelectMembershipsFromRoomAndMembership(
+					ctx, roomNID, tables.MembershipStateLeaveOrBan, localOnly,
+				)
+				if err != nil {
+					return eventNIDs, err
+				}
+				eventNIDs = append(eventNIDs, filteredEventNIDs...)
```
kegsay, line 385: Annoyingly it will be possible for the filter to want bans but not leaves, in which case this will return the wrong data. You need to check this case. In the common case where both leave and ban are present then no filtering needs to be done.

kegsay (changes requested):
I'm surprised `Can filter rooms/{roomId}/members` is still failing with this PR?

neilalexander:
Closing this for now because it is stale. If you get the bandwidth to continue working on this then please let us know and we'll happy revisit it.

PR #1885, opened 2021-06-25
Fixed log printing bug

kilosliu:
I have modified the code of the `/internal/sqlutil/trace.go:132`, log print here should be a function call, but actually it passed variables of function type.

Signed-off-by: `Kilos kilosliu@workly.ai>`


neilalexander (approved):
Changes in `trace.go` are fine, bit curious about what happened to the line endings in `go.sum` and `.gitignore`? :-)

kilosliu:
> Changes in `trace.go` are fine, bit curious about what happened to the line endings in `go.sum` and `.gitignore`? :-)

In the past I have added some folders to the `.gitignore`, but I finally removed them, and I have no idea about the `go.sum`, I have sync it with this repo many times but it has still a few diffirence

PR #1888, opened 2021-06-28
Use a custom FIFO queue for the RS input API

neilalexander:
Currently the RS input API queues up work for a room by using a buffered channel. The problem is that if the channel is full, we'll block the entire loop until it isn't full anymore and other work for other rooms gets stuck behind that.

This adds a custom FIFO queue (which I originally wrote for Pinecone) which will grow as needed and won't block. That way room events are always going to be queued instantly into their worker queues.

### DIFF
--- a/roomserver/internal/input/input_fifo.go
+++ b/roomserver/internal/input/input_fifo.go
@@ -13,7 +13,7 @@ type fifoQueue struct {
 
 func newFIFOQueue() *fifoQueue {
 	q := &fifoQueue{
-		notifs: make(chan struct{}),
+		notifs: make(chan struct{}, 1),
 	}
 	return q
 }
--- a/roomserver/internal/input/input_fifo.go
+++ b/roomserver/internal/input/input_fifo.go
@@ -1,0 +1,60 @@
+package input
+
+import (
+	"sync"
+)
+
+type fifoQueue struct {
+	frames []*inputTask
+	count  int
+	mutex  sync.Mutex
+	notifs chan struct{}
+}
+
+func newFIFOQueue() *fifoQueue {
+	q := &fifoQueue{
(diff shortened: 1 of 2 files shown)


kegsay:
This leaves memory growth unbounded though which feels bad.

neilalexander:
Memory usage is unbounded anyway because we can stack up HTTP requests in the federation API. This is about choosing a trade-off — potentially use more memory in the roomserver input API to reduce the chance of blocking, or definitely use more memory in the federation API with requests waiting for the roomserver to unblock.

roomserver/internal/input/input_fifo.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func newFIFOQueue() *fifoQueue {
+	q := &fifoQueue{
+		notifs: make(chan struct{}, 1),
+	}
+	return q
+}
+
+func (q *fifoQueue) push(frame *inputTask) bool {
```
kegsay, line 21: Return arg seems pointless, it's always `true`.

roomserver/internal/input/input_fifo.go
```diff
@@ -0,0 +1,60 @@
...
+		// grown significantly if the queue was hammered for some reason
+		q.frames = nil
+	}
+	return frame, true
+}
+
+func (q *fifoQueue) wait() <-chan struct{} {
+	q.mutex.Lock()
+	defer q.mutex.Unlock()
+	if q.count > 0 {
```
kegsay, line 54: This races in that you can call `wait()`, have it return and then try to `pop()` only for nothing to be there. You need to make this clear in the docs.

kegsay (changes requested):

PR #1890, opened 2021-06-29
Reduce memory usage in federation /send endpoint

neilalexander:
This PR ensures that we're aggressively using `cacheAndReturn` on all paths that bring in new events to ensure that we're a) populating the cache to begin with and b) deduplicating every time an event is already in the cache.

This reduces the memory footprint of the federation API substantially in early testing.


kegsay (approved):
This cache is txn scoped though, so the benefits may not be as good as you'd like.

neilalexander:
Indeed, but it helps a lot in cases where we already know *most* of what we set out to find. Some caches scoped to the room rather than the transaction would probably be even more helpful but not sure how to upper-bound that yet.

PR #1895, opened 2021-06-30
Asynchronous roomserver input from federation API /send

neilalexander:
I have no idea if this will improve things or make a lot of duplicate work, but this allows the RS input API to asynchronously process input events rather than blocking on the input API call.

Note that there's no persistence here yet - if the process crashes or restarts, the queued work will currently be lost.


kegsay:
> the queued work will currently be lost.

We cannot merge anything that makes Dendrite not durable :(

neilalexander:
> We cannot merge anything that makes Dendrite not durable :(

Indeed. If we get NATS merged then I'll update this to persist using Jetstream.

neilalexander:
Stale, ultimately will be cured in NATS instead.

PR #1901, opened 2021-07-06
Add dendrite-upgrade-test

kegsay:
This binary can be run under CI to assert that we can upgrade from the last versioned release to HEAD. This should help catch DB migration errors like https://github.com/matrix-org/dendrite/pull/1839

Currently we need to specify the version, but I can add a `lastVer` keyword or something for CI purposes. To try it out:
```
go build ./cmd/dendrite-upgrade-tests
./dendrite-upgrade-tests --from v0.3.11
```

Requires a running Docker, doesn't support docker-in-docker yet.


cmd/dendrite-upgrade-tests/main.go
```diff
@@ -0,0 +1,416 @@
...
+func downloadArchive(cli *http.Client, tmpDir, archiveURL string, dockerfile []byte) (*bytes.Buffer, error) {
+	resp, err := cli.Get(archiveURL)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != 200 {
+		return nil, fmt.Errorf("got HTTP %d", resp.StatusCode)
+	}
+	os.Mkdir(tmpDir, os.ModePerm)
```
neilalexander, line 106: We might want to check for errors here, in case there's a permissions problem.

cmd/dendrite-upgrade-tests/tests.go
```diff
@@ -0,0 +1,189 @@
...
+	r := struct {
+		RoomID string `json:"room_id"`
+	}{}
+	err = client.MakeRequest("GET", u, nil, &r)
+	if err != nil {
+		return fmt.Errorf("failed to /directory: %s", err)
+	}
+	log.Println("    public room exists: OK")
+
+	history, err := client.Messages(r.RoomID, client.Store.LoadNextBatch(client.UserID), "", 'b', 100)
```
neilalexander, line 172: Does it make sense to check that `r.RoomID` is set?

neilalexander (approved):
Overall this is really impressive. No idea if the couple inline comments below are useful changes to make for sanity.

cmd/dendrite-upgrade-tests/main.go
kegsay: If there is a permissions problem it'll fail when trying to write. I originally ignored the error to handle cases where we partially ran before (and hence the directory already existed), but this isn't the case anymore now. Will add checks.

cmd/dendrite-upgrade-tests/tests.go
kegsay: Yes, it does!

PR #1902, opened 2021-07-07
Fix concurrent map reads/writes on t.hadEvents

neilalexander:
This PR ensures that `t.hadEvents` in `/send` doesn't panic due to concurrent map reads/writes.


kegsay:
Rather than very carefully wrapping r/ws with the mutex, can we instead make functions which exposes read/write operations please? That will make it harder to miss a lock somewhere.

neilalexander:
If we do that then the only safe way to read a map returned from those functions is to make a copy of it, which seems like a nice way to burn allocations. There aren't *too* many cases where we need to hit this map.

PR #1904, opened 2021-07-07
db migration: handle create events with no state blocks from v0.1.0

kegsay:
Handles the very specific case of:
 - Create event has a state snapshot nid
 - Said state snapshot nid refers to 0 state blocks


neilalexander (approved):
I think this looks OK!

PR #1909, opened 2021-07-09
Add more optimised code path for checking if we're in a room

neilalexander:
Asking the roomserver for all of the room state events is an expensive operation, but we don't strictly need to do that when we only care about finding out if the local server is in the room. The memberships table contains that information.

This adds a new code path for a much cheaper check using the memberships table.

It also adds code to the federation API to not bother processing events if the server thinks we aren't in the room.

### DIFF
--- a/roomserver/internal/query/query.go
+++ b/roomserver/internal/query/query.go
@@ -36,6 +36,7 @@ import (
 type Queryer struct {
 	DB         storage.Database
 	Cache      caching.RoomServerCaches
+	ServerName gomatrixserverlib.ServerName
 	ServerACLs *acls.ServerACLs
 }
 
@@ -140,6 +141,18 @@ func (r *Queryer) QueryMissingAuthPrevEvents(
 
 	response.RoomExists = !info.IsStub
 	response.RoomVersion = info.RoomVersion
+
+	joined, err := r.DB.GetLocalServerInRoom(ctx, info.RoomNID)
+	if err != nil {
+		return fmt.Errorf("r.DB.GetLocalServerInRoom: %w", err)
+	}
+	response.RoomJoined = joined
+
+	// If we're not joined to the room then there's no point in hitting
+	// the database further to work out which events we're missing.
+	if !joined {
+		return nil
+	}
 
 	for _, authEventID := range request.AuthEventIDs {
 		if nids, err := r.DB.EventNIDs(ctx, []string{authEventID}); err != nil || len(nids) == 0 {
@@ -327,6 +340,16 @@ func (r *Queryer) QueryServerJoinedToRoom(
 		return nil
 	}
 	response.RoomExists = true
+
+	if request.ServerName == r.ServerName {
+		var joined bool
+		joined, err = r.DB.GetLocalServerInRoom(ctx, info.RoomNID)
+		if err != nil {
+			return fmt.Errorf("r.DB.GetLocalServerInRoom: %w", err)
+		}
+		response.IsInRoom = joined
+		return nil
+	}
 
 	eventNIDs, err := r.DB.GetMembershipEventNIDsForRoom(ctx, info.RoomNID, true, false)
 	if err != nil {
--- a/roomserver/internal/query/query.go
+++ b/roomserver/internal/query/query.go
@@ -142,14 +142,6 @@ func (r *Queryer) QueryMissingAuthPrevEvents(
 	response.RoomExists = !info.IsStub
 	response.RoomVersion = info.RoomVersion
 
-	if response.RoomExists {
-		joined, err := r.DB.GetLocalServerInRoom(ctx, info.RoomNID)
-		if err != nil {
-			return fmt.Errorf("r.DB.GetLocalServerInRoom: %w", err)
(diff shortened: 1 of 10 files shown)


roomserver/internal/query/query.go
```diff
@@ -141,6 +142,18 @@ func (r *Queryer) QueryMissingAuthPrevEvents(
...
 
+	joined, err := r.DB.GetLocalServerInRoom(ctx, info.RoomNID)
+	if err != nil {
+		return fmt.Errorf("r.DB.GetLocalServerInRoom: %w", err)
+	}
+	response.RoomJoined = joined
+
+	// If we're not joined to the room then there's no point in hitting
+	// the database further to work out which events we're missing.
+	if !joined {
```
kegsay, line 153: This feels like a bad smell. `QueryMissingAuthPrevEvents` should do as it is told, regardless of the current room state if it has the data. Otherwise any caller to `QueryMissingAuthPrevEvents` needs to consider this which makes it hard to work with.

roomserver/storage/postgres/membership_table.go
```diff
@@ -324,3 +335,19 @@ func (s *membershipStatements) UpdateForgetMembership(
 	)
 	return err
 }
+
+func (s *membershipStatements) SelectLocalServerInRoom(ctx context.Context, roomNID types.RoomNID) (bool, error) {
+	rows, err := s.selectLocalServerInRoomStmt.QueryContext(ctx, roomNID)
```
kegsay, line 340: Why not `QueryRowContext`?

roomserver/storage/postgres/membership_table.go
```diff
@@ -124,6 +124,15 @@ var selectKnownUsersSQL = "" +
...
 	") AND membership_nid = " + fmt.Sprintf("%d", tables.MembershipStateJoin) + " AND event_state_key LIKE $2 LIMIT $3"
 
+// selectLocalServerInRoomSQL is an optimised case for checking if we, the local server,
+// are in the room by using the target_local column of the membership table. Normally when
+// we want to know if a server is in a room, we have to unmarshal the entire room state which
+// is expensive. The presence of a single row from this query suggests we're still in the
+// room, no rows returned suggests we aren't.
+const selectLocalServerInRoomSQL = "" +
+	"SELECT room_nid FROM roomserver_membership WHERE target_local = true AND membership_nid = 3 AND room_nid = $1 " +
+	"FETCH FIRST 1 ROWS ONLY"
```
kegsay, line 134: Why not `LIMIT 1`?

roomserver/storage/sqlite3/membership_table.go
```diff
@@ -100,6 +100,14 @@ var selectKnownUsersSQL = "" +
 	"  SELECT DISTINCT room_nid FROM roomserver_membership WHERE target_nid=$1 AND membership_nid = " + fmt.Sprintf("%d", tables.MembershipStateJoin) +
 	") AND membership_nid = " + fmt.Sprintf("%d", tables.MembershipStateJoin) + " AND event_state_key LIKE $2 LIMIT $3"
 
+// selectLocalServerInRoomSQL is an optimised case for checking if we, the local server,
+// are in the room by using the target_local column of the membership table. Normally when
+// we want to know if a server is in a room, we have to unmarshal the entire room state which
+// is expensive. The presence of a single row from this query suggests we're still in the
+// room, no rows returned suggests we aren't.
+const selectLocalServerInRoomSQL = "" +
+	"SELECT room_nid FROM roomserver_membership WHERE target_local = 1 AND membership_nid = 3 AND room_nid = $1 LIMIT 1"
```
kegsay, line 109: Feels icky to have `membership_nid=3`, if we ever shuffle the nids for new memberships (e.g knocking) then this is a footgun as it's easy to forget to update this. Just use a paramterised query arg please.

kegsay (changes requested):

roomserver/internal/query/query.go
neilalexander: We can leave it just to do the work and burn the effort, or we could add a request parameter to tell it whether to bother if we aren't joined to the room?

roomserver/storage/postgres/membership_table.go
neilalexander: Changed to `QueryRowContext`.

roomserver/storage/postgres/membership_table.go
neilalexander: I had in mind that there was an optimisation around `FETCH FIRST 1 ROWS ONLY` vs `LIMIT 1`, but looking into it I don't think it applies in PostgreSQL. Switched it to `LIMIT 1`.

roomserver/storage/sqlite3/membership_table.go
neilalexander: Done.

roomserver/internal/query/query.go
kegsay: It should be caller dependent really. Who is calling this API and when?

kegsay (approved):

PR #1912, opened 2021-07-09
Replace some helpers.IsServerCurrentlyInRoom calls with query API

neilalexander:
This replaces a couple of instances where we call `helpers.IsServerCurrentlyInRoom` with calls to `query.QueryServerJoinedToRoom` instead, so we can take advantage of optimisations when querying the local server name.

The remaining user of `helpers.IsServerCurrentlyInRoom` is `ScanEventTree`, so I've left it alone for now rather than removing it altogether.

### DIFF
--- a/roomserver/internal/helpers/helpers.go
+++ b/roomserver/internal/helpers/helpers.go
@@ -50,6 +50,10 @@ func UpdateToInviteMembership(
 	return updates, nil
 }
 
+// IsServerCurrentlyInRoom checks if a server is in a given room, based on the room
+// memberships. If the servername is not supplied then the local server will be
+// checked instead using a faster code path.
+// TODO: This should probably be replaced by an API call.
 func IsServerCurrentlyInRoom(ctx context.Context, db storage.Database, serverName gomatrixserverlib.ServerName, roomID string) (bool, error) {
 	info, err := db.RoomInfo(ctx, roomID)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


roomserver/internal/helpers/helpers.go
```diff
@@ -50,6 +50,10 @@ func UpdateToInviteMembership(
 	return updates, nil
 }
 
+// IsServerCurrentlyInRoom checks if a server is in a given room, based on the room
+// memberships. If the servername is not supplied then the local server will be
+// checked instead using a faster code path.
```
kegsay, line 55: It feels like a bug to call `IsServerCurrentlyInRoom` without.. a server? Why have this weird API shape?

roomserver/internal/perform/perform_join.go
```diff
@@ -205,7 +207,14 @@ func (r *Joiner) performJoinRoomByID(
 
 	// Force a federated join if we aren't in the room and we've been
 	// given some server names to try joining by.
-	serverInRoom, _ := helpers.IsServerCurrentlyInRoom(ctx, r.DB, r.ServerName, req.RoomIDOrAlias)
+	inRoomReq := &api.QueryServerJoinedToRoomRequest{
+		RoomID: req.RoomIDOrAlias,
+	}
```
kegsay, line 212: Missing server name.

neilalexander: That's acceptable — in the API comments:
```
	// Server name of the server to find. If not specified, we will
	// default to checking if the local server is joined.
```

roomserver/internal/helpers/helpers.go
neilalexander: Mostly because nothing in `IsServerCurrentlyInRoom` knows the current server name to compare against, and then it also has the same behaviour as `QueryServerJoinedToRoomRequest`.

kegsay: Bleh, I don't like it. I guess we can fix that later.

roomserver/internal/perform/perform_join.go
kegsay: :S okay

PR #1917, opened 2021-07-13
Fix failing complement test

kegsay:
Specifically `TestBannedUserCannotSendJoin`

### DIFF
--- a/federationapi/routing/join.go
+++ b/federationapi/routing/join.go
@@ -271,14 +271,24 @@ func SendJoin(
 
 	// Check if the user is already in the room. If they're already in then
 	// there isn't much point in sending another join event into the room.
+	// Also check to see if they are banned: if they are then we reject them.
 	alreadyJoined := false
+	isBanned := false
 	for _, se := range stateAndAuthChainResponse.StateEvents {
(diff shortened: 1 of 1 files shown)


federationapi/routing/join.go
```diff
@@ -271,17 +271,27 @@ func SendJoin(
...
 			alreadyJoined = (membership == gomatrixserverlib.Join)
+			isBanned = (membership == gomatrixserverlib.Ban)
 			break
 		}
 	}
 
+	if isBanned {
+		return util.JSONResponse{
+			Code: http.StatusForbidden,
+			JSON: jsonerror.Forbidden("user is banned"),
```
neilalexander, line 291: If this is a message that will get passed through to clients then we might want to make it friendlier.

neilalexander (approved):
LGTM overall.

federationapi/routing/join.go
kegsay: Friendlier how? I'm saying they are banned. Also, this requires explicit fudginess for it to work (you need to make_join with an unbanned user, then swap the state_key over to a banned user), so I don't expect this to happen in normal operation.

PR #1918, opened 2021-07-13
bugfix: retire invites even when we cannot talk to the remote server to make/send_leave

### LINKED ISSUE #1743: Cannot reject invites over federation
I'm on dendrite.matrix.org.

### Description
I cannot reject invites

### Steps to reproduce

- Invite a user on a server running Dendrite
- Revoke their invite
- They will not be able to reject the invite

Edit: after doing a bit more testing only invites over federation have this issue. Inviting/disinviting other users on dendrite.matrix.org (for example) works great.
Important: After doing a few more tests, it looks like revoking their invite is a necessary step in these tests (rejecting invites before having them revoked seems to work just fine)
<!--

``` (three backticks, on a line on their own), so that they are formatted legibly.

-->

### PR DESCRIPTION
**Pull request by kegsay**
Also modify the leave response in /sync to include a fake event as this is ultimately
what clients (and sytest) will use to determine leave-ness.

Fixes #1743 #1727 #1828 #1873

With sytests.

### DIFF
--- a/syncapi/streams/stream_invite.go
+++ b/syncapi/streams/stream_invite.go
@@ -3,7 +3,7 @@ package streams
 import (
 	"context"
 	"crypto/sha256"
-	"encoding/hex"
+	"encoding/base64"
 	"strconv"
 	"time"
 
@@ -64,7 +64,7 @@ func (p *InviteStreamProvider) IncrementalSync(
 			h := sha256.Sum256(append([]byte(roomID), []byte(strconv.FormatInt(int64(to), 10))...))
 			lr.Timeline.Events = append(lr.Timeline.Events, gomatrixserverlib.ClientEvent{
 				// fake event ID which muxes in the to position
-				EventID:        "$" + hex.EncodeToString(h[:]),
+				EventID:        "$" + base64.RawURLEncoding.EncodeToString(h[:]),
 				OriginServerTS: gomatrixserverlib.AsTimestamp(time.Now()),
 				RoomID:         roomID,
 				Sender:         req.Device.UserID,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: syncapi/streams/stream_invite.go [hunk 1]
```diff
@@ -56,6 +59,16 @@ func (p *InviteStreamProvider) IncrementalSync(
 	for roomID := range retiredInvites {
 		if _, ok := req.Response.Rooms.Join[roomID]; !ok {
 			lr := types.NewLeaveResponse()
+			lr.Timeline.Events = append(lr.Timeline.Events, gomatrixserverlib.ClientEvent{
+				// fake event ID which muxes in the to position
+				EventID:        "$stub-retired-invite-" + roomID + "-" + strconv.FormatInt(int64(to), 10),
```
neilalexander, line 64: We should probably strip `roomID` so that it only contains the alphanumeric characters and not `!`, `:` etc. I don't think `!` is a valid character for the event ID in any room version and `:` has a special meaning as a separator in room version 1. Otherwise we run the risk of sending something that will confuse clients.

neilalexander:
We probably also need to make sure that the `syncapi_invites` table marks the event as `deleted` so that it doesn't come back in the next full sync.

Are we going to do anything with the `roomserver_invites` or just leave that for consistency with other federated servers?

kegsay:
The retiring invites logic remains the same as before, so it will already be doing:

> that the syncapi_invites table marks the invite as deleted so that it doesn't come back in the next full sync.

I will leave the `roomserver_invites` table for consistency with other federated servers.

neilalexander:
> The retiring invites logic remains the same as before, so it will already be doing

Lovely!

> I will leave the `roomserver_invites` table for consistency with other federated servers

This probably makes sense.

PR #1921, opened 2021-07-15
Update dendrite-demo-yggdrasil to Yggdrasil 0.4

neilalexander:
This updates the `dendrite-demo-yggdrasil` to Yggdrasil v0.4:
* QUIC is replaced with uTP, like the Pinecone demo
* DHT lookups are no longer needed as we can route using the public key
* Dynamic reconfiguration (setting peers, multicast etc at runtime) is currently not supported so that is commented out


cmd/dendrite-demo-yggdrasil/yggconn/node.go
```diff
@@ -163,64 +125,33 @@ func (n *Node) Stop() {
...
-func (n *Node) SigningPrivateKey() ed25519.PrivateKey {
-	privBytes, _ := hex.DecodeString(n.config.SigningPrivateKey)
-	return ed25519.PrivateKey(privBytes)
+func (n *Node) PrivateKey() ed25519.PrivateKey {
+	sk := make(ed25519.PrivateKey, ed25519.PrivateKeySize)
+	sb, err := hex.DecodeString(n.config.PrivateKey)
+	if err == nil {
+		copy(sk, sb[:])
+	} else {
+		panic(err)
```
kegsay, line 137: Is this appropriate?

kegsay (approved):

cmd/dendrite-demo-yggdrasil/yggconn/node.go
neilalexander: I think so, as we have bigger problems if our own private key is borked. Since we use this function to populate the Dendrite portion of the config, it should result in a fairly speedy panic.

neilalexander:
Complement CI failures seem to be happening on `master` too.

PR #1922, opened 2021-07-15
Efficient /state_ids

neilalexander:
Beforehand, `/state_ids` was still pulling all of the state and auth events out of the database and unmarshalling them. This isn't really necessary because the `roomserver_events` table already knows both the event ID and the auth event NIDs, so we can look those up instead. This adds the necessary roomserver query shape to do just that.

### DIFF
--- a/roomserver/api/query.go
+++ b/roomserver/api/query.go
@@ -250,6 +250,28 @@ type QueryStateAndAuthChainResponse struct {
 	AuthChainEvents []*gomatrixserverlib.HeaderedEvent `json:"auth_chain_events"`
 }
 
+// QueryStateAndAuthChainIDsRequest is a request to QueryStateAndAuthChainIDs
+type QueryStateAndAuthChainIDsRequest struct {
+	// The room ID to query the state in.
+	RoomID string `json:"room_id"`
+	// The list of prev events for the event. Used to calculate the state at
+	// the event.
+	PrevEventIDs []string `json:"prev_event_ids"`
+}
+
+// QueryStateAndAuthChainIDsResponse is a response to QueryStateAndAuthChainIDs
+type QueryStateAndAuthChainIDsResponse struct {
+	// Does the room exist on this roomserver?
+	// If the room doesn't exist this will be false and StateEvents will be empty.
+	RoomExists bool `json:"room_exists"`
+	// The room version of the room.
+	RoomVersion gomatrixserverlib.RoomVersion `json:"room_version"`
+	// The state and auth chain event IDs that were requested.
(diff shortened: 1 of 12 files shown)


roomserver/api/query.go
```diff
@@ -250,6 +250,28 @@ type QueryStateAndAuthChainResponse struct {
...
+type QueryStateAndAuthChainIDsResponse struct {
+	// Does the room exist on this roomserver?
+	// If the room doesn't exist this will be false and StateEvents will be empty.
+	RoomExists bool `json:"room_exists"`
+	// The room version of the room.
+	RoomVersion gomatrixserverlib.RoomVersion `json:"room_version"`
+	// The state and auth chain event IDs that were requested.
+	// The lists will be in an arbitrary order.
+	StateEvents     []string `json:"state_event_ids"`
+	AuthChainEvents []string `json:"auth_chain_event_ids"`
```
kegsay, line 272: These aren't events, they are event IDs. `AuthChainEventIDs` and `StateEventIDs` please.

roomserver/internal/query/query.go
```diff
@@ -526,6 +526,70 @@ func (r *Queryer) QueryStateAndAuthChain(
...
+	if err != nil {
+		return fmt.Errorf("r.DB.EventIDs: %w", err)
+	}
+
+	for _, eventID := range stateEventIDs {
+		response.StateEvents = append(response.StateEvents, eventID)
+	}
+
+	for _, eventID := range authEventIDs {
+		response.AuthChainEvents = append(response.AuthChainEvents, eventID)
```
kegsay, line 587: These are just the auth events for the state events, not the auth chain.

roomserver/storage/sqlite3/events_table.go
```diff
@@ -145,6 +149,7 @@ func prepareEventsTable(db *sql.DB) (tables.Events, error) {
 		{&s.bulkSelectEventIDStmt, bulkSelectEventIDSQL},
 		{&s.bulkSelectEventNIDStmt, bulkSelectEventNIDSQL},
 		//{&s.selectRoomNIDForEventNIDStmt, selectRoomNIDForEventNIDSQL},
+		//{&s.bulkSelectEventAuthEventNIDsStmt, bulkSelectEventAuthEventNIDsSQL},
```
kegsay, line 152: Spurious, please remove

kegsay (changes requested):
It isn't returning the auth chain.

Devkey19:
Bug

neilalexander:
Stale.

PR #1931, opened 2021-07-19
Fix failing Complement tests

kegsay:
Requires https://github.com/matrix-org/complement/pull/166

### DIFF
--- a/federationapi/routing/join.go
+++ b/federationapi/routing/join.go
@@ -237,7 +237,7 @@ func SendJoin(
 			JSON: jsonerror.BadJSON("missing content.membership key"),
 		}
 	}
-	if membership != "join" {
+	if membership != gomatrixserverlib.Join {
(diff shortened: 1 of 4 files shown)


federationapi/routing/join.go
```diff
@@ -228,6 +229,21 @@ func SendJoin(
...
 
+	// Check that this is in fact a join event
+	membership, err := event.Membership()
+	if err != nil {
+		return util.JSONResponse{
+			Code: http.StatusBadRequest,
+			JSON: jsonerror.BadJSON("missing content.membership key"),
+		}
+	}
+	if membership != "join" {
```
neilalexander, line 240: Stylistic nitpick but we could use `gomatrixserverlib.Join` here.

neilalexander (approved):
Overall looks fine.

PR #1933, opened 2021-07-21
Optimise QueryServerJoinedToRoom

neilalexander:
This adds new queries into the membership table for determining whether a user belonging to a specific server is in a room. Simple `JOIN`s here are much more efficient than pulling membership events out of the database — the query planner pretty much just uses index scans so the queries are very fast.

This also removes the `ServerNames` field in `QueryServerJoinedToRoomResponse` as nothing uses it.

### DIFF
--- a/roomserver/storage/sqlite3/membership_table.go
+++ b/roomserver/storage/sqlite3/membership_table.go
@@ -117,7 +117,7 @@ const selectLocalServerInRoomSQL = "" +
 const selectServerInRoomSQL = "" +
 	"SELECT room_nid FROM roomserver_membership" +
 	" JOIN roomserver_event_state_keys ON roomserver_membership.target_nid = roomserver_event_state_keys.event_state_key_nid" +
-	" WHERE membership_nid = $1 AND room_nid = $2 AND event_state_key LIKE '%:$3' LIMIT 1"
+	" WHERE membership_nid = $1 AND room_nid = $2 AND event_state_key LIKE '%:' || $3 LIMIT 1"
 
 type membershipStatements struct {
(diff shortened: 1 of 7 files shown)


roomserver/storage/sqlite3/membership_table.go
```diff
@@ -108,6 +109,16 @@ var selectKnownUsersSQL = "" +
...
 
+// selectServerMembersInRoomSQL is an optimised case for checking for server members in a room.
+// The JOIN is significantly leaner than the previous case of looking up event NIDs and reading the
+// membership events from the database, as the JOIN query amounts to little more than two index
+// scans which are very fast. The presence of a single row from this query suggests the server is
+// in the room, no rows returned suggests they aren't.
+const selectServerInRoomSQL = "" +
+	"SELECT room_nid FROM roomserver_membership" +
+	" JOIN roomserver_event_state_keys ON roomserver_membership.target_nid = roomserver_event_state_keys.event_state_key_nid" +
+	" WHERE membership_nid = $1 AND room_nid = $2 AND event_state_key LIKE '%:' || $3 LIMIT 1"
```
kegsay, line 120: Are we certain that this won't misfire for things like server `foo.bar` and user IDs like `@alice:foo.barbaz`?

kegsay (approved):

roomserver/storage/sqlite3/membership_table.go
neilalexander: The `:` is a literal separator and `$3` is the exact server name. There's no wildcard either side of the server name so it should be fine and seems to do the right thing in my tests.

PR #1934, opened 2021-07-21
Various alias fixes

neilalexander:
This PR is a rather motley collection of fixes for aliases:

* Adds the `/room/{roomID}/aliases` endpoint
* Stops sending `m.room.aliases` events into that room, since that event no longer has any meaning in the spec
* Sends `m.room.canonical_alias` when creating the room, if an alias is specified
* No longer tries to send state events in response to changing room aliases (since we need to rethink this)
* Fixes a bunch of behaviour around the `/directory/{roomID}` endpoints for `PUT`, `DELETE` etc

Makes a couple more sytests happy and actually means that some of them that were passing before are now passing for the right reasons.

### DIFF
--- a/clientapi/routing/aliases.go
+++ b/clientapi/routing/aliases.go
@@ -1,0 +1,89 @@
+// Copyright 2017 Vector Creations Ltd
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package routing
+
+import (
+	"fmt"
+	"net/http"
+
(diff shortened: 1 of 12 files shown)


roomserver/storage/shared/storage.go
```diff
@@ -857,6 +857,9 @@ func (d *Database) GetStateEvent(ctx context.Context, roomID, evType, stateKey s
 	if err != nil {
 		return nil, err
 	}
+	if roomInfo == nil {
```
kegsay, line 860: We probably want to check for `IsStub` as well?

clientapi/routing/aliases.go
```diff
@@ -0,0 +1,94 @@
...
+		RoomID:      roomID,
+		StateTuples: []gomatrixserverlib.StateKeyTuple{stateTuple},
+	}
+	stateRes := &api.QueryCurrentStateResponse{}
+	if err := rsAPI.QueryCurrentState(req.Context(), stateReq, stateRes); err != nil {
+		util.GetLogger(req.Context()).WithError(err).Error("rsAPI.QueryCurrentState failed")
+		return util.ErrorResponse(fmt.Errorf("rsAPI.QueryCurrentState: %w", err))
+	}
+
+	if historyVisEvent, ok := stateRes.StateEvents[stateTuple]; ok {
```
kegsay, line 47: If there is no history visibility event we probably want to err on the side of rejection rather than allowing this to fall through.

kegsay (approved):
Otherwise LGTM

PR #1945, opened 2021-07-27
Key Backups (2/3) : Add E2E backup key tables

kegsay:
It's now possible to insert keys into a backup, but not query them.

Commented out the table creation to allow for the schema to change and land PRs onto `master` without having to do SQL deltas.

### DIFF
--- a/userapi/storage/accounts/postgres/storage.go
+++ b/userapi/storage/accounts/postgres/storage.go
@@ -418,6 +418,7 @@ func (d *Database) GetKeyBackup(
 	return
 }
 
+// nolint:nakedret
 func (d *Database) UpsertBackupKeys(
 	ctx context.Context, version, userID string, uploads []api.InternalKeyBackupSession,
 ) (count int64, etag string, err error) {
@@ -484,7 +485,6 @@ func (d *Database) UpsertBackupKeys(
 		} else {
 			etag = oldETag
 		}
-
 		return nil
 	})
 	return
--- a/userapi/storage/accounts/postgres/storage.go
+++ b/userapi/storage/accounts/postgres/storage.go
@@ -19,6 +19,7 @@ import (
 	"database/sql"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"strconv"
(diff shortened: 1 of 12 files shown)


userapi/storage/accounts/postgres/storage.go
```diff
@@ -377,7 +383,7 @@ func (d *Database) CreateKeyBackup(
 	ctx context.Context, userID, algorithm string, authData json.RawMessage,
 ) (version string, err error) {
 	err = sqlutil.WithTransaction(d.db, func(txn *sql.Tx) error {
```
neilalexander, line 385: Should these all be `d.writer` too?

neilalexander (approved):

userapi/storage/accounts/postgres/storage.go
```diff
@@ -396,18 +402,109 @@ func (d *Database) DeleteKeyBackup(
...
+	// "if the keys have different values for is_verified, then it will keep the key that has is_verified set to true"
+	if uploaded.IsVerified && !existing.IsVerified {
+		return true
+	}
+	// "if they have the same values for is_verified, then it will keep the key with a lower first_message_index"
+	if uploaded.FirstMessageIndex < existing.FirstMessageIndex {
+		return true
+	}
+	// "and finally, is is_verified and first_message_index are equal, then it will keep the key with a lower forwarded_count"
+	if uploaded.ForwardedCount < existing.ForwardedCount {
```
ara4n, line 506: I think there's a bug here. if uploaded.FirstMessageIndex > existing.FirstMessageIndex we should never replace (irrespective of forwarded count) as we'll overwrite the existing keys which have a lower index than the uploaded.FirstMessageIndex.

kegsay: Right I see now. I'll add a complement test for this then. I see now that Synapse has something similar but with `else if`.

kegsay: https://github.com/matrix-org/complement/pull/179

PR #1952, opened 2021-08-02
Syslog integration

sambhavsaggi:
~~This is 100% a work in progress, as I am much more comfortable with Rust than I am with Go. If anyone has any recommendation changes, feel free to send.~~ Ready to review!

### Pull Request Checklist

To use, switch the "file" logging parameter to "syslog" and use the following params:
```
address: localhost:514
protocol: tcp
```

Signed-off-by: Sambhav Saggi <samb@disroot.org>


sambhavsaggi:
Alright, so I've tested this and yes, it does send information to syslog. I did this with a quick Debian container with rsyslog attacked and ports 514 forwarded with TCP.

sambhavsaggi:
It doesn't look like UDP logging is working, so that's something to fix.

sambhavsaggi:
It looks like TCP works and everything is alright, so I'm gonna take this out of draft

internal/log.go
```diff
@@ -173,6 +178,34 @@ func setupFileHook(hook config.LogrusHook, level logrus.Level, componentName str
...
+	}
+
+}
+
+func setupSyslogHook(hook config.LogrusHook) {
+	syslogHook, err := lSyslog.NewSyslogHook(hook.Params["protocol"].(string), hook.Params["address"].(string), syslog.LOG_INFO, "")
+	if err == nil {
+		logrus.AddHook(syslogHook)
+	}
+}
```
S7evinK, line 207: I think the hook should accept the same parameters as `setupFileHook` and use the `componentName` as the syslog tag. Also, maybe use the `logLevelHook`, like the file hook?

S7evinK (changes requested):
Beside the comment, looks good.

internal/log.go
sambhavsaggi: Fixed!

S7evinK (approved):

kegsay (approved):
Thanks!

PR #1953, opened 2021-08-02
Cross-signing groundwork

neilalexander:
Depends on matrix-org/gomatrixserverlib#274.

This adds support for:

* uploading cross-signing keys
* uploading self cross-signing signatures (although without all of the necessary checks yet)
* query cross-signing keys for other users

Still to do in future PRs:

* all of the necessary verification for signature uploads
* include device signatures in device queries
* uploading other cross-signing signatures
* EDU support for sending/receiving cross-signing key/signature updates over federation
* notifying sync about key/signature changes


keyserver/internal/internal.go
```diff
@@ -344,6 +378,23 @@ func (a *KeyInternalAPI) queryRemoteKeys(
 				res.DeviceKeys[userID][deviceID] = keyJSON
 			}
 		}
+
+		for userID, body := range result.MasterKeys {
+			switch b := body.CrossSigningBody.(type) {
+			case *gomatrixserverlib.CrossSigningKey:
```
kegsay, line 384: Why do we need to type cast here? Why can't `body` just specify the correct types?

keyserver/internal/internal.go
```diff
@@ -393,9 +444,6 @@ func (a *KeyInternalAPI) queryRemoteKeysOnServer(
 			continue
 		}
 	}
-	if len(devKeys) == 0 {
-		return
-	}
```
kegsay, line 398: Why remove this?

keyserver/internal/cross_signing.go
```diff
@@ -0,0 +1,333 @@
+package internal
```
kegsay, line 1: Needs copyright.

keyserver/internal/cross_signing.go
```diff
@@ -0,0 +1,333 @@
...
+	}
+
+	// Does the key ID match the key value? Iterates exactly once
+	for keyID, keyData := range key.Keys {
+		b64 := keyData.Encode()
+		tokens := strings.Split(string(keyID), ":")
+		if len(tokens) != 2 {
+			return fmt.Errorf("key ID is incorrectly formatted")
+		}
+		if tokens[1] != b64 {
```
kegsay, line 28: spec: feels weirdly redundant if the key ID is the base 64 of the key data, doesn't it?

keyserver/internal/cross_signing.go
```diff
@@ -0,0 +1,333 @@
...
+	if userID != key.UserID {
+		return fmt.Errorf("key has a user ID mismatch")
+	}
+
+	// Does the key contain the correct purpose?
+	useful := false
+	for _, usage := range key.Usage {
+		if usage == purpose {
+			useful = true
+		}
```
kegsay, line 43: Probably should `break` after `true`?

keyserver/internal/cross_signing.go
```diff
@@ -0,0 +1,333 @@
...
+	}
+
+	// If the user hasn't given a new master key, then let's go and get their
+	// existing keys from the database.
+	var masterKey gomatrixserverlib.Base64Bytes
+	if !hasMasterKey {
+		existingKeys, err := a.DB.CrossSigningKeysForUser(ctx, req.UserID)
+		if err != nil {
+			res.Error = &api.KeyError{
+				Err: "User-signing key sanity check failed: " + err.Error(),
```
kegsay, line 90: You mean loading from the DB.

keyserver/internal/cross_signing.go
```diff
@@ -0,0 +1,333 @@
...
+		if !hasMasterKey {
+			res.Error = &api.KeyError{
+				Err:            "No master key was found, either in the database or in the request!",
+				IsMissingParam: true,
+			}
+			return
+		}
+	} else {
+		for _, keyData := range req.MasterKey.Keys { // iterates once, see sanityCheckKey
+			masterKey = keyData
```
kegsay, line 105: I would prefer it if `var masterKey` was defined earlier and we do this when we set `hasMasterKey = true`. At the moment the comment is clear why this works, but it would be better to just pull it up to where the sanity check is done IMO.

keyserver/internal/cross_signing.go
```diff
@@ -0,0 +1,333 @@
...
+	// Work out which things we need to verify the signatures for.
+	toVerify := make(map[gomatrixserverlib.CrossSigningKeyPurpose]gomatrixserverlib.CrossSigningKey, 3)
+	toStore := api.CrossSigningKeyMap{}
+	if len(req.MasterKey.Keys) > 0 {
+		toVerify[gomatrixserverlib.CrossSigningKeyPurposeMaster] = req.MasterKey
+	}
+	if len(req.SelfSigningKey.Keys) > 0 {
+		toVerify[gomatrixserverlib.CrossSigningKeyPurposeSelfSigning] = req.SelfSigningKey
+	}
+	if len(req.SelfSigningKey.Keys) > 0 {
```
kegsay, line 119: This is the same as `:116`. You presumably mean user signing?

keyserver/internal/cross_signing.go
```diff
@@ -0,0 +1,333 @@
...
+			switch key := keyOrDevice.CrossSigningBody.(type) {
+			case *gomatrixserverlib.CrossSigningKey:
+				if key.UserID == req.UserID {
+					if _, ok := selfSignatures[userID]; !ok {
+						selfSignatures[userID] = map[gomatrixserverlib.KeyID]gomatrixserverlib.CrossSigningForKeyOrDevice{}
+					}
+					selfSignatures[userID][keyID] = keyOrDevice
+				} else {
+					if _, ok := selfSignatures[userID]; !ok {
+						otherSignatures[userID] = map[gomatrixserverlib.KeyID]gomatrixserverlib.CrossSigningForKeyOrDevice{}
```
kegsay, line 177: Shouldn't this be `_, ok := otherSignatures[userID]`?

keyserver/internal/cross_signing.go
```diff
@@ -0,0 +1,333 @@
...
+				}
+
+			case *gomatrixserverlib.DeviceKeys:
+				if key.UserID == req.UserID {
+					if _, ok := selfSignatures[userID]; !ok {
+						selfSignatures[userID] = map[gomatrixserverlib.KeyID]gomatrixserverlib.CrossSigningForKeyOrDevice{}
+					}
+					selfSignatures[userID][keyID] = keyOrDevice
+				} else {
+					if _, ok := selfSignatures[userID]; !ok {
```
kegsay, line 189: Same here. Shouldn't this be `otherSignatures`?

keyserver/internal/cross_signing.go
```diff
@@ -0,0 +1,333 @@
...
+			if err != nil {
+				logrus.WithError(err).Errorf("Failed to get cross-signing signatures for user %q key %q", userID, keyID)
+				continue
+			}
+
+			appendSignature := func(originUserID string, originKeyID gomatrixserverlib.KeyID, signature gomatrixserverlib.Base64Bytes) {
+				if key.Signatures == nil {
+					key.Signatures = api.CrossSigningSigMap{}
+				}
+				if _, ok := key.Signatures[originUserID]; !ok {
```
kegsay, line 301: There's an awful lot of boilerplate around ensuring maps exist. It makes it hard to see the actually useful bits here. Perhaps we can have custom `New` constructors or something to ease this somewhat, or more functions on the type e.g
```go
func (k *gomatrixserverlib.CrossSigningKey) SetSignature(userID, keyID, signature) { ... }
```

keyserver/internal/cross_signing.go
```diff
@@ -0,0 +1,333 @@
...
+
+			for originUserID, forOrigin := range sigs {
+				for originKeyID, signature := range forOrigin {
+					switch {
+					case req.UserID != "" && originUserID == req.UserID:
+						// Include signatures that we created
+						appendSignature(originUserID, originKeyID, signature)
+					case originUserID == userID:
+						// Include signatures that were created by the person whose key
+						// we are processing
```
kegsay, line 315: Right, this is something we should make clearer. We have a lot of user IDs flying around, and it isn't clear that `userID` means "the person whose key we are prcessing".

keyserver/storage/postgres/cross_signing_keys_table.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/matrix-org/dendrite/keyserver/storage/tables"
+	"github.com/matrix-org/gomatrixserverlib"
+)
+
+var crossSigningKeysSchema = `
+CREATE TABLE IF NOT EXISTS keyserver_cross_signing_keys (
+    user_id TEXT NOT NULL,
+	key_type TEXT NOT NULL,
+	key_data TEXT NOT NULL,
+	PRIMARY KEY (user_id, key_type)
```
kegsay, line 34: In Dendrite fashion, we should probably enumify the `key_type`.

keyserver/storage/postgres/cross_signing_keys_table.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	defer internal.CloseAndLogIfError(ctx, rows, "selectCrossSigningKeysForUserStmt: rows.close() failed")
+	r = api.CrossSigningKeyMap{}
+	for rows.Next() {
+		var keyType gomatrixserverlib.CrossSigningKeyPurpose
+		var keyData gomatrixserverlib.Base64Bytes
+		if err := rows.Scan(&keyType, &keyData); err != nil {
+			return nil, err
+		}
+		r[keyType] = keyData
```
kegsay, line 85: We don't need to `make()` first?

keyserver/storage/postgres/cross_signing_keys_table.go
```diff
@@ -0,0 +1,97 @@
...
+	key_data TEXT NOT NULL,
+	PRIMARY KEY (user_id, key_type)
+);
+`
+
+const selectCrossSigningKeysForUserSQL = "" +
+	"SELECT key_type, key_data FROM keyserver_cross_signing_keys" +
+	" WHERE user_id = $1"
+
+const insertCrossSigningKeysForUserSQL = "" +
```
kegsay, line 42: Technically `upsert`.

keyserver/storage/postgres/cross_signing_keys_table.go
```diff
@@ -0,0 +1,97 @@
...
+
+func NewPostgresCrossSigningKeysTable(db *sql.DB) (tables.CrossSigningKeys, error) {
+	s := &crossSigningKeysStatements{
+		db: db,
+	}
+	_, err := db.Exec(crossSigningKeysSchema)
+	if err != nil {
+		return nil, err
+	}
+	if s.selectCrossSigningKeysForUserStmt, err = db.Prepare(selectCrossSigningKeysForUserSQL); err != nil {
```
kegsay, line 61: You may find [this](https://github.com/matrix-org/dendrite/commit/ed4097825bc65f2332bcdc975ed201841221ff7c#diff-422d0daadceb3c7a9cbb103effe3ff6bbae6aabdca854ad938eaf3ba0e033bc3R80) easier.

keyserver/storage/postgres/cross_signing_sigs_table.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+var crossSigningSigsSchema = `
+CREATE TABLE IF NOT EXISTS keyserver_cross_signing_sigs (
+    origin_user_id TEXT NOT NULL,
+	origin_key_id TEXT NOT NULL,
+	target_user_id TEXT NOT NULL,
+	target_key_id TEXT NOT NULL,
+	signature TEXT NOT NULL,
+	PRIMARY KEY (origin_user_id, target_user_id, target_key_id)
```
kegsay, line 36: It's unclear why we should be uniquing on the target key ID and not both the origin and target?

keyserver/storage/postgres/cross_signing_sigs_table.go
```diff
@@ -0,0 +1,106 @@
...
+);
+`
+
+const selectCrossSigningSigsForTargetSQL = "" +
+	"SELECT origin_user_id, origin_key_id, signature FROM keyserver_cross_signing_sigs" +
+	" WHERE target_user_id = $1 AND target_key_id = $2"
+
+const insertCrossSigningSigsForTargetSQL = "" +
```
kegsay, line 44: Technically upsert.

kegsay (approved):
Numerous small bits but overall LGTM

keyserver/internal/internal.go
neilalexander: Because of [this](https://github.com/matrix-org/gomatrixserverlib/blob/master/crosssigning.go#L57). If there's a `device_id` field in the JSON then we actually unmarshal into a `gomatrixserverlib.DeviceKeys` (signing a device), otherwise we unmarshal into a `gomatrixserverlib.CrossSigningKey` (signing a key).

keyserver/internal/cross_signing.go
neilalexander: It does rather :-(

kegsay: https://github.com/matrix-org/gomatrixserverlib/issues/182#issuecomment-891082647

PR #1969, opened 2021-08-10
Add type field to DeviceMessage

neilalexander:
This should make it easier to expand device updates for cross-signing.

### DIFF
--- a/keyserver/api/api.go
+++ b/keyserver/api/api.go
@@ -53,9 +53,17 @@ func (k *KeyError) Error() string {
 	return k.Err
 }
 
+type DeviceMessageType int
+
+const (
+	TypeDeviceKeyUpdate DeviceMessageType = iota
+	TypeCrossSigningUpdate
+)
+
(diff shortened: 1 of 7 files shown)


keyserver/api/api.go
```diff
@@ -53,9 +53,17 @@ func (k *KeyError) Error() string {
...
+
+const (
+	TypeDeviceKeyUpdate DeviceMessageType = iota
+	TypeCrossSigningUpdate
+)
+
 // DeviceMessage represents the message produced into Kafka by the key server.
 type DeviceMessage struct {
-	DeviceKeys
+	Type        DeviceMessageType `json:"Type,omitempty"`
```
kegsay, line 65: Capital T?

neilalexander: Pretty much because none of the fields had JSON key name annotations before, so they were capitalised as the field names were written. This is just in-keeping with that.

kegsay (approved):

PR #1974, opened 2021-08-13
Cross-signing fixes, notifications via sync, federation

neilalexander:


eduserver/api/types.go
```diff
@@ -0,0 +1,47 @@
+package api
```
S7evinK, line 1: Missing Copyright :)

keyserver/consumers/cross_signing.go
```diff
@@ -0,0 +1,94 @@
+package consumers
```
S7evinK, line 1: Also missing copyright. :P

S7evinK (commented):
Since I've not yet an idea about signing keys (unfortunately).. Looks good code wise. :)

federationapi/routing/send.go
```diff
@@ -502,6 +502,22 @@ func (t *txnReq) processEDUs(ctx context.Context) {
 					}
 				}
 			}
+		case eduserverAPI.MSigningKeyUpdate:
+			var updatePayload eduserverAPI.CrossSigningKeyUpdate
+			if err := json.Unmarshal(e.Content, &updatePayload); err != nil {
+				util.GetLogger(ctx).WithError(err).WithFields(logrus.Fields{
+					"user_id": updatePayload.UserID,
+				}).Error("Failed to send signing key update to edu server")
```
kegsay, line 510: More like failed to unmarshal?

keyserver/consumers/cross_signing.go
```diff
@@ -0,0 +1,108 @@
...
+	process *process.ProcessContext,
+	cfg *config.Dendrite,
+	kafkaConsumer sarama.Consumer,
+	keyDB storage.Database,
+	keyAPI api.KeyInternalAPI,
+) *OutputCrossSigningKeyUpdateConsumer {
+	consumer := internal.ContinualConsumer{
+		Process:        process,
+		ComponentName:  "keyserver/crosssigning",
+		Topic:          cfg.Global.Kafka.TopicFor(config.TopicOutputKeyChangeEvent),
```
kegsay, line 49: I find this confusing. This is the keyserver consuming this topic, but it's unclear what produces on this topic. I would expect the keyserver...? So we're producing and consuming from the same component?

federationsender/consumers/keychange.go
```diff
@@ -133,6 +146,49 @@ func (t *KeyChangeConsumer) onMessage(msg *sarama.ConsumerMessage) error {
 	return t.queues.SendEDU(edu, t.serverName, destinations)
 }
 
+func (t *KeyChangeConsumer) onCrossSigningMessage(m api.DeviceMessage) error {
+	output := m.CrossSigningKeyUpdate
+	_, host, err := gomatrixserverlib.SplitID('@', output.UserID)
+	if err != nil {
+		logrus.WithError(err).Errorf("eduserver output log: user ID parse failure")
```
kegsay, line 153: I'd rather log the consumer here not the producer, as it's a failure in the consumer code. E.g `fedsender:`

federationsender/consumers/keychange.go
```diff
@@ -133,6 +146,49 @@ func (t *KeyChangeConsumer) onMessage(msg *sarama.ConsumerMessage) error {
...
+		return nil
+	}
+
+	// Pack the EDU and marshal it
+	edu := &gomatrixserverlib.EDU{
+		Type:   eduserverAPI.MSigningKeyUpdate,
+		Origin: string(t.serverName),
+	}
+	if edu.Content, err = json.Marshal(output); err != nil {
+		return err
```
kegsay, line 185: Why do we return an `err` here but not elsewhere? Returning an error will cause this component to panic iirc which isn't what we want I think.

keyserver/internal/cross_signing.go
```diff
@@ -45,6 +46,41 @@ func sanityCheckKey(key gomatrixserverlib.CrossSigningKey, userID string, purpos
...
+		case "curve25519":
+			if len(keyData) != curve25519.PointSize {
+				return fmt.Errorf("curve25519 key is not the correct length")
+			}
+		default:
+			// We can't enforce the key length to be correct for an
+			// algorithm that we don't recognise, so instead we'll
+			// just make sure that it isn't incredibly excessive.
+			if len(keyData) > 4096 {
+				return fmt.Errorf("unknown key type is too long")
```
kegsay, line 63: Log the length maybe?

keyserver/internal/cross_signing.go
```diff
@@ -222,6 +217,28 @@ func (a *KeyInternalAPI) PerformUploadDeviceKeys(ctx context.Context, req *api.P
...
+			UserID: req.UserID,
+		}
+		if mk, ok := byPurpose[gomatrixserverlib.CrossSigningKeyPurposeMaster]; ok {
+			update.MasterKey = &mk
+		}
+		if ssk, ok := byPurpose[gomatrixserverlib.CrossSigningKeyPurposeSelfSigning]; ok {
+			update.SelfSigningKey = &ssk
+		}
+		if update.MasterKey == nil && update.SelfSigningKey == nil {
+			return
```
kegsay, line 233: So it's not an error to have a master key but no self-signing key or vice versa? Currently we will produce notifications if:
 - master key is blank but self-signing key isn't
 - self-signing key is blank but master key isn't
 - master key and self-signing key are not blank

keyserver/internal/cross_signing.go
```diff
@@ -290,10 +307,25 @@ func (a *KeyInternalAPI) PerformUploadDeviceSignatures(ctx context.Context, req
...
+				}
+				return
+			}
+		}
+	}
 }
 
 func (a *KeyInternalAPI) processSelfSignatures(
-	ctx context.Context, _ string, queryRes *api.QueryKeysResponse,
+	ctx context.Context, _ string,
```
kegsay, line 328: Remove `_ string` maybe?

keyserver/consumers/cross_signing.go
neilalexander: I've added some comments to this:
```
	// The keyserver both produces and consumes on the TopicOutputKeyChangeEvent
	// topic. We will only produce events where the UserID matches our server name,
	// and we will only consume events where the UserID does NOT match our server
	// name (because the update came from a remote server).
```

neilalexander: Largely because I felt it was easier to reason about the code if it was all using the same paths, and we have to consume that topic anyway because of the device list updates, so might as well reuse it.

keyserver/internal/cross_signing.go
neilalexander: Yes, as a notification that contains only one or other key will just update that key, rather than overwriting both.

Issue #1963, opened 2021-08-05
Customize Password Requirements


Tsangares:
Currently I do not see any password configuration parameters for new users registering to the dendrite server. On the synapse server there was options on the strength. Ultimately, I think if its a strong 8 character password I think that's enough.

I am trying to onboard friends and family to use matrix over signal or whatsapp when possible and having a crazy intense password will limit the ability for onboarding new users. 

The feature could be simple, in the yaml config just parametrize each of the password requirements and allow boolean toggles on and off and a int field for the minimum password length.

kegsay:
Basically this is just adding https://github.com/matrix-org/dendrite/blob/1ed732cc783e079feeaf637e23120b027fb7d70b/clientapi/routing/register.go#L60 into the config in `client_api`.

dmizelle:
hey @kegsay and @Tsangares -- I put in a PR to try and take a swing at this. This is my first PR to dendrite, so its totally possible to have botched something. Let me know what you think. :tada:

Azanul:
According to [this comment](https://github.com/matrix-org/dendrite/pull/1977#issuecomment-1337728101), this issue isn't required anymore.

Frey0-0:
Hello @Tsangares @kegsay is thia issue still open? If yes then I would like to work on it :) I am a new contributor to dendrite

Tsangares:
Ask @kegsay what they meant about "sidecar" feature.

TruncatedDinoSour:
+1

PR #1977, opened 2021-08-13
Add Password Complexity Configuration

dmizelle:
A potential solution to #1963.

This commit does the following:

1. Moves the values for minimum and maximum password length into the
ClientAPI configuration struct.
2. Introduces a new struct representing the password complexity
requirements defined in dendrite-config.yml, with four options. Defaults
are compatible with what users probably expect out of synapse.
  * Minimum length, default of 8
  * Maximum length, default of 512
  * Minimum number of symbols, default of 0
  * Requiring mixed case toggle, default of false
3. Adds tests for the logic of validating passwords.

Signed-off-by: Devon Mizelle <dev@devon.so>

### Pull Request Checklist

### DIFF
--- a/clientapi/routing/register.go
+++ b/clientapi/routing/register.go
@@ -232,7 +232,7 @@ func validatePassword(password string, cfg config.PasswordRequirements) *util.JS
 			Code: http.StatusBadRequest,
 			JSON: jsonerror.WeakPassword(fmt.Sprintf("'password' >%d characters", cfg.MaxPasswordLength)),
 		}
-	} else if len(password) >= 0 && len(password) < cfg.MinPasswordLength {
+	} else if len(password) > 0 && len(password) < cfg.MinPasswordLength {
 		return &util.JSONResponse{
 			Code: http.StatusBadRequest,
 			JSON: jsonerror.WeakPassword(fmt.Sprintf("password too weak: min %d chars", cfg.MinPasswordLength)),
@@ -541,9 +541,9 @@ func Register(
 		if resErr = validateUsername(r.Username); resErr != nil {
 			return *resErr
 		}
-	}
-	if resErr = validatePassword(r.Password, cfg.PasswordRequirements); resErr != nil {
-		return *resErr
+		if resErr = validatePassword(r.Password, cfg.PasswordRequirements); resErr != nil {
+			return *resErr
+		}
 	}
 
 	logger := util.GetLogger(req.Context())
--- a/clientapi/routing/register.go
+++ b/clientapi/routing/register.go
@@ -57,8 +57,6 @@ var (
 )
 
 const (
-	minPasswordLength = 8   // http://matrix.org/docs/spec/client_server/r0.2.0.html#password-based
-	maxPasswordLength = 512 // https://github.com/matrix-org/synapse/blob/v0.20.0/synapse/rest/client/v2_alpha/register.py#L161
 	maxUsernameLength = 254 // http://matrix.org/speculator/spec/HEAD/intro.html#user-identifiers TODO account for domain
 	sessionIDLength   = 24
 )
@@ -111,6 +109,10 @@ var (
 	// sessions stores the completed flow stages for all sessions. Referenced using their sessionID.
 	sessions           = newSessionsDict()
 	validUsernameRegex = regexp.MustCompile(`^[0-9a-z_\-=./]+$`)
+
+	passwordSymbols   = regexp.MustCompile(`[^0-9a-zA-Z]`)
+	passwordUppercase = regexp.MustCompile(`[A-Z]`)
+	passwordLowercase = regexp.MustCompile(`[a-z]`)
 )
 
 // registerRequest represents the submitted registration request.
@@ -225,17 +227,38 @@ func validateApplicationServiceUsername(username string) *util.JSONResponse {
 }
 
 // validatePassword returns an error response if the password is invalid
-func validatePassword(password string) *util.JSONResponse {
+func validatePassword(password string, cfg config.PasswordRequirements) *util.JSONResponse {
 	// https://github.com/matrix-org/synapse/blob/v0.20.0/synapse/rest/client/v2_alpha/register.py#L161
-	if len(password) > maxPasswordLength {
+	if len(password) > cfg.MaxPasswordLength {
 		return &util.JSONResponse{
 			Code: http.StatusBadRequest,
-			JSON: jsonerror.BadJSON(fmt.Sprintf("'password' >%d characters", maxPasswordLength)),
+			JSON: jsonerror.WeakPassword(fmt.Sprintf("'password' >%d characters", cfg.MaxPasswordLength)),
 		}
-	} else if len(password) > 0 && len(password) < minPasswordLength {
+	} else if len(password) > 0 && len(password) < cfg.MinPasswordLength {
 		return &util.JSONResponse{
 			Code: http.StatusBadRequest,
-			JSON: jsonerror.WeakPassword(fmt.Sprintf("password too weak: min %d chars", minPasswordLength)),
+			JSON: jsonerror.WeakPassword(fmt.Sprintf("password too weak: min %d chars", cfg.MinPasswordLength)),
+		}
+	}
+
+	if cfg.MinNumberSymbols > 0 {
+		matches := passwordSymbols.FindAllStringIndex(password, -1)
+		if len(matches) < cfg.MinNumberSymbols {
+			return &util.JSONResponse{
+				Code: http.StatusBadRequest,
+				JSON: jsonerror.WeakPassword(fmt.Sprintf("password too weak: minimum %d symbols", cfg.MinNumberSymbols)),
+			}
+		}
+	}
+
+	if cfg.RequireMixedCase {
+		lowercase := passwordLowercase.FindAllStringIndex(password, -1)
+		uppercase := passwordUppercase.FindAllStringIndex(password, -1)
+		if len(lowercase) == 0 || len(uppercase) == 0 {
+			return &util.JSONResponse{
+				Code: http.StatusBadRequest,
+				JSON: jsonerror.WeakPassword("password must have uppercase and lowercase letters"),
+			}
 		}
 	}
 	return nil
@@ -511,7 +534,7 @@ func Register(
 			return *resErr
 		}
 	}
-	if resErr = validatePassword(r.Password); resErr != nil {
+	if resErr = validatePassword(r.Password, cfg.PasswordRequirements); resErr != nil {
 		return *resErr
 	}
 
@@ -935,7 +958,7 @@ func RegisterAvailable(
 	}
 }
 
-func handleSharedSecretRegistration(userAPI userapi.UserInternalAPI, sr *SharedSecretRegistration, req *http.Request) util.JSONResponse {
+func handleSharedSecretRegistration(userAPI userapi.UserInternalAPI, sr *SharedSecretRegistration, passwordRequirements config.PasswordRequirements, req *http.Request) util.JSONResponse {
 	ssrr, err := NewSharedSecretRegistrationRequest(req.Body)
 	if err != nil {
 		return util.JSONResponse{
@@ -959,7 +982,7 @@ func handleSharedSecretRegistration(userAPI userapi.UserInternalAPI, sr *SharedS
 	if resErr := validateUsername(ssrr.User); resErr != nil {
 		return *resErr
 	}
-	if resErr := validatePassword(ssrr.Password); resErr != nil {
+	if resErr := validatePassword(ssrr.Password, passwordRequirements); resErr != nil {
 		return *resErr
 	}
 	deviceID := "shared_secret_registration"
--- a/clientapi/routing/register.go
+++ b/clientapi/routing/register.go
@@ -28,6 +28,7 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"unicode"
 
 	"github.com/matrix-org/dendrite/internal/eventutil"
 	"github.com/matrix-org/dendrite/setup/config"
@@ -109,10 +110,6 @@ var (
 	// sessions stores the completed flow stages for all sessions. Referenced using their sessionID.
 	sessions           = newSessionsDict()
 	validUsernameRegex = regexp.MustCompile(`^[0-9a-z_\-=./]+$`)
-
-	passwordSymbols   = regexp.MustCompile(`[^0-9a-zA-Z]`)
-	passwordUppercase = regexp.MustCompile(`[A-Z]`)
-	passwordLowercase = regexp.MustCompile(`[a-z]`)
 )
 
 // registerRequest represents the submitted registration request.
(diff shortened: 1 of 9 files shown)


clientapi/routing/register.go
```diff
@@ -935,7 +958,7 @@ func RegisterAvailable(
 	}
 }
 
-func handleSharedSecretRegistration(userAPI userapi.UserInternalAPI, sr *SharedSecretRegistration, req *http.Request) util.JSONResponse {
+func handleSharedSecretRegistration(userAPI userapi.UserInternalAPI, sr *SharedSecretRegistration, passwordRequirements config.PasswordRequirements, req *http.Request) util.JSONResponse {
```
dmizelle, line 961: I'm a little concerned here. I'm not really sure if this shared secret registration should share the same requirements as user passwords. I'm happy to have it do the defaults instead (8 minlength & 512minlength and nothing else) regardless of what the user specifies in the config.

clientapi/routing/register_test.go
```diff
@@ -208,3 +212,123 @@ func TestValidationOfApplicationServices(t *testing.T) {
...
+		MinNumberSymbols:  5,
+	}
+	var testCases = []struct {
+		name     string
+		config   config.PasswordRequirements
+		password string
+		expected *util.JSONResponse
+	}{
+		{
+			"default reject too short",
```
S7evinK, line 234: Minor nitpick: the tests are "hard" to read, without the fields specified.

clientapi/routing/register.go
S7evinK: I think it's fine to use the same requirements, as shared secret registration isn't any different to a normal registration.

S7evinK (approved):
LGTM.

clientapi/routing/register_test.go
neilalexander: Second this — can you please name the fields in the values?

clientapi/routing/register.go
```diff
@@ -225,17 +227,38 @@ func validateApplicationServiceUsername(username string) *util.JSONResponse {
...
 	// https://github.com/matrix-org/synapse/blob/v0.20.0/synapse/rest/client/v2_alpha/register.py#L161
-	if len(password) > maxPasswordLength {
+	if len(password) > cfg.MaxPasswordLength {
 		return &util.JSONResponse{
 			Code: http.StatusBadRequest,
-			JSON: jsonerror.BadJSON(fmt.Sprintf("'password' >%d characters", maxPasswordLength)),
+			JSON: jsonerror.WeakPassword(fmt.Sprintf("'password' >%d characters", cfg.MaxPasswordLength)),
 		}
-	} else if len(password) > 0 && len(password) < minPasswordLength {
+	} else if len(password) > 0 && len(password) < cfg.MinPasswordLength {
```
kegsay, line 237: This fails when `len(password) == 0` as it'll fallthrough.

clientapi/routing/register.go
```diff
@@ -225,17 +227,38 @@ func validateApplicationServiceUsername(username string) *util.JSONResponse {
...
+			JSON: jsonerror.WeakPassword(fmt.Sprintf("password too weak: min %d chars", cfg.MinPasswordLength)),
+		}
+	}
+
+	if cfg.MinNumberSymbols > 0 {
+		matches := passwordSymbols.FindAllStringIndex(password, -1)
+		if len(matches) < cfg.MinNumberSymbols {
+			return &util.JSONResponse{
+				Code: http.StatusBadRequest,
+				JSON: jsonerror.WeakPassword(fmt.Sprintf("password too weak: minimum %d symbols", cfg.MinNumberSymbols)),
```
kegsay, line 249: "Symbols" is a bit of a stretch when we are really talking about non-alphanumerics. E.g `ë` will be treated as a symbol which doesn't seem right.

clientapi/routing/register.go
kegsay: Agreed, let's have a single consistent ruleset for passwords.

clientapi/routing/register_test.go
```diff
@@ -208,3 +212,123 @@ func TestValidationOfApplicationServices(t *testing.T) {
...
+			&util.JSONResponse{
+				Code: http.StatusBadRequest,
+				JSON: jsonerror.WeakPassword(fmt.Sprintf("password too weak: min %d chars", defaults.MinPasswordLength)),
+			},
+		},
+		{
+			"default reject too long",
+			*defaults,
+			// len 600
+			"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
```
kegsay, line 246: `strings.Repeat("a", 600)` would be clearer.

clientapi/routing/register_test.go
```diff
@@ -208,3 +212,123 @@ func TestValidationOfApplicationServices(t *testing.T) {
...
+				t.Errorf("expected password to be validated. got error: %s", response.JSON.(*jsonerror.MatrixError).Err)
+			} else if test.expected != nil && response == nil {
+				t.Errorf("expected password to fail, but was validated")
+			} else if test.expected != nil && test.expected.Code != response.Code {
+				t.Errorf("expected error code %d, got %d", test.expected.Code, response.Code)
+			} else if test.expected != nil && response != nil {
+				matrixError := response.JSON.(*jsonerror.MatrixError)
+				expectedError := test.expected.JSON.(*jsonerror.MatrixError)
+
+				if expectedError.Err != matrixError.Err {
```
kegsay, line 324: Please don't do this. Asserting error strings is really a pain, we have error codes for a reason, use that instead.

clientapi/routing/register_test.go
```diff
@@ -208,3 +212,123 @@ func TestValidationOfApplicationServices(t *testing.T) {
...
+			"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
+			&util.JSONResponse{
+				Code: http.StatusBadRequest,
+				JSON: jsonerror.WeakPassword(fmt.Sprintf("'password' >%d characters", defaults.MaxPasswordLength)),
+			},
+		},
+		{
+			"default allow long enough",
+			*defaults,
+			"thisisalongenoughpassword",
```
kegsay, line 255: This is brittle. If we change the default length of passwords then this test will fail because we're assuming the values. I see little benefit in testing this. By all means, check that:
```go
defaults := &config.PasswordRequirements{}
defaults.Defaults()
```
sets some sensible values (e.g min > 0 and max > min) but try to not bake in assumptions about the values.

clientapi/routing/register_test.go
```diff
@@ -208,3 +212,123 @@ func TestValidationOfApplicationServices(t *testing.T) {
...
+				Code: http.StatusBadRequest,
+				JSON: jsonerror.WeakPassword("password must have uppercase and lowercase letters"),
+			},
+		},
+		{
+			"custom settings allow",
+			*custom,
+			"$0me_$up3r_$trong_P@ass",
+			nil,
+		},
```
kegsay, line 306: Add a test for an empty password please.

kegsay (changes requested):

dmizelle:
👋 thanks for the review y'all. I'll work through these later today.

clientapi/routing/register_test.go
dmizelle: Does something like this work?

```
                        "default reject too short", // test name
                        *defaults,                  // password requirements
                        "foobar",                   // password value to test
                        &util.JSONResponse{ // expected response
                                Code: http.StatusBadRequest,
                                JSON: jsonerror.WeakPassword(fmt.Sprintf("password too weak: min %d chars", defaults.MinPasswordLength)),
                        },
```

clientapi/routing/register.go
dmizelle: oh my, good catch. oops.

clientapi/routing/register_test.go
dmizelle: removed this test. added a test in `setup/config/config_clientapi_test.go` to make sure the values are set correctly.

clientapi/routing/register_test.go
dmizelle: what, I shouldn't scream in my tests? j/k :)

clientapi/routing/register_test.go
dmizelle: oh! now i understand. okay, will do.

clientapi/routing/register.go
dmizelle: There goes my anglophone-centric way of thinking here. :stuck_out_tongue:  I actually changed up this logic a bit to try and take into account non-latin characters / characters with symbols. Will mention you in the line where i switched it up.

clientapi/routing/register.go
```diff
@@ -225,17 +224,49 @@ func validateApplicationServiceUsername(username string) *util.JSONResponse {
...
+
+	var (
+		symbolCount    = 0
+		uppercaseCount = 0
+		lowercaseCount = 0
+	)
+
+	for _, letter := range password {
+		switch {
+		case unicode.IsSymbol(letter) || unicode.IsPunct(letter):
```
dmizelle, line 250: @kegsay -- ended up changing this to use the `unicode` package instead, as I feel like its less anglocentric here. would love to get your thoughts.

dmizelle:
Looks like I broke sytests on my branch. Will take a look later tonight.

clientapi/routing/register.go
dmizelle: hey, so i cant actually deny empty passwords here. the most recent failure for sytest on buildkite comes from the AS tests:

https://buildkite.com/matrix-dot-org/dendrite/builds/5406

It looks like that in synapse itself, we "allow" empty passwords here:

https://github.com/matrix-org/synapse/blob/642a42eddece60afbbd5e5a6659fa9b939238b4a/synapse/handlers/password_policy.py#L52-L60

From what I understand (and I'm still learning, so excuse any dumb assumptions :) ) is that Application Services actually create users (specifically around IRC bridging) that have empty passwords.

I _think_ we should be denying account registrations with empty passwords, but that seems like it might get a little weird with application services.

I could edit the sytest tests to send a password along, but I'm sort of afraid that by merging this in with checking with length 0 passwords, that it could break application services in the future that might do this. :\

I hope I'm understanding correctly and that I haven't word salad'd :laughing:

dmizelle:
(put a reply in one of the review comments above around AS and creating users)

clientapi/routing/register.go
kegsay: We should allow empty passwords if it is an application service doing the registration, else we should reject.

dmizelle: @kegsay i keep coming back to trying to get this one to pass, but my gorilla brain can't really figure out how to differentiate between registration attempts that are application services and ones that are not.

is there a header or something that is set?

i want to figure out a way how to do this without breaking backwards compatibility (saying "hey, you now need to set XYZ thing to say you are an application service") but im sort of barking up weird trees.

S7evinK: [r.Type == authtypes.LoginTypeApplicationService](https://github.com/matrix-org/dendrite/blob/eadbfa7eea28beb5d706650ff9a158b68f37f521/clientapi/routing/register.go#L525) which could be passed to `validatePassword`. (Tested and now only `GET /register yields a set of flows` is still failing)

clientapi/routing/register.go
kegsay: Seems reasonable.

clientapi/routing/register.go
dmizelle: Oh hell yeah! Okay, I'm coming off a busy weekend, but I'll push a new commit soon.

Thanks a ton for the tip! <3

danpe:
Amazing PR @dmizelle! Can we add another option to **restrict sequential and repetitive characters (e.g. 12345 or aaaaaa).** 
We can use the following regex for it: `/([a-zA-Z0-9])\1{2,}/`

kegsay:
We recently updated our [contributing guidelines](https://github.com/matrix-org/dendrite/blob/main/docs/CONTRIBUTING.md#contribution-types). This PR is being closed because it isn't a feature we want to maintain going forwards. If there is a serious need for this, you can have a sidecar process to enforce password length and complexity.

TruncatedDinoSour:
> We recently updated our [contributing guidelines](https://github.com/matrix-org/dendrite/blob/main/docs/CONTRIBUTING.md#contribution-types). This PR is being closed because it isn't a feature we want to maintain going forwards. If there is a serious need for this, you can have a sidecar process to enforce password length and complexity.

what even is a 'sidecar' process lmao

PR #1978, opened 2021-08-14
Update golangci-lint

S7evinK:
Updates golangci-lint and fixes new found issues.
Not exactly sure why it is complaining about `quic`.

### Pull Request Checklist

### DIFF
--- a/cmd/dendrite-demo-yggdrasil/yggconn/node.go
+++ b/cmd/dendrite-demo-yggdrasil/yggconn/node.go
@@ -26,7 +26,7 @@ import (
 	"os"
 	"strings"
 
-	"github.com/lucas-clemente/quic-go" //nolint: typecheck
+	"github.com/lucas-clemente/quic-go"
 	"github.com/matrix-org/gomatrixserverlib"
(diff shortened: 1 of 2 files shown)


cmd/dendrite-demo-yggdrasil/yggconn/node.go
```diff
@@ -26,7 +26,7 @@ import (
 	"os"
 	"strings"
 
-	"github.com/lucas-clemente/quic-go"
```
kegsay, line 29: I think the problem is it doesn't statically know that `quic-go` is `quic`. So:
```go
quic "github.com/lucas-clemente/quic-go"
```
should fix it.

kegsay (approved):
Otherwise LGTM

cmd/dendrite-demo-yggdrasil/yggconn/node.go
S7evinK: Also tried that before, Goland complained about redundant alias.
`typecheck` is even disabled in the `golangci-lint` config.

Somehow, after reverting that commit, the linter seems to be happy again.

PR #1979, opened 2021-08-18
Delete device keys/signatures from key server when deleting devices

neilalexander:
This stops us from getting ghost keys/signatures in the key server after the user has deleted their device from the Security tab in Element Web.


kegsay (approved):
Weird that `Invited user can reject invite over federation for empty room` is flakey though.

PR #1982, opened 2021-08-22
Add missing HTTP mode for userapi

S7evinK:
### Pull Request Checklist


kegsay (commented):
:S I'm confused. How has this ever worked in HTTP mode then?

S7evinK:
Exactly my thought.
Probably because [UserAPIClient](https://github.com/matrix-org/dendrite/blob/329c9e671c763ce3f7adc8c734246eb53b47df94/setup/base.go#L232) (which uses `userapi/inthttp/client.go`) is only used when running in polylith mode?

neilalexander:
Avoided the race by adding a flag to the device creation that avoids sending device list updates. It isn't clear that we need to bother sending device list updates for appservices.

PR #2025, opened 2021-10-21
sytest: update blacklist and whitelist

kegsay:
Should make CI happy again


kegsay:
Merging this as this diff is fine, it's sytest that isn't picking up the blacklist.

PR #2045, opened 2021-11-11
Restricted joins

neilalexander:
This PR adds support for restricted joins, both locally and in `/make_join`.

### DIFF
--- a/federationapi/routing/join.go
+++ b/federationapi/routing/join.go
@@ -144,17 +144,22 @@ func MakeJoin(
 	provider := gomatrixserverlib.NewAuthEvents(stateEvents)
 
 	// Check the join rules. If it's a restricted join then there are special rules.
-	var joinRuleEvent *gomatrixserverlib.Event
-	var joinRules gomatrixserverlib.JoinRuleContent
-	if joinRuleEvent, err = provider.JoinRules(); err != nil {
+	joinRules := gomatrixserverlib.JoinRuleContent{
+		JoinRule: gomatrixserverlib.Public, // Default join rule if not specified.
+	}
+	joinRuleEvent, err := provider.JoinRules()
+	if err != nil {
 		return util.JSONResponse{
 			Code: http.StatusNotFound,
-			JSON: jsonerror.NotFound("Room join rules do not exist"),
+			JSON: jsonerror.NotFound("Failed to retrieve join rules"),
 		}
-	} else if err = json.Unmarshal(joinRuleEvent.Content(), &joinRules); err != nil {
-		return util.JSONResponse{
-			Code: http.StatusBadRequest,
-			JSON: jsonerror.Unknown("Failed to unmarshal room join rules"),
+	}
+	if joinRuleEvent != nil {
+		if err = json.Unmarshal(joinRuleEvent.Content(), &joinRules); err != nil {
+			return util.JSONResponse{
+				Code: http.StatusBadRequest,
+				JSON: jsonerror.Unknown("Failed to unmarshal room join rules"),
+			}
 		}
 	}
 
--- a/federationapi/routing/join.go
+++ b/federationapi/routing/join.go
@@ -144,9 +144,11 @@ func MakeJoin(
 	provider := gomatrixserverlib.NewAuthEvents(stateEvents)
 
 	// Check the join rules. If it's a restricted join then there are special rules.
-	joinRules := gomatrixserverlib.JoinRuleContent{
-		JoinRule: gomatrixserverlib.Public, // Default join rule if not specified.
-	}
+	// We have to do this in two steps in order to satisfy the Complement tests. The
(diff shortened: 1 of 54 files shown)


roomserver/internal/perform/perform_join.go
```diff
@@ -318,6 +338,137 @@ func (r *Joiner) performJoinRoomByID(
...
+			Msg:  fmt.Sprintf("Unable to retrieve the join rules: %s", err),
+		}
+	}
+
+	// First unmarshal the join rule itself. It might seem strange that this is
+	// a two-step process, but the Complement tests specifically populate the
+	// 'allow' field with a nonsense value that won't unmarshal and therefore
+	// trying to unmarshal a gomatrixserverlib.JoinRuleContent fails entirely.
+	// We need to get the join rule first to check if the room is restricted
+	// though, regardless of what the 'allow' key contains.
```
kegsay, line 360: Do we still need to contort for this?

roomserver/internal/perform/perform_join.go
```diff
@@ -318,6 +338,137 @@ func (r *Joiner) performJoinRoomByID(
...
+			continue
+		}
+		_, domain, err := gomatrixserverlib.SplitID('@', userID)
+		if err != nil || domain != r.ServerName {
+			continue
+		}
+		if powerLevels.UserLevel(userID) < powerLevels.Invite {
+			continue
+		}
+		if err := eb.SetContent(map[string]string{
```
kegsay, line 457: It feels a bit awkward to side-effect success by calling `SetContent`. It would be slightly nicer if we returned the content to set OR an error rather than just an error. At present this code is fine, but if we touch `eb` earlier and don't revert it we risk polluting the `EventBuilder` with partial data across multiple rooms.

federationapi/routing/join.go
```diff
@@ -158,6 +201,173 @@ func MakeJoin(
...
+			continue
+		}
+
+		logrus.Info("Inspecting room ", allowed.RoomID)
+
+		// Now have a look and see if any of the joined users match the
+		// user who has initiated this join.
+		found := false
+		for _, member := range queryRes.JoinEvents {
+			if *member.StateKey == userID {
```
kegsay, line 264: Doesn't this mean that the user is *already joined* to the room and are calling `/make_join` again for some reason? It doesn't make sense why they would do this.

federationapi/routing/join.go
```diff
@@ -158,6 +201,173 @@ func MakeJoin(
...
+			JSON: jsonerror.UnableToGrantJoin("None of the users from this homeserver have the power to invite"),
+		}
+	case ableToAuthoriseJoin && !foundUserInAnyRoom:
+		// We found ourselves in some of the allowed rooms, but none of them
+		// seemed to contain the joining user.
+		return util.JSONResponse{
+			Code: http.StatusForbidden,
+			JSON: jsonerror.Forbidden("You are not joined to any allowed rooms"),
+		}
+	default:
```
kegsay, line 361: I don't like all the error handling happening at the bottom here, as it's hard to unpick when `ableToAuthoriseJoin` and `foundUserInAnyRoom` are set or not set. I understand we need to loop over all members before giving up which is probably why it's not a simple if statement, but it still feels a bit hard to parse since these flags are set far away from where they ultimately get processed.

kegsay (approved):

federationapi/routing/join.go
neilalexander: This is checking the *allowed* rooms, so not the room that the user is trying to join, but their membership in one of the other rooms that should decide if they are allowed or not:
```
	// Let's see if we can validate the user being in
	// any of the allowed rooms.
	for _, allowed := range joinRules.Allow {
```

roomserver/internal/perform/perform_join.go
neilalexander: I'm not really a fan of this either but the concern was that either a) we try checking the auth of the event here and have to pull out a bunch of events just to have to do it again in the input API, or that b) we might miss some other more delicate parts of event auth or soft state checking by trying to reinvent the wheel. I figured it was better to use our tried-and-tested code paths for that, given this is basically a security feature.

roomserver/internal/perform/perform_join.go
neilalexander: With the tests the way they are currently, yes.

federationapi/routing/join.go
neilalexander: It honestly isn't clear to me why `M_UNABLE_TO_AUTHORISE_JOIN` and `M_UNABLE_TO_GRANT_JOIN` need to even be separate error codes. The ultimate result is the same — we have to try another server. This is basically just attempting to return the *right* error code based on what MSC3083 suggests.

neilalexander:
Back to the drawing board.

PR #2048, opened 2021-11-16
Don't persist transaction IDs in the roomserver

neilalexander:
It doesn't appear as if we ever use this information in the roomserver, so it seems pointless to store it there. It's just taking up database I/Os and disk space.

This PR does not remove the transaction ID from the RS input requests or output events because we still need to pass them through the roomserver — the sync API expects to consume the transaction ID for populating the `"transaction_id"` in the `"unsigned"` key of events in the sync response (from memory this is so that clients can do the right thing with local echoes).

The client API also tracks some recent transactions in memory to prevent a client submitting the same thing more than once, but that seems to be largely unrelated to this.


kegsay (commented):
I think this mostly exists so that if the roomserver dies whilst processing an event it can detect if it had processed it before. Some earlier work would only call txn.commit() right at the end of processing to guarantee this. How do we propose to handle this if not this way?

neilalexander:
The event gets written into the database and *then* we take out a row-level lock and open a transaction to run the latest events and membership updaters. If we see that an event has already been persisted and it's an outlier, we won't process it again. If it's a new event, we *want* to process it again because we need to see through the whole process of updating the forward extremities and generating output events, else other components will miss out on the interrupted event. AFAICT, this transaction check is useless and might actually result in us not reprocessing something that we should have done.

kegsay (approved):

PR #2055, opened 2021-11-22
Merge `federationapi`, `federationsender`, `signingkeyserver` components

neilalexander:
This PR merges the `federationsender` and `signingkeyserver` components into the `federationapi` component, as a part of our mission to reduce the number of microservices where they don't necessarily make sense.


build/docker/postgres/create_db.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
 
-for db in userapi_accounts userapi_devices mediaapi syncapi roomserver signingkeyserver keyserver federationsender appservice naffka; do
```
kegsay, line 3: Can we nuke the other DBs please like `signingkeyserver`?

kegsay (approved):
Probably fine if CI is happy. Can we run this on d.n.d as well as a sanity check?

PR #2058, opened 2021-11-24
Don't populate config defaults where it doesn't make sense

neilalexander:
The various `Defaults()` functions are designed to populate sane defaults before we unmarshal the config, so that any obviously missing but useful fields are populated. We also use this for `cmd/generate-config` and for the various P2P demos, which mostly want to run with defaults.

The problem is that we can accidentally populate defaults that aren't that sane for regular monolith/polylith installs. For example, the federation refactoring in matrix-org/dendrite#2055 would not complain about the missing `database` section from the `federation_api` config section because it was being defaulted.

This adds a "generate" flag to the various `Defaults()` functions for specifying if we want a *full* configuration generating, or whether to omit things that we obviously want to fail if they are missing. We definitely don't want to populate things like database connection strings if they are missing as we would rather panic and signal to the user that there's a problem in the config instead.

### DIFF
--- a/build/gobind-pinecone/monolith.go
+++ b/build/gobind-pinecone/monolith.go
@@ -260,7 +260,7 @@ func (m *DendriteMonolith) Start() {
 
 	prefix := hex.EncodeToString(pk)
 	cfg := &config.Dendrite{}
-	cfg.Defaults()
+	cfg.Defaults(true)
 	cfg.Global.ServerName = gomatrixserverlib.ServerName(hex.EncodeToString(pk))
 	cfg.Global.PrivateKey = sk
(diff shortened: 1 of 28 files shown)


kegsay (approved):

kegsay:
But sytest is unhappy:
```


[server]: time="2021-11-24T11:40:09Z" level=error msg="Configuration error: missing config key \"federation_api.database.connection_string\""
--


WARN: Error starting server-0: Unable to start dendrite monolith: Process died without becoming connectable
```

neilalexander:
Looks like it did exactly what we wanted then because matrix-org/sytest#1171 wasn't merged yet :-)

PR #2061, opened 2021-11-28
Fix EDU Typing handling in syncapi and add push_server.push_gateway_tls_insecure_skip_verify_for_testing.

tommie:
Fixes for push notifications; #1842.

Signed-off-by: `Tommie Gannert <tommie@gannert.se>`

### DIFF
--- a/setup/base/base.go
+++ b/setup/base/base.go
@@ -21,7 +21,6 @@ import (
 	"io"
 	"net"
 	"net/http"
-	"net/url"
 	"os"
 	"os/signal"
 	"syscall"
@@ -79,7 +78,6 @@ type BaseDendrite struct {
 	SynapseAdminMux        *mux.Router
 	UseHTTPAPIs            bool
 	apiHttpClient          *http.Client
-	httpClient             *http.Client
 	Cfg                    *config.Dendrite
 	Caches                 *caching.Caches
 	DNSCache               *gomatrixserverlib.DNSCache
@@ -183,13 +181,6 @@ func NewBaseDendrite(cfg *config.Dendrite, componentName string, options ...Base
 			},
 		},
 	}
-	client := http.Client{Timeout: HTTPClientTimeout}
-	if cfg.FederationAPI.Proxy.Enabled {
-		client.Transport = &http.Transport{Proxy: http.ProxyURL(&url.URL{
-			Scheme: cfg.FederationAPI.Proxy.Protocol,
-			Host:   fmt.Sprintf("%s:%d", cfg.FederationAPI.Proxy.Host, cfg.FederationAPI.Proxy.Port),
-		})}
-	}
 
 	// Ideally we would only use SkipClean on routes which we know can allow '/' but due to
 	// https://github.com/gorilla/mux/issues/460 we have to attach this at the top router.
@@ -219,7 +210,6 @@ func NewBaseDendrite(cfg *config.Dendrite, componentName string, options ...Base
 		InternalAPIMux:         mux.NewRouter().SkipClean(true).PathPrefix(httputil.InternalPathPrefix).Subrouter().UseEncodedPath(),
 		SynapseAdminMux:        mux.NewRouter().SkipClean(true).PathPrefix("/_synapse/").Subrouter().UseEncodedPath(),
 		apiHttpClient:          &apiClient,
-		httpClient:             &client,
 	}
 }
 
--- a/setup/config/config_pushserver.go
+++ b/setup/config/config_pushserver.go
@@ -1,0 +1,16 @@
+package config
+
+type PushServer struct {
+	Matrix *Global `yaml:"-"`
+
+	// DisableTLSValidation disables the validation of X.509 TLS certs
+	// on remote Push gateway endpoints. This is not recommended in
+	// production!
+	DisableTLSValidation bool `yaml:"disable_tls_validation"`
+}
+
+func (c *PushServer) Defaults(generate bool) {
+}
+
+func (c *PushServer) Verify(configErrs *ConfigErrors, isMonolith bool) {
+}
--- a/setup/config/config.go
+++ b/setup/config/config.go
@@ -60,6 +60,7 @@ type Dendrite struct {
 	FederationAPI FederationAPI `yaml:"federation_api"`
 	KeyServer     KeyServer     `yaml:"key_server"`
 	MediaAPI      MediaAPI      `yaml:"media_api"`
+	PushServer    PushServer    `yaml:"push_server"`
 	RoomServer    RoomServer    `yaml:"room_server"`
 	SyncAPI       SyncAPI       `yaml:"sync_api"`
 	UserAPI       UserAPI       `yaml:"user_api"`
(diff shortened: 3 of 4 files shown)


tommie:
* FYI: @PiotrKozimor 
* Reviewer: @kegsay

setup/base/base.go
```diff
@@ -183,12 +183,21 @@ func NewBaseDendrite(cfg *config.Dendrite, componentName string, options ...Base
...
+	}
 	if cfg.FederationAPI.Proxy.Enabled {
-		client.Transport = &http.Transport{Proxy: http.ProxyURL(&url.URL{
+		transport.Proxy = http.ProxyURL(&url.URL{
 			Scheme: cfg.FederationAPI.Proxy.Protocol,
 			Host:   fmt.Sprintf("%s:%d", cfg.FederationAPI.Proxy.Host, cfg.FederationAPI.Proxy.Port),
-		})}
+		})
+	}
+	client := http.Client{
```
neilalexander, line 198: I’m not *entirely* sure why we have this HTTP client at any rate — I think anything that previously used it is probably now using `apiHttpClient` instead. Is it really used?

Also this will probably only really do what you expect in monolith mode, ideally the component that is producing the HTTP client calls (I guess in your case, the push server) should be instantiating a HTTP client for the purpose. That said, having a function that returns a suitable HTTP client based on the config (proxy, TLS validation etc) would probably be a useful way to simplify this as well as in other places like the appservice API.

tommie: You're right. It was replaced with `apiHttpClient` 4d ago: https://github.com/matrix-org/dendrite/commit/ec716793eb86e7723ecfaa356e148334bc3921eb#diff-794b74a8e5bab56b4fdac37a8aeb4fbc1d6eaa5218341d75761abad8af1d9729L269. Federation was the only case I know of using it.

So I'm guessing that means `client` should be removed now? (I thought it was neat there was a separate client for external requests, though. You have far less control (and probably higher security requirements) over those requests. So made sense to adjust timeouts and security params separately.)

BTW, the reason I made this a separate PR is because I need to modify Sytest to enable this option before I can post the Pushgateway PR. Here's the end-goal: https://github.com/tommie/dendrite/blob/serving/setup/base.go#L308

> Also this will probably only really do what you expect in monolith mode

Even in polyliths, the base is still shared. Why is it not doing what I'd expect? Calling code here: https://github.com/tommie/dendrite/blob/serving/cmd/dendrite-polylith-multi/personalities/pushserver.go#L25

I don't understand your preferred approach. Even now, the `apiHttpClient` is injected into e.g. Federation API. Are you suggesting doing something different? (It sounds like you want to inject a client factory; an indirection I don't understand the purpose of.)

tommie:
Reading the `apiHttpClient` code, it seems it forces non-TLS and is HTTP/2-only, which means we can't use it for Push gateways.

I think the conclusion from our comments is that I should just add the config option, so I can enable it in Sytest. Then the later code will instantiate its own `http.Client` inside the future `BaseDendrite.PushgatewayHTTPClient()`, linked to earlier.

I've added a commit to drop `httpClient` as a cleanup.

setup/base/base.go
neilalexander: > So I'm guessing that means client should be removed now? 

Well, if there's a good reason for using a shared client like this then I guess we can keep it, but it isn't really clear that a shared client is any better than components just building a `http.Client` with timeout and proxy values that make sense for that specific workload. It might just be clearer to remove it.

Note that `apiHttpClient` is specifically different because it forces H2C instead of plain HTTP, which we use when talking *between* components in polylith mode, but we strictly do not use it to make external HTTP calls.

> BTW, the reason I made this a separate PR is because I need to modify Sytest to enable this option before I can post the Pushgateway PR. Here's the end-goal: https://github.com/tommie/dendrite/blob/serving/setup/base.go#L308

This feels misplaced. The various `FooHTTPClient()` functions here are for returning shaped clients for talking to *internal APIs* so that components can communicate with each other. Surely the only thing talking to push gateways directly is this new push server? In which case why would the `base` package care about the client used?

> Even in polyliths, the base is still shared

Yes, you're right, `NewBaseDendrite` gets called before the polylith personalities are attached.

> I don't understand your preferred approach. Even now, the apiHttpClient is injected into e.g. Federation API. 

I don't think this is true (at least it shouldn't be). The `apiHttpClient` isn't really injected anywhere — it's used to return a HTTP-powered API implementation for a given component's API surface. The federation API makes its own `http.Client` indirectly via the `gomatrixserverlib.FederationClient`, the media API uses a `gomatrixserverlib.Client` etc.

tommie: Ah, thanks for the explanation. Makes sense.

In that case, I think my we/I have three options:

1. Create the global config option here, enable it in Sytest, then submit the Push gateway code with Sytest whitelisting.
1. Create the pushserver-specific config option here, enable it in Sytest, then submit the Push gateway code with Sytest whitelisting.
2. Create the Push gateway code with a pushgateway-specific config option to disable TLS verification, enable it in Sytest, then submit Sytest whitelisting.

Since I'd love for the big PR to pass Sytests rather than do that as a follow-up, I'd prefer the (1) or (2) flows. Given that we won't have a global HTTP client, (2) makes the most sense.

Updated code pushed.

setup/base/base.go
```diff
@@ -183,13 +181,6 @@ func NewBaseDendrite(cfg *config.Dendrite, componentName string, options ...Base
 			},
 		},
 	}
-	client := http.Client{Timeout: HTTPClientTimeout}
-	if cfg.FederationAPI.Proxy.Enabled {
```
kegsay, line 187: Why are you removing this?

setup/config/config_pushserver.go
```diff
@@ -0,0 +1,16 @@
+package config
+
+type PushServer struct {
+	Matrix *Global `yaml:"-"`
+
+	// PushGatewayTLSInsecureSkipVerify controls whether the Push
+	// gateway client should verify X.509 certificates, or allow
+	// anything. Only for testing, with e.g. self-signed certificates.
+	PushGatewayTLSInsecureSkipVerify bool `yaml:"push_gateway_tls_insecure_skip_verify_for_testing"`
```
kegsay, line 9: Please be consistent with other flags. See federation sender which has a similar option: https://github.com/matrix-org/sytest/pull/1172/files#diff-a50fea42e82ac587405319c40b1e270ccab6f6af970d1444bd547822330c5bd9R175

kegsay (changes requested):

setup/base/base.go
tommie: Per discussion with @neilalexander above, this seems unused since a few days ago.

neilalexander:
I've cherry-picked the typing stream position fix and the removal of the unused `http.Client` straight onto `master` now.

I think it's probably a good idea to close this PR and roll the `disable_tls_validation` bit into whichever push notifications branch, and then update matrix-org/sytest#1172 accordingly.

Otherwise, we'll end up with a push server config section on `master` when none of that code is committed yet.

tommie:
Ok. Thanks!

Issue #2059, opened 2021-11-26
Login API should casefold usernames, matching registration API


timmc:
### Background information

- **Dendrite version or git SHA**: v0.5.1
- **Monolith or Polylith?**: monolith
- **SQLite3 or Postgres?**: Postgres
- **Running in Docker?**: no
- **`go version`**: 1.17.3

### Description

User can register with mixed-case names e.g. `Test` but then cannot log in.

### Steps to reproduce

- Register with username `Test`
- Try to log in with username `Test`

The registration code [case-folds the username to lowercase](https://github.com/matrix-org/dendrite/blob/v0.5.1/clientapi/routing/register.go#L899), so the registered username is actually `test`. However, the login code doesn't perform this case-folding, so login fails.

References:

- Spec forbids uppercase letters: https://spec.matrix.org/v1.1/appendices/#user-identifiers -- but homeserver is free to downcase
- Previous issue that requested downcasing in login and registration, although only the registration side was actually done: https://github.com/matrix-org/dendrite/issues/325
- `ParseUsernameParam` may be a good place to make this change, dunno

timmc:
Looks like this is taken care of by that PR. :100:

PR #2065, opened 2021-12-01
Squash username to lowercase at login

BernardZhao:
Signed-off-by: Bernard Zhao <bernard.zhao.us@gmail.com>

This should fix the issue described in #2059.

### DIFF
--- a/clientapi/auth/password.go
+++ b/clientapi/auth/password.go
@@ -17,6 +17,7 @@ package auth
 import (
 	"context"
 	"net/http"
+	"strings"
 
 	"github.com/matrix-org/dendrite/clientapi/jsonerror"
 	"github.com/matrix-org/dendrite/clientapi/userutil"
@@ -48,7 +49,8 @@ func (t *LoginTypePassword) Request() interface{} {
 
 func (t *LoginTypePassword) Login(ctx context.Context, req interface{}) (*Login, *util.JSONResponse) {
 	r := req.(*PasswordRequest)
-	username := r.Username()
+	// Squash username to all lowercase letters
+	username := strings.ToLower(r.Username())
 	if username == "" {
 		return nil, &util.JSONResponse{
 			Code: http.StatusUnauthorized,


spaetz:
Sorry, if this is a dumb question, I am not familiar with the flows of logic on login. Does this take care of all forms of login, e.g. via m.login.token or only the simple user/password one?

BernardZhao:
> Sorry, if this is a dumb question, I am not familiar with the flows of logic on login. Does this take care of all forms of login, e.g. via m.login.token or only the simple user/password one?

I'm not too familiar with it either, but I assume from the comment [here](https://github.com/matrix-org/dendrite/blob/e3679799ea6f6387d36fb1a2f938f8ea9711ad2d/clientapi/auth/user_interactive.go#L37-L39) that they go through this same flow.

clientapi/auth/password.go
```diff
@@ -48,7 +49,8 @@ func (t *LoginTypePassword) Request() interface{} {
 
 func (t *LoginTypePassword) Login(ctx context.Context, req interface{}) (*Login, *util.JSONResponse) {
 	r := req.(*PasswordRequest)
-	username := r.Username()
+	// Squash username to all lowercase letters
+	username := strings.ToLower(r.Username())
```
neilalexander, line 53: I'm not really convinced this is the right thing to do in isolation, as I suspect what would happen is something like this:

1. User goes through the `/register` flow as `Mary`
2. Database entry gets created in `account_accounts` etc as `Mary`
3. User goes through the `/login` flow as `Mary`
4. This line changes `Mary` down to `mary`
5. There's no entry `mary` in `account_accounts` as the query is case-sensitive to avoid a table scan
6. Login fails because the user is not found

neilalexander (changes requested):

clientapi/auth/password.go
neilalexander: Actually, looking at it, we seem to enforce `validUsernameRegex` on registration, so this may not be a problem after all.

neilalexander (approved):
It looks like the tests are happy and closer inspection reveals the registration flow already does the right thing for preventing upper-case characters on registration, so I'll approve this after all.

clientapi/auth/password.go
bones-was-here: The `create-account` binary provided with dendrite, does not enforce the use of all lower case (it also doesn't honour bcrypt_cost in dendrite.yaml).  One user created their account with an upper case letter, which has been in use for some time now and exists in many events with that upper case letter.  Is this going to break federation of rooms?  Is there a way to recover?

bones-was-here: Filed a proper issue since this is merged #2073

PR #2085, opened 2022-01-06
Fix #2084 - incorrect /event_auth response

### LINKED ISSUE #2084: incorrect response to `/_matrix/federation/v1/event_auth`
`GET /_matrix/federation/v1/event_auth/{roomId}/{eventId}` should return the events in the auth chain for `eventId`. Dendrite currently returns the auth chains for *all* of the events in the state of the room.

kegsay:
We now very strictly only return the auth chain for the event given. I've also made it so that we don't load the entire current state (!!!!) for this operation. This will save on allocations / memory spikes when other servers call `/event_auth` on Dendrite, which it sounds like it could be reasonably frequently in busy IRC rooms where auth events are constantly changing. According to vdh:

> Synapse does it whenever it receives an event where we don't already have all the auth_events

So this will be done less frequently than `/get_missing_events` but still frequently in high event loss rooms (this may be particularly important for P2P rooms).

### DIFF
--- a/roomserver/api/query.go
+++ b/roomserver/api/query.go
@@ -226,7 +226,7 @@ type QueryStateAndAuthChainRequest struct {
 	PrevEventIDs []string `json:"prev_event_ids"`
 	// The list of auth events for the event. Used to calculate the auth chain
 	AuthEventIDs []string `json:"auth_event_ids"`
-	// If true, the auth chain events for the authh event IDs given will be fetched only. Prev event IDs are ignored.
+	// If true, the auth chain events for the auth event IDs given will be fetched only. Prev event IDs are ignored.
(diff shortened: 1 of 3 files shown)


roomserver/api/query.go
```diff
@@ -226,6 +226,10 @@ type QueryStateAndAuthChainRequest struct {
 	PrevEventIDs []string `json:"prev_event_ids"`
 	// The list of auth events for the event. Used to calculate the auth chain
 	AuthEventIDs []string `json:"auth_event_ids"`
+	// If true, the auth chain events for the authh event IDs given will be fetched only. Prev event IDs are ignored.
```
neilalexander, line 229: authh

neilalexander (approved):

PR #2086, opened 2022-01-07
NATS JetStream tweaks

neilalexander:
This fixes a few small bugs:

* Using durable named consumers now means that interest-based policies won't be confused when Dendrite goes away (as the ephemeral consumers would have disappeared too)
* Namespaces are now configured properly using the `topic_prefix`, unlike before

If you get an error like `level=fatal msg="Unable to add stream" error="subjects overlap with an existing stream"`, you may need to clear out your JetStream `store_dir` and restart.


kegsay (approved):
Modulo failing CI

Issue #2078, opened 2021-12-23
NGINX Monolith sample config


idi-moe:
https://github.com/matrix-org/dendrite/blob/a47b12dc7d692e0ddd4aaa0801dafc9bb462aad9/docs/nginx/monolith-sample.conf#L26

Should "monolith" in this line not be "localhost"? I changed mine during install.

Forchapeatl:
Thank you for bringing this up.

kegsay:
https://github.com/matrix-org/dendrite/pull/2087

PR #2087, opened 2022-01-11
Update monolith-sample.conf

Forchapeatl:
*[ ] I have  replaced the undefined 'monolith' value  with 'server_name'  (my.hostname.com) value in reference  this [issue](https://github.com/matrix-org/dendrite/issues/2078)

### Pull Request Checklist

Signed-off-by: `Forcha Pearl <forchapearl1@gmail.com.org>`


spaetz:
I applaud the direction of changes, but to be honest, I find "my.hostname.com" also quite vague and prone to misunderstandings. Often, my.hostname.com will be "localhost" rather than a full blown FQDN and my.hostname.com could be the public facing nginx hostname or the dendrite hostname or...

Perhaps a little clearer with something like http://HOSTNAME.OR.IP.OF.DENDRITE_SERVER.COM/_matrix

Forchapeatl:
> I applaud the direction of changes, but to be honest, I find "my.hostname.com" also quite vague and prone to misunderstandings. Often, my.hostname.com will be "localhost" rather than a full blown FQDN and my.hostname.com could be the public facing nginx hostname or the dendrite hostname or...
> 
> Perhaps a little clearer with something like http://HOSTNAME.OR.IP.OF.DENDRITE_SERVER.COM/_matrix

Thank you. I understand , the problem is the server_name  value used in the  [nignix.conf](https://github.com/matrix-org/dendrite/blob/a47b12dc7d692e0ddd4aaa0801dafc9bb462aad9/docs/nginx/monolith-sample.conf#L26) docs was "my.hostname.com" . Or should I just set it to default 'localhost'?

genofire:
Or just `[::1]` (or for legacy users `127.0.0.1`)

idi-moe:
Replacing it with `my.hostname.com` is a _really_ bad idea because this would possibly imply to users that this should be set to their domain (eg. matrix.example.com), which simply won't work with 99% of router setups and will fail to resolve or have the router just accept the packet. It should be replaced with `localhost`, or at the very least add a comment in the config file explaining to change it to localhost in the event that nginx and dendrite are on the same machine. An alternative and more robust fix would be to add:
```
#change IP to location of monolith server
upstream monolith{
    server 127.0.0.1:8008;
}
```

and change the location section to:
```
location /_matrix {
        proxy_pass http://monolith;
    }
```

kegsay (approved):
Thanks for the comments and review on this. I agree that the new style here is a good solution. Merging!

PR #2096, opened 2022-01-19
BREAKING: Remove Partitioned Stream Positions

kegsay:
We don't use Kafka anymore so remove the partitioned stream position.

This was used by the device list stream position. The device list position
now corresponds to the `Offset`, and the partition is always 0, in prep
for removing reliance on Kafka topics for device list changes.

When clients try to `/sync` with an older token with the partitioned stream position they will receive a 401 which should force clients to logout. This will prevent them tight-looping on bad sync tokens.


neilalexander (approved):

r3k2:
The breaking means it will break our current server if we pull this? want to make sure before I do this. thanks

kegsay:
It _was_ a breaking change in that it was intended to soft-logout clients due to the change in the sync token format. However, upon testing on Element-Web, it seems 4xx errors on `/sync` do not cause clients to soft-logout :( so there's backwards compat code in place to handle token format migrations. If you pull, I recommend you pull https://github.com/matrix-org/dendrite/pull/2098 - this is just waiting for review.

PR #2102, opened 2022-01-21
Send created room events as new events

neilalexander:
... rather than outliers, so the state is rolled forward correctly.

Also don't complain if create events don't have a state snapshot NID of 0, because that's an acceptable case.


kegsay (approved):

neilalexander:
This landed in #2104.

PR #2104, opened 2022-01-21
Roomserver/federation input refactor

neilalexander:
This PR moves the responsibility for checking and fetching missing auth events, prev events and state away from the Federation API and into the room server. Benefits are:

* multiple incoming federation requests no longer race and use more resources than necessary fetching the same things;
* the roomserver has direct access to information about which events and state snapshots are missing;
* the roomserver acts as a point of linearisation, so we don't end up doing more work than necessary;
* the federation API is now much much simpler than it was, so much less likely to need sharding.

Overall this seems to smooth out incoming federation significantly and has unstuck a bunch of rooms on `dendrite.matrix.org`.


roomserver/internal/input/input_events.go
```diff
@@ -228,10 +302,115 @@ func (r *Inputer) processRoomEvent(
 	return nil
 }
 
+func (r *Inputer) checkForMissingAuthEvents(
```
kegsay, line 305: This needs documentation. It's not clear by the function name what this is trying to do. It doesn't return anything either, so I'm assuming it has some side-effects.

roomserver/internal/input/input_events.go
```diff
@@ -113,11 +172,37 @@ func (r *Inputer) processRoomEvent(
...
-			logrus.WithFields(logrus.Fields{
-				"event_id": event.EventID(),
-				"type":     event.Type(),
-				"room":     event.RoomID(),
-			}).WithError(err).Info("Error authing soft-failed event")
+			logger.WithError(err).Info("Error authing soft-failed event")
+		}
+	}
+
+	missingPrev := !input.HasState && len(missingRes.MissingPrevEventIDs) > 0
```
kegsay, line 179: Needs comments. Specifically, if we _are_ missing prev events so `len(missingRes.MissingPrevEventIDs) > 0` but we `HasState`, why does this get out of the following code? It's not clear why this makes sense.

roomserver/internal/input/input_events.go
```diff
@@ -157,24 +237,18 @@ func (r *Inputer) processRoomEvent(
 		return fmt.Errorf("r.DB.RoomInfo missing for room %s", event.RoomID())
 	}
 
-	if stateAtEvent.BeforeStateSnapshotNID == 0 {
+	if !missingPrev && stateAtEvent.BeforeStateSnapshotNID == 0 {
```
kegsay, line 240: Why?

roomserver/internal/input/input_events.go
```diff
@@ -87,24 +101,69 @@ func (r *Inputer) processRoomEvent(
...
+
+	// First of all, check that the auth events of the event are known.
+	// If they aren't then we will ask the federation API for them.
 	isRejected := false
-	authEventNIDs, rejectionErr := helpers.CheckAuthEvents(ctx, r.DB, headered, input.AuthEventIDs)
-	if rejectionErr != nil {
-		logrus.WithError(rejectionErr).WithField("event_id", event.EventID()).WithField("auth_event_ids", input.AuthEventIDs).Error("helpers.CheckAuthEvents failed for event, rejecting event")
+	authEvents := gomatrixserverlib.NewAuthEvents(nil)
+	knownEvents := map[string]*types.Event{}
+	if err = r.checkForMissingAuthEvents(ctx, logger, headered, &authEvents, knownEvents, serverRes.ServerNames); err != nil {
```
kegsay, line 147: Given `checkForMissingAuthEvents` is called in exactly one place (here), and given we init empty `knownEvents` and `authEvents`, can we please just make this `fetchAuthEvents(ctx, logger, headered, serverNames) (authEvents, knownEvents, error)`. It's not even right to say _missing_ here as we _don't know_ if they are missing as we blindly call this function irrespective of `if len(missingRes.MissingAuthEventIDs) > 0 || len(missingRes.MissingPrevEventIDs) > 0 {` on `:127`.

roomserver/internal/input/input_events.go
```diff
@@ -228,10 +302,115 @@ func (r *Inputer) processRoomEvent(
...
+		return nil
+	}
+
+	unknown := map[string]struct{}{}
+
+	authEvents, err := r.DB.EventsFromIDs(ctx, authEventIDs)
+	if err != nil {
+		return fmt.Errorf("r.DB.EventsFromIDs: %w", err)
+	}
+	for _, event := range authEvents {
```
kegsay, line 324: It's not safe to do this because `EventsFromIDs` calls `BulkSelectEventNID` which omits missing events:
```go
// bulkSelectEventNIDs returns a map from string event ID to numeric event ID.
// If an event ID is not in the database then it is omitted from the map.
func (s *eventStatements) BulkSelectEventNID(ctx context.Context, eventIDs []string) (map[string]types.EventNID, error) {
```

Sidenote: we should be really clear on this as it's a massive footgun in other places too, probably having something like `SelectFoo(args...) (result, hasMissing, error)` would be a better pattern.

You need to loop `authEventIDs`.

roomserver/internal/input/input_events.go
```diff
@@ -228,10 +302,115 @@ func (r *Inputer) processRoomEvent(
...
+			known[event.EventID()] = &ev
+			if err = auth.AddEvent(event.Event); err != nil {
+				return fmt.Errorf("auth.AddEvent: %w", err)
+			}
+		} else {
+			unknown[event.EventID()] = struct{}{}
+		}
+	}
+
+	if len(unknown) > 0 {
```
kegsay, line 336: Invert and outdent please e.g

```go
if len(unknown) == 0 {
    return nil
}
```

roomserver/internal/input/input_events.go
```diff
@@ -228,10 +302,115 @@ func (r *Inputer) processRoomEvent(
...
+			res, err = r.FSAPI.GetEventAuth(ctx, serverName, event.RoomID(), event.EventID())
+			if err != nil {
+				logger.WithError(err).Warnf("Failed to get event auth from federation for %q: %s", event.EventID(), err)
+				continue
+			}
+			found = true
+			break
+		}
+		if !found {
+			return fmt.Errorf("no servers provided event auth")
```
kegsay, line 349: Might be nice to log which servers we tried, it'd help for debugging. Likewise, the event ID.

roomserver/internal/input/input_events.go
```diff
@@ -228,10 +302,115 @@ func (r *Inputer) processRoomEvent(
...
+
+			// Check the signatures of the event.
+			// TODO: It really makes sense for the federation API to be doing this,
+			// because then it can attempt another server if one serves up an event
+			// with an invalid signature. For now this will do.
+			if err := event.VerifyEventSignatures(ctx, r.FSAPI.KeyRing()); err != nil {
+				return fmt.Errorf("event.VerifyEventSignatures: %w", err)
+			}
+
+			// Otherwise, we need to store, and that means we need to know the
```
kegsay, line 370: The "Otherwise" throws me off here... what?

roomserver/internal/input/input_events.go
```diff
@@ -228,10 +302,115 @@ func (r *Inputer) processRoomEvent(
...
+			return fmt.Errorf("no servers provided event auth")
+		}
+
+		for _, event := range gomatrixserverlib.ReverseTopologicalOrdering(
+			res.AuthEvents,
+			gomatrixserverlib.TopologicalOrderByAuthEvents,
+		) {
+			// If we already know about this event then we don't need to store
+			// it or do anything further with it.
+			if ev, ok := known[event.EventID()]; ok && ev != nil {
```
kegsay, line 358: If an event is in `known` does that imply they are verified as well? Might be worth making that crystal clear here so people know _not_ to add events to this map if they haven't been checked.

roomserver/internal/input/input_events.go
```diff
@@ -228,10 +302,115 @@ func (r *Inputer) processRoomEvent(
...
+				continue
+			}
+			found = true
+			break
+		}
+		if !found {
+			return fmt.Errorf("no servers provided event auth")
+		}
+
+		for _, event := range gomatrixserverlib.ReverseTopologicalOrdering(
```
kegsay, line 352: It would be nicer to call this `authEvent` or something for clarity.

roomserver/internal/input/input_events.go
```diff
@@ -228,10 +302,115 @@ func (r *Inputer) processRoomEvent(
...
+		} else {
+			unknown[event.EventID()] = struct{}{}
+		}
+	}
+
+	if len(unknown) > 0 {
+		var res gomatrixserverlib.RespEventAuth
+		var found bool
+		for _, serverName := range servers {
+			res, err = r.FSAPI.GetEventAuth(ctx, serverName, event.RoomID(), event.EventID())
```
kegsay, line 340: This is the auth chain right? Might be worth mentioning chain here.

roomserver/internal/input/input_missing.go
```diff
@@ -0,0 +1,709 @@
...
+	federation      fedapi.FederationInternalAPI
+	roomsMu         *internal.MutexByRoom
+	servers         []gomatrixserverlib.ServerName
+	hadEvents       map[string]bool
+	hadEventsMutex  sync.Mutex
+	haveEvents      map[string]*gomatrixserverlib.HeaderedEvent
+	haveEventsMutex sync.Mutex
+}
+
+func (t *missingStateReq) processEventWithMissingState(
```
kegsay, line 35: It looks like this is the main entry point to this struct (as it's the place where we call it from in this PR, but it's not really clear from the function name or visibility. It helps readability to know where to start reading. Can you add comments to this function or something to indicate that this is the entry point?

roomserver/internal/input/input_missing.go
```diff
@@ -0,0 +1,709 @@
...
+		)
+		return nil, missingPrevEventsError{
+			eventID: e.EventID(),
+			err:     err,
+		}
+	}
+
+	// security: how we handle failures depends on whether or not this event will become the new forward extremity for the room.
+	// There's 2 scenarios to consider:
+	// - Case A: We got pushed an event and are now fetching missing prev_events. (isInboundTxn=true)
```
kegsay, line 406: `isInboundTxn` no longer exists.

roomserver/internal/input/input_missing.go
```diff
@@ -0,0 +1,709 @@
...
+	// - Case B: We are fetching missing prev_events already and now fetching some more  (isInboundTxn=false)
+	// In Case B, we know for sure that the event we are currently processing will not become the new forward extremity for the room,
+	// as it was called in response to an inbound txn which had it as a prev_event.
+	// In Case A, the event is a forward extremity, and could eventually become the _only_ forward extremity in the room. This is bad
+	// because it means we would trust the state at that event to be the state for the entire room, and allows rooms to be hijacked.
+	// https://github.com/matrix-org/synapse/pull/3456
+	// https://github.com/matrix-org/synapse/blob/229eb81498b0fe1da81e9b5b333a0285acde9446/synapse/handlers/federation.py#L335
+	// For now, we do not allow Case B, so reject the event.
```
kegsay, line 414: We don't reject anything on the next line.

roomserver/internal/input/input_missing.go
```diff
@@ -0,0 +1,709 @@
...
+	// This will either:
+	// - fill in the gap completely then process event `e` returning no backwards extremity
+	// - fail to fill in the gap and tell us to terminate the transaction err=not nil
+	// - fail to fill in the gap and tell us to fetch state at the new backwards extremity, and to not terminate the transaction
+	newEvents, err := t.getMissingEvents(ctx, e, roomVersion)
+	if err != nil {
+		return fmt.Errorf("t.getMissingEvents: %w", err)
+	}
+	if len(newEvents) == 0 {
+		return nil
```
kegsay, line 63: It's not clear if this is a success or error case. The maps won't be populated which... means what exactly?

kegsay (approved):
Needs more comments overall, but seems like it should work with the exception of not getting all auth events from the DB when querying.

clientapi/routing/createroom.go
```diff
@@ -452,20 +452,18 @@ func createRoom(
...
-			},
-			ev.Headered(roomVersion),
-			nil,
-			false,
-		); err != nil {
-			util.GetLogger(req.Context()).WithError(err).Error("SendEventWithState failed")
+		inputs := make([]roomserverAPI.InputRoomEvent, 0, len(builtEvents))
+		for _, event := range builtEvents {
+			inputs = append(inputs, roomserverAPI.InputRoomEvent{
+				Kind:         roomserverAPI.KindNew,
```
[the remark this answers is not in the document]
neilalexander: It is, yes, that PR was cherry-picked from this branch.

roomserver/api/wrapper.go
```diff
@@ -63,6 +65,7 @@ func SendEventWithState(
 		ires = append(ires, InputRoomEvent{
 			Kind:         KindOutlier,
 			Event:        outlier.Headered(event.RoomVersion),
+			Origin:       origin,
 			AuthEventIDs: outlier.AuthEventIDs(),
```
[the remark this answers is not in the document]
neilalexander: I think this is probably bad function naming. `SendEvents` is typically used when we want to send something outbound to the federation and `BuildEvent` has already been used. `SendEventWithState` is normally used when we've received an event and a bunch of state (after a federated room join) and therefore we don't want to send anything over federation as a result. We can just rename those functions so it is more clear what they do.

setup/jetstream/nats.go
```diff
@@ -33,6 +33,7 @@ func Prepare(cfg *config.JetStream) (nats.JetStreamContext, sarama.Consumer, sar
 			StoreDir:         string(cfg.StoragePath),
 			NoSystemAccount:  true,
 			AllowNewAccounts: false,
+			MaxPayload:       16 * 1024 * 1024,
```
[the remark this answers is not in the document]
neilalexander: You're right. I'll do that.

federationapi/routing/threepid.go
```diff
@@ -89,7 +89,7 @@ func CreateInvitesFrom3PIDInvites(
 	}
 
 	// Send all the events
-	if err := api.SendEvents(req.Context(), rsAPI, api.KindNew, evs, cfg.Matrix.ServerName, nil, false); err != nil {
+	if err := api.SendEvents(req.Context(), rsAPI, api.KindNew, evs, "TODO", cfg.Matrix.ServerName, nil, false); err != nil {
```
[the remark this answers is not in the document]
neilalexander: Yes, because for some reason this endpoint is `MakeExternalAPI` instead of `MakeFedAPI` and therefore there's no `gomatrixserverlib.FederationRequest` to tell us what the real origin of the request was, and I haven't looked into why yet.

kegsay:
For aud trail: The failing tests are flakey failures being resolved by https://github.com/matrix-org/sytest/pull/1185

PR #2109, opened 2022-01-25
Increase NATS maximum message size to 16MB

neilalexander:
Currently joining rooms with huge auth chains or state can cause us to hit this limit on the output events. This is by no means a permanent fix — we need to refactor how state is communicated via the roomserver output events — but this should alleviate the symptoms a bit in the meantime.


spaetz:
It definitely unblocked some rooms for me and is obviously the right thing to do in the short term. In the long-term we should avoid passing arbitarily big blobs via the IPC mechanism, I guess. Should this size be made configurable rather than be hard coded?

kegsay (approved):
LGTM -  though it is just a stopgap.

PR #2115, opened 2022-01-27
Don't flood Sentry with context cancelled/deadline exceeded errors

neilalexander:
Also don't block the room server worker itself on hooks for new events.


neilalexander:
We really should just have wrapper functions for Sentry that check all of the usual error conditions that we don't care about, e.g. host not in room, context cancelled, deadline exceeded, etc.

PR #2134, opened 2022-01-31
Use background contexts during federated join for clarity

neilalexander:
I think the `cancel()` sometimes happens prematurely and since it's all just a background context anyway, we'd might as well just be explicit about it.


kegsay:
CI is failing.

PR #2136, opened 2022-01-31
Support CA certificates in CI

kegsay:


build/scripts/Complement.Dockerfile
```diff
@@ -13,9 +13,14 @@ RUN go build ./cmd/dendrite-monolith-server
...
-RUN ./generate-keys --private-key matrix_key.pem --tls-cert server.crt --tls-key server.key
+RUN ./generate-keys --private-key matrix_key.pem
 
 ENV SERVER_NAME=localhost
 EXPOSE 8008 8448
 
-CMD sed -i "s/server_name: localhost/server_name: ${SERVER_NAME}/g" dendrite.yaml && ./dendrite-monolith-server --tls-cert server.crt --tls-key server.key --config dendrite.yaml
+# At runtime, generate TLS cert based on the CA now mounted at /ca
+# At runtime, replace the SERVER_NAME with what we are told
+CMD ./generate-keys --server $SERVER_NAME --tls-cert server.crt --tls-key server.key --tls-authority-cert /ca/ca.crt --tls-authority-key /ca/ca.key && \
```
neilalexander, line 23: It might be safer just to allow `generate-config` to take a parameter for the server name?

internal/test/config.go
```diff
@@ -205,3 +206,73 @@ func NewTLSKey(tlsKeyPath, tlsCertPath string) error {
...
+
+// NewTLSKey generates a new RSA TLS key and certificate and writes it to a file.
+func NewTLSKey(tlsKeyPath, tlsCertPath string) error {
+	priv, template, err := generateTLSTemplate(nil)
+	if err != nil {
+		return err
+	}
+
+	// Self-signed certificate: template == parent
+	derBytes, err := x509.CreateCertificate(rand.Reader, template, template, &priv.PublicKey, priv)
```
neilalexander, line 218: I'm not sure if it actually matters if it is just for testing, but it may be a good idea to `rand.Seed()` first.

neilalexander (approved):
Overall looks fine, just a couple of minor things I noticed.

internal/test/config.go
kegsay: There is no `rand.Seed()`. It just reads from `/dev/urandom`

build/scripts/Complement.Dockerfile
kegsay: `generate-keys` also needs the server name, so you're saying that I should add a `--server` to `generate-config` instead of `sed`ing like this?

neilalexander: Yep, just in case we ever change the format or the default value in the config.

PR #2137, opened 2022-02-01
PerformInvite: bugfix and rejig control flow

kegsay:
Local clients would not be notified of invites to rooms
Dendrite had already joined in all cases due to not returning
an `api.OutputNewInviteEvent` for local invites. We now do this.

This was an easy mistake to make due to the control flow of the
function which doesn't handle the happy case at the end of the
function and instead forks the function depending on if the
invite was via federation or not. This has now been changed to
handle the federated invite as if it were an error (in that we
check it, do it and bail out) rather than outstay our welcome.
This ends up with the local invite being the happy case, which
now both sends an `InputRoomEvent` to the roomserver _and_ a
`api.OutputNewInviteEvent` is returned.

### DIFF
--- a/roomserver/internal/perform/perform_invite.go
+++ b/roomserver/internal/perform/perform_invite.go
@@ -27,6 +27,7 @@ import (
 	"github.com/matrix-org/dendrite/roomserver/types"
 	"github.com/matrix-org/dendrite/setup/config"
 	"github.com/matrix-org/gomatrixserverlib"
+	"github.com/matrix-org/util"
 	log "github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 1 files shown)


roomserver/internal/perform/perform_invite.go
```diff
@@ -54,18 +55,23 @@ func (r *Inviter) PerformInvite(
...
+	isTargetLocal := domain == r.Cfg.Matrix.ServerName
+	isOriginLocal := event.Origin() == r.Cfg.Matrix.ServerName
+
+	logger := util.GetLogger(ctx).WithFields(map[string]interface{}{
+		"inviter": event.Sender(),
+		"invitee": *event.StateKey(),
+		"room_id": roomID,
+	})
+	logger.WithFields(log.Fields{
 		"event_id":         event.EventID(),
```
neilalexander, line 68: Can we move some of these fields into the `logger` definition so that later logging calls have some useful context?

neilalexander (approved):
Aside from a potential logging improvement, this looks fine to me.

PR #2138, opened 2022-02-02
Remove sarama/saramajetstream dependencies

S7evinK:
This should remove all left traces of sarama/saramajetstream.
Not exactly sure about the delivery policy for each JetStream subscription, though.

### Pull Request Checklist


federationapi/consumers/keychange.go
```diff
@@ -149,24 +144,25 @@ func (t *KeyChangeConsumer) onDeviceKeyMessage(m api.DeviceMessage) error {
 		Keys:              m.KeyJSON,
 	}
 	if edu.Content, err = json.Marshal(event); err != nil {
-		return err
+		return true
```
kegsay, line 147: Don't you mean false?

S7evinK: Looking at https://github.com/matrix-org/dendrite/blob/a09d71d231861f8825a4f8f1dfd79311c4c236a6/federationapi/consumers/eduserver.go#L124-L127
we seem to return true in other places. Maybe we should really change this to false, so we can try again. But this could result in a loop, if JetStream keeps on redelivering the same data.

neilalexander: `true` is correct — if we fail to unmarshal the JSON once, we'll fail to unmarshal it every time, so we'd might as well just ACK it and move on.

kegsay (approved):

neilalexander (approved):

PR #2139, opened 2022-02-02
Adding `Jetstream` service to docker-compose.monolith.yml

nkitsaini:
### Pull Request Checklist

Signed-off-by: `Ankit Saini <nnkitsaini@gmail.com>`

### DIFF
--- a/build/docker/docker-compose.monolith.yml
+++ b/build/docker/docker-compose.monolith.yml
@@ -20,6 +20,20 @@ services:
     networks:
       - internal
 
+  jetstream:
+    hostname: jetstream
+    image: nats:latest
+    command: |
+      --jetstream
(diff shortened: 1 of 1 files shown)


neilalexander (commented):
It's not immediately clear we need this in the monolith example, given that we have a built-in NATS Server in the monolith. If we do want to keep it , it might be worth clarifying in some comments above this section that this is optional and only actually gets used if JetStream `addresses` are specified in the Dendrite config.

nkitsaini:
Ah sorry, that's my bad. Didn't realize there's an internal implementation (should've read the config!). I think it's good in current state. One difference is that now user needs to modify config to run `monolith`, but I think it's not that big of a deal.

Thanks for replying though. I'll close the PR for now.

PR #2140, opened 2022-02-02
Use pull consumers

neilalexander:
This PR updates the Sync API, Federation API and Appservice API to use pull consumers instead of push consumers so that we can strictly guarantee ordering and so that we can reduce the number of duplicates delivered to us. This should reduce flakiness in some places.


kegsay (approved):
LGTM - scary about the context footguns though!

PR #2141, opened 2022-02-02
Full roomserver input transactional isolation

neilalexander:
This PR does the following things:

* renames the "latest events updater" to the "room updater"
* uses the room updater to perform all database tasks during the RS input process
* only acknowledges messages from NATS when the transaction is successfully committed
* fudges around the storage code so that all roomserver tables now take an optional `*sql.Tx`
* fudges some more storage code so that `*RoomUpdater` satisfies new subset interfaces needed for state res and checking auth/soft fail
* other various tweaks to fix bugs

### DIFF
--- a/roomserver/internal/input/input.go
+++ b/roomserver/internal/input/input.go
@@ -157,18 +157,15 @@ func (r *Inputer) processRoomEventUsingUpdater(
 	}
 	commit, err := r.processRoomEvent(ctx, updater, inputRoomEvent)
 	if commit {
-		if err = updater.Commit(); err != nil {
-			return false, fmt.Errorf("updater.Commit: %w", err)
+		if cerr := updater.Commit(); err != nil {
+			return true, fmt.Errorf("updater.Commit: %w", cerr)
 		}
 	} else {
 		if rerr := updater.Rollback(); rerr != nil {
-			return true, fmt.Errorf("updater.Rollback: %w", err)
+			return true, fmt.Errorf("updater.Rollback: %w", rerr)
 		}
 	}
-	if err != nil {
-		return true, err
-	}
-	return false, nil
+	return false, err
 }
 
 // InputRoomEvents implements api.RoomserverInternalAPI
--- a/roomserver/internal/input/input.go
+++ b/roomserver/internal/input/input.go
@@ -19,12 +19,14 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"sync"
 	"time"
 
 	"github.com/Arceliar/phony"
 	"github.com/getsentry/sentry-go"
 	fedapi "github.com/matrix-org/dendrite/federationapi/api"
+	"github.com/matrix-org/dendrite/internal/sqlutil"
 	"github.com/matrix-org/dendrite/roomserver/acls"
 	"github.com/matrix-org/dendrite/roomserver/api"
 	"github.com/matrix-org/dendrite/roomserver/internal/query"
@@ -101,7 +103,7 @@ func (r *Inputer) Start() error {
 				_ = msg.InProgress() // resets the acknowledgement wait timer
 				defer eventsInProgress.Delete(index)
 				defer roomserverInputBackpressure.With(prometheus.Labels{"room_id": roomID}).Dec()
-				if err := r.processRoomEvent(context.Background(), &inputRoomEvent); err != nil {
+				if err := r.processRoomEventUsingUpdater(context.Background(), roomID, &inputRoomEvent); err != nil {
 					if !errors.Is(err, context.DeadlineExceeded) && !errors.Is(err, context.Canceled) {
 						sentry.CaptureException(err)
 					}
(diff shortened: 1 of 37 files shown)


roomserver/storage/sqlite3/rooms_table.go
```diff
@@ -274,13 +282,18 @@ func (s *roomStatements) BulkSelectRoomIDs(ctx context.Context, roomNIDs []types
...
 
-func (s *roomStatements) BulkSelectRoomNIDs(ctx context.Context, roomIDs []string) ([]types.RoomNID, error) {
+func (s *roomStatements) BulkSelectRoomNIDs(ctx context.Context, txn *sql.Tx, roomIDs []string) ([]types.RoomNID, error) {
 	iRoomIDs := make([]interface{}, len(roomIDs))
 	for i, v := range roomIDs {
 		iRoomIDs[i] = v
 	}
 	sqlQuery := strings.Replace(bulkSelectRoomNIDsSQL, "($1)", sqlutil.QueryVariadic(len(roomIDs)), 1)
-	rows, err := s.db.QueryContext(ctx, sqlQuery, iRoomIDs...)
+	sqlPrep, err := s.db.Prepare(sqlQuery)
```
kegsay, line 291: It's not really worth doing this, as we don't hold onto `sqlPrep`. It will end up adding more work and increasing latency.

roomserver/storage/shared/storage.go
```diff
@@ -349,37 +410,49 @@ func (d *Database) GetMembership(ctx context.Context, roomNID types.RoomNID, req
...
+
+func (d *Database) events(
+	ctx context.Context, txn *sql.Tx, eventNIDs []types.EventNID,
+) ([]types.Event, error) {
+	eventJSONs, err := d.EventJSONTable.BulkSelectEventJSON(ctx, txn, eventNIDs)
 	if err != nil {
 		return nil, err
 	}
-	eventIDs, _ := d.EventsTable.BulkSelectEventID(ctx, eventNIDs)
+	eventIDs, _ := d.EventsTable.BulkSelectEventID(ctx, nil, eventNIDs)
```
kegsay, line 450: Why are being inconsistent here and passing `nil` but 5 lines later we pass `txn`?

roomserver/storage/shared/room_updater.go
```diff
@@ -0,0 +1,262 @@
...
+}
+
+func rollback(txn *sql.Tx) {
+	if txn == nil {
+		return
+	}
+	txn.Rollback() // nolint: errcheck
+}
+
+func NewRoomUpdater(ctx context.Context, d *Database, txn *sql.Tx, roomInfo *types.RoomInfo) (*RoomUpdater, error) {
```
kegsay, line 28: This leaks `sql.Tx` outside the storage layer, no?

roomserver/internal/input/input.go
```diff
@@ -131,6 +137,37 @@ func (r *Inputer) Start() error {
...
+// commit the transaction) and an error signifying anything else that may
+// have gone wrong.
+func (r *Inputer) processRoomEventUsingUpdater(
+	ctx context.Context,
+	roomID string,
+	inputRoomEvent *api.InputRoomEvent,
+) (bool, error) {
+	roomInfo, err := r.DB.RoomInfo(ctx, roomID)
+	if err != nil {
+		return false, fmt.Errorf("r.DB.RoomInfo: %w", err)
```
kegsay, line 152: It's not clear why this should be `false` and not `true`.

roomserver/internal/input/input_events.go
```diff
@@ -67,14 +68,15 @@ var processRoomEventDuration = prometheus.NewHistogramVec(
 // nolint:gocyclo
 func (r *Inputer) processRoomEvent(
 	ctx context.Context,
+	updater *shared.RoomUpdater,
 	input *api.InputRoomEvent,
-) (err error) {
+) (commit bool, err error) {
```
kegsay, line 73: I would really prefer this to be a dedicated type and not a bool. It's unclear as it stands because we also have functions which return `bool, error` but mean very different things: `retry, err := r.processRoomEventUsingUpdater(context.Background(), roomID, &inputRoomEvent)` something like:
```
type Retry bool
var RetryEvent Retry = true
var DoNotRetry Retry = false
```
then use `Retry` instead of `bool`. Ditto for the commit stuff. It makes it a lot clearer to see `Commit, nil` or `Rollback, err`, etc.
```

kegsay (approved):
Overall this is an improvement on what we had before. I think the API design needs to be a bit clearer rather than just sending around `bool`s, but otherwise LGTM.

roomserver/storage/shared/storage.go
neilalexander: Genuine bug, will fix.

roomserver/storage/shared/room_updater.go
neilalexander: Well, sort of, but this is `shared.NewRoomUpdater` rather than `(storage.Database).GetRoomUpdater()` — it's this way because `NewRoomUpdater` needs to be exported to the `sqlite3` package so that it can override the `GetRoomUpdater` and `MembershipUpdater` functions with ones that pass in a `nil` txn.

FWIW the membership and latest events updater have always been this way.

PR #2152, opened 2022-02-07
Use `KindOld` instead of `KindNew` for fetched missing events

neilalexander:
This should reduce backpressure on the roomserver input quite a bit because `KindNew` requires far more checks. If we filled the gap then we can skip all the effort of checking for missing previous events and we don't want to run the latest events updater for these events either.


neilalexander:
Replaced by #2154.

PR #2159, opened 2022-02-08
Allow events to be un-rejected

neilalexander:
Today, if an event is rejected, that is a permanent action. This PR changes it so that if we re-process the event later and it doesn't end up rejected after all, we can deal with that case by updating it in the database to un-reject it and then continue processing it fully.


kegsay (changes requested):
Please add a complement test for this and xref it with this PR. Without a test we can't know if this works / if we break it.

neilalexander:
matrix-org/complement#301 merged to help prevent regressions in this area, although the effects of this PR are mostly to ensure that the roomserver database is consistent with the output events generated.

PR #2160, opened 2022-02-08
Remove OutputKeyChangeEvent consumer on keyserver

S7evinK:
This removes the `OutputKeyChangeEvent` consumer on the keyserver to make it easier to reason about, as it was consuming/producing on the same topic.


neilalexander (approved):
This is a really nice simplification and certainly easier to understand without consuming and producing on the same topic. LGTM if tests pass.

kegsay (approved):

PR #2162, opened 2022-02-09
Fix sytest `uploading signed devices gets propagated over federation`

S7evinK:
Fixes the failing `uploading signed devices gets propagated over federation` test, still not ideal I think.

### DIFF
--- a/federationapi/consumers/roomserver.go
+++ b/federationapi/consumers/roomserver.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+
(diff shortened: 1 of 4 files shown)


kegsay:
Assigning to @neilalexander as I'm probably not best placed to know if these changes make sense. Why do you think they aren't ideal?

S7evinK:
We're sending a federated request, even though we _should_ have the keys already.

neilalexander (approved):
I think this should be OK, I've been aware of a problem where we might hit federation unnecessarily anyway but I'd rather that for now than to drop updates.

PR #2163, opened 2022-02-09
Fix fetching missing state

neilalexander:
Beforehand, if we were missing prev events or state, we were attempting to roll forward from the state at the backwards extremity of our `/get_missing_events` result. The problem is that sometimes this can fail, so we get to the end of our new events and are still either missing events or state (because something got rejected). 

In that case, we need to also check after doing so that we actually solved the problem and found the state before the new events' prev events, and if not, go through the `/state_ids` dance for that too.

In effect, this PR is making a flow that looks much more like this:

Untitled 10

### DIFF
--- a/roomserver/internal/input/input_events.go
+++ b/roomserver/internal/input/input_events.go
@@ -299,7 +299,7 @@ func (r *Inputer) processRoomEvent(
 		return rollbackTransaction, fmt.Errorf("updater.RoomInfo missing for room %s", event.RoomID())
 	}
 
-	if !missingPrev && stateAtEvent.BeforeStateSnapshotNID == 0 {
+	if input.HasState || (!missingPrev && stateAtEvent.BeforeStateSnapshotNID == 0) {
 		// We haven't calculated a state for this event yet.
 		// Lets calculate one.
 		err = r.calculateAndSetState(ctx, updater, input, roomInfo, &stateAtEvent, event, isRejected)
--- a/roomserver/internal/input/input_events.go
+++ b/roomserver/internal/input/input_events.go
@@ -255,9 +255,16 @@ func (r *Inputer) processRoomEvent(
 				hadEvents:  map[string]bool{},
 				haveEvents: map[string]*gomatrixserverlib.HeaderedEvent{},
 			}
-			if err := missingState.processEventWithMissingState(ctx, event, headered.RoomVersion); err != nil {
+			if override, err := missingState.processEventWithMissingState(ctx, event, headered.RoomVersion); err != nil {
 				isRejected = true
 				rejectionErr = fmt.Errorf("missingState.processEventWithMissingState: %w", err)
+			} else if override != nil {
+				missingPrev = false
+				input.HasState = true
+				input.StateEventIDs = make([]string, 0, len(override.StateEvents))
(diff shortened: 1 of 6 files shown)


roomserver/internal/query/query.go
```diff
@@ -150,7 +150,7 @@ func (r *Queryer) QueryMissingAuthPrevEvents(
 
 	for _, prevEventID := range request.PrevEventIDs {
 		state, err := r.DB.StateAtEventIDs(ctx, []string{prevEventID})
-		if err != nil || len(state) == 0 {
+		if err != nil || len(state) == 0 || state[0].BeforeStateSnapshotNID == 0 {
```
kegsay, line 153: What about the create event?

roomserver/internal/input/input_events.go
```diff
@@ -255,13 +255,32 @@ func (r *Inputer) processRoomEvent(
 				hadEvents:  map[string]bool{},
 				haveEvents: map[string]*gomatrixserverlib.HeaderedEvent{},
 			}
-			if err := missingState.processEventWithMissingState(ctx, event, headered.RoomVersion); err != nil {
+			if override, err := missingState.processEventWithMissingState(ctx, event, headered.RoomVersion); err != nil {
```
kegsay, line 258: I don't like the name `override`, it's not clear what that means. Comments are fab further down, so perhaps `stateSnapshot` or something?

roomserver/internal/input/input_missing.go
```diff
@@ -83,16 +84,142 @@ func (t *missingStateReq) processEventWithMissingState(
...
+	hadEvents := map[string]bool{}
+	t.hadEventsMutex.Lock()
+	for k, v := range t.hadEvents {
+		hadEvents[k] = v
+	}
+	t.hadEventsMutex.Unlock()
+
+	sendOutliers := func(resolvedState *gomatrixserverlib.RespState) error {
+		var outliers []*gomatrixserverlib.Event
+		outliers, err = resolvedState.Events()
```
kegsay, line 123: We'll need to use `OrderAuthAndStateEvents` going forwards.

roomserver/internal/input/input_missing.go
```diff
@@ -83,16 +84,142 @@ func (t *missingStateReq) processEventWithMissingState(
...
+		}
+	}
+
+	// Then send all of the newer backfilled events, of which will all be newer
+	// than the backward extremity, into the roomserver without state. This way
+	// they will automatically fast-forward based on the room state at the
+	// extremity in the last step.
+	for _, newEvent := range newEvents {
+		_, err = t.inputer.processRoomEvent(ctx, t.db, &api.InputRoomEvent{
+			Kind:         api.KindOld,
```
kegsay, line 183: Thinking of names for these... can we maybe do:
 - `KindOld` -> `KindTimeline`
 - `KindNew` -> `KindFwdsExtremity`?

kegsay (approved):
LGTM!

roomserver/internal/input/input_missing.go
neilalexander: Suggest we do this in a separate PR — let's raise an issue and I can add it to my todo list.

PR #2164, opened 2022-02-09
Unmarshal events at the Dendrite level not GMSL level

kegsay:
This relies on https://github.com/matrix-org/gomatrixserverlib/pull/288

Previously, when you made federation requests you had to supply a room version up front. This value would be set in a private field on the response type and would be used to automatically unmarshal the event JSON into the right `Event`.

This was problematic because it meant:
 - You couldn't mock out the response types because you couldn't set the room version (only GMSL could, and it only set it on the federation client functions).
 - We parsed the event JSON at deserialisation time. This implies that we want that work to be intrinsically linked to deserialisation, which is just not the case. Polylith mode will have serialisation boundaries between components, meaning we would be repeatedly (de)serialising events.

The net result is that we should have a better API shape in GMSL and do less work in Dendrite (fewer sig checks and hash checks).

Follow-up work is to add the event ID to the headered event to avoid hashing costs at API boundaries (we never cache the event ID and always calculate it).


neilalexander (approved):
Looks OK to me! Suspect there'll be merge conflicts with the work in #2163.

PR #2165, opened 2022-02-09
Fix transaction issues in events table in PSQL

neilalexander:
This fixes a problem where `InsertEvent` doesn't use the database transaction, so it can get stuck if a transaction tries to do something with the same rows as another transaction.


kegsay (approved):
What about SQLite?

neilalexander:
SQLite doesn't use transactions so it should be no change.

PR #2166, opened 2022-02-09
Allow user to forget a room, even if they never were a member

S7evinK:
The spec doesn't say anything about forgetting a room one has never been part of. So remove this part.
Also returns a better message if the user is still a member.


S7evinK:
Test is at https://github.com/matrix-org/complement/pull/302, Synapse seems to be happy. Dendrite is unhappy, because this PR isn't merged yet. :)

kegsay (approved):

PR #2171, opened 2022-02-10
Add account type

S7evinK:
This adds an `account_type` column to the database and to the `Device` & `Account` type, so it can be checked on requests.
Currently the only way to create an admin account is by using the `cmd/create-account/main.go` utility.


clientapi/routing/admin_whois.go
```diff
@@ -47,8 +47,7 @@ func GetAdminWhois(
 	req *http.Request, userAPI api.UserInternalAPI, device *api.Device,
 	userID string,
 ) util.JSONResponse {
-	if userID != device.UserID {
-		// TODO: Still allow if user is admin
+	if device.AccountType != api.AccountTypeAdmin && userID != device.UserID {
```
kegsay, line 50: I think it would be clearer inverted. Access controls checks are [easy to mess-up](https://owasp.org/Top10/A01_2021-Broken_Access_Control/) so splitting this out to something like the following would be nicer I think:
```go
allowed := device.AccountType == api.AccountTypeAdmin || userID == device.UserID
if !allowed {
    return ....
}
```

userapi/storage/accounts/sqlite3/deltas/2022021012490600_add_account_type.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func LoadAddAccountType(m *sqlutil.Migrations) {
+	m.AddMigration(UpAddAccountType, DownAddAccountType)
+}
+
+func UpAddAccountType(tx *sql.Tx) error {
+	_, err := tx.Exec(`
+	ALTER TABLE account_accounts RENAME TO account_accounts_tmp;
+CREATE TABLE account_accounts (
```
kegsay, line 23: Why not just add the column? https://www.sqlite.org/lang_altertable.html

userapi/storage/accounts/sqlite3/accounts_table.go
```diff
@@ -39,14 +40,16 @@ CREATE TABLE IF NOT EXISTS account_accounts (
     -- Identifies which application service this account belongs to, if any.
     appservice_id TEXT,
     -- If the account is currently active
-    is_deactivated BOOLEAN DEFAULT 0
+    is_deactivated BOOLEAN DEFAULT 0,
+	-- The account_type (user = 1, guest = 2, admin = 3, appservice = 4)
+	account_type INTEGER DEFAULT 2
```
kegsay, line 45: Please do not use defaults. They are error-prone. Dendrite should _always_ be setting this value correctly.

userapi/storage/accounts/postgres/deltas/2022021013023800_add_account_type.go
```diff
@@ -0,0 +1,32 @@
...
+
+	"github.com/matrix-org/dendrite/internal/sqlutil"
+)
+
+func LoadAddAccountType(m *sqlutil.Migrations) {
+	m.AddMigration(UpAddAccountType, DownAddAccountType)
+}
+
+func UpAddAccountType(tx *sql.Tx) error {
+	_, err := tx.Exec("ALTER TABLE account_accounts ADD COLUMN IF NOT EXISTS account_type INT DEFAULT 2;")
```
kegsay, line 15: Do not use DEFAULT please.

userapi/internal/api.go
```diff
@@ -375,6 +374,15 @@ func (a *UserInternalAPI) QueryAccessToken(ctx context.Context, req *api.QueryAc
...
 	}
+	localPart, _, err := gomatrixserverlib.SplitID('@', device.UserID)
+	if err != nil {
+		return err
+	}
+	acc, err := a.AccountDB.GetAccountByLocalpart(ctx, localPart)
+	if err != nil {
+		return err
+	}
+	device.AccountType = acc.AccountType
```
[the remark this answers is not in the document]
kegsay: I wouldn't worry about it for now. We can profile and improve it if it becomes an issue.

kegsay (changes requested):

clientapi/routing/admin_whois.go
S7evinK: Definitely makes it easier to read.

userapi/storage/accounts/sqlite3/deltas/2022021012490600_add_account_type.go
S7evinK: I just copied the migration, I also thought there was something that prevented `ADD COLUMN`. Changed it now.

userapi/storage/accounts/postgres/accounts_table.go
```diff
@@ -39,16 +40,18 @@ CREATE TABLE IF NOT EXISTS account_accounts (
     -- Identifies which application service this account belongs to, if any.
     appservice_id TEXT,
     -- If the account is currently active
-    is_deactivated BOOLEAN DEFAULT FALSE
+    is_deactivated BOOLEAN DEFAULT FALSE,
+	-- The account_type (user = 1, guest = 2, admin = 3, appservice = 4)
+	account_type SMALLINT
```
kegsay, line 45: Add NOT NULL please.

userapi/storage/accounts/sqlite3/accounts_table.go
```diff
@@ -39,14 +40,16 @@ CREATE TABLE IF NOT EXISTS account_accounts (
     -- Identifies which application service this account belongs to, if any.
     appservice_id TEXT,
     -- If the account is currently active
-    is_deactivated BOOLEAN DEFAULT 0
+    is_deactivated BOOLEAN DEFAULT 0,
+	-- The account_type (user = 1, guest = 2, admin = 3, appservice = 4)
+	account_type INTEGER
```
kegsay, line 45: NOT NULL please.

userapi/storage/accounts/sqlite3/deltas/2022021012490600_add_account_type.go
```diff
@@ -0,0 +1,43 @@
...
+	goose.AddMigration(UpAddAccountType, DownAddAccountType)
+}
+
+func LoadAddAccountType(m *sqlutil.Migrations) {
+	m.AddMigration(UpAddAccountType, DownAddAccountType)
+}
+
+func UpAddAccountType(tx *sql.Tx) error {
+	// initially set every account to useracount, change appserver accounts afterwards
+	_, err := tx.Exec(`ALTER TABLE account_accounts ADD COLUMN account_type INTEGER;`)
```
kegsay, line 22: NOT NULL

kegsay (approved):
Needs NOT NULL but otherwise LGTM

clientapi/routing/register.go
```diff
@@ -759,13 +760,16 @@ func completeRegistration(
 			JSON: jsonerror.BadJSON("missing password"),
 		}
 	}
-
+	accType := userapi.AccountTypeUser
+	if appserviceID != "" {
+		accType = userapi.AccountTypeAppService
```
kegsay, line 765: If you look [here](https://github.com/matrix-org/dendrite/blob/a84f50f4fb0963e15b6aca60f658bed45b779127/clientapi/routing/register.go#L966) you can see we call `completeRegistration` during shared secret auth but ignore the [admin flag[(https://github.com/matrix-org/dendrite/blob/a84f50f4fb0963e15b6aca60f658bed45b779127/clientapi/routing/register_secret.go#L26). We should really move the `accType` to be an arg to `completeRegistration`.

kegsay (approved):
SHIPIT

Issue #2173, opened 2022-02-10
Roomserver can deadlock due to multiple txns


kegsay:
When the roomserver processes events, it requires _two_ connections as it makes _two_ transactions (annoyingly). One to process the event and then another for the latest events / membership updater. This was always known to be a wart but this can explode spectacularly on large/busy servers. This has been present since Day One in Dendrite.

In one scenario, the `max_open_conns` was set to `10` for the roomserver. However, the server was very busy such that >10 events in different rooms arrived at once, causing all 10 connections to be obtained. These connections could not complete because they then required a 2nd connection for the latest events / membership updater, which it could never get, hence it deadlocks.

This can manifest in metrics with `dendrite_roomserver_input_backpressure` being present and not changing even for a newly created room(!).

The workaround for now is to increase the connection count, but this merely delays the problem....

We need to guarantee that we only use a single in-flight txn when processing an event to ensure we cannot deadlock.

neilalexander:
Fixed by #2178.

PR #2178, opened 2022-02-11
Ensure only one transaction is used for RS input per room

neilalexander:
This PR makes the following changes:

* All code on the RS input path now uses the room updater, so it happens strictly within a single transaction
* `QueryMissingAuthPrevEvents` is now gone, as an equivalent function has now been moved into `shared.RoomUpdater`
* The `haveEvents` cache in `missingStateReq` now tracks `Event` instead of `HeaderedEvent`, as it removes a lot of wholly unnecessary wrapping and unwrapping

This should go some way towards helping #2173 and @r3k2's problems, although is likely not the final solution.

### DIFF
--- a/roomserver/internal/input/input_events.go
+++ b/roomserver/internal/input/input_events.go
@@ -128,9 +128,17 @@ func (r *Inputer) processRoomEvent(
 		}
 	}
 
+	// Don't waste time processing the event if the room doesn't exist.
+	// A room entry locally will only be created in response to a create
+	// event.
+	isCreateEvent := event.Type() == gomatrixserverlib.MRoomCreate && event.StateKeyEquals("")
+	if !updater.RoomExists() && !isCreateEvent {
+		return rollbackTransaction, fmt.Errorf("room does not exist")
+	}
+
 	var missingAuth, missingPrev bool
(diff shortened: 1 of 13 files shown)


roomserver/internal/input/input_events.go
```diff
@@ -128,20 +128,24 @@ func (r *Inputer) processRoomEvent(
 		}
 	}
 
-	missingRes := &api.QueryMissingAuthPrevEventsResponse{}
+	// Don't waste time processing the event if the room doesn't exist.
+	// A room entry locally will only be created in response to a create
+	// event.
+	isCreateEvent := event.Type() == gomatrixserverlib.MRoomCreate && event.StateKeyEquals("")
+	if !updater.RoomExists() && !isCreateEvent {
+		return rollbackTransaction, fmt.Errorf("room does not exist")
```
kegsay, line 136: Be more specific please. State the event that we got for the room it purports to be in.

kegsay (commented):
Without a test I cannot say with any confidence that there is only 1 transaction used for `InputRoomEvents`. Can you add one please?

kegsay (approved):

PR #2180, opened 2022-02-11
Implement server notices

S7evinK:
This adds the unspecced server notices.
Synapse Docs:
[admin api server_notices.md](https://github.com/matrix-org/synapse/blob/b65acead428653b988351ae8d7b22127a22039cd/docs/admin_api/server_notices.md) (for the request)
[server_notices](https://github.com/matrix-org/synapse/blob/b65acead428653b988351ae8d7b22127a22039cd/docs/server_notices.md) (for server side implementation details)

> When the user is first sent a server notice, they will get an invitation to a room (typically called 'Server Notices', though this is configurable in homeserver.yaml). They will be unable to reject this invitation - attempts to do so will receive an error.
> 
> Once they accept the invitation, they will see the notice message in the room history; it will appear to have come from the 'server notices user' (see below).
> 
> The user is prevented from sending any messages in this room by the power levels.
> 
> Having joined the room, the user can leave the room if they want. Subsequent server notices will then cause a new room to be created.

Due to not having the history visibility yet, a message is sent, but the user can't see it. :/

### DIFF
--- a/clientapi/routing/server_notices.go
+++ b/clientapi/routing/server_notices.go
@@ -1,0 +1,311 @@
+// Copyright 2022 The Matrix.org Foundation C.I.C.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package routing
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/matrix-org/gomatrix"
+	"github.com/matrix-org/gomatrixserverlib"
+	"github.com/matrix-org/gomatrixserverlib/tokens"
+	"github.com/matrix-org/util"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/sirupsen/logrus"
+
+	appserviceAPI "github.com/matrix-org/dendrite/appservice/api"
+	"github.com/matrix-org/dendrite/clientapi/httputil"
+	"github.com/matrix-org/dendrite/clientapi/jsonerror"
+	"github.com/matrix-org/dendrite/internal/eventutil"
+	"github.com/matrix-org/dendrite/internal/transactions"
+	"github.com/matrix-org/dendrite/roomserver/api"
+	"github.com/matrix-org/dendrite/setup/config"
+	userapi "github.com/matrix-org/dendrite/userapi/api"
+	"github.com/matrix-org/dendrite/userapi/storage/accounts"
+)
+
+// Unspecced server notice request
+// https://github.com/matrix-org/synapse/blob/develop/docs/admin_api/server_notices.md
+type sendServerNoticeRequest struct {
+	UserID  string `json:"user_id,omitempty"`
+	Content struct {
+		MsgType string `json:"msgtype,omitempty"`
+		Body    string `json:"body,omitempty"`
+	} `json:"content,omitempty"`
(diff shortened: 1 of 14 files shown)


setup/config/config_global.go
```diff
@@ -123,6 +127,30 @@ func (c *Metrics) Defaults(generate bool) {
...
+	DisplayName string `yaml:"display_name"`
+	// The avatar of this user
+	AvatarURL string `yaml:"avatar"`
+	// The roomname to be used when creating messages
+	RoomName string `yaml:"room_name"`
+}
+
+func (c *ServerNotices) Defaults(generate bool) {
+	if generate {
+		c.LocalPart = "server"
```
kegsay, line 145: See before: maybe use `_server`?

clientapi/routing/server_notices.go
```diff
@@ -0,0 +1,331 @@
...
+
+	senderDevice, err := getSenderDevice(ctx, userAPI, accountsDB, cfgClient)
+	if err != nil {
+		logrus.WithError(err).Error("unable to get device")
+		return util.ErrorResponse(err)
+	}
+	// create a new room for the user
+	if len(commonRooms) == 0 {
+		powerLevelContent := eventutil.InitialPowerLevelsContent(senderUserID)
+		powerLevelContent.Users[r.UserID] = -10
```
kegsay, line 150: Not 0? Please quote synapse for things like this.

clientapi/routing/server_notices.go
```diff
@@ -0,0 +1,331 @@
...
+// It returns an userapi.Device, which is used for building the event
+func getSenderDevice(
+	ctx context.Context,
+	userAPI userapi.UserInternalAPI,
+	accountDB accounts.Database,
+	cfg *config.ClientAPI,
+) (*userapi.Device, error) {
+	var accRes userapi.PerformAccountCreationResponse
+	// create account if it doesn't exist
+	err := userAPI.PerformAccountCreation(ctx, &userapi.PerformAccountCreationRequest{
```
kegsay, line 280: This will be called every time this endpoint is invoked which surely isn't what's intended?

kegsay (changes requested):
Mostly LGTM - how has this been tested?

clientapi/routing/server_notices.go
S7evinK: Right, forgot about moving this to `routing.go`.

S7evinK:
> Mostly LGTM - how has this been tested?

Since there's no spec for this, mostly manually with a local sqlite database and running curl with the required data.

kegsay:
Hmm, we can do better than that I think. I'll see if I can hook up admin user support in Complement and then we can actually test this in CI.

kegsay:
I'll work to add admin support in Complement today so we can test then merge this.

kegsay:
Tests can now be added to Complement.

clientapi/routing/leaveroom.go
```diff
@@ -41,7 +41,7 @@ func LeaveRoomByID(
 		if leaveRes.Code != 0 {
 			return util.JSONResponse{
 				Code: leaveRes.Code,
-				JSON: jsonerror.Forbidden(leaveRes.Message),
+				JSON: jsonerror.LeaveServerNoticeError(),
```
kegsay, line 44: There's no guarantee that this is because of server notices though?

S7evinK: The `PerformLeaveResponse` struct just got the `Code` in [2e1e0c81a19830cbf2055a6d15c5d7424c2d0087](https://github.com/matrix-org/dendrite/pull/2180/commits/2e1e0c81a19830cbf2055a6d15c5d7424c2d0087#diff-7c6dd331dc0262e2eb97b598ef5d4a8cbb70050a4291249d05af21dfc26e0086), so right now it should be guaranteed.

kegsay: That's not very extensible or kind to the next person who wants to add an additional code :) it would be nicer to just proxy back the JSON blob to send back so instead of `Message string` having `Message interface{}` which is JSON serialisable should work.

kegsay (approved):
This is looking fantastic, and the passing tests gives me much confidence this PR does what we think it should do. Excellent work!

PR #2186, opened 2022-02-14
Merge both user API databases into one

neilalexander:
This folds the device database into the account database, so that the user API just has a single database.

This also adds code to ensure that old device list entries in the key server are cleaned up if the corresponding device entries in the user API no longer exist.


keyserver/storage/interface.go
```diff
@@ -53,7 +53,7 @@ type Database interface {
 
 	// DeviceKeysForUser returns the device keys for the device IDs given. If the length of deviceIDs is 0, all devices are selected.
 	// If there are some missing keys, they are omitted from the returned slice. There is no ordering on the returned slice.
-	DeviceKeysForUser(ctx context.Context, userID string, deviceIDs []string) ([]api.DeviceMessage, error)
+	DeviceKeysForUser(ctx context.Context, userID string, deviceIDs []string, includeEmpty bool) ([]api.DeviceMessage, error)
```
kegsay, line 56: Flags like this aren't clear from callsites which makes it harder to know which to use. `includeEmpty` is also somewhat vague. What is an empty device?

keyserver/internal/internal.go
```diff
@@ -554,10 +554,60 @@ func (a *KeyInternalAPI) populateResponseWithDeviceKeysFromDatabase(
...
+		return
+	}
+
+	// Work out whether we have device keys in the keyserver for devices that
+	// no longer exist in the user API. This is mostly an exercise to ensure
+	// that we keep some integrity between the two.
+	var toClean []gomatrixserverlib.KeyID
+	for _, k := range existingKeys {
+		if _, ok := existingDeviceMap[k.DeviceID]; !ok {
+			toClean = append(toClean, gomatrixserverlib.KeyID(k.DeviceID))
```
kegsay, line 592: KeyID != DeviceID I thought..? Why are we typecasting?

keyserver/internal/internal.go
```diff
@@ -554,10 +554,60 @@ func (a *KeyInternalAPI) populateResponseWithDeviceKeysFromDatabase(
...
+		}
+	}
+
+	if len(toClean) > 0 {
+		if err = a.DB.DeleteDeviceKeys(ctx, req.UserID, toClean); err != nil {
+			res.Error = &api.KeyError{
+				Err: fmt.Sprintf("failed to clean device keys: %s", err.Error()),
```
kegsay, line 599: Should this be fatal? I don't think it should?

keyserver/storage/sqlite3/device_keys_table.go
```diff
@@ -52,6 +52,9 @@ const selectDeviceKeysSQL = "" +
 const selectBatchDeviceKeysSQL = "" +
 	"SELECT device_id, key_json, stream_id, display_name FROM keyserver_device_keys WHERE user_id=$1 AND key_json <> ''"
 
+const selectBatchDeviceKeysWithEmptiesSQL = "" +
+	"SELECT device_id, key_json, stream_id, display_name FROM keyserver_device_keys WHERE user_id=$1"
```
kegsay, line 56: Given `key_json TEXT NOT NULL` why do we need this and why is `selectBatchDeviceKeysSQL` not suitable?

userapi/storage/postgres/deltas/20201001204705_last_seen_ts_ip.go
```diff
@@ -5,13 +5,8 @@ import (
 	"fmt"
 
 	"github.com/matrix-org/dendrite/internal/sqlutil"
-	"github.com/pressly/goose"
 )
 
-func LoadFromGoose() {
```
kegsay, line 11: This reminds me: we need to just nuke these from all the deltas and remove `/cmd/goose`. In practice, no one uses it and hence we shouldn't carry the baggage for it. The intention was to allow people to rollback DB changes using goose.

userapi/storage/postgres/storage.go
```diff
@@ -47,14 +49,23 @@ type Database struct {
...
 	keyBackups            keyBackupStatements
 	serverName            gomatrixserverlib.ServerName
 	bcryptCost            int
 	openIDTokenLifetimeMS int64
 }
 
+const (
+	// The length of generated device IDs
+	deviceIDByteLength   = 6
+	loginTokenByteLength = 32
```
kegsay, line 64: Is it possible to merge `storage/postgres/storage.go` and `storage/sqlite/storage.go` into a `storage/shared/storage.go` like we do elsewhere? There's a lot of logic here which is duplicated which is sadness making.

kegsay (approved):

keyserver/storage/interface.go
neilalexander: Honestly, I was surprised at the behaviour. It seems we create device key entries that just contain an empty key JSON when the device gets created. Then for some reason we filter them out until they get populated later by a key upload by the client. My guess is that it was so that the two tables stay roughly in sync between the user API and the key server, but I don't know why we bother with that — your guess is as good as mine.

keyserver/internal/internal.go
neilalexander: Because apparently the function signature to delete device IDs is `DeleteDeviceKeys(ctx context.Context, userID string, deviceIDs []gomatrixserverlib.KeyID) error` because that bleeds through from `SelectCrossSigningSigsForTarget`, `DeleteCrossSigningSigsForTarget` etc where the device ID is a "key ID". We can probably just stringify it right the way through in a future PR.

keyserver/storage/sqlite3/device_keys_table.go
neilalexander: It's `NOT NULL` but that doesn't prevent it from being `EMPTY` which apparently it is when the device is first created (as earlier comment). I agree it's a weird shape but it's pre-existing.

userapi/storage/postgres/storage.go
neilalexander: Yep, I'll do so in a follow up PR though to keep it easier to follow.

userapi/storage/postgres/deltas/20201001204705_last_seen_ts_ip.go
neilalexander: Agreed.

keyserver/storage/interface.go
kegsay: iirc we need some way to track deleted devices I think, which might be what this is.

keyserver/storage/sqlite3/device_keys_table.go
kegsay: Yeah I think I remember why now - deleted devices need to be propogated over federation and hence need a stream ID etc, so an empty device is the tombstone.

userapi/storage/postgres/storage.go
kegsay: ⭐

PR #2188, opened 2022-02-15
Add consent tracking

S7evinK:
This adds the unspecified consent tracking.
Most parts are documented [here](https://github.com/matrix-org/synapse/blob/develop/docs/consent_tracking.md)

- When creating a new account, the user now has to accept the specified terms an services (or what ever you call it via the config)
- If enabled after the account was created, the user is blocked from sending any messages/joining rooms and will receive a message to accept the terms before being able to do anything
- Asking the user to accept the policy (e.g. via server notices) can be done with #2180. (once merged, needs additional work)


clientapi/routing/server_notices.go
```diff
@@ -266,7 +262,36 @@ func SendServerNotice(
...
+	userRooms := api.QueryRoomsForUserResponse{}
+	if err := rsAPI.QueryRoomsForUser(ctx, &api.QueryRoomsForUserRequest{
+		UserID:         userID,
+		WantMembership: "join",
+	}, &userRooms); err != nil {
+		return nil, err
+	}
+	allUserRooms = append(allUserRooms, userRooms.RoomIDs...)
+	// get invites for specified user
+	if err := rsAPI.QueryRoomsForUser(ctx, &api.QueryRoomsForUserRequest{
```
neilalexander, line 279: My feeling is that duplicate API calls like this will end up getting expensive on a busy server — we may just want to modify `rsAPI.QueryRoomsForUser` so that if we don't specify `WantMembership`, we just get all rooms regardless of membership.

docs/templates/privacy/1.0.gohtml
```diff
@@ -0,0 +1,26 @@
+<!doctype html>
+<html lang="en">
+<head>
+    <title>Matrix.org Privacy policy</title>
```
neilalexander, line 4: We may want to populate this file with more neutral server names / wording, i.e. not `matrix.org`.

internal/httputil/httpapi.go
```diff
@@ -49,7 +55,10 @@ type BasicAuth struct {
 
 // MakeAuthAPI turns a util.JSONRequestHandler function into an http.Handler which authenticates the request.
 func MakeAuthAPI(
-	metricsName string, userAPI userapi.UserInternalAPI,
+	metricsName string,
+	userAPI userapi.UserInternalAPI,
+	userConsentCfg config.UserConsentOptions,
+	requireConsent bool,
```
neilalexander, line 61: From some call sites in the client API it's not really clear what this bool does. Can we type & const it somewhere?
```
type Consent bool

const (
  ConsentRequired Consent = true
  ConsentNotRequired Consent = false
)
```

setup/flags.go
```diff
@@ -43,7 +43,8 @@ func ParseFlags(monolith bool) *config.Dendrite {
 	}
 
 	cfg, err := config.Load(*configPath, monolith)
-
+	// TODO: just for testing
```
neilalexander, line 46: Let's nuke this if we can and just specify consent-based URL stuff in the consent section of the config.

userapi/storage/postgres/deltas/2022021414375800_add_policy_version.go
```diff
@@ -0,0 +1,30 @@
...
+
+	"github.com/matrix-org/dendrite/internal/sqlutil"
+)
+
+func LoadAddPolicyVersion(m *sqlutil.Migrations) {
+	m.AddMigration(UpAddPolicyVersion, DownAddPolicyVersion)
+}
+
+func UpAddPolicyVersion(tx *sql.Tx) error {
+	_, err := tx.Exec("ALTER TABLE account_accounts ADD COLUMN IF NOT EXISTS policy_version TEXT;" +
```
neilalexander, line 15: Might be wiser to split this out into two `tx.Exec`s.

userapi/storage/sqlite3/deltas/2022021414375800_add_policy_version.go
```diff
@@ -0,0 +1,30 @@
...
+
+	"github.com/matrix-org/dendrite/internal/sqlutil"
+)
+
+func LoadAddPolicyVersion(m *sqlutil.Migrations) {
+	m.AddMigration(UpAddPolicyVersion, DownAddPolicyVersion)
+}
+
+func UpAddPolicyVersion(tx *sql.Tx) error {
+	_, err := tx.Exec("ALTER TABLE account_accounts ADD COLUMN policy_version TEXT;" +
```
neilalexander, line 15: Same in these.

setup/base/base.go
```diff
@@ -205,6 +205,7 @@ func NewBaseDendrite(cfg *config.Dendrite, componentName string, options ...Base
 		PublicKeyAPIMux:        mux.NewRouter().SkipClean(true).PathPrefix(httputil.PublicKeyPathPrefix).Subrouter().UseEncodedPath(),
 		PublicMediaAPIMux:      mux.NewRouter().SkipClean(true).PathPrefix(httputil.PublicMediaPathPrefix).Subrouter().UseEncodedPath(),
 		PublicWellKnownAPIMux:  mux.NewRouter().SkipClean(true).PathPrefix(httputil.PublicWellKnownPrefix).Subrouter().UseEncodedPath(),
+		PublicConsentAPIMux:    mux.NewRouter().SkipClean(true).PathPrefix("/_matrix").Subrouter().UseEncodedPath(),
```
neilalexander, line 208: This feels just a bit off — it's really a mux for the whole `/_matrix` namespace. Are we actually required to use `/_matrix/consent` specifically as the endpoint?

neilalexander (commented):
Overall looking good, just a few minor things.

clientapi/routing/server_notices.go
S7evinK: Added a new `MemberShipStateAll` const.

setup/base/base.go
S7evinK: Since it's not in the spec, we are free to decide where it serves. I just followed what Synapse did. :P

neilalexander: Maybe it's OK for it to live in the `PublicClientAPIMux` as `/_matrix/client/consent` then, and we don't have to pass around a router for the entirety of `/_matrix`, since that seems like a potential footgun waiting to happen.

dendrite-config.yaml
```diff
@@ -80,6 +80,35 @@ global:
     # The roomname to be used when creating messages
     room_name: "Server Alerts"
 
+  # Consent tracking configuration
+  user_consent:
+    # If the user consent tracking is enabled or not
+    enabled: false
+    # The base URL this homeserver will serve clients on, e.g. https://matrix.org
+    base_url: http://localhost
```
kegsay, line 88: I wonder if this should be elsewhere given it's surely specified in other places? Well-known?

dendrite-config.yaml
```diff
@@ -80,6 +80,35 @@ global:
...
+    # Randomly generated string to be used to calculate the HMAC
+    form_secret: "superSecretRandomlyGeneratedSecret"
```
kegsay, line 90: Can you put the suggested `pwgen` command here as a comment?

docs/templates/privacy/1.0.gohtml
```diff
@@ -0,0 +1,26 @@
...
+    <p>
+        Please give your consent to keep using this homeserver.
+    </p>
+    {{ if not .PublicVersion }}
+        <!-- The variables used here are only provided when the 'u' param is given to the homeserver -->
+        <form method="post" action="consent">
+            <input type="hidden" name="v" value="{{ .Version }}"/>
+            <input type="hidden" name="u" value="{{ .User }}"/>
+            <input type="hidden" name="h" value="{{ .UserHMAC }}"/>
+            <input type="submit" value="Sure thing!"/>
```
kegsay, line 21: This should probably be more formal - e.g "I consent"

internal/httputil/httpapi.go
```diff
@@ -47,9 +53,19 @@ type BasicAuth struct {
...
+	ConsentNotRequired Consent = false
+)
+
 // MakeAuthAPI turns a util.JSONRequestHandler function into an http.Handler which authenticates the request.
 func MakeAuthAPI(
-	metricsName string, userAPI userapi.UserInternalAPI,
+	metricsName string,
+	userAPI userapi.UserInternalAPI,
+	userConsentCfg config.UserConsentOptions,
+	requireConsent Consent,
```
kegsay, line 68: I would be very tempted to use functional options here so we don't need to constantly provide these 2 options.

internal/httputil/httpapi.go
```diff
@@ -89,6 +115,56 @@ func MakeAuthAPI(
...
+	res := &userapi.QueryPolicyVersionResponse{}
+	err = userAPI.QueryPolicyVersion(ctx, &userapi.QueryPolicyVersionRequest{
+		LocalPart: localPart,
+	}, res)
+	if err != nil {
+		return nil
+	}
+
+	// user hasn't accepted any policy, block access.
+	if userConsentCfg.Version != res.PolicyVersion {
```
kegsay, line 133: How does this work for rollbacks? E.g 1.2 != 1.0 but they have consented to a newer version?

internal/httputil/httpapi.go
```diff
@@ -89,6 +115,56 @@ func MakeAuthAPI(
...
+		}
+		msg := &bytes.Buffer{}
+		c := struct {
+			ConsentURL string
+		}{
+			ConsentURL: uri,
+		}
+		if err = userConsentCfg.TextTemplates.ExecuteTemplate(msg, "blockEventsError", c); err != nil {
+			logrus.Infof("error consent message: %+v", err)
+			return jsonerror.Unknown("unable to get consent URL")
```
kegsay, line 146: Fix up the error messages on this please. Failure to execute template would be better.

internal/httputil/httpapi.go
```diff
@@ -89,6 +115,56 @@ func MakeAuthAPI(
...
+	mac := hmac.New(sha256.New, []byte(config.FormSecret))
+	_, err := mac.Write([]byte(username))
+	if err != nil {
+		return "", err
+	}
+	hmac := hex.EncodeToString(mac.Sum(nil))
+
+	return fmt.Sprintf(
+		"%s/_matrix/client/consent?u=%s&h=%s&v=%s",
+		config.BaseURL, username, hmac, config.Version,
```
kegsay, line 164: Do we need to query encode any of this?

setup/config/config_global.go
```diff
@@ -225,3 +234,80 @@ func (c *DNSCacheOptions) Verify(configErrs *ConfigErrors, isMonolith bool) {
...
+}
+
+func (c *UserConsentOptions) Defaults() {
+	c.Enabled = false
+	c.RequireAtRegistration = false
+	c.SendServerNoticeToGuest = false
+	c.PolicyName = "Privacy Policy"
+	c.Version = "1.0"
+	c.TemplateDir = "./templates/privacy"
+	c.BaseURL = "http://localhost"
```
kegsay, line 278: I wouldn't set this otherwise people could naively set "enabled: true" in the config and then it won't fall over with a missing base URL.

setup/config/testdata/privacy/1.0.gohtml
```diff
@@ -0,0 +1,26 @@
+<!doctype html>
+<html lang="en">
+<head>
+    <title>Matrix.org Privacy policy</title>
```
kegsay, line 4: Don't refer to matrix.org in the default privacy policy please.

userapi/storage/sqlite3/accounts_table.go
```diff
@@ -94,23 +114,27 @@ func NewSQLiteAccountsTable(db *sql.DB, serverName gomatrixserverlib.ServerName)
...
 	createdTimeMS := time.Now().UnixNano() / 1000000
 	stmt := s.insertAccountStmt
 
 	var err error
 	if accountType != api.AccountTypeAppService {
-		_, err = sqlutil.TxStmt(txn, stmt).ExecContext(ctx, localpart, createdTimeMS, hash, nil, accountType)
+		_, err = sqlutil.TxStmt(txn, stmt).ExecContext(ctx, localpart, createdTimeMS, hash, nil, accountType, policyVersion)
 	} else {
-		_, err = sqlutil.TxStmt(txn, stmt).ExecContext(ctx, localpart, createdTimeMS, hash, appserviceID, accountType)
+		_, err = sqlutil.TxStmt(txn, stmt).ExecContext(ctx, localpart, createdTimeMS, hash, appserviceID, accountType, "")
```
kegsay, line 137: I would continue to pass in `policyVersion` here - else we're baking in too many places that AS users don't accept policies.

roomserver/storage/tables/interface.go
```diff
@@ -113,6 +113,7 @@ type Invites interface {
 type MembershipState int64
 
 const (
+	MemberShipStateAll        MembershipState = 0
```
kegsay, line 116: No capital S - `MembershipStateAll`

clientapi/routing/server_notices.go
```diff
@@ -85,41 +85,38 @@ func SendServerNotice(
...
-		return util.ErrorResponse(err)
-	}
-	allUserRooms = append(allUserRooms, userRooms.RoomIDs...)
-	// get left rooms for specified user
-	if err := rsAPI.QueryRoomsForUser(ctx, &api.QueryRoomsForUserRequest{
-		UserID:         r.UserID,
-		WantMembership: "leave",
-	}, &userRooms); err != nil {
-		return util.ErrorResponse(err)
+	allUserRooms, err := getAllUserRooms(ctx, rsAPI, serverNoticeRequest.UserID)
```
kegsay, line 116: As discussed, it would be nicer if we just tracked a single room ID field for the server notice room per-user and ensure the user is invited/joined to that room always as having to pull out thousands of rooms to then diff with shared rooms in common isn't great

clientapi/routing/consent_tracking.go
```diff
@@ -0,0 +1,226 @@
...
+	consentOpts := cfgClient.Matrix.UserConsentOptions
+
+	mac := hmac.New(sha256.New, []byte(consentOpts.FormSecret))
+	_, err := mac.Write([]byte(userID))
+	if err != nil {
+		return "", err
+	}
+	userMAC := mac.Sum(nil)
+
+	return fmt.Sprintf("%s/_matrix/client/consent?u=%s&h=%s&v=%s", cfgClient.Matrix.UserConsentOptions.BaseURL, userID, userMAC, consentOpts.Version), nil
```
kegsay, line 211: Do we need to query encode?

clientapi/routing/consent_tracking.go
```diff
@@ -0,0 +1,226 @@
...
+
+	if len(res.OutdatedUsers) > 0 {
+		logrus.WithField("count", len(res.OutdatedUsers)).Infof("Sending server notice to users who have not yet accepted the policy")
+	}
+
+	for _, userID := range res.OutdatedUsers {
+		if userID == cfgClient.Matrix.ServerNotices.LocalPart {
+			continue
+		}
+		userID = fmt.Sprintf("@%s:%s", userID, cfgClient.Matrix.ServerName)
```
kegsay, line 157: Hang on, we are already dealing with a user ID? Unless you mean `res.OutdatedUsers` are in fact localparts? If so, this needs to be made clear in variable names and field names. "Users" is not a helpful term as it's neither "user ID" nor "localpart".

clientapi/routing/consent_tracking.go
```diff
@@ -0,0 +1,226 @@
...
+	// The data used to populate the /consent request
+	data := constentTemplateData{
+		User:     req.FormValue("u"),
+		Version:  req.FormValue("v"),
+		UserHMAC: req.FormValue("h"),
+	}
+	switch req.Method {
+	case http.MethodGet:
+		// display the privacy policy without a form
+		data.PublicVersion = data.User == "" || data.UserHMAC == "" || data.Version == ""
```
kegsay, line 59: `PublicVersion` isn't a great name. AFAICT it seems to control whether there is an accept button or not. So perhaps "read only" would be better.

clientapi/routing/consent_tracking.go
```diff
@@ -0,0 +1,226 @@
...
+			&userapi.UpdatePolicyVersionResponse{},
+		); err != nil {
+			_, err = writer.Write([]byte("unable to update database"))
+			if err != nil {
+				logrus.WithError(err).Error("unable to write to database")
+			}
+			return &internalError
+		}
+		// display the privacy policy without a form
+		data.PublicVersion = false
```
kegsay, line 114: Why do this here though? I guess to hide the form after they hit accept?

kegsay (changes requested):
Some more stuff to do here, mostly around naming and some around API design. We need to be using functional options for making auth APIs.

dendrite-config.yaml
S7evinK: I feel it's better to have this here, so it's clear that it belongs to the consent tracking, as an alternative I can move the `base_url` to global

internal/httputil/httpapi.go
S7evinK: Not sure how we'd want to fix this, as the version can be a random string. We could use [semver](https://pkg.go.dev/golang.org/x/mod/semver), but that would require prefixing the version with a leading "v"

> Package semver implements comparison of semantic version strings. In this package, semantic version strings must begin with a leading "v", as in "v1.0.0".

userapi/storage/postgres/accounts_table.go
```diff
@@ -92,24 +122,25 @@ func NewPostgresAccountsTable(db *sql.DB, serverName gomatrixserverlib.ServerNam
...
+	ctx context.Context, txn *sql.Tx, localpart, hash, appserviceID, policyVersion string, accountType api.AccountType,
 ) (*api.Account, error) {
 	createdTimeMS := time.Now().UnixNano() / 1000000
 	stmt := sqlutil.TxStmt(txn, s.insertAccountStmt)
 
-	var err error
-	if accountType != api.AccountTypeAppService {
-		_, err = stmt.ExecContext(ctx, localpart, createdTimeMS, hash, nil, accountType)
-	} else {
-		_, err = stmt.ExecContext(ctx, localpart, createdTimeMS, hash, appserviceID, accountType)
```
kegsay, line 111: Why did you remove this?

userapi/storage/postgres/accounts_table.go
```diff
@@ -178,3 +209,83 @@ func (s *accountsStatements) SelectNewNumericLocalpart(
...
 	return id + 1, err
 }
+
+// selectPrivacyPolicy gets the current privacy policy a specific user accepted
+func (s *accountsStatements) SelectPrivacyPolicy(
+	ctx context.Context, txn *sql.Tx, localPart string,
+) (policy string, err error) {
+	stmt := s.selectPrivacyPolicyStmt
+	if txn != nil {
+		stmt = sqlutil.TxStmt(txn, stmt)
```
kegsay, line 219: You don't need to `txn != nil` here, as `sqlutil.TxStmt(...)` does that.

userapi/storage/postgres/accounts_table.go
```diff
@@ -178,3 +209,83 @@ func (s *accountsStatements) SelectNewNumericLocalpart(
...
+	}
+	err = stmt.QueryRowContext(ctx, localPart).Scan(&policy)
+	return
```
kegsay, line 222: Handle `sql.ErrNoRows`. We probably don't want to produce an error here.

userapi/storage/postgres/accounts_table.go
```diff
@@ -178,3 +209,83 @@ func (s *accountsStatements) SelectNewNumericLocalpart(
...
+}
+
+// batchSelectPrivacyPolicy queries all users which didn't accept the current policy version
+func (s *accountsStatements) BatchSelectPrivacyPolicy(
+	ctx context.Context, txn *sql.Tx, policyVersion string,
+) (userIDs []string, err error) {
+	stmt := s.batchSelectPrivacyPolicyStmt
+	if txn != nil {
+		stmt = sqlutil.TxStmt(txn, stmt)
```
kegsay, line 231: Likewise, no need for the conditional.

userapi/storage/postgres/accounts_table.go
```diff
@@ -178,3 +209,83 @@ func (s *accountsStatements) SelectNewNumericLocalpart(
...
+	stmt := s.selectServerNoticeRoomStmt
+	if txn != nil {
+		stmt = sqlutil.TxStmt(txn, stmt)
+	}
+
+	roomIDNull := sql.NullString{}
+	row := stmt.QueryRowContext(ctx, localpart)
+	err = row.Scan(&roomIDNull)
+	if err != nil {
+		return "", err
```
kegsay, line 273: Handle `sql.ErrNoRows`. We probably don't want to produce an error here.

userapi/storage/sqlite3/accounts_table.go
```diff
@@ -94,24 +124,25 @@ func NewSQLiteAccountsTable(db *sql.DB, serverName gomatrixserverlib.ServerName)
...
+	ctx context.Context, txn *sql.Tx, localpart, hash, appserviceID, policyVersion string, accountType api.AccountType,
 ) (*api.Account, error) {
 	createdTimeMS := time.Now().UnixNano() / 1000000
 	stmt := s.insertAccountStmt
 
-	var err error
-	if accountType != api.AccountTypeAppService {
-		_, err = sqlutil.TxStmt(txn, stmt).ExecContext(ctx, localpart, createdTimeMS, hash, nil, accountType)
-	} else {
-		_, err = sqlutil.TxStmt(txn, stmt).ExecContext(ctx, localpart, createdTimeMS, hash, appserviceID, accountType)
```
kegsay, line 113: I don't understand why you removed this.

userapi/storage/sqlite3/accounts_table.go
```diff
@@ -183,3 +214,87 @@ func (s *accountsStatements) SelectNewNumericLocalpart(
...
+
+func (s *accountsStatements) SelectPrivacyPolicy(
+	ctx context.Context, txn *sql.Tx, localPart string,
+) (policy string, err error) {
+	stmt := s.selectPrivacyPolicyStmt
+	if txn != nil {
+		stmt = sqlutil.TxStmt(txn, stmt)
+	}
+	err = stmt.QueryRowContext(ctx, localPart).Scan(&policy)
+	return
```
kegsay, line 228: Same deal, handle `sql.ErrNoRows`.

kegsay (changes requested):

PR #2192, opened 2022-02-17
Raise waitTime for the DeviceListUpdater on failure

S7evinK:
Raises the `waitTime` for network related failures, otherwise you could get spammed with `failed to query device keys for some users` since the `waitTime` is set to 2 seconds.

### DIFF
--- a/keyserver/internal/device_list_update.go
+++ b/keyserver/internal/device_list_update.go
@@ -367,6 +367,9 @@ func (u *DeviceListUpdater) processServer(serverName gomatrixserverlib.ServerNam
 					waitTime = fcerr.RetryAfter
 				} else if fcerr.Blacklisted {
 					waitTime = time.Hour * 8
+				} else {
+					// For all other errors (DNS resolution, network etc.) wait 1 hour.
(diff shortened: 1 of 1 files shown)


keyserver/internal/device_list_update.go
```diff
@@ -367,6 +367,9 @@ func (u *DeviceListUpdater) processServer(serverName gomatrixserverlib.ServerNam
 					waitTime = fcerr.RetryAfter
 				} else if fcerr.Blacklisted {
 					waitTime = time.Hour * 8
+				} else {
+					// For all other errors (DNS resolution, network etc.) wait 1 hour.
+					waitTime = time.Hour
```
kegsay, line 372: How can this be hit at all when the error type is a `fedsenderapi.FederationClientError`?

S7evinK: The `FederationClientError` has a `RetryAfter` of `0` and `Blacklisted` is `false`. So no path here is hit.
This should probably be fixed in the `FederationClient` later on, this is just a quick fix.

kegsay (approved):

PR #2195, opened 2022-02-17
Don't re-send sent events in `add_state_events`

neilalexander:
The roomserver output events contain `add_state_events`, which were often including events that had already been sent to the output stream before. If that's the case, downstream components shouldn't need to see them additional times — referring to them by event ID in `adds_state_event_ids` and `removes_state_event_ids` should be enough.

### DIFF
--- a/roomserver/storage/postgres/events_table.go
+++ b/roomserver/storage/postgres/events_table.go
@@ -108,9 +108,6 @@ const updateEventStateSQL = "" +
 const selectEventSentToOutputSQL = "" +
 	"SELECT sent_to_output FROM roomserver_events WHERE event_nid = $1"
 
-const bulkSelectEventFilteredBySentToOutputSQL = "" +
-	"SELECT event_nid FROM roomserver_events WHERE event_nid = ANY($1) AND sent_to_output = $2"
-
 const updateEventSentToOutputSQL = "" +
 	"UPDATE roomserver_events SET sent_to_output = TRUE WHERE event_nid = $1"
 
@@ -130,6 +127,9 @@ const bulkSelectEventIDSQL = "" +
 const bulkSelectEventNIDSQL = "" +
 	"SELECT event_id, event_nid FROM roomserver_events WHERE event_id = ANY($1)"
 
+const bulkSelectUnsentEventNIDSQL = "" +
+	"SELECT event_id, event_nid FROM roomserver_events WHERE event_id = ANY($1) AND sent_to_output = FALSE"
+
 const selectMaxEventDepthSQL = "" +
 	"SELECT COALESCE(MAX(depth) + 1, 0) FROM roomserver_events WHERE event_nid = ANY($1)"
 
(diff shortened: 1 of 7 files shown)


roomserver/storage/shared/membership_updater.go
```diff
@@ -136,7 +136,7 @@ func (u *MembershipUpdater) SetToJoin(senderUserID string, eventID string, isUpd
 		}
 
 		// Look up the NID of the new join event
-		nIDs, err := u.d.eventNIDs(u.ctx, u.txn, []string{eventID})
+		nIDs, err := u.d.eventNIDs(u.ctx, u.txn, []string{eventID}, false)
```
kegsay, line 139: Please alias this, as it's hard to know what `false` means from call-sites. Especially since `false` means 'all' not 'only sent' events.
```go
var (
    NoFilter = false
    FilterUnsentOnly = true
)
```
or something.

roomserver/storage/postgres/events_table.go
```diff
@@ -460,8 +465,13 @@ func (s *eventStatements) BulkSelectEventID(ctx context.Context, txn *sql.Tx, ev
 
 // bulkSelectEventNIDs returns a map from string event ID to numeric event ID.
 // If an event ID is not in the database then it is omitted from the map.
-func (s *eventStatements) BulkSelectEventNID(ctx context.Context, txn *sql.Tx, eventIDs []string) (map[string]types.EventNID, error) {
-	stmt := sqlutil.TxStmt(txn, s.bulkSelectEventNIDStmt)
+func (s *eventStatements) BulkSelectEventNID(ctx context.Context, txn *sql.Tx, eventIDs []string, onlyUnsent bool) (map[string]types.EventNID, error) {
```
kegsay, line 468: Questioning whether it would just be better to make this private and have two functions: `BulkSelectEventNID` and `BulkSelectUnsentEventNID`. It would be clearer and not need a flag which is unintuitive. I don't mind having this boolean inside a private function call, so you don't need to duplicate code.

roomserver/storage/sqlite3/events_table.go
```diff
@@ -496,13 +501,18 @@ func (s *eventStatements) BulkSelectEventID(ctx context.Context, txn *sql.Tx, ev
...
 	iEventIDs := make([]interface{}, len(eventIDs))
 	for k, v := range eventIDs {
 		iEventIDs[k] = v
 	}
-	selectOrig := strings.Replace(bulkSelectEventNIDSQL, "($1)", sqlutil.QueryVariadic(len(iEventIDs)), 1)
+	var selectOrig string
+	if onlyUnsent {
+		selectOrig = strings.Replace(bulkSelectUnsentEventNIDSQL, "($1)", sqlutil.QueryVariadic(len(iEventIDs)), 1)
+	} else {
+		selectOrig = strings.Replace(bulkSelectEventNIDSQL, "($1)", sqlutil.QueryVariadic(len(iEventIDs)), 1)
```
kegsay, line 514: This reminds me we need to chunkify this on sqlite (and even postgres actually...) - https://github.com/matrix-org/sliding-sync/blob/main/sqlutil/sql.go#L49 is what I ended up doing for sliding sync and it worked well. Not for now though.

kegsay (approved):
I question the API shape here. Having 2 functions is clearer than having behaviour flip based on a boolean which itself is inconsistent. Otherwise LGTM.

PR #2197, opened 2022-02-17
Listen for /v3 on CSAPI

kegsay:
See https://github.com/matrix-org/complement/pull/305

### DIFF
--- a/clientapi/routing/routing.go
+++ b/clientapi/routing/routing.go
@@ -117,6 +117,11 @@ func Setup(
 		).Methods(http.MethodGet, http.MethodPost, http.MethodOptions)
 	}
 
+	// You can't just do PathPrefix("/(r0|v3)") because regexps only apply when inside named path variables.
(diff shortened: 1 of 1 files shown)


clientapi/routing/routing.go
```diff
@@ -117,15 +117,16 @@ func Setup(
 		).Methods(http.MethodGet, http.MethodPost, http.MethodOptions)
 	}
 
-	r0mux := publicAPIMux.PathPrefix("/r0").Subrouter()
+	v3mux := publicAPIMux.PathPrefix("/{version:(?:r0|v3)}/").Subrouter()
```
neilalexander, line 120: Can we please just stick a comment here explaining the regex? 

> "make a named path var called version and give it the matcher for a non-capturing group (because everything in the path var is 'captured' so having additional named capture groups makes no sense) which then selects r0 OR v3, with a trailing slash to stop v333 matching

neilalexander (approved):

PR #2199, opened 2022-02-18
Allow preventing guest registration

neilalexander:
This PR adds the new config option `guests_disabled` in the client API and also ensures that guests can't be created if `registration_disabled` is set too.

### DIFF
--- a/dendrite-config.yaml
+++ b/dendrite-config.yaml
@@ -142,6 +142,10 @@ client_api:
   # using the registration shared secret below.
   registration_disabled: false
 
+  # Prevents new guest accounts from being created. Guest registration is also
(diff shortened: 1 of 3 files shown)


dendrite-config.yaml
```diff
@@ -142,6 +142,10 @@ client_api:
   # using the registration shared secret below.
   registration_disabled: false
 
+  # Prevents new guest accounts from being created. Guest registration is also
+  # disabled implicitly by setting 'registration_disabled' above.
+  guests_disabled: false
```
kegsay, line 147: Maybe default to true?

kegsay (approved):

PR #2200, opened 2022-02-18
Remove event type and state key caches

neilalexander:
These caches may create races if a transaction that assigns them is rolled back, and honestly, they don't seem to make any measurable difference to event processing time anyway.

### DIFF
--- a/internal/caching/cache_roomservernids.go
+++ b/internal/caching/cache_roomservernids.go
@@ -7,14 +7,6 @@ import (
 )
 
 const (
-	RoomServerStateKeyNIDsCacheName       = "roomserver_statekey_nids"
-	RoomServerStateKeyNIDsCacheMaxEntries = 1024
-	RoomServerStateKeyNIDsCacheMutable    = false
-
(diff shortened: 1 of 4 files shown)


kegsay (commented):
Is removal really the right thing to do here? Not just write the nids before getting into a transaction? Is it really not making a performance impact?

kegsay (approved):
LGTM because it helps fix an actual problem, but I'm concerned there will be a performance impact.

neilalexander:
From Matrix:

> The main problem is that I'm not actually sure the cache helps much. We store things into the caches when we assign new state key NIDs and event types, but it's quite possible they never get used again. Then something like state res comes along for a big room like HQ and evicts probably most if not all of the existing cache entries to make way for the ones it wants to load at that point in time, and then we have to pull things out of the database for BAU event processing again afterwards
> (well, I suppose that applies much more so to the event state key NID cache than the event type cache which is relatively stable)
> So I think at the moment it may be worth just removing it and seeing if there is a real world performance penalty for it, and then if we think there is, we do something about it properly that's not just a dumb LRU cache that suffers churn and mass evictions every time state res happens in a big room

PR #2207, opened 2022-02-21
Implement `/context`

S7evinK:
I'm not totally sure if that's the expected behavior (e.g. when filtering events and applying filters), Sytests seems to be happy, though.
Also for the pagination tokens: They are in the spec, but I don't see a way they are actually used when making a request to `/context`.


syncapi/storage/postgres/output_room_events_table.go
```diff
@@ -130,6 +130,28 @@ const selectStateInRangeSQL = "" +
...
+const selectContextAfterEventSQL = "" +
+	"SELECT id, headered_event_json FROM syncapi_output_room_events WHERE room_id = $1 AND id > $2" +
+	" AND ( $4::text[] IS NULL OR     sender  = ANY($4)  )" +
+	" AND ( $5::text[] IS NULL OR NOT(sender  = ANY($5)) )" +
+	" AND ( $6::text[] IS NULL OR     type LIKE ANY($6)  )" +
+	" AND ( $7::text[] IS NULL OR NOT(type LIKE ANY($7)) )" +
+	" ORDER BY id ASC LIMIT $3"
+
+const selectEventIDsAfterSQL = "" +
+	"SELECT event_id FROM syncapi_output_room_events WHERE room_id = $1 AND id > $2"
```
neilalexander, line 153: I would expect this to be `LIMIT`ed?

syncapi/storage/sqlite3/output_room_events_table.go
```diff
@@ -85,19 +88,36 @@ const selectStateInRangeSQL = "" +
...
+// WHEN, ORDER BY and LIMIT are appended by prepareWithFilters
 
 const deleteEventsForRoomSQL = "" +
 	"DELETE FROM syncapi_output_room_events WHERE room_id = $1"
 
+const selectContextEventSQL = "" +
+	"SELECT id, headered_event_json FROM syncapi_output_room_events WHERE room_id = $1 AND event_id = $2"
+
+const selectContextBeforeEventSQL = "" +
+	"SELECT headered_event_json FROM syncapi_output_room_events WHERE room_id = $1 AND id < $2"
```
neilalexander, line 101: Might be good to just pop the `// WHEN, ORDER BY and LIMIT are appended by prepareWithFilters` comment on these too, for clarity.

syncapi/routing/context.go
```diff
@@ -0,0 +1,189 @@
...
+	filter := &gomatrixserverlib.RoomEventFilter{Limit: 10}
+
+	l := req.URL.Query().Get("limit")
+	f := req.URL.Query().Get("filter")
+	if l != "" {
+		limit, err := strconv.Atoi(l)
+		if err != nil {
+			return nil, err
+		}
+		// not in the spec, but feels like a good idea to have an upper bound limit
```
neilalexander, line 176: We normally annotate these things as `// NOTSPEC:` just so that we can find them later :-)

syncapi/routing/context.go
```diff
@@ -0,0 +1,189 @@
...
+		Senders:                 filter.Senders,
+		Types:                   filter.Types,
+		LazyLoadMembers:         filter.LazyLoadMembers,
+		IncludeRedundantMembers: filter.IncludeRedundantMembers,
+		NotRooms:                filter.NotRooms,
+		Rooms:                   filter.Rooms,
+		ContainsURL:             filter.ContainsURL,
+	}
+
+	state, _ := syncDB.CurrentState(ctx, roomID, &stateFilter, nil)
```
neilalexander, line 85: We should probably comment somewhere here about the fact that this is checking the *current* room history visibility rules, rather than the state at the time of the event. I think for now that's probably acceptable since we don't fully implement history visibility anyway, but we might want to `TODO` that for fixing later.

syncapi/storage/postgres/output_room_events_table.go
S7evinK: Removed, I was playing around with using the returned event ids as the excludeEventsIDs for `syncDB.CurrentState`

neilalexander (approved):
LGTM!

Issue #2181, opened 2022-02-11
stuck in a #dendrite:matrix.org related loop, heavy resource usage


bones-was-here:
### Background information

- **Dendrite version or git SHA**: 0.6.3 and 5106cc807cf22a95420b24f6bfdd5c9ac8aa06de
- **Monolith or Polylith?**: monolith
- **SQLite3 or Postgres?**: postgres
- **Running in Docker?**: no
- **`go version`**: 1.17.6
- **Client used (if applicable)**: element, curl

### Description

 - **What** is the problem: dendrite is stuck in a loop printing the same errors
 - **Who** is affected: 
 - **How** is this bug manifesting: steadily increasing ram and jetstream storage size
 - **When** did this first appear: some time after updating to 0.6.3

### Steps to reproduce

```
level=error msg="syncapi: failed to QuerySharedUsers for key change event from key server" error="found 2577 users but only have state key nids for 2350 of them"
level=error msg="syncapi: failed to QuerySharedUsers for key change event from key server" error="found 2448 users but only have state key nids for 2221 of them"
```
- it has something to do with #dendrite:matrix.org because that's the only room with that many users that this server has participated in
- these print at about 8 lines per second continuously
- restarting and deleting ~/jetstream have no effect
- ~/jetstream is growing steadily
- joining and leaving #dendrite:matrix.org has no effect
- forgetting the room has no effect #2176 

<!--
Describe how what happens differs from what you expected.

If you can identify any relevant log snippets from server logs, please include
those (please be careful to remove any personal or private data). Please surround them with
``` (three backticks, on a line on their own), so that they are formatted legibly.

Alternatively, please send logs to @kegan:matrix.org or @neilalexander:matrix.org
with a link to the respective Github issue, thanks!
-->

**Comment by bones-was-here** [comment]
Eventually it panics, restarts, and continues the loop

**Comment by bones-was-here** [comment]
It eventually stopped after many hours and several GB of jetstream dir.  Now I can restart it and delete jetstream and it continues behaving.  Have not attempted to rejoin #dendrite:matrix.org :)

**Comment by neilalexander** [comment]
Can you please run this query against your roomserver database?
```
SELECT COUNT(target_nid) FROM roomserver_membership AS m WHERE NOT EXISTS (
	SELECT event_state_key_nid FROM roomserver_event_state_keys AS s
	WHERE m.target_nid = s.event_state_key_nid
);
```

bones-was-here:
[...]

PR #2213, opened 2022-02-21
Remove error when state keys are missing for user NIDs

neilalexander:
There is still an actual bug here somewhere in the membership updater, but this check does more harm than good, since it means that the key consumers don't actually distribute notifications to *anyone* via `/sync`. It's better just to deal with this silently for now.

Created issue #2214 to track the root of the problem.


neilalexander:
related: #2181

kegsay (approved):

PR #2216, opened 2022-02-22
Remote banned user is kicked and may not rejoin until unbanned

S7evinK:


clientapi/routing/state.go
```diff
@@ -235,7 +235,7 @@ func OnIncomingStateTypeRequest(
 		}
 		// If the user has never been in the room then stop at this point.
 		// We won't tell the user about a room they have never joined.
-		if !membershipRes.HasBeenInRoom {
+		if !membershipRes.HasBeenInRoom || membershipRes.Membership == "ban" {
```
neilalexander, line 238: Please use the `gomatrixserverlib.Ban` const here instead.

neilalexander (approved):

PR #2219, opened 2022-02-22
Add `DeviceKeysEqual`

neilalexander:
This just makes it a bit clearer for working out when we should be emitting device list updates via output key change events.


kegsay (changes requested):
Please can we not have a 3-way result (true, false, error) for what really should be 2-way (true, false). I would just remove the `error` response part.

PR #2221, opened 2022-02-22
Set powerlevel.invite to 0, if `PowerLevelContentOverride` is empty

S7evinK:
Synapse sets the default `powerlevel_content_override` to `{ "invite": 0 }` for `private_chat` and `trusted_private_chat`, but overwrites them when they are set by the user. ([source](https://github.com/matrix-org/synapse/blob/6d14b3dabfe38c6ae487d0f663e294056b6cc056/synapse/handlers/room.py#L120-L133))

This solves a timeout in Sytest when trying to invite to a room with the "wrong" powerlevel.invite of 50

### DIFF
--- a/clientapi/routing/createroom.go
+++ b/clientapi/routing/createroom.go
@@ -227,6 +227,13 @@ func createRoom(
 		HistoryVisibility: historyVisibilityShared,
 	}
 
+	// NOTSPEC: The powerlevel for invite should be set to 50, according to the spec, but synapse sets it to 0 for invite.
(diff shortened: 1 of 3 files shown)


neilalexander (commented):
I’m wondering whether it is possible to fix the test rather than introduce not-spec behaviour into Dendrite if we can avoid it?

S7evinK:
> I’m wondering whether it is possible to fix the test rather than introduce not-spec behaviour into Dendrite if we can avoid it?

Yep, probably the better idea. Will create a PR for Sytest tomorrow.

Issue #2214, opened 2022-02-21
Membership updater sometimes populates `sender_nid` and `target_nid`s for which no state key entries are assigned


neilalexander:
I suspect this may be something to do with invites.

```
SELECT * FROM roomserver_membership AS m WHERE NOT EXISTS (
	SELECT event_state_key_nid FROM roomserver_event_state_keys AS s
	WHERE m.sender_nid = s.event_state_key_nid
);
```

```
SELECT * FROM roomserver_membership AS m WHERE NOT EXISTS (
	SELECT event_state_key_nid FROM roomserver_event_state_keys AS s
	WHERE m.target_nid = s.event_state_key_nid
);
```

PR #2224, opened 2022-02-23
Relax roomserver input transactional isolation

neilalexander:
This was originally done in #2141 to ensure crash safety and to restrict the amount of database resources that a single room can take up while processing input events.

Unfortunately, it also created other problems, like when assigning state event NIDs — I'm pretty sure this is probably the cause of #2214. 

This PR therefore relaxes it so that we only perform event state calculation, latest events and membership updates in room-locked transactions instead.

### DIFF
--- a/roomserver/storage/shared/storage.go
+++ b/roomserver/storage/shared/storage.go
@@ -677,8 +677,12 @@ func (d *Database) GetPublishedRooms(ctx context.Context) ([]string, error) {
 func (d *Database) MissingAuthPrevEvents(
 	ctx context.Context, e *gomatrixserverlib.Event,
 ) (missingAuth, missingPrev []string, err error) {
+	authEventNIDs, err := d.EventNIDs(ctx, e.AuthEventIDs())
+	if err != nil {
+		return nil, nil, fmt.Errorf("d.EventNIDs: %w", err)
+	}
 	for _, authEventID := range e.AuthEventIDs() {
-		if nids, err := d.EventNIDs(ctx, []string{authEventID}); err != nil || len(nids) == 0 {
+		if _, ok := authEventNIDs[authEventID]; !ok {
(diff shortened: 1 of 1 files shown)


roomserver/internal/input/input.go
```diff
@@ -127,11 +112,8 @@ func (r *Inputer) Start() error {
 						"event_id": inputRoomEvent.Event.EventID(),
 						"type":     inputRoomEvent.Event.Type(),
 					}).Warn("Roomserver failed to process async event")
-				}
-				switch action {
-				case retryLater:
-					_ = msg.Nak()
-				case doNotRetry:
+					_ = msg.Term()
```
kegsay, line 115: What does term mean again?

kegsay (approved):

roomserver/storage/shared/storage.go
```diff
@@ -674,6 +674,25 @@ func (d *Database) GetPublishedRooms(ctx context.Context) ([]string, error) {
 	return d.PublishedTable.SelectAllPublishedRooms(ctx, nil, true)
 }
 
+func (d *Database) MissingAuthPrevEvents(
+	ctx context.Context, e *gomatrixserverlib.Event,
+) (missingAuth, missingPrev []string, err error) {
+	for _, authEventID := range e.AuthEventIDs() {
+		if nids, err := d.EventNIDs(ctx, []string{authEventID}); err != nil || len(nids) == 0 {
```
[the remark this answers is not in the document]
neilalexander: For `EventNIDs` we probably can, since that should just omit anything missing, but we can't for `StateAtEventIDs` since that errors if anything is missing.

roomserver/internal/input/input.go
neilalexander: "Terminate" — don't redeliver us this message ever again.

PR #2225, opened 2022-02-24
Remember parameters on registration

S7evinK:
This remembers registration parameters using the session ID.
After registration or a timeout (currently set to 10min) the session ID is cleaned up.

### DIFF
--- a/clientapi/routing/register.go
+++ b/clientapi/routing/register.go
@@ -72,14 +72,19 @@ func init() {
 // sessionsDict keeps track of completed auth stages for each session.
 // It shouldn't be passed by value because it contains a mutex.
 type sessionsDict struct {
-	sync.Mutex
+	sync.RWMutex
 	sessions map[string][]authtypes.LoginType
+	params   map[string]registerRequest
+	timer    map[string]*time.Timer
 }
 
-// GetCompletedStages returns the completed stages for a session.
-func (d *sessionsDict) GetCompletedStages(sessionID string) []authtypes.LoginType {
-	d.Lock()
-	defer d.Unlock()
+// defaultTimeout is the timeout used to
+const defaultTimeOut = time.Minute * 10
+
+// getCompletedStages returns the completed stages for a session.
+func (d *sessionsDict) getCompletedStages(sessionID string) []authtypes.LoginType {
+	d.RLock()
+	defer d.RUnlock()
 
 	if completedStages, ok := d.sessions[sessionID]; ok {
 		return completedStages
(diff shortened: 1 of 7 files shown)


clientapi/routing/register.go
```diff
@@ -88,28 +93,75 @@ func (d *sessionsDict) GetCompletedStages(sessionID string) []authtypes.LoginTyp
 	return make([]authtypes.LoginType, 0)
 }
 
+// addParams adds a registerRequest to a sessionID and starts a timer to delete that registerRequest
+func (d *sessionsDict) addParams(sessionID string, r registerRequest) {
+	d.Lock()
+	d.Unlock()
```
neilalexander, line 99: Missing `defer`.

clientapi/routing/register.go
```diff
@@ -88,28 +93,75 @@ func (d *sessionsDict) GetCompletedStages(sessionID string) []authtypes.LoginTyp
...
 
-// AddCompletedSessionStage records that a session has completed an auth stage.
-func AddCompletedSessionStage(sessionID string, stage authtypes.LoginType) {
-	sessions.Lock()
-	defer sessions.Unlock()
+func (d *sessionsDict) startTimer(duration time.Duration, sessionID string) {
+	d.RLock()
+	defer d.RUnlock()
+	if _, ok := d.timer[sessionID]; !ok {
+		go func() {
```
neilalexander, line 137: Rather than trying to re-engineer, it's probably better to use `d.timer[sessionID] = timer.AfterFunc(time.Duration, f func())`.

Then if the timer already exists and the function is called again, stop and reset the timer?

clientapi/routing/register.go
```diff
@@ -88,28 +93,75 @@ func (d *sessionsDict) GetCompletedStages(sessionID string) []authtypes.LoginTyp
...
+// deleteSession cleans up a given session, either because the registration completed
+// successfully, or because a given timeout (default: 5min) was reached.
+func (d *sessionsDict) deleteSession(sessionID string) {
+	d.Lock()
+	defer d.Unlock()
+	delete(d.params, sessionID)
+	delete(d.sessions, sessionID)
+	// stop the timer, e.g. because the registration was completed
+	if t, ok := d.timer[sessionID]; ok {
+		t.Stop()
```
neilalexander, line 120: See notes at https://pkg.go.dev/time#Timer.Stop — you need to make sure the channel is drained.

clientapi/routing/register.go
```diff
@@ -72,14 +72,19 @@ func init() {
...
+	params   map[string]registerRequest
+	timer    map[string]*time.Timer
 }
 
-// GetCompletedStages returns the completed stages for a session.
-func (d *sessionsDict) GetCompletedStages(sessionID string) []authtypes.LoginType {
-	d.Lock()
-	defer d.Unlock()
+// defaultTimeout is the timeout used to
+const defaultTimeOut = time.Minute * 10
```
neilalexander, line 82: Comment is incomplete here, also 10 minutes to complete the registration flow seems rather high? A comment below suggests the default is 5 minutes.

neilalexander (changes requested):

clientapi/routing/register.go
S7evinK: Changed to `time.AfterFunc`, but at one point it just deadlocks when trying to drain the channel. It's commented in the code.

neilalexander:
```
if !t.Stop() {
    select {
    case <-t.C:
    default:
     }
}
```
... should solve it.

neilalexander (approved):

PR #2226, opened 2022-02-24
State resolution v2 micro-optimisations

neilalexander:
This makes two minor tweaks that should help state resolution v2 performance slightly in big rooms:

* Only populate auth events once into the `authEvents` slice, which should:
  * reduce the number of iterations needed to calculate the auth difference
  * reduce the chance we'll outgrow the capacity bounds of the slice requiring a reallocation of the whole slice
  * avoid duplicate entries appearing in the auth difference which adds burden to the state res algorithm itself

* Refactors `isInAllAuthLists` to give up the moment we find an auth event that doesn't appear in all auth event lists rather than running unnecessary iterations

### DIFF
--- a/roomserver/state/state.go
+++ b/roomserver/state/state.go
@@ -851,15 +851,20 @@ func (v *StateResolution) resolveConflictsV2(
 		if err != nil {
 			return nil, err
 		}
+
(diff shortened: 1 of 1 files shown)


kegsay (approved):
LGTM!

PR #2227, opened 2022-02-25
Update `Events` call-sites which now don't return an error, update `parsedRespState` to sort

neilalexander:
So that earlier events should satisfy auth for later ones, which should hopefully make federated joins more reliable.


kegsay (commented):
Doesn't `outliers, err := state.Events(event.RoomVersion)` do this already?

PR #2232, opened 2022-02-28
MSC2946: Spaces Summary (round 2)

kegsay:
There was an early implemention of MSC2946 done on Dendrite but this is not what landed in the specification.

This PR updates the early implementation to a version which works with spaces today.

Given MSC2946 merged, we should probably migrate this code to the client API eventually. Or roomserver?


go.mod
```diff
@@ -39,7 +40,7 @@ require (
 	github.com/matrix-org/go-http-js-libp2p v0.0.0-20200518170932-783164aeeda4
 	github.com/matrix-org/go-sqlite3-js v0.0.0-20210709140738-b0d1ba599a6d
 	github.com/matrix-org/gomatrix v0.0.0-20210324163249-be2af5ef2e16
-	github.com/matrix-org/gomatrixserverlib v0.0.0-20220225115648-d2a338a15438
+	github.com/matrix-org/gomatrixserverlib v0.0.0-20220228124130-5585f65cec09
```
neilalexander, line 43: Whatever this commit is, I assume it needs to be PR'd into `gomatrixserverlib` and then this PR updated with the squashed commit ID.

neilalexander (approved):

go.mod
kegsay: Ah yes, I'll update this now.

PR #2234, opened 2022-03-01
Limit `JoinedUsersSetInRooms` to interested users

neilalexander:
Beforehand, `JoinedUsersSetInRooms` would pull out all memberships to all relevant rooms, even though we already had a good idea of the users we cared about either from the result of `QueryKeyChanges` or from join/leave events in the sync response from the PDU stream.

This PR limits the database work so that we filter on only users we care about, so that we're no longer passing massive amounts of data we don't care about across the database boundary and therefore will no longer waste massive amounts of CPU time to calculate key changes in sync responses.

### DIFF
--- a/roomserver/storage/interface.go
+++ b/roomserver/storage/interface.go
@@ -151,7 +151,7 @@ type Database interface {
 	// GetBulkStateContent returns all state events which match a given room ID and a given state key tuple. Both must be satisfied for a match.
 	// If a tuple has the StateKey of '*' and allowWildcards=true then all state events with the EventType should be returned.
 	GetBulkStateContent(ctx context.Context, roomIDs []string, tuples []gomatrixserverlib.StateKeyTuple, allowWildcards bool) ([]tables.StrippedEvent, error)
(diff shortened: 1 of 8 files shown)


roomserver/storage/interface.go
```diff
@@ -152,7 +152,7 @@ type Database interface {
 	// If a tuple has the StateKey of '*' and allowWildcards=true then all state events with the EventType should be returned.
 	GetBulkStateContent(ctx context.Context, roomIDs []string, tuples []gomatrixserverlib.StateKeyTuple, allowWildcards bool) ([]tables.StrippedEvent, error)
 	// JoinedUsersSetInRooms returns all joined users in the rooms given, along with the count of how many times they appear.
-	JoinedUsersSetInRooms(ctx context.Context, roomIDs []string) (map[string]int, error)
+	JoinedUsersSetInRooms(ctx context.Context, roomIDs, userIDs []string) (map[string]int, error)
```
kegsay, line 155: Need to update docs for `userIDs`.

kegsay (approved):
Very nice improvement - it looks like sqlite has problems in CI though.

PR #2236, opened 2022-03-01
Add canonical alias support

S7evinK:
For the changes in [roomserver/internal/alias.go](https://github.com/matrix-org/dendrite/compare/s7evink/canonicalalias?expand=1#diff-d96a304ebb23d290cee289680b2d30cb810a7af62a7fe42395ff2b7e3cdf6f92), I'm not sure if that's the correct thing to do.


roomserver/internal/alias.go
```diff
@@ -183,6 +186,65 @@ func (r *RoomserverInternalAPI) RemoveRoomAlias(
...
+		newEvent, err := builder.Build(time.Now(), r.ServerName, r.Cfg.Matrix.KeyID, r.Cfg.Matrix.PrivateKey, ev.RoomVersion)
+		if err != nil {
+			return err
+		}
+
+		inputEvent := api.InputRoomEvent{
+			Kind:          api.KindNew,
+			Event:         newEvent.Headered(ev.RoomVersion),
+			Origin:        r.ServerName,
+			StateEventIDs: stateEventIDs,
```
neilalexander, line 236: It's best not to specify `StateEventIDs` and `HasState` here — we only use those if we want to force the roomserver to create a new state snapshot, which isn't necessary in this case.

This means we should be able to remove all of the `r.QueryLatestEventsAndState` bit.

roomserver/internal/alias.go
```diff
@@ -183,6 +186,65 @@ func (r *RoomserverInternalAPI) RemoveRoomAlias(
 		}
 	}
 
+	ev, err := r.DB.GetStateEvent(ctx, roomID, gomatrixserverlib.MRoomCanonicalAlias, "")
+	if err != nil && err != sql.ErrNoRows {
```
neilalexander, line 190: There's no case for if the error is `sql.ErrNoRows` — should there be?

neilalexander (changes requested):

neilalexander (approved):

clientapi/routing/sendevent.go
```diff
@@ -120,6 +122,40 @@ func SendEvent(
...
+		requestAliases := append(aliasReq.AltAliases, aliasReq.Alias)
+		for _, alias := range aliasRes.Aliases {
+			for _, altAlias := range requestAliases {
+				if altAlias == alias {
+					found++
+				}
+			}
+		}
+		// check that we found at least the same amount of existing aliases as are in the request
+		if aliasReq.Alias != "" && found < len(requestAliases) {
```
NickCao, line 151: I think the check here is too strict. Consider the following usecase: we have a room federated between example.com and example.org, a user from example.com adds a canonical alias to the room, and then user b from example.org tries to add another alias, this event would be rejected. As local alias are not federated per https://spec.matrix.org/v1.2/client-server-api/#room-aliases. `GetAliasesForRoomID` request to example.org would not be able to list the alias set on example.com. I think the correct way to check this is to go through the full alias resolution process to make sure all the aliases in the event are resolved to the room in question.

PR #2241, opened 2022-03-01
Fix read receipts

neilalexander:
We were rewriting the `content` map with an empty value instead of reusing the previous one in some cases.

Also we should start with a `lastPos` of 0 so that it is easier to spot at the call site if there were no receipts in the range.

### DIFF
--- a/syncapi/storage/postgres/receipt_table.go
+++ b/syncapi/storage/postgres/receipt_table.go
@@ -96,7 +96,7 @@ func (r *receiptStatements) UpsertReceipt(ctx context.Context, txn *sql.Tx, room
 }
 
 func (r *receiptStatements) SelectRoomReceiptsAfter(ctx context.Context, roomIDs []string, streamPos types.StreamPosition) (types.StreamPosition, []api.OutputReceiptEvent, error) {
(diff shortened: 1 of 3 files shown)


syncapi/storage/postgres/receipt_table.go
```diff
@@ -96,7 +96,7 @@ func (r *receiptStatements) UpsertReceipt(ctx context.Context, txn *sql.Tx, room
 }
 
 func (r *receiptStatements) SelectRoomReceiptsAfter(ctx context.Context, roomIDs []string, streamPos types.StreamPosition) (types.StreamPosition, []api.OutputReceiptEvent, error) {
-	lastPos := streamPos
```
kegsay, line 99: By doing this, if there are no receipts, we will return `0`. Is this intentional?

kegsay (approved):

syncapi/storage/postgres/receipt_table.go
neilalexander: Yes, because the call site for `RoomReceiptsAfter` specifically checks for that case.

PR #2244, opened 2022-03-03
Clean old notifications regularly

neilalexander:
We'll keep highlights for a month and non-highlights for a day, to stop the `userapi_notifications` table from growing indefinitely.

We'll also allow storing events even if no pushers are present, because apparently Element Web expects to work that way.

### DIFF
--- a/sytest-blacklist
+++ b/sytest-blacklist
@@ -24,7 +24,6 @@ Local device key changes get to remote servers with correct prev_id
 
 # Flakey
 Local device key changes appear in /keys/changes
-/context/ with lazy_load_members filter works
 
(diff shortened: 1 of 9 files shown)


kegsay (commented):
What does this mean in practice? We will fail to correctly populate the `unread_notifications` section of `/sync` after these times? That feels suboptimal...

S7evinK (approved):

neilalexander:
I suppose old notifications may not be counted correctly after that time, but I'm not sure yet if there's a better way to manage that. We need to do *something* though to stop the `userapi_notifications` table from growing indefinitely.

PR #2249, opened 2022-03-04
Add opt-in anonymous stats reporting

S7evinK:
As discussed, we're not sending stats about message count until we find a viable solution. 

The queries are basically those used by Synapse, as defined in [metrics.py](https://github.com/matrix-org/synapse/blob/8533c8b03d8916e3805c7d0e0020226017680147/synapse/storage/databases/main/metrics.py), updated to Dendrites DB layout. The reporting interval is also taken from [Synapse](https://github.com/matrix-org/synapse/blob/8533c8b03d8916e3805c7d0e0020226017680147/synapse/app/phone_stats_home.py#L208-L210), so we're open to change the values here.

Requires https://github.com/matrix-org/panopticon/pull/32 to be merged and deployed first.


userapi/storage/postgres/stats_table.go
```diff
@@ -0,0 +1,348 @@
...
+	SELECT users.localpart, platform, users.created_ts, MAX(uip.last_seen_ts)
+	FROM account_accounts users
+	INNER JOIN
+	(SELECT 
+		localpart, last_seen_ts,
+		CASE
+	    	WHEN user_agent LIKE '%%Android%%' THEN 'android'
+    	    WHEN user_agent LIKE '%%iOS%%' THEN 'ios'
+        	WHEN user_agent LIKE '%%Electron%%' THEN 'electron'
+        	WHEN user_agent LIKE '%%Mozilla%%' THEN 'web'
```
kegsay, line 61: Inconsistent checks here. You don't lowercase it here but do elsewhere.

userapi/storage/postgres/stats_table.go
```diff
@@ -0,0 +1,348 @@
...
+	var platform string
+	var count int64
+	var result = map[string]int64{
+		"ios":      0,
+		"android":  0,
+		"web":      0,
+		"electron": 0,
+		"all":      0,
+	}
+	for rows.Next() {
```
kegsay, line 316: Missing `defer rows.Close()`

userapi/storage/postgres/stats_table.go
```diff
@@ -0,0 +1,348 @@
...
+		diff.Milliseconds(),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	var platform string
+	var count int64
+	var result = make(map[string]int64)
+	for rows.Next() {
```
kegsay, line 274: Missing `defer rows.Close()`

userapi/storage/sqlite3/stats_table.go
```diff
@@ -0,0 +1,360 @@
+// Copyright 2022 The Matrix.org Foundation C.I.C.
```
kegsay, line 1: Needs tests. For SQLite you can pretty easily use `:memory:`. I'm not confident these queries work correctly without tests.

kegsay (changes requested):
Needs tests.

userapi/storage/postgres/stats_table.go
S7evinK: That's intentional, that's why there is `countR30UsersSQL` and `countR30UsersV2SQL`, those are the queries used by Synapse.
Idea behind this: Make the values Dendrite sends more comparable to Synapse.

dendrite-config.yaml
```diff
@@ -85,6 +85,14 @@ global:
     # Whether outbound presence events are allowed, e.g. sending presence events to other servers
     enable_outbound: false
 
+  # Whether this instance sends anonymous usage stats
+  report_stats: false
+
+  # The endpoint to report the anonymized homeserver usage statistics to.
+  # Defaults to https://matrix.org/report-usage-stats/push
+  #
+  report_stats_endpoint: https://matrix.org/report-usage-stats/push
```
kegsay, line 94: Might be neater to have it be:
```yaml
report_stats:
    enabled: true
    endpoint: ...
```

This matches the grouping on other things like `server_notices`.

userapi/storage/postgres/stats_table.go
kegsay: Bleh. I guess.. Can you comment this then?

userapi/storage/tables/stats_table_test.go
```diff
@@ -0,0 +1,319 @@
...
+		}
+
+		t.Run(fmt.Sprintf("want %s database engine", wantType), func(t *testing.T) {
+			_, gotDB, err := statsDB.UserStatistics(ctx, nil)
+			if err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
+
+			if wantType != gotDB.Engine { // can't use DeepEqual, as the Version might differ
+				t.Errorf("UserStatistics() gotDB = %+v, want SQLite", gotDB.Engine)
```
kegsay, line 147: Actually you want `wantType`

kegsay (approved):

S7evinK:
Before merging we'll need to ensure https://github.com/matrix-org/panopticon/pull/32 is merged and redeployed. Will see who can help with this tomorrow.

kegsay (approved):
SHIPIT

PR #2263, opened 2022-03-09
Better mapping of stream positions to topological positions in `/messages`

### LINKED ISSUE #2257: `/messages` `to` parameter must also allow `s`/sync tokens
As per [the spec](https://spec.matrix.org/v1.2/client-server-api/#get_matrixclientv3roomsroomidmessages);

> `to`: The token to stop returning events at. This token can be obtained from a prev_batch or next_batch token returned by the /sync endpoint, or from an end token returned by a previous request to this endpoint.

neilalexander:
This PR updates `/messages` with the following:

1. Adds `db.StreamToTopologicalPosition` for finding the closest topological position in a given room for a stream position
2. Uses `db.GetEventsInTopologicalRange` in all cases, with mapped values from above if needed
3. Removes `db.GetEventsInStreamingRange` as it is no longer needed

It should mean that streaming tokens can be supplied to either `"from"` or `"to"` in a `/messages` request and it should do the right thing, so it should also fix #2257.


syncapi/routing/messages.go
```diff
@@ -41,19 +41,17 @@ type messagesReq struct {
...
-	fromStream       *types.StreamingToken
 	device           *userapi.Device
 	wasToProvided    bool
 	backwardOrdering bool
 	filter           *gomatrixserverlib.RoomEventFilter
 }
 
 type messagesResp struct {
-	Start       string                          `json:"start"`
-	StartStream string                          `json:"start_stream,omitempty"` // NOTSPEC: so clients can hit /messages then immediately /sync with a latest sync token
```
kegsay, line 53: Cerulean uses this to load quickly. https://github.com/matrix-org/cerulean/blob/main/src/Client.js#L401

syncapi/routing/messages.go
```diff
@@ -125,16 +111,42 @@ func OnIncomingMessagesRequest(
...
-				Code: http.StatusBadRequest,
-				JSON: jsonerror.InvalidArgumentValue("Invalid to parameter: " + err.Error()),
+			var streamToken types.StreamingToken
+			if streamToken, err = types.NewStreamTokenFromString(toQuery); err != nil {
+				return util.JSONResponse{
+					Code: http.StatusBadRequest,
+					JSON: jsonerror.InvalidArgumentValue("Invalid to parameter: " + err.Error()),
+				}
+			} else {
+				to, err = db.StreamToTopologicalPosition(req.Context(), roomID, streamToken.PDUPosition, !backwardOrdering)
```
kegsay, line 145: Maybe de-dupe this as we do exactly the same thing with `from`?

syncapi/storage/postgres/output_room_events_topology_table.go
```diff
@@ -50,14 +50,14 @@ const insertEventInTopologySQL = "" +
 const selectEventIDsInRangeASCSQL = "" +
 	"SELECT event_id FROM syncapi_output_room_events_topology" +
 	" WHERE room_id = $1 AND (" +
-	"(topological_position > $2 AND topological_position < $3) OR" +
+	"(topological_position > $2 AND topological_position <= $3) OR" +
```
kegsay, line 53: Why? This very specifically was not `<=` else you can get out-of-order events when you have multiple events with the same topological token but different stream positions, hence the line below.

kegsay (changes requested):

PR #2264, opened 2022-03-09
Update database migrations, remove goose

S7evinK:
As of now, it can happen that later migrations are run before earlier ones, since we don't guarantee any component startup order. This can result in skipped migrations, which in turn result in Dendrite crashing on startup.

This PR does the following:
- Removes goose as a dependency
- Implements a simple database migration mechanism:
  - Currently only upgrades
  - Migrations are executed on a per table basis where possible (state block refactor being one of the migrations where that's not possible)

Since, for example, the state block refactor migration can only run once, there's still code to check if the migration was already executed.

The migrations should be executed after table creation, to ensure the migration is run before preparing SQL statements.

Also adds the possibility to do direct upgrade tests from e.g. version 0.3.0 to HEAD.
Tested successfully with 
```bash
for x in "0.3.0" "0.4.0" "0.5.0" "0.6.0" "0.6.1" "0.6.2" "0.6.3" "0.6.4" "0.6.5"; do 
  ./dendrite-upgrade-tests -tmp=/tmp/dendrite -direct -from ${x} -head . > /tmp/dendrite/upgrade_${x}.out 2>&1; 
done
```


keyserver/storage/postgres/key_changes_table.go
```diff
@@ -55,7 +57,24 @@ func NewPostgresKeyChangesTable(db *sql.DB) (tables.KeyChanges, error) {
...
-	return s, err
+	if err != nil {
+		return s, err
+	}
+
+	// TODO: Remove when we are sure we are not having goose artifacts in the db
+	// This forces an error, which indicates the migration is already applied, since the
+	// column partition was removed from the table
+	var count int
+	err = db.QueryRow("SELECT partition FROM keyserver_key_changes LIMIT 1;").Scan(&count)
```
kegsay, line 68: Can we not just select from the goose_db_version table for the migration version?

kegsay (changes requested):
Please read the existing goose_db_version table:
```
select * from goose_db_version;
 id |    version_id    | is_applied |           tstamp           
----+------------------+------------+----------------------------
  1 |                0 | t          | 2022-01-20 18:37:51.794326
  2 | 2021020411080000 | t          | 2022-01-20 18:37:51.81805
  3 | 2022012016470000 | t          | 2022-01-20 18:48:25.99
  4 | 2022021013023800 | t          | 2022-02-17 12:54:21.335007
(4 rows)
```
And use that rather than using the presence/absence of some columns as a proxy.

internal/sqlutil/migrate.go
```diff
@@ -1,130 +1,129 @@
...
+	txn, err := m.db.BeginTx(ctx, nil)
+	if err != nil {
+		return fmt.Errorf("unable to begin transaction: %w", err)
 	}
-	for {
-		current, err := goose.EnsureDBVersion(db)
+	defer func() {
 		if err != nil {
-			return fmt.Errorf("runDeltas: Failed to EnsureDBVersion: %w", err)
+			_ = txn.Rollback()
```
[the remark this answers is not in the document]
S7evinK: Changed.

keyserver/storage/postgres/key_changes_table.go
S7evinK: Checking `goose_db_version` will not work for users starting with Dendrite releases once this is merged or users manually removing the table., so I'd rather keep this check in. It's not ideal, for sure.

internal/sqlutil/migrate.go
```diff
@@ -1,130 +1,140 @@
...
+	"github.com/matrix-org/dendrite/internal"
 )
 
-type Migrations struct {
-	registeredGoMigrations map[int64]*goose.Migration
+const createDBMigrationsSQL = "" +
+	"CREATE TABLE IF NOT EXISTS db_migrations (" +
+	" version TEXT PRIMARY KEY," +
+	" time TEXT," +
+	" dendrite_version TEXT" +
```
kegsay, line 31: All need NOT NULL.

internal/sqlutil/migrate.go
```diff
@@ -1,130 +1,140 @@
...
+	");"
+
+const insertVersionSQL = "" +
+	"INSERT INTO db_migrations (version, time, dendrite_version)" +
+	" VALUES ($1, $2, $3) " +
+	" ON CONFLICT(version) DO UPDATE SET dendrite_version = $4, time = $5"
```
kegsay, line 37: This implies it is allowed to exeucte the same update > 1 time? That should surely be an error?

internal/sqlutil/migrate.go
```diff
@@ -1,130 +1,140 @@
...
-// AddNamedMigration : Add a named migration.
-func (m *Migrations) AddNamedMigration(filename string, up func(*sql.Tx) error, down func(*sql.Tx) error) {
-	v, _ := goose.NumericComponent(filename)
-	migration := &goose.Migration{Version: v, Next: -1, Previous: -1, Registered: true, UpFn: up, DownFn: down, Source: filename}
-
-	if existing, ok := m.registeredGoMigrations[v]; ok {
-		panic(fmt.Sprintf("failed to add migration %q: version conflicts with %q", filename, existing.Source))
+// AddMigrations adds new migrations to the list.
+// De-duplicates migrations by their version
+func (m *Migrator) AddMigrations(migrations ...Migration) {
```
kegsay, line 71: Should be clear that if a migration already exists with this version string it is not added (first wins). In addition, you need to be extremely clear that the order of the migrations in this list will be the order that they get applied. It's only implied at present via "to the list" wording. You also need to be clear that repeated calls to this function append.

internal/sqlutil/migrate.go
```diff
@@ -1,130 +1,140 @@
...
 
-		next, err := migrations.Next(current)
-		if err != nil {
-			if err == goose.ErrNoNextVersion {
-				return nil
+	return WithTransaction(m.db, func(txn *sql.Tx) error {
+		for i := range m.migrations {
+			now := time.Now().UTC().Format(time.RFC3339)
+			migration := m.migrations[i]
+			if !executedMigrations[migration.Version] {
```
kegsay, line 98: Invert and outdent please:
```go
if executedMigrations[migration.Version] {
    continue
}
// ... rest of code that was inside the if statement.
```

internal/sqlutil/migrate.go
```diff
@@ -1,130 +1,140 @@
...
-	// populate next and previous for each migration
-	for i, m := range migrations {
-		prev := int64(-1)
-		if i > 0 {
-			prev = migrations[i-1].Version
-			migrations[i-1].Next = m.Version
-		}
-		migrations[i].Previous = prev
+// ExecutedMigrations returns a map with already executed migrations
+func (m *Migrator) ExecutedMigrations(ctx context.Context) (map[string]bool, error) {
```
kegsay, line 120: It also makes the migrations table if it doesn't exist, worth mentioning.

keyserver/storage/postgres/cross_signing_sigs_table.go
```diff
@@ -66,6 +67,16 @@ func NewPostgresCrossSigningSigsTable(db *sql.DB) (tables.CrossSigningSigs, erro
 	if err != nil {
 		return nil, err
 	}
+
+	m := sqlutil.NewMigrator(db)
+	m.AddMigrations(sqlutil.Migration{
+		Version: "",
```
kegsay, line 73: Why is this blank?

keyserver/storage/postgres/key_changes_table.go
kegsay: If users have manually removed the goose db version table then uh, that's on them. For Dendrite users starting in the next release, we just need to make sure it's non-fatal that if the table doesn't exist then we know that they have been migrated already.

keyserver/storage/sqlite3/cross_signing_sigs_table.go
```diff
@@ -65,6 +66,15 @@ func NewSqliteCrossSigningSigsTable(db *sql.DB) (tables.CrossSigningSigs, error)
 	if err != nil {
 		return nil, err
 	}
+	m := sqlutil.NewMigrator(db)
+	m.AddMigrations(sqlutil.Migration{
+		Version: "",
```
kegsay, line 71: Blank again.

keyserver/storage/sqlite3/key_changes_table.go
```diff
@@ -53,7 +55,24 @@ func NewSqliteKeyChangesTable(db *sql.DB) (tables.KeyChanges, error) {
 		db: db,
 	}
 	_, err := db.Exec(keyChangesSchema)
-	return s, err
+	if err != nil {
+		return s, err
+	}
+	// TODO: Remove when we are sure we are not having goose artefacts in the db
+	// This forces an error, which indicates the migration is already applied, since the
+	// column partition was removed from the table
```
kegsay, line 63: Yeah, if you're relying on this sort of behaviour then just do that on the goose_db_version, it's more explicit.

userapi/storage/postgres/devices_table.go
```diff
@@ -120,6 +121,15 @@ func NewPostgresDevicesTable(db *sql.DB, serverName gomatrixserverlib.ServerName
 	if err != nil {
 		return nil, err
 	}
+	m := sqlutil.NewMigrator(db)
+	m.AddMigrations(sqlutil.Migration{
+		Version: "add last_seen_ts",
+		Up:      deltas.UpLastSeenTSIP,
+	})
+	err = m.Up(context.Background())
```
kegsay, line 129: It does seem a bit silly to add a `context.Context` and just never really make use of it.

kegsay (approved):
A few comments but otherwise LGTM

Issue #1997, opened 2021-08-30
Dendrite cannot load conduit media


ShadowJonathan:
### Background information

- **Dendrite version or git SHA**: 0.5.0
- **Monolith or Polylith?**: N/A
- **SQLite3 or Postgres?**: N/A
- **Running in Docker?**: N/A
- **`go version`**: N/A

Cannot acquire details, homeserver in question is `jp.cube64128.xyz`

These same symptoms are present on `dendrite.matrix.org`

### Description

Dendrite fails without a proper error (also see https://github.com/matrix-org/dendrite/issues/1994) when loading some media from a Conduit server.

The following MXC from `conduit.rs` loads correctly on matrix.org and any other synapse server, but fails to load on dendrite;

`mxc://conduit.rs/E649aeQG37N7VGvioBB5lrlVzt8VgPud`

<details>
<summary>Synapse</summary>

`https://matrix.org/_matrix/media/r0/download/conduit.rs/E649aeQG37N7VGvioBB5lrlVzt8VgPud`
`https://matrix.jboi.nl/_matrix/media/r0/download/conduit.rs/E649aeQG37N7VGvioBB5lrlVzt8VgPud`

</details>

<details>
<summary>Dendrite</summary>

`https://jp.cube64128.xyz/_matrix/media/r0/download/conduit.rs/E649aeQG37N7VGvioBB5lrlVzt8VgPud`
`https://dendrite.matrix.org/_matrix/media/r0/download/conduit.rs/E649aeQG37N7VGvioBB5lrlVzt8VgPud`

</details>

No logs, but seeing as how this occurs on `dendrite.matrix.org`, this should be easy to trace.

If there are any details on what conduit could be doing wrong here, please close this issue with it and ping @timokoesters with the details

database64128:
My dendrite instance was built on commit `221ed52` from #1993 with **Go 1.17**.

- **Monolith or Polylith?**: Monolith
- **SQLite3 or Postgres?**: Postgres
- **Running in Docker?**: No

ShadowJonathan:
I think that the EOF fallthrough here (as noted [here](https://github.com/matrix-org/dendrite/issues/1994#issuecomment-908255352))shadows an actual request error;

https://github.com/matrix-org/dendrite/blob/802f1c96f804f7a146e4e12e25b20c980a6af870/mediaapi/fileutils/fileutils.go#L133

Issue #1994, opened 2021-08-30
Dendrite response incorrectly to invalid media


ShadowJonathan:
### Background information

- **Dendrite version or git SHA**: 0.5.0
- **Monolith or Polylith?**: N/A
- **SQLite3 or Postgres?**: N/A
- **Running in Docker?**: N/A
- **`go version`**: N/A

Cannot acquire details, homeserver in question is `jp.cube64128.xyz`

These same symptoms are present on `dendrite.matrix.org`

### Description

Dendrite returns a confusing error message to non-existent media;

`{"errcode":"M_NOT_FOUND","error":"Failed to download: fileutils.GetPathFromBase64Hash: Invalid filePath (Base64Hash too short - min 3 characters): \"\""}`

[this MXC](https://jp.cube64128.xyz/_matrix/media/r0/download/matrix.org/DRevoaEiuzbkOznknySKuMmE) is valid, while [this one](https://jp.cube64128.xyz/_matrix/media/r0/download/matrix.org/DRevoaEiuzbkOznknySKuMmF) isn't, replacing the latter with `matrix.org` or any synapse server returns a proper message of "not found"

ShadowJonathan:
I'm pretty sure the EOF fallthrough here and the overwriting of `err` later down there is causing this;
https://github.com/matrix-org/dendrite/blob/802f1c96f804f7a146e4e12e25b20c980a6af870/mediaapi/fileutils/fileutils.go#L133

PR #2266, opened 2022-03-10
Return error from remote download requests

S7evinK:
We were returning a nil error if we didn't find the remote file, which would result in useless database entries (filesize 0, no hash, etc.), causing behaviors like #1994 and #1997
Should I add a "migration" to delete those database entries?

### DIFF
--- a/mediaapi/routing/download.go
+++ b/mediaapi/routing/download.go
@@ -556,7 +556,8 @@ func (r *downloadRequest) getRemoteFile(
 				cfg.MaxThumbnailGenerators,
 			)
 			if err != nil {
-				return fmt.Errorf("r.fetchRemoteFileAndStoreMetadata: %w", err)
+				r.Logger.WithError(err).Errorf("r.fetchRemoteFileAndStoreMetadata: failed to fetch remote file")
+				return err
 			}
 		} else {
 			// If we have a record, we can respond from the local file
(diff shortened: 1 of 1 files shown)


mediaapi/routing/download.go
```diff
@@ -720,14 +721,13 @@ func (r *downloadRequest) fetchRemoteFile(
 	r.Logger.Debug("Fetching remote file")
 
 	// create request for remote file
-	resp, err := r.createRemoteRequest(ctx, client)
+	resp, err := client.CreateMediaDownloadRequest(ctx, r.MediaMetadata.Origin, string(r.MediaMetadata.MediaID))
+	if err != nil || resp.StatusCode != http.StatusOK {
+		return "", false, fmt.Errorf("file with media ID %q could not be downloaded from %s", r.MediaMetadata.MediaID, r.MediaMetadata.Origin)
+	}
 	if err != nil {
 		return "", false, err
```
kegsay, line 729: Redundant `err != nil` check.

kegsay (approved):
We should probably fall back to the original intended behaviour. Going through history, it looks like the code used to be correct:
 - https://github.com/matrix-org/dendrite/commit/923013aa5598e057fdcf64edc8a553bda590160a#diff-546eb2a9d675e4dceb7297ac659d423bc38ffab854a0c600d4a87177f9a983b0R486
 - Returned correctly (proxying the 404) - https://github.com/matrix-org/dendrite/commit/923013aa5598e057fdcf64edc8a553bda590160a#diff-546eb2a9d675e4dceb7297ac659d423bc38ffab854a0c600d4a87177f9a983b0R385
 - A refactoring exercise introduced the bug - https://github.com/matrix-org/dendrite/commit/08b9940dde5b316c40e7f6ae976cc40cadccde42#diff-546eb2a9d675e4dceb7297ac659d423bc38ffab854a0c600d4a87177f9a983b0R632

PR #2268, opened 2022-03-10
Fix issue where Room name is shown as "Empty room" to an invited user

tak-slashtalk:
### Pull Request Checklist

Issue:
The stateEvent "m.room.name" is being filtered out for a user who just joined a room. As a result, the user only sees "Empty room" as the name of the room.

The fix is to add a condition to the exclusion check, and allow this state event to get back to the client.

Filed tracking issue: https://github.com/matrix-org/dendrite/issues/2269

Ran the sytest locally. 3 additional tests are now passing.

* [x ] Pull request includes a [sign off](https://github.com/matrix-org/dendrite/blob/main/docs/CONTRIBUTING.md#sign-off)


syncapi/streams/stream_pdu.go
```diff
@@ -276,7 +276,7 @@ func (p *PDUStreamProvider) getJoinResponseForCompleteSync(
 	if !wantFullState {
 		excludingEventIDs = make([]string, 0, len(recentStreamEvents))
 		for _, event := range recentStreamEvents {
-			if event.StateKey() != nil {
+			if event.StateKey() != nil && event.Type() != "m.room.name" {
```
neilalexander, line 279: I'm a bit confused as to how/why this is a solution. The idea of this block of code (despite the seemingly incomplete comment) is so that if state events already appear in the `"timeline"` section of the room response, we don't then repeat them a second time in the `"state"` section too. 

Have you been able to verify that the `m.room.name` event was indeed missing from the timeline on a complete sync?

tak-slashtalk: Yes, I did. it's missing from timeline when the client gets the response. This is because for a recently created room, the recentStreamEvents contains m.room.name and it is excluded here. Then in line 298-299, it is also excluded from recentEvents because the user just joined the room:

			if membership == "join" {
				joinEventIndex = i

So at this point, m.room.name is filtered from both stateEvents and recentEvents. Consequently, an invited user who just joined this room will not see the m.room.name event. The room name is always "Empty room"

tak-slashtalk:
Hi Neil,
I hope the explanation is clear. I can add comments around the code to
explain why state events like "m.room.name" shouldn't be removed if that
helps.

neilalexander:
Right, so I now see why it's happening, although I still think the solution probably isn't optimal. I'd really rather not add special cases for `m.room.name` (or any other event type) if we can avoid it.

This seems like an ordering problem. We're trying to truncate `recentStreamEvents` *before* we work out which state events we want to retrieve and exclude — meanwhile the room name event (amongst probably others) fall down the gap. It seems like the better solution would be to switch the order around — truncate `recentStreamEvents` first, *then* go and work out which state events are in the timeline and then request the rest of the state.

In that case, if the room name event came before the join event, then it won't appear in `"timeline"` but it will appear in `"state"`, which is the correct behaviour.

Can you try #2281 and see if that also solves your issue?

syncapi/streams/stream_pdu.go
tak-slashtalk: @neilalexander any thoughts on how to fix this differently?

neilalexander: Did you see my other comment on whether #2281 solves your problem?

tak-slashtalk: Yes, this works better. Thanks for the fix.

PR #2270, opened 2022-03-11
Don't range entire state for `/sync`

neilalexander:
Currently we pull out the state deltas for *all rooms* between any two given sync positions from the database. This can be a *vast* amount of data, predominantly for rooms that the user has never interacted with and doesn't care about.

This PR modifies the path instead so that we only do this for rooms that the user has ever interacted with (including `leave`) so that we can reduce the amount of unnecessary work to generate the sync response.

This isn't a perfect solution by any means but it is certainly an improvement.

### DIFF
--- a/syncapi/storage/postgres/current_room_state_table.go
+++ b/syncapi/storage/postgres/current_room_state_table.go
@@ -77,6 +77,9 @@ const DeleteRoomStateForRoomSQL = "" +
 const selectRoomIDsWithMembershipSQL = "" +
 	"SELECT DISTINCT room_id FROM syncapi_current_room_state WHERE type = 'm.room.member' AND state_key = $1 AND membership = $2"
 
+const selectRoomIDsWithAnyMembershipSQL = "" +
+	"SELECT DISTINCT room_id, membership FROM syncapi_current_room_state WHERE type = 'm.room.member' AND state_key = $1"
+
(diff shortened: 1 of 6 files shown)


syncapi/storage/postgres/current_room_state_table.go
```diff
@@ -194,6 +201,31 @@ func (s *currentRoomStateStatements) SelectRoomIDsWithMembership(
...
+	rows, err := stmt.QueryContext(ctx, userID)
+	if err != nil {
+		return nil, err
+	}
+	defer internal.CloseAndLogIfError(ctx, rows, "selectRoomIDsWithAnyMembership: rows.close() failed")
+
+	result := map[string]string{}
+	for rows.Next() {
+		var roomID string
+		var membership string
```
kegsay, line 220: Should we be declaring this outside the for loop?

S7evinK (approved):

kegsay (approved):
So long as this worked for invited rooms which tend to be weird state-wise then LGTM!

syncapi/storage/postgres/current_room_state_table.go
neilalexander: I'll look into optimisations like that in a separate PR sometime, since they will need doing throughout the storage packages.

PR #2273, opened 2022-03-12
Use accountdb with provided config

### LINKED ISSUE #2272: create-account program ignores bcrypt_cost setting
### Background information

- **Dendrite version or git SHA**: 01d71a767dbd11fd7b17b948db6d3052f577c796
- **Running in Docker?**: no
- **`go version`**: 1.17.7

### Description

 - **What** is the problem: `create-account -reset-password` does not honour the `bcrypt_cost` setting in `dendrite.yaml`.
 - **Who** is affected: passwords created or reset with `create-account`
 - **How** is this bug manifesting: the default cost of 10 is used regardless of the setting in the config, even when the full path to the config is specified with `-config`
 - **When** did this first appear: I think it has always been like this

S7evinK:
Fixes #2272
Also uses the same regex for validating usernames as `/register` to make sure we only allow lowercase usernames.

### DIFF
--- a/cmd/create-account/main.go
+++ b/cmd/create-account/main.go
@@ -21,6 +21,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"regexp"
 	"strings"
 
 	"github.com/matrix-org/dendrite/setup/base"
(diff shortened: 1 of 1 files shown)


kegsay (approved):
Assuming CI passes.

PR #2277, opened 2022-03-14
Fix /context with lazy_load_members

S7evinK:
This moves the membership events to the end of the list, so they are send correctly when a limit would remove them from the list again.


S7evinK:
Needs https://github.com/matrix-org/sytest/pull/1214, as Sytest will then wait for the messages to come down /sync before hitting /context

kegsay (approved):

PR #2278, opened 2022-03-15
Get rooms where user is either a member or send an event

S7evinK:
This should make `Room creation reports m.room.create to myself` more reliable.
We were hitting `/sync` before the membership event made it to the database, resulting in no data returned but the PDU stream being updated (not updating the PDU stream resulted in more tests failing..). Further `/sync` requests wouldn't get the `m.room.create` event and the test would fail.


S7evinK:
This breaks other things..

PR #2279, opened 2022-03-15
Ensure we return a non-empty list

S7evinK:
Flakey test `Inbound federation can query room alias directory`.
If we're not excluding ourselves, we should be added to the list of servers even if the storage layer doesn't know about it yet.


neilalexander (commented):
What problem is this trying to solve exactly?

S7evinK:
As discussed, will fix this in Sytest.

PR #2285, opened 2022-03-18
Fixes for `create-account`

S7evinK:
Updates `create-account` to verify the given username exists/not exists, doesn't exceed 255 characters.
If the password is empty, falls back to asking the user for the password.


cmd/create-account/main.go
```diff
@@ -83,7 +81,14 @@ func main() {
 		os.Exit(1)
 	}
 
-	pass := getPassword(password, pwdFile, pwdStdin, askPass, os.Stdin)
+	if len(fmt.Sprintf("@%s:%s", *username, cfg.Global.ServerName)) > 255 {
+		logrus.Fatalln("Username can not be longer than 255 characters.")
```
kegsay, line 85: Needs better message here as they may not realise the server name is part of the 255. Log the entire username in the error please.

cmd/create-account/main_test.go
```diff
@@ -8,53 +8,55 @@ import (
...
-		want string
+		name    string
+		args    args
+		want    string
+		wantErr bool
 	}{
-		{
-			name: "no password defined",
-			args: args{},
-			want: "",
```
kegsay, line 34: I don't understand why you are removing the ability to create passwordless accounts?

cmd/create-account/main_test.go
```diff
@@ -8,53 +8,55 @@ import (
...
 		t.Errorf("unable to write to buffer: %+v", err)
 	}
 	tests := []struct {
-		name string
-		args args
-		want string
+		name    string
+		args    args
+		want    string
+		wantErr bool
```
kegsay, line 28: You've added `wantErr` but no test seems to use it?

kegsay (changes requested):
I don't like the API shape. I like that we default to asking for a password from stdin as it is more secure. I don't like that it's now impossible to make a passwordless account.

kegsay (approved):
Tests could be more comprehensive (eg inputs are cli vars and output is created/updated/error enum) to test more logic, but this is good enough for now.

PR #2290, opened 2022-03-21
Refactor appservice & client API to use userapi internal

S7evinK:
Removes the direct dependency on AccountDB in the client API.
Also tries to split up the huge `UserInternalAPI` interface into smaller pieces, making it, hopefully, easier to write tests.


neilalexander (approved):
LGTM once CI sorted.

PR #2292, opened 2022-03-22
Fix #2287 by trying to fetch account by lowercase localpart

### LINKED ISSUE #2287: Error in SQL when using incorrect login
### Background information
- **Dendrite version or git SHA**: 0.6.5
- **Running in Docker?**: No, installed from Alpine, updated from 0.5
- **`go version`**: I believe it is compiled with v1.17
- **Client used (if applicable)**: Element Firefox

### Description

 - **What** is the problem: there are SQL issues when login whit a capitalised user name
 - **Who** is affected: -
 - **How** is this bug manifesting: Using Element from app.element.io, I tried login to my user account. The username is `irvise`, but I input `Irvise`. After a long sync time, I decided to take a look at the logs and noticed that there were a lot of SQL errors.
 - **When** did this first appear: just now, I have no previous experience seeing this error. I am opening this issue as @neilalexander recommended to do so in the #dendrite channel.

### Steps to reproduce

Try to log in in a new client. Capitalise the first letter of your username. Dendrite should have this SQL error in the logs
```
time="2022-03-20T20:01:58.373234324Z" level=error msg="userAPI.QueryAccessToken failed" func=[github.com/matrix-org/dendrite/clientapi/auth.VerifyUserFromRequest](http://github.com/matrix-org/dendrite/clientapi/auth.VerifyUserFromRequest) file="
[github.com/matrix-org/dendrite/clientapi/auth/auth.go:70](http://github.com/matrix-org/dendrite/clientapi/auth/auth.go:70)" error="sql: no rows in result set" [req.id](http://req.id/)=I5g6MEbMACpg req.method=GET req.path=/_matrix/client/r0/pushrules/
```
many repetitions of this error appear. It only stops if the login attempt is stopped by the client.

If more information is needed, then I hope to be able to provide it.

Regards,
Fer

S7evinK:
When logging in we're returning the `user_id` as passed by the `/login` request (e.g. `@Alice:hs1`) which is then used by clients to do further requests.
When we're verifying the `AccessToken`, we are were using this `user_id` to check for the existence of the account, which would fail in case the `localpart` is lowercase in the database.

This first checks for the lowercase `localpart`, which should be the majority of the cases, and if not found tries it as passed by the request.

### DIFF
--- a/userapi/storage/shared/storage.go
+++ b/userapi/storage/shared/storage.go
@@ -301,18 +301,10 @@ func (d *Database) GetAccountByLocalpart(ctx context.Context, localpart string,
 ) (*api.Account, error) {
 	// try to get the account with lowercase localpart (majority)
 	acc, err := d.Accounts.SelectAccountByLocalpart(ctx, strings.ToLower(localpart))
-	switch err {
-	case sql.ErrNoRows: // try with localpart as passed by the request
-		acc, err = d.Accounts.SelectAccountByLocalpart(ctx, localpart)
-		if err != nil {
-			return nil, err
-		}
-		return acc, nil
-	case nil:
-		return acc, nil
-	default:
-		return nil, err
+	if err == sql.ErrNoRows {
+		acc, err = d.Accounts.SelectAccountByLocalpart(ctx, localpart) // try with localpart as passed by the request
 	}
+	return acc, err
 }
 
 // SearchProfiles returns all profiles where the provided localpart or display name
--- a/userapi/storage/shared/storage.go
+++ b/userapi/storage/shared/storage.go
@@ -23,6 +23,7 @@ import (
 	"errors"
 	"fmt"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/matrix-org/gomatrixserverlib"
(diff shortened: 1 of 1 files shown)


userapi/storage/shared/storage.go
```diff
@@ -298,7 +299,20 @@ func (d *Database) CheckAccountAvailability(ctx context.Context, localpart strin
...
+	case sql.ErrNoRows: // try with localpart as passed by the request
+		acc, err = d.Accounts.SelectAccountByLocalpart(ctx, localpart)
+		if err != nil {
+			return nil, err
+		}
+		return acc, nil
+	case nil:
+		return acc, nil
+	default:
+		return nil, err
```
kegsay, line 314: Feels convoluted. Why not just:
```go
acc, err := d.Accounts.SelectAccountByLocalpart(ctx, strings.ToLower(localpart))
if err == sql.ErrNoRows {
    acc, err = d.Accounts.SelectAccountByLocalpart(ctx, localpart)
}
return acc, err
```
?

kegsay (approved):

PR #2293, opened 2022-03-22
Per-room consumers

neilalexander:
This refactors the roomserver input API (yet again) to have per-room consumers, filtered on subject. This is an important step in being able to shard room servers.

It works by:

1. Having an ephemeral headers-only consumer which peeks into the input stream, looking for room IDs;
2. Having durable consumers per-room which filter on the subject, so that we can retrieve the actual events.

This means we no longer need to pull everything out into memory in order to sort them, so we will only ever have one event in memory at a time for each room. This should help to reduce memory usage when the homeserver is busy processing things like joins or missing state/events.

### DIFF
--- a/roomserver/internal/input/input.go
+++ b/roomserver/internal/input/input.go
@@ -144,10 +144,11 @@ func (r *Inputer) Start() error {
 		func(m *nats.Msg) {
 			roomID := m.Header.Get(jetstream.RoomID)
 			r.startWorkerForRoom(roomID)
+			_ = m.Ack()
 		},
 		nats.HeadersOnly(),
 		nats.DeliverAll(),
-		nats.AckNone(),
+		nats.AckAll(),
 		nats.BindStream(r.InputRoomEventTopic),
 	)
 	return err
--- a/roomserver/internal/input/input.go
+++ b/roomserver/internal/input/input.go
@@ -120,6 +120,7 @@ func (w *worker) next() {
 		if len(msgs) != 1 {
 			return
 		}
+		defer w.Act(nil, w.next)
 	case context.DeadlineExceeded:
 		logrus.Infof("Stream for room %q idle, shutting down", w.roomID)
 		if err = w.subscription.Unsubscribe(); err != nil {
@@ -140,7 +141,6 @@ func (w *worker) next() {
 	}
 
 	msg := msgs[0]
-
 	var inputRoomEvent api.InputRoomEvent
 	if err = json.Unmarshal(msg.Data, &inputRoomEvent); err != nil {
 		_ = msg.Term()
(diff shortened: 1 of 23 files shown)


setup/jetstream/nats.go
```diff
@@ -75,14 +76,29 @@ func setupNATS(cfg *config.JetStream, nc *natsclient.Conn) (natsclient.JetStream
...
+			subjects = []string{name, name + ".>"}
+		}
+		if info != nil {
+			switch {
+			case !reflect.DeepEqual(info.Config.Subjects, subjects):
+				fallthrough
+			case info.Config.Retention != stream.Retention:
+				fallthrough
+			case info.Config.Storage != stream.Storage:
+				if err = s.DeleteStream(name); err != nil {
```
kegsay, line 95: This will clean up the old key change consumer right?

roomserver/internal/input/input.go
```diff
@@ -46,6 +47,7 @@ var keyContentFields = map[string]string{
 }
 
 type Inputer struct {
```
kegsay, line 49: A block level comment explaining the stream magic would be welcome. Specifically, please have some sort of visual aid e.g:
```
Roomserver input stream

          A BC _   (where each room consumer is up to in the stream)
          v vv v
    ABAABCAABCAA   (letter = subject for each message)
    ---newer--->

 - B and C are all caught up.
 - A is behind and stuck processing an earlier message.
 - _ is the ephemeral consumer which checks the NATS Header for the room ID for the message and
   ensures that a consumer for this room ID has been spun up. See Inputer.Start()
```

kegsay: Assuming my understanding is correct?

kegsay (approved):

roomserver/internal/input/input.go
```diff
@@ -57,105 +59,208 @@ type Inputer struct {
...
+	}
+}
+
+// Start creates an ephemeral non-durable consumer on the roomserver
+// input topic. It is configured to deliver us headers only because we
+// don't actually care about the contents of the message at this point,
+// we only care about the `room_id` field. Once a message arrives, we
+// will look to see if we have a worker for that room which has its
+// own consumer. If we don't, we'll start one.
+func (r *Inputer) Start() error {
```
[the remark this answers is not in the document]
neilalexander: We would need one ephemeral consumer per shard, so that each shard can peek through the stream and pick out the room IDs it is interested in.

setup/jetstream/nats.go
neilalexander: Yes, it will.

roomserver/internal/input/input.go
neilalexander: Yep, that's spot on. I'll adapt it into a block level comment.

PR #2297, opened 2022-03-24
Move CI to Github Actions

S7evinK:
Moves all buildsteps from buildkite to Github Actions, which seems to be running a bit faster (even without cache).
Using "Matrix"-tests, we can easily test for different go versions or different setups, e.g. GOOS/GOARCH, upgrade tests from different versions.
[Example result](https://github.com/S7evinK/dendrite/actions/runs/2033122448)

Linter & wasm failures don't stop the runs right now. (golangci-lint is a newer version than we used on buildkite, so finds more things.)


neilalexander (approved):

.github/workflows/dendrite.yml
```diff
@@ -0,0 +1,317 @@
...
+          - label: PostgreSQL, full HTTP APIs
+            postgres: postgres
+            api: full-http
+    container:
+      image: matrixdotorg/sytest-dendrite:latest
+      volumes:
+        - ${{ github.workspace }}:/src
+      env:
+        POSTGRES: ${{ matrix.postgres && 1}}
+        API: ${{ matrix.api && 1 }}
```
kegsay, line 220: Do we really need the `&& 1`?

kegsay (approved):

PR #2299, opened 2022-03-24
Upload Docker images for releases to both Docker Hub and GitHub Container Registry

neilalexander:
This PR will also ensure that our Docker images are uploaded to GitHub Container Registry, and should hopefully show up [here](https://github.com/orgs/matrix-org/packages?repo_name=dendrite). Docker Hub is annoyingly inconsistent so it would be nice to have an alternative.

It also adds a new tag `:main`, which will be updated every time the CI pipeline on the `main` branch completes successfully, for those who want to track `main` more closely.


S7evinK (approved):
Looks plausible to me.

PR #2306, opened 2022-03-26
Remove eduserver

S7evinK:
Removes the EDU server, as it's not really necessary anymore.
Moves all needed parts to the syncapi or federationapi.


dendrite-config.yaml
```diff
@@ -187,12 +187,6 @@ client_api:
     threshold: 5
     cooloff_ms: 500
 
-# Configuration for the EDU server.
```
neilalexander, line 190: Also need to update the equivalent Docker sample configs too.

neilalexander (approved):

Issue #1304, opened 2020-08-24
Implement Room Upgrades


Kegsay:
Spec: https://matrix.org/docs/spec/client_server/r0.6.1#id160
Sytests:
```
    × /upgrade creates a new room
    × /upgrade should preserve room visibility for public rooms
    × /upgrade should preserve room visibility for private rooms
    × /upgrade copies >100 power levels to the new room
    × /upgrade copies the power levels to the new room
    × /upgrade preserves the power level of the upgrading user in old and new rooms
    × /upgrade copies important state to the new room
    × /upgrade copies ban events to the new room
    × local user has push rules copied to upgraded room
    × remote user has push rules copied to upgraded room
    × /upgrade moves aliases to the new room
    × /upgrade moves remote aliases to the new room
    × /upgrade preserves direct room state
    × /upgrade preserves room federation ability
    × /upgrade restricts power levels in the old room
    × /upgrade restricts power levels in the old room when the old PLs are unusual
    × /upgrade to an unknown version is rejected
    × /upgrade is rejected if the user can't send state events
    × /upgrade of a bogus room fails gracefully
    × Cannot send tombstone event that points to the same room
    × Local and remote users' homeservers remove a room from their public directory on upgrade
    × local user has tags copied to the new room
    × remote user has tags copied to the new room
```

DavidSpenler:
I can work on this.

neilalexander:
Updated with the remaining tests.

tombh:
Is this a good issue to subscribe to to get notifications for progress towards stable v9 rooms?

neilalexander:
@tombh No, this PR is about the room upgrade mechanism specifically. I have been working on v8 and v9 rooms today though.

tombh:
Oh great! Is there somewhere I can subscribe for updates? (I'd rather not subscribe to the whole repo)

PR #2307, opened 2022-03-26
Added /upgrade endpoint

DavidSpenler:
(This addresses issue #1304)

The added changes cause all tests to pass except 6:
` /upgrade copies >100 power levels to the new room` - current code disallows creating power levels greater than 100 right now
`/upgrade moves remote aliases to the new room`, `Local and remote users' homeservers remove a room from their public directory on upgrade` - current APIs to change aliases and publishing rooms do not federate these changes right now
`/upgrade preserves direct room state` - using current APIs, this would require an order O(n) procedure to update room state for all participating users 
`local user has tags copied to the new room`, `remote user has tags copied to the new room` - another procedure which without specific APIs would require an O(n) time to operate for all participating users.
### Pull Request Checklist

This 
Signed-off-by: `David Spenler <davidspenler@gmail.com>`


neilalexander (commented):
I’ve glanced through and this looks like a good starting point!

However, given the amount of tromboning back and forth between the client API and the roomserver here (reading state, building events etc), it would really make more sense for this to be a `PerformRoomUpgrade` API call in the roomserver itself. 

That way you have direct access to the database in all the relevant places, will use far less cross-component API calls and all the client API needs to do is to call `PerformRoomUpgrade`.

S7evinK (commented):
Besides the comments, looks fine to me.

clientapi/routing/upgrade_room.go
```diff
@@ -0,0 +1,92 @@
...
+	asAPI appserviceAPI.AppServiceQueryAPI,
+) util.JSONResponse {
+	var r upgradeRoomRequest
+	if rErr := httputil.UnmarshalJSONRequest(req, &r); rErr != nil {
+		return *rErr
+	}
+
+	// Validate that the room version is supported
+	if _, err := version.SupportedRoomVersion(gomatrixserverlib.RoomVersion(r.NewVersion)); err != nil {
+		return util.JSONResponse{
```
kegsay, line 54: We never check that the new version is in fact "newer" than the old version?

roomserver/internal/perform/perform_upgrade.go
```diff
@@ -0,0 +1,691 @@
...
+	}
+
+	oldCreateEvent := state[gomatrixserverlib.StateKeyTuple{EventType: gomatrixserverlib.MRoomCreate, StateKey: ""}]
+	oldMembershipEvent := state[gomatrixserverlib.StateKeyTuple{EventType: gomatrixserverlib.MRoomMember, StateKey: userID}]
+	oldPowerLevelsEvent := state[gomatrixserverlib.StateKeyTuple{EventType: gomatrixserverlib.MRoomPowerLevels, StateKey: ""}]
+	oldJoinRulesEvent := state[gomatrixserverlib.StateKeyTuple{EventType: gomatrixserverlib.MRoomJoinRules, StateKey: ""}]
+
+	// Create the new room create event. Using a map here instead of CreateContent
+	// means that we preserve any other interesting fields that might be present
+	// in the create event (such as for the room types MSC).
```
kegsay, line 368: Excellent comment.

roomserver/internal/perform/perform_upgrade.go
```diff
@@ -0,0 +1,691 @@
...
+	}
+	newCreateContent["creator"] = userID
+	newCreateContent["room_version"] = newVersion
+	newCreateContent["predecessor"] = gomatrixserverlib.PreviousRoom{
+		EventID: tombstoneEvent.EventID(),
+		RoomID:  roomID,
+	}
+	newCreateEvent := fledglingEvent{
+		Type:    gomatrixserverlib.MRoomCreate,
+		Content: newCreateContent,
```
kegsay, line 384: I'd prefer if it we explicitly set `StateKey: ""` rather than relying on zero values.

roomserver/internal/perform/perform_upgrade.go
```diff
@@ -0,0 +1,691 @@
...
+
+	// Make a copy of the existing power level content.
+	tempPowerLevelContent := *powerLevelContent
+	powerLevelsOverridden := false
+
+	// At this point, the "Users", "Events" and "Notifications" keys are all
+	// pointing to the map of the original PL content, so we will specifically
+	// override the users map with a new one and duplicate the values deeply,
+	// so that we can modify them without modifying the original.
+	tempPowerLevelContent.Users = make(map[string]int64, len(powerLevelContent.Users))
```
kegsay, line 629: Whilst this may be marginally slowly, I'd prefer it if we just did a JSON round-trip to deep-copy it as it's less difficult to get wrong when new fields are added e.g:
```go
b, err := json.Marshal(powerLevelContent)
if err != nil {
    return nil, err
}
var tempPowerLevelsContent gomatrixserverlib.PowerLevelsContent
if err := json.Unmarshal(b, &tempPowerLevelsContent); err != nil {
    return nil, err
}
```

roomserver/internal/perform/perform_upgrade.go
```diff
@@ -0,0 +1,691 @@
...
+	}
+
+	// We might need to temporarily give ourselves a higher power level
+	// than we had in the old room in order to be able to send all of
+	// the relevant state events. This function will return whether we
+	// had to override the power level events or not — if we did, we
+	// need to send the original power levels again later on.
+	tempPowerLevelsEvent, powerLevelsOverridden := createTemporaryPowerLevels(powerLevelContent, userID)
+
+	joinRulesContent, err := oldJoinRulesEvent.JoinRule()
```
kegsay, line 416: I really don't know why we are validating all these things here. This increases the chance of us failing to read content bodies and failing to upgrade, which feels bad.

For example, say there was a bug in room version X in join rules which means it returns an error when calling `.JoinRule()`. We _cannot upgrade this room_ because we check these things here. Unless it is absolutely required, we should really be making the minimum amount of assumptions here given these rooms may be broken (hence why they're being upgraded). Why not just unmarshal into `map[string]interface{}` for the content when you don't parse out fields?

kegsay: This would presumably also fix the sytest about PLs >100.

roomserver/internal/perform/perform_upgrade.go
```diff
@@ -0,0 +1,691 @@
...
+	}
+
+	// Duplicate all of the old state events into the new room.
+	for tuple, event := range state {
+		if _, ok := override[tuple]; ok {
+			// Don't duplicate events we have overridden already. They
+			// are already in `eventsToMake`.
+			continue
+		}
+		if membership, merr := event.Membership(); merr == nil && membership != gomatrixserverlib.Ban {
```
kegsay, line 449: It's icky that we're doing exactly the same check in two places. It would be better if we did a single pass over the old state and bucket them into "stuff to send" and "stuff to ignore".

roomserver/internal/perform/perform_upgrade.go
```diff
@@ -0,0 +1,691 @@
...
+		if err = json.Unmarshal(event.Content(), &newEvent.Content); err != nil {
+			logrus.WithError(err).Error("Failed to unmarshal old event")
+			continue
+		}
+		eventsToMake = append(eventsToMake, newEvent)
+	}
+
+	// If we sent a temporary power level event into the room before,
+	// override that now by restoring the original power levels.
+	if powerLevelsOverridden {
```
kegsay, line 468: I don't understand why this is necessary. We are _creating the room_ ergo we will have the power to do anything we like. Why aren't we just making the room, sending all the state bar the PL event, then send the PL event (which may drop our privileges?)

roomserver/internal/perform/perform_upgrade.go
```diff
@@ -0,0 +1,691 @@
...
+	}
+	return eventsToMake, nil
+}
+
+func (r *Upgrader) sendInitialEvents(ctx context.Context, evTime time.Time, userID, newRoomID, newVersion string, eventsToMake []fledglingEvent) *api.PerformError {
+	var err error
+	var builtEvents []*gomatrixserverlib.HeaderedEvent
+	authEvents := gomatrixserverlib.NewAuthEvents(nil)
+	for i, e := range eventsToMake {
+		depth := i + 1 // depth starts at 1
```
kegsay, line 482: Surely this sort of stuff should be shoved into GMSL or eventutil? We do the same thing in /createRoom.

roomserver/internal/perform/perform_upgrade.go
```diff
@@ -0,0 +1,691 @@
...
+	}
+
+	// if the old room is published (was public), publish the new room
+	if len(pubQueryRes.RoomIDs) == 1 {
+		publishNewRoom(ctx, r.URSAPI, roomID, newRoomID)
+	}
+	return nil
+}
+
+func publishNewRoom(
```
kegsay, line 250: Bad naming. It doesn't _just_ publish the new room, it removes the old one too.

roomserver/internal/perform/perform_upgrade.go
```diff
@@ -0,0 +1,691 @@
...
+		if err = URSAPI.SetRoomAlias(ctx, &setAliasReq, &setAliasRes); err != nil {
+			return &api.PerformError{
+				Msg: "api.SetRoomAlias failed",
+			}
+		}
+	}
+	return nil
+}
+
+func (r *Upgrader) clearOldCanonicalAliasEvent(ctx context.Context, evTime time.Time, userID, roomID string) *api.PerformError {
```
kegsay, line 212: We never seem to check if there even _is_ a canonical alias event. Surely we need to before calling this function?

roomserver/internal/perform/perform_upgrade.go
```diff
@@ -0,0 +1,691 @@
...
+}
+
+func (r *Upgrader) restrictOldRoomPowerLevels(ctx context.Context, evTime time.Time, userID, roomID string) *api.PerformError {
+	powerLevelContent, pErr := r.getRoomPowerLevels(ctx, roomID)
+	if pErr != nil {
+		return pErr
+	}
+
+	restrictedPowerLevelContent := &gomatrixserverlib.PowerLevelContent{}
+	*restrictedPowerLevelContent = *powerLevelContent
```
kegsay, line 152: Needs comments.

roomserver/internal/perform/perform_upgrade.go
```diff
@@ -0,0 +1,691 @@
...
+
+	restrictedDefaultPowerLevel := int64(50)
+	if restrictedPowerLevelContent.UsersDefault+1 > restrictedDefaultPowerLevel {
+		restrictedDefaultPowerLevel = restrictedPowerLevelContent.UsersDefault + 1
+	}
+	restrictedPowerLevelContent.EventsDefault = restrictedDefaultPowerLevel
+	restrictedPowerLevelContent.Invite = restrictedDefaultPowerLevel
```
kegsay, line 159: All of this needs comments. We cannot just set random PL numbers without justification as it affects auth.

kegsay (changes requested):

clientapi/routing/upgrade_room.go
neilalexander: Nothing requires the room version to be newer. You can "upgrade" to the same or an older version.

roomserver/internal/perform/perform_upgrade.go
neilalexander: I think the second check is actually not needed, so I'll remove it.

roomserver/internal/perform/perform_upgrade.go
neilalexander: I think because of this: https://github.com/matrix-org/gomatrixserverlib/blob/0980b7f341e0c70adc2784eb8a978578835bbd85/eventauth.go#L266-L271

roomserver/internal/perform/perform_upgrade.go
neilalexander: Possibly, but then we also have to refactor that part of `/createRoom`, which feels like it's sprawling out of the scope of this PR.

PR #2309, opened 2022-03-28
Allow stored session parameters to be overwritten in the registration request

neilalexander:
This changes the priority of the parameter cache around a bit so that we prefer values provided in the request over those in the cache. This fixes a bug with Element iOS registration where the registration can fail because of a `"missing password"` error if the initial request in the flow didn't contain the password.

### DIFF
--- a/clientapi/routing/register.go
+++ b/clientapi/routing/register.go
@@ -557,8 +557,6 @@ func Register(
 		return handleGuestRegistration(req, r, cfg, userAPI)
 	}
 
-	// Retrieve or generate the sessionID
-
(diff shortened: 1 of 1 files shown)


clientapi/routing/register.go
```diff
@@ -525,21 +525,40 @@ func Register(
 	userAPI userapi.UserRegisterAPI,
 	cfg *config.ClientAPI,
 ) util.JSONResponse {
+	reqBody, err := ioutil.ReadAll(req.Body)
```
S7evinK, line 528: Missing `defer req.Body.Close()`

S7evinK (approved):

PR #2311, opened 2022-03-28
User directory for nearby Pinecone peers (P2P demo)

neilalexander:
This interfaces `UserDirectoryProvider` so that we can use a custom implementation that allows federated queries of the user directory. This is specifically used for the Pinecone P2P demos, for finding nearby peers to start chats with.


clientapi/routing/userdirectory.go
```diff
@@ -67,8 +68,16 @@ func SearchUserDirectory(
...
 
-		userID := fmt.Sprintf("@%s:%s", user.Localpart, serverName)
+		var userID string
+		if user.ServerName != "" {
+			userID = fmt.Sprintf("@%s:%s", user.Localpart, user.ServerName)
+		} else {
+			userID = fmt.Sprintf("@%s:%s", user.Localpart, serverName)
+		}
 		if _, ok := results[userID]; !ok {
+			if userID == "_server" {
```
kegsay, line 78: This needs commenting to explain who sets this and why.

kegsay (approved):

PR #2312, opened 2022-03-31
Add presence module V2

S7evinK:
Add the presence module.
Since `presence` is stored in the syncapi, this also adds a first "Request/Reply" NATS "handler", where the ClientAPI requests the current presence status of a specific user.

Todo:
- [x] Add config to disable presence
- [x] Move direct access to Jetstream to a producer


clientapi/producers/syncapi.go
```diff
@@ -173,3 +175,21 @@ func (p *SyncAPIProducer) SendTyping(
...
+	ctx context.Context, userID, presence string, statusMsg *string,
+) error {
+	m := nats.NewMsg(p.TopicPresenceEvent)
+	m.Header.Set(jetstream.UserID, userID)
+	m.Header.Set("presence", presence)
+	nilMsg := statusMsg == nil
+	if !nilMsg {
+		m.Header.Set("status_msg", *statusMsg)
+	}
+	m.Header.Set("status_msg_nil", strconv.FormatBool(nilMsg))
```
neilalexander, line 189: Do we need to do this? I would naively expect that a lack of the `status_msg` header on the receiving side could be considered as a `nil` status message.

clientapi/routing/presence.go
```diff
@@ -0,0 +1,138 @@
...
+	req *http.Request,
+	device *api.Device,
+	natsClient *nats.Conn,
+	presenceTopic string,
+	userID string,
+) util.JSONResponse {
+	msg := nats.NewMsg(presenceTopic)
+	msg.Header.Set(jetstream.UserID, userID)
+
+	presence, err := natsClient.RequestMsg(msg, time.Second*10)
```
neilalexander, line 99: It might actually be better just to use HTTP here for now. Then when we come to change things from HTTP to NATS request/reply later, we can do so inside the API packages and not have to worry about other places throughout the codebase.

dendrite-config.yaml
```diff
@@ -68,6 +68,9 @@ global:
   # to other servers and the federation API will not be exposed.
   disable_federation: false
 
+  # Disable presence. Dendrite will not handle presence events.
+  disable_presence: true
```
neilalexander, line 72: Unless there's a considerable performance hit from having presence enabled, do we want to leave it enabled in our sample config?

federationapi/consumers/presence.go
```diff
@@ -0,0 +1,139 @@
...
+		log.WithError(err).Error("failed to get joined hosts")
+		return true
+	}
+	if len(joined) == 0 {
+		return true
+	}
+
+	newStatusMsg := &statusMsg
+	if nilStatusMsg {
+		newStatusMsg = nil
```
neilalexander, line 106: Same here as other comment — let's just check for the absence of `"status_msg"` instead.

federationapi/producers/syncapi.go
```diff
@@ -142,3 +144,21 @@ func (p *SyncAPIProducer) SendTyping(
...
+func (p *SyncAPIProducer) SendPresence(
+	ctx context.Context, userID, presence string, statusMsg *string, lastActiveAgo int64,
+) error {
+	m := nats.NewMsg(p.TopicPresenceEvent)
+	m.Header.Set(jetstream.UserID, userID)
+	m.Header.Set("presence", presence)
+	if statusMsg != nil {
+		m.Header.Set("status_msg", *statusMsg)
+	}
+	m.Header.Set("status_msg_nil", strconv.FormatBool(statusMsg == nil))
```
neilalexander, line 157: Here too.

setup/jetstream/streams.go
```diff
@@ -89,4 +91,9 @@ var streams = []*nats.StreamConfig{
 		Retention: nats.InterestPolicy,
 		Storage:   nats.FileStorage,
 	},
+	{
+		Name:      OutputPresenceEvent,
+		Retention: nats.InterestPolicy,
+		Storage:   nats.FileStorage,
```
neilalexander, line 97: We probably want `MemoryStorage` for this particular stream, since I don't imagine we care about sending delayed presence updates if Dendrite were to die. We'd just end up generating new ones.

syncapi/sync/requestpool.go
```diff
@@ -80,6 +92,52 @@ func (rp *RequestPool) cleanLastSeen() {
...
+	}
+}
+
+// updatePresence sends presence updates to the SyncAPI and FederationAPI
+func (rp *RequestPool) updatePresence(presence string, userID string) {
+	if rp.cfg.Matrix.DisablePresence {
+		return
+	}
+	if presence == "" {
+		presence = "online"
```
neilalexander, line 115: If these terms like `"online"`, `"unavailable"` have special meaning from the spec, let's define them as `const` somewhere.

neilalexander (commented):
Overall this is excellent, just a few minor things.

setup/jetstream/streams.go
neilalexander: We could also potentially set a `MaxAge`, like `OutputTypingEvent`, so that if an update becomes stale after a certain period of time, we just drop it.

clientapi/routing/presence.go
S7evinK: tbh, I didn't want to add a HTTP API, as the SyncAPI doesn't have one yet and it felt like more overhead.

dendrite-config.yaml
S7evinK: It's disabled by default :)

setup/jetstream/streams.go
S7evinK: Sounds like a good idea.
We probably also need to have a look at the FederationAPI later, as presence events could pile up for servers and we'd send them old events.

S7evinK: The question is, what is a good `MaxAge` for the stream?

clientapi/routing/presence.go
neilalexander: I suppose that's a good point — maybe worth leaving it as it is then rather than adding an API for now.

setup/jetstream/streams.go
neilalexander: I suppose however long we consider a presence update to be "stale". Is that in the order of minutes or so? Does the spec say how often we should be sending things?

S7evinK: The spec just says 
> The server will automatically set a user’s presence to unavailable if their last active time was over a threshold value (e.g. 5 minutes).

that's the 5min cleanup period which changes the presence.

clientapi/producers/syncapi.go
S7evinK: Hm.. Having tested a bit and switching everything to something like
```go
var statusMsg *string = nil
if data, ok := m.Header["status_msg"]; ok && len(data) > 0 {
  statusMsg = &data[0]
}
```
(`m.Header.Get("status_msg")` returns empty strings..) and
```go
if  statusMsg != nil {
   m.Header.Set("status_msg", *statusMsg)
}
```
It doesn't seem to work as expected.

Issue #2275, opened 2022-03-13
Please publish binaries with releases


08d2:
It would be helpful if published releases e.g. [0.6.5](https://github.com/matrix-org/dendrite/releases/tag/v0.6.5) included binaries for supported platforms, even if that's just linux/amd64 for now. I appreciate that you're publishing Docker images, but binaries are more widely usable.

ealib:
The availability of precompiled Dendrite binaries would be particularly important for Windows users, of course.

ealib:
For reference only, other projects written in Go that release pre-compiled binaries for various platforms supported by the toolchain:

* [Caddy](https://github.com/caddyserver/caddy/releases)
* [Gitea](https://github.com/go-gitea/gitea/releases)

Furthermore, for full integration in Windows, where background processes have to comply with certain rules in order to be compatible with the *Service Control Manager*, the easiest way seems to be to use the [Minimal windows service stub](https://pkg.go.dev/github.com/kardianos/minwinsvc).

ealib:
I forgot that the very [NATS](https://nats.io/) publishes

* [NATS server pre-built binaries](https://github.com/nats-io/nats-server/releases)
* [NATS Windows Service](https://docs.nats.io/running-a-nats-service/introduction/windows_srv)

kegsay:
We are not planning on supporting binaries with releases at this time. We already have docker images with different architectures. We are a very small team and are sadly unable to maintain a comprehensive release setup, as the cost of this would be too great as we'd need to ensure binaries don't have regressions _in addition_ to the Docker releases which we actively use and test.

We may revisit this in the future.

PR #2319, opened 2022-04-04
build.sh to build.cmd

ealib:
Convert and adapt the build script for Windows users.

Binaries are for `windows-amd64` and are built in `.\bin` as in the original script.

This PR does not require testing because it simply increases the tools available for binaries generation: the binaries themselves represent the passing of the test.

See this also as an interim step for #2275

### Pull Request Checklist

Signed-off-by: `Emanuele Aliberti <dev@mtka.eu>`


build.cmd
```diff
@@ -0,0 +1,53 @@
...
+    set FLAGS=-X github.com/matrix-org/dendrite/internal.branch=%BRANCH% -X github.com/matrix-org/dendrite/internal.build=%BUILD%
+
+:CHECK_BIN
+    if exist "bin" goto :ALL_SET
+    mkdir "bin"
+
+:ALL_SET
+    set CGO_ENABLED=1
+    set GOOS=windows
+    set GOARCH=amd64
```
neilalexander, line 39: This seems like it will do the wrong thing on 32-bit Windows?

ealib: I removed hardwired `GOOS` and `GOARCH` to let the Go toolchain determine what the characteristics of the host are.

ealib: I updated `docs/INSTALL.md` too.

neilalexander (approved):
Thanks for this!

PR #2320, opened 2022-04-04
Slower federation warm-up

neilalexander:
Right now, at startup, the federation API tries to wake up destination queues and the key server tries to refresh stale device lists almost instantly. This can create a sudden spike in federation activity and CPU usage.

This PR adjusts things so that we wake up a bit more gradually — federation queues wake up after 5 seconds and the stale updater after 10 seconds — and each server is stepped by a second, to create much less of a sudden shock of traffic.

### DIFF
--- a/federationapi/queue/queue.go
+++ b/federationapi/queue/queue.go
@@ -119,11 +119,11 @@ func NewOutgoingQueues(
 		} else {
 			log.WithError(err).Error("Failed to get EDU server names for destination queue hydration")
 		}
-		offset := time.Duration(5)
+		offset := time.Second * 5
 		for serverName := range serverNames {
 			if queue := queues.getQueue(serverName); queue != nil {
-				time.AfterFunc(time.Second*offset, queue.wakeQueueIfNeeded)
-				offset++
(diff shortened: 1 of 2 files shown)


federationapi/queue/queue.go
```diff
@@ -104,28 +104,28 @@ func NewOutgoingQueues(
...
-					queue.wakeQueueIfNeeded()
-				}
+		} else {
+			log.WithError(err).Error("Failed to get EDU server names for destination queue hydration")
+		}
+		offset := time.Second * 5
+		for serverName := range serverNames {
+			if queue := queues.getQueue(serverName); queue != nil {
+				time.AfterFunc(offset, queue.wakeQueueIfNeeded)
+				offset += time.Second
```
kegsay, line 126: This probably isn't good enough if we have a few thousand servers (not entirely unrealistic) as this would wait 33min for 2000 servers. It would be better if it was stochastic over a large but reasonable timeframe (e.g random number between 0-120 and then wake after that many seconds).

kegsay (approved):

PR #2321, opened 2022-04-05
Add Are We Synapse Yet to GHA

S7evinK:
Even if already run from Sytest, I think this makes it easier to get a fast overview (without having to scroll down).
Are We Synapse Yet runs in verbose mode and groups the tests.

### DIFF
--- a/.github/workflows/dendrite.yml
+++ b/.github/workflows/dendrite.yml
@@ -233,7 +233,14 @@ jobs:
       - name: Summarise results.tap
         if: ${{ always() }}
         run: /sytest/scripts/tap_to_gha.pl /logs/results.tap
-
+      - name: Sytest List Maintenance
+        if: ${{ always() }}
+        run: /src/show-expected-fail-tests.sh /logs/results.tap /src/sytest-whitelist /src/sytest-blacklist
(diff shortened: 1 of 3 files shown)


.github/workflows/dendrite.yml
```diff
@@ -233,7 +233,14 @@ jobs:
       - name: Summarise results.tap
         if: ${{ always() }}
         run: /sytest/scripts/tap_to_gha.pl /logs/results.tap
-
+      - name: Sytest List Maintenance
+        if: ${{ always() }}
```
kegsay, line 237: What's the deal with this? I've not seen this before.

kegsay (approved):
Looks great!

.github/workflows/dendrite.yml
S7evinK: The list maintenance has already been in Buildkite, but somehow got lost over time.
Not sure if the `if: ${{ always() }}` should be left in or if we want this to fail the build.

neilalexander: FWIW I don't think things that are purely informational should fail the pipeline.

PR #2322, opened 2022-04-05
Fix full state for new joins check

neilalexander:
The sync API could trip over no-op membership events and end up sending the full room state instead of the delta. This should fix that.

### DIFF
--- a/syncapi/storage/shared/syncserver.go
+++ b/syncapi/storage/shared/syncserver.go
@@ -690,10 +690,12 @@ func (d *Database) GetStateDeltas(
 
 	allRoomIDs := make([]string, 0, len(memberships))
 	joinedRoomIDs := make([]string, 0, len(memberships))
+	joinedRoomMap := make(map[string]struct{}, len(memberships))
 	for roomID, membership := range memberships {
 		allRoomIDs = append(allRoomIDs, roomID)
 		if membership == gomatrixserverlib.Join {
 			joinedRoomIDs = append(joinedRoomIDs, roomID)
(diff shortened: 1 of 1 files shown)


syncapi/storage/shared/syncserver.go
```diff
@@ -739,13 +741,8 @@ func (d *Database) GetStateDeltas(
...
 	for roomID, stateStreamEvents := range state {
 		for _, ev := range stateStreamEvents {
-			// TODO: Currently this will incorrectly add rooms which were ALREADY joined but they sent another no-op join event.
-			//       We should be checking if the user was already joined at fromPos and not proceed if so. As a result of this,
-			//       dupe join events will result in the entire room state coming down to the client again. This is added in
-			//       the 'state' part of the response though, so is transparent modulo bandwidth concerns as it is not added to
-			//       the timeline.
 			if membership := getMembershipFromEvent(ev.Event, userID); membership != "" {
-				if membership == gomatrixserverlib.Join {
+				if _, ok := joinedRoomMap[roomID]; !ok && membership == gomatrixserverlib.Join {
```
kegsay, line 745: Are we sure that `joinedRoomMap` is getting joined rooms from _before_ the `state` array?

kegsay (approved):
Otherwise LGTM

syncapi/storage/shared/syncserver.go
neilalexander: Actually, that's a rather good point. No, it probably won't.

PR #2332, opened 2022-04-08
Handle QueryAccountAvailabilityPath in internal user API

PiotrKozimor:
Seems like this one was skipped, I couldn't find test cases in sytest for `/register/available`. It would be perfect to write them, but it's out of this PR scope.

### Pull Request Checklist


neilalexander (approved):
Good catch — thanks for this!

PR #2337, opened 2022-04-08
Remove libp2p demos

neilalexander:
We don't maintain these anymore, the version of libp2p is old and upgrading it would require API changes.


kegsay (changes requested):
We need to have a WASM compatible thing running. I'm surprised CI isn't whining given I thought we tested dendritejs at least built.

neilalexander:
I think the WASM tests `dendritejs-pinecone`, not `dendritejs`, but the WASM tests are already unhappy anyway apparently.

PR #2339, opened 2022-04-08
Correctly use provided filters

S7evinK:
Updates the filters, so they can be used "correctly".
Also simplifies `SelectAccountDataInRange`.


syncapi/routing/context.go
```diff
@@ -60,7 +60,11 @@ func Context(
 			Headers: nil,
 		}
 	}
-	filter.Rooms = append(filter.Rooms, roomID)
+	if filter.Rooms != nil {
+		roomsFilter := *filter.Rooms
+		roomsFilter = append(roomsFilter, roomID)
+		filter.Rooms = &roomsFilter
```
neilalexander, line 66: `*filter.Rooms = append(*filter.Rooms, roomID)` will work if `filter.Rooms != nil`.

syncapi/streams/stream_pdu.go
```diff
@@ -423,8 +423,12 @@ func (p *PDUStreamProvider) addIgnoredUsersToFilter(ctx context.Context, req *ty
 		return err
 	}
 	req.IgnoredUsers = *ignores
+	var userList []string = nil
```
neilalexander, line 426: `userList := make([]string, 0, len(ignores.List))` to reduce the number of re-allocations.

neilalexander (approved):

PR #2343, opened 2022-04-09
Add hardened systemd unit file

S7evinK:
Add a hardened systemd service, based on the work of [jassu](https://matrix.to/#/@jassu:kapsi.fi)

→ Overall exposure level for dendrite.service: 1.1 OK

With this, there's no need to create a user for the service, as it uses `DynamicUser`. When switching from the "normal" service file, it is required to change the ownership of media/jetstream files.


kegsay (commented):
How can we test this?

S7evinK:
Unfortunately, I don't think we can test this.

kegsay (commented):
I'm uncomfortable telling people to use this when we cannot test this to ensure it doesn't bitrot away.

spaetz:
> I'm uncomfortable telling people to use this when we cannot test this to ensure it doesn't bitrot away.

OTOH, we don't really test the current dendrite.service file either, so this is not much worse than we have now :-). Wouldn't a CI image that starts dendrite via this systemd file with a simple check of _matrix/server/versions be sufficient?

neilalexander (changes requested):
I think on the whole I'd be much more comfortable with adding a `Hardening` page to the documentation that explains how to do this and what the options do, and allow users to pick what's right for them. Chucking a sample unit file over the fence at people that's this big and with so many side effects feels like danger to me.

S7evinK:
Closing this for now, will re-visit this later.

PR #2346, opened 2022-04-12
Implement lazy loading on `/sync`

S7evinK:
Sync API                 :  91% (10/11 tests)
    ✅ Lazy loading parameters in the filter are strictly boolean
    ✅ The only membership state included in an initial sync is for all the senders in the timeline
    ✅ The only membership state included in an incremental sync is for senders in the timeline
    ❌ The only membership state included in a gapped incremental sync is for senders in the timeline
    ✅ Gapped incremental syncs include all state changes
    ✅ Old leaves are present in gapped incremental syncs
    ✅ Leaves are present in non-gapped incremental syncs
    ✅ Old members are included in gappy incr LL sync if they start speaking
    ✅ Members from the gap are included in gappy incr LL sync
    ✅ We don't send redundant membership state across incremental syncs by default
    ✅ We do send redundant membership state across incremental syncs if asked

`The only membership state included in a gapped incremental sync is for senders in the timeline` is also blacklisted on Synapse, as it conflicts with `Gapped incremental syncs include all state changes` afaict.

### DIFF
--- a/internal/caching/cache_lazy_load_members.go
+++ b/internal/caching/cache_lazy_load_members.go
@@ -3,6 +3,8 @@ package caching
 import (
 	"fmt"
 	"time"
+
+	userapi "github.com/matrix-org/dendrite/userapi/api"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


internal/caching/cache_lazy_load_members.go
```diff
@@ -0,0 +1,71 @@
...
+
+type LazyLoadCache struct {
+	// Mapping from userID/deviceID to InMemoryLRUCachePartition
+	userCaches map[string]*InMemoryLRUCachePartition
+}
+
+// NewLazyLoadCache creates a new LazyLoadCache.
+func NewLazyLoadCache() *LazyLoadCache {
+	return &LazyLoadCache{
+		userCaches: make(map[string]*InMemoryLRUCachePartition),
```
neilalexander, line 25: Should this be unbounded? I was somewhat expecting that we'd have an LRU cache containing LRU partitions.

ara4n:
Yup, from memory: `❌ The only membership state included in a gapped incremental sync is for senders in the timeline` was a thinko on my side: if you don't include all membership changes during a gapped incremental sync, your membership list just gets out of sync.  Given /members doesn't have a way to report incremental changes (in v2 sync at least), you'd have had to do a full new /members after every gappy sync to keep the membership list in sync, which is obviously inefficient. So instead, we gave up suppressing member changes in gaps for now when in LL load.

cc @kegsay given this is something that LL in sliding sync could/should fix, by providing a way for clients to incrementally sync membership details when needed rather than being forced into it.

neilalexander (approved):

PR #2348, opened 2022-04-13
Precompute values for `userIDSet` in sync notifier

neilalexander:
This reduces the CPU burn on the hot paths that use `values()` by using precomputed arrays rather than calculating them each time. We now compute the slices on modify rather than on access.

### DIFF
--- a/syncapi/notifier/notifier.go
+++ b/syncapi/notifier/notifier.go
@@ -262,7 +262,10 @@ func (n *Notifier) SharedUsers(userID string) []string {
 func (n *Notifier) _sharedUsers(userID string) []string {
 	n._sharedUserMap[userID] = struct{}{}
 	for roomID, users := range n.roomIDToJoinedUsers {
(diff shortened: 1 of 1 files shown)


syncapi/notifier/notifier.go
```diff
@@ -282,8 +282,8 @@ func (n *Notifier) IsSharedUser(userA, userB string) bool {
 	defer n.lock.RUnlock()
 	var okA, okB bool
 	for _, users := range n.roomIDToJoinedUsers {
-		_, okA = users[userA]
-		_, okB = users[userB]
+		okA = users.isIn(userA)
+		okB = users.isIn(userB)
```
S7evinK, line 286: Just a minor thing that I already had seen yesterday, no need to access the map twice if `userA` isn't already in there.
```suggestion
		okA = users.isIn(userA)
		if !okA {
		    continue
		}
		okB = users.isIn(userB)
```

S7evinK (approved):

Issue #1765, opened 2021-02-14
Anything that isn't an image won't upload and seems to be caught by the thumbnailer


w3bb:
### Background information

- **0.3.9**
- **Monolith**
- **Postgres**
- **Using the Alpine package, no docker**
- **Go version N/A**

### Description

When uploading a video/generic file, Dendrite will refuse to upload the file because of a thumbnailer error.

### Steps to reproduce

- Install the Dendrite package on Alpine
- Set Dendrite up
- Upload an mp4/webm/zip file

Dendrite should upload files that aren't images, and it really shouldn't be running the thumbnailer on things that aren't images.

``` time="<snip>" level=error msg="Failed to read src file" func=github.com/matrix-org/dendrite/mediaapi/thumbnailer.GenerateThumbnails file="github.com/matrix-org/dendrite/mediaapi/thumbnailer/thumbnailer_nfnt.go:53" Origin=spiderden.net error="image: unknown format" <snipped the rest> ```

It looks identical for for the webms, zips, and mp4s I tested. Uploading on my usual homeserver running Synapse doesn't fail.

w3bb:
It still occurs with the 0.3.10

Cool-fire:
@w3bb Tried to reproduce this issue in my homeserver running dendrite and element client. I am able to upload any type of file which is small in size but getting Upload failed with large sized files and not seeing logs for large sized files so might be issue with the client as it might not be accepting file size more than a threshold.

Also agree that GenerateThumbnails is running for every file and it is throwing error if the file is not Image.

neilalexander:
It's worth noting that the thumbnailer runs *after* the upload has been persisted and a thumbnailer error is not fatal (we log it but don't do anything else).

Is this still happening on recent versions of Dendrite, e.g. v0.4.0?

Salastil:
> It's worth noting that the thumbnailer runs _after_ the upload has been persisted and a thumbnailer error is not fatal (we log it but don't do anything else).
> 
> Is this still happening on recent versions of Dendrite, e.g. v0.4.0?

I can confirm this is occurring on my install that I deployed yesterday. Same file types as the original issue, only images work.

spaetz:
Sure, this is not simply a max file-size issue too? There is a max setting in dendrite, and your front end proxy certainly also has a max limit.

S7evinK:
This should be solved by #2352.
Like mentioned, even with this error message, the file should have been uploaded just fine.

PR #2352, opened 2022-04-14
Refactor media storage layer, add tests

S7evinK:
Refactors the media storage layer to use the commonly used layout.
Adds tests to make sure everything is working as expected (found one issue earlier - wrong statement used when querying metadata by hash).

And.. Tries not to generate thumbnails if the uploaded file is not an image, to avoid unnecessary `warn`/`error` logs like mentioned in #1765 and other places.


kegsay (approved):
💯 

Excellent!

PR #2358, opened 2022-04-19
Change `pushkey_ts` to be seconds

### LINKED ISSUE #2354: pushkey_ts shouldn't be in ms
### Description

 - **What** is the problem: Dendrite is sending `pushkey_ts` fields with ms accuracy, but [the spec](https://spec.matrix.org/v1.2/push-gateway-api/#post_matrixpushv1notify) specifically talks about seconds.
 - **Who** is affected: Users using Fluffychat on Dendrite accounts won't be getting push notifications, because Hedwig, the software used as a push gateway for Fluffychat, [uses 32 bit integers for this field](https://gitlab.com/famedly/company/backend/services/hedwig/-/blob/main/src/models.rs#L69).
 - **How** is this bug manifesting: We get `Apr 16 10:30:13.595 ERROR matrix_hedwig::error: invalid value: integer `1648417058456`, expected u32 at line 1 column 321` in our logs.
 - **When** did this first appear: When #1842 was merged

https://github.com/globekeeper/dendrite/blob/main/internal/pushgateway/pushgateway.go#L53 is the offending line here, as it's using the same timestamp data type as all the other matrix things, but other timestamps in the spec are using milliseconds instead of seconds. It'd probably be sensible to add a complement/sytest test here, that validates that the timestamp is roughly what's expected.

As for the actually affected users: We'll be fixing this in Hedwig, by using a u64 instead of a u32, so this is more about behaving according to spec and less about fixing something for the users.

neilalexander:
This should fix #2354 by changing the `pushkey_ts` field to be seconds [as per the spec](https://github.com/matrix-org/matrix-spec/blob/ca466b5a572a867de79ba9077e6445265b841c96/data/api/push-gateway/push_notifier.yaml#L185-L189). The `gomatrixserverlib.Timestamp` time expects to be milliseconds, not seconds.

### DIFF
--- a/userapi/internal/api.go
+++ b/userapi/internal/api.go
@@ -653,7 +653,7 @@ func (a *UserInternalAPI) PerformPusherSet(ctx context.Context, req *api.Perform
 		return a.DB.RemovePusher(ctx, req.Pusher.AppID, req.Pusher.PushKey, req.Localpart)
 	}
 	if req.Pusher.PushKeyTS == 0 {
-		req.Pusher.PushKeyTS = gomatrixserverlib.AsTimestamp(time.Now())
+		req.Pusher.PushKeyTS = int64(time.Now().Second())
 	}
 	return a.DB.UpsertPusher(ctx, req.Pusher, req.Localpart)
 }
--- a/internal/pushgateway/pushgateway.go
+++ b/internal/pushgateway/pushgateway.go
@@ -3,8 +3,6 @@ package pushgateway
 import (
 	"context"
 	"encoding/json"
-
-	"github.com/matrix-org/gomatrixserverlib"
 )
 
 // A Client is how interactions with a Push Gateway is done.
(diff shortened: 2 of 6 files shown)


kegsay (approved):

userapi/internal/api.go
```diff
@@ -653,7 +653,7 @@ func (a *UserInternalAPI) PerformPusherSet(ctx context.Context, req *api.Perform
 		return a.DB.RemovePusher(ctx, req.Pusher.AppID, req.Pusher.PushKey, req.Localpart)
 	}
 	if req.Pusher.PushKeyTS == 0 {
-		req.Pusher.PushKeyTS = gomatrixserverlib.AsTimestamp(time.Now())
+		req.Pusher.PushKeyTS = int64(time.Now().Second())
```
kegsay, line 656: This is the number of seconds in a minute, not unix seconds. `.Unix()` instead.

PR #2382, opened 2022-04-27
Don't send account data or receipts for left/forgotten rooms

neilalexander:
This should stop so many forgotten rooms from reappearing as "Empty room" after a cache clear & complete sync.


S7evinK (approved):
LGTM, if CI is happy

PR #2384, opened 2022-04-27
Add UserAPI storage tests

S7evinK:
Adds UserAPI storage tests.
Fixes a minor issue - getting a new numeric localpart could fail if a user registered with a numeric localpart earlier.
Also splits up the huge storage interface.


userapi/storage/postgres/devices_table.go
```diff
@@ -235,16 +235,20 @@ func (s *devicesStatements) SelectDevicesByID(ctx context.Context, deviceIDs []s
...
-		var displayName sql.NullString
-		if err := rows.Scan(&dev.ID, &localpart, &displayName); err != nil {
+		if err := rows.Scan(&dev.ID, &localpart, &displayName, &lastseents); err != nil {
 			return nil, err
 		}
 		if displayName.Valid {
 			dev.DisplayName = displayName.String
 		}
+		if lastseents.Valid {
+			dev.LastSeenTS = lastseents.Int64
```
neilalexander, line 250: Presumably if `lastseents.Valid` is not true, the next loop iteration will have the value of the last iteration that was?

neilalexander (approved):
LGTM overall if passing.

PR #2395, opened 2022-04-27
Improve test reliability and debugability

brianathere:
Fix TestThumbnailsStorage (sqlite) failing when media results come back in non-deterministic order
Silence expected error when tests are run multiple times against the same Postgres database

### Pull Request Checklist


test/db.go
```diff
@@ -39,13 +42,18 @@ func createLocalDB(dbName string) {
...
-		createDB.Stdout = os.Stdout
-		createDB.Stderr = os.Stderr
+		createDB.Stdout = io.Writer(&outBuff)
+		createDB.Stderr = io.Writer(&errBuff)
 	}
 	err := createDB.Run()
 	if err != nil && !Quiet {
-		fmt.Println("createLocalDB returned error:", err)
+		// Silence the output if it's just saying the database already exists
+		match, _ := regexp.MatchString("(?m)^createdb: error: database creation failed: ERROR:  database(.*)already exists$", errBuff.String())
```
S7evinK, line 53: Not sure if intended, but there are two whitespaces between `ERROR:` and `database`. Might be better to replace that with a regex.

brianathere: Yes, that's the exact string Postgres returns and (from looking at old posts on Stackoverflow) has been the stable response over many years. I considered adding an actual check if the database exists, but that's more complicated than just filtering the error. 

This portion of the change was to reduce the number of places ERROR shows up in the test run even when it passes. Since this "error" doesn't interfere with the tests succeeding it simply removes it from the log. If any other "error" occurs the behavior is unchanged. I wanted to make the match as exact as reasonable to avoid accidentally filtering other "errors" that may be helpful for debugging.

S7evinK (approved):

neilalexander:
We've cherrypicked in the media API test fix — thanks for this! — but we're having a bit of internal debate about the `createdb` bit, so will come back to that.

PR #2399, opened 2022-04-28
Add housekeeping function to delete old/expired EDUs

S7evinK:
This adds an `expires_at` field to the `federationsender_queue_edus` table which is used to determine which EDUs can be deleted from the database.
Reason: Keeping old events, specially receipts, presence and typing events, can bloat the database and are rarely useful after a given time.


federationapi/storage/shared/storage_edus.go
```diff
@@ -20,10 +20,19 @@ import (
...
 	"errors"
 	"fmt"
+	"time"
 
 	"github.com/matrix-org/gomatrixserverlib"
 )
 
+// expireEDUTypes contains EDUs which can/should be expired after a given time
+// if the target server isn't reachable for some reason.
+var expireEDUTypes = map[string]struct{}{
```
neilalexander, line 30: Suggest we might want to make this `map[string]time.Duration` instead, so that we can specify different expiry times for different EDU types?

For example, a read receipt is useful to keep around for longer, but a typing notification isn't really useful if it doesn't get sent within under a minute, or a presence update in under 5/10 minutes.

federationapi/federationapi.go
```diff
@@ -166,5 +168,16 @@ func NewInternalAPI(
...
 		logrus.WithError(err).Panic("failed to start presence consumer")
 	}
+
+	var cleanExpiredEDUs func()
+	cleanExpiredEDUs = func() {
+		logrus.Infof("Cleaning expired EDUs")
+		if err := federationDB.DeleteExpiredEDUs(base.Context()); err != nil {
+			logrus.WithError(err).Error("Failed to clean expired EDUs")
+		}
+		time.AfterFunc(time.Hour, cleanExpiredEDUs)
```
kegsay, line 178: Please attach this to some sort of context so we can bail when the application needs to quit. This will help with unit tests.

federationapi/storage/postgres/queue_edus_table.go
```diff
@@ -61,6 +63,12 @@ const selectQueueEDUCountSQL = "" +
 const selectQueueServerNamesSQL = "" +
 	"SELECT DISTINCT server_name FROM federationsender_queue_edus"
 
+const selectExpiredEDUsSQL = "" +
+	"SELECT DISTINCT json_nid FROM federationsender_queue_edus WHERE expires_at IS NOT NULL AND expires_at <= $1"
+
+const deleteExpiredEDUsSQL = "" +
+	"DELETE FROM federationsender_queue_edus WHERE expires_at IS NOT NULL AND expires_at <= $1"
```
kegsay, line 70: Presumably then we'll never get rid of EDUs stored prior to this PR as the expires value will be NULL? It might be worth doing something like `DEFAULT now()` or something?

federationapi/storage/shared/storage_edus.go
```diff
@@ -33,13 +41,20 @@ func (d *Database) AssociateEDUWithDestination(
 	receipt *Receipt,
 	eduType string,
 ) error {
+	var expiresAt *gomatrixserverlib.Timestamp
+	if duration, ok := expireEDUTypes[eduType]; ok {
```
kegsay, line 45: I think all EDUs should probably expire given they are ephemeral. Maybe default to 24h or something?

federationapi/storage/sqlite3/deltas/2022042812473400_addexpiresat.go
```diff
@@ -0,0 +1,101 @@
...
+
+	"github.com/matrix-org/dendrite/internal/sqlutil"
+)
+
+func LoadAddExpiresAt(m *sqlutil.Migrations) {
+	m.AddMigration(upAddexpiresat, downAddexpiresat)
+}
+
+func upAddexpiresat(tx *sql.Tx) error {
+	_, err := tx.Exec("ALTER TABLE federationsender_queue_edus RENAME TO federationsender_queue_edus_old;")
```
kegsay, line 29: Why? SQLite supports adding columns? https://www.sqlite.org/lang_altertable.html

federationapi/storage/tables/interface.go
```diff
@@ -34,12 +34,15 @@ type FederationQueuePDUs interface {
...
 type FederationQueueEDUs interface {
-	InsertQueueEDU(ctx context.Context, txn *sql.Tx, eduType string, serverName gomatrixserverlib.ServerName, nid int64) error
+	InsertQueueEDU(ctx context.Context, txn *sql.Tx, eduType string, serverName gomatrixserverlib.ServerName, nid int64, expiresAt *gomatrixserverlib.Timestamp) error
 	DeleteQueueEDUs(ctx context.Context, txn *sql.Tx, serverName gomatrixserverlib.ServerName, jsonNIDs []int64) error
 	SelectQueueEDUs(ctx context.Context, txn *sql.Tx, serverName gomatrixserverlib.ServerName, limit int) ([]int64, error)
 	SelectQueueEDUReferenceJSONCount(ctx context.Context, txn *sql.Tx, jsonNID int64) (int64, error)
 	SelectQueueEDUCount(ctx context.Context, txn *sql.Tx, serverName gomatrixserverlib.ServerName) (int64, error)
 	SelectQueueEDUServerNames(ctx context.Context, txn *sql.Tx) ([]gomatrixserverlib.ServerName, error)
+	SelectExpiredEDUs(ctx context.Context, txn *sql.Tx, expiredBefore gomatrixserverlib.Timestamp) ([]int64, error)
+	DeleteExpiredEDUs(ctx context.Context, txn *sql.Tx, expiredBefore gomatrixserverlib.Timestamp) error
```
kegsay, line 44: Please add unit tests for new code. This is easy to do now with the test helper functions.

kegsay (changes requested):

neilalexander (approved):
LGTM!

PR #2401, opened 2022-04-28
Add `/_dendrite/admin/evacuateRoom/{roomID}`

neilalexander:
This is probably the first real Dendrite admin API endpoint which allows you to forcibly remove all local users from your server from a given room.


S7evinK (approved):

r3k2:
nice!

PR #2409, opened 2022-05-02
Return `null` if MaxFileSizeBytes is 0

### LINKED ISSUE #2408: max_file_size_bytes = 0 counts as limited file size
### Background information

- **Dendrite version or git SHA**: 0.8.2
- **Running in Docker?**: Yes
- **`go version`**: -
- **Client used (if applicable)**: Element Web

### Description

 - **What** is the problem:
When `max_file_size_bytes` to `0`, uploading file doesn't count as unlimited as intended. Fetching file from other servers counts as unlimited.

 - **Who** is affected: -
 

Set `max_file_size_bytes` to `0` in `dendrite.yaml` config file

### Steps to reproduce

- Set `max_file_size_bytes` to `0` in `dendrite.yaml` config file
- Restart the server
- Try to upload the file

<!--

``` (three backticks, on a line on their own), so that they are formatted legibly.

-->

### PR DESCRIPTION
**Pull request by S7evinK**
This should fix #2408.

> The maximum size an upload can be in bytes. Clients SHOULD use this as a guide when uploading content. If not listed or null, the size limit should be treated as unknown.

### DIFF
--- a/mediaapi/routing/routing.go
+++ b/mediaapi/routing/routing.go
@@ -74,7 +74,8 @@ func Setup(
 			return *r
 		}
 		respondSize := cfg.MaxFileSizeBytes
-		if *cfg.MaxFileSizeBytes == 0 {
+		// Shouldn't be nil, as this is checked at startup.
+		if cfg.MaxFileSizeBytes != nil && *cfg.MaxFileSizeBytes == 0 {
 			respondSize = nil
 		}
 		return util.JSONResponse{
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: mediaapi/routing/routing.go [hunk 1]
```diff
@@ -73,9 +73,13 @@ func Setup(
 		if r := rateLimits.Limit(req); r != nil {
 			return *r
 		}
+		respondSize := cfg.MaxFileSizeBytes
+		if *cfg.MaxFileSizeBytes == 0 {
```
neilalexander, line 77: Need to check not `nil` here also to avoid a potential panic.

neilalexander (changes requested):

mediaapi/routing/routing.go
S7evinK: That should be done at startup, will add a comment.

neilalexander: If it can't be `nil`, why is it a pointer type to begin with? :-) Let's simplify that.

neilalexander (approved):

PR #2410, opened 2022-05-02
Cleanup configs, use `time.Duration` throughout the config

S7evinK:
After #2411 was merged, this now mainly cleans up the configs and uses `time.Duration`s instead of different time units spread across the config.


nakoo:
What about to delete `build/docker/config/dendrite.yaml` file and ask user to follow these files? I also want to mention there're some inconsistencies of these config files.

1. Push Server API only includes in `build/docker/config/dendrite.yaml`.

https://github.com/matrix-org/dendrite/blob/979a551f1e2aeb9f3417df5e52a7279230b7a3ba/build/docker/config/dendrite.yaml#L315-L324

2. `app_service_api` would be `appservice_api`.

There're so many places that used `app_service_api` instead of `appservice_api` like below.

https://github.com/matrix-org/dendrite/blob/979a551f1e2aeb9f3417df5e52a7279230b7a3ba/build/docker/config/dendrite.yaml#L119-L128

S7evinK:
Closed in favor of #2456 
Changing the different time units will be done when we need to change the config version anyway, as this will break current configs.

PR #2416, opened 2022-05-04
syncapi: define specific interfaces for internal HTTP communications

kegsay:
This will make it easier to test the syncapi. If we like this style of interface declaration, I'll update the remaining components to match.


kegsay:
@S7evinK sez:

> Pretty much prefer this over the previous interfaces

neilalexander (approved):
Looks OK to me, suggest we try it and see how it feels.

PR #2424, opened 2022-05-05
Use sync position supplied from the user stream listener

neilalexander:
This should stop us from racing with the current position, since `ApplyUpdates` will only apply increasing updates.


kegsay:
Why would we ever go backwards?

neilalexander:
That said, I was rather hoping this would help to fix some of the flakiness in sytest and it hasn't.

PR #2437, opened 2022-05-09
Don't try to re-fetch the event if it is listed in `adds_state_event_ids`

neilalexander:
This stops the sync API from trying to fetch the event in the output message itself if it references itself in `adds_state_event_ids` and otherwise treating it like it is missing.

### DIFF
--- a/syncapi/consumers/roomserver.go
+++ b/syncapi/consumers/roomserver.go
@@ -161,7 +161,12 @@ func (s *OutputRoomEventConsumer) onNewRoomEvent(
 	// in the sync API database or we'll need to ask the roomserver.
 	knownEventIDs := make(map[string]bool, len(msg.AddsStateEventIDs))
 	for _, eventID := range msg.AddsStateEventIDs {
-		knownEventIDs[eventID] = eventID == ev.EventID()
+		if eventID == ev.EventID() {
+			knownEventIDs[eventID] = true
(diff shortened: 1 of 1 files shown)


kegsay (approved):

syncapi/consumers/roomserver.go
```diff
@@ -154,41 +154,71 @@ func (s *OutputRoomEventConsumer) onNewRoomEvent(
...
+	if len(missingEventIDs) > 0 {
+		alreadyKnown, err := s.db.Events(ctx, msg.AddsStateEventIDs)
 		if err != nil {
 			return fmt.Errorf("s.db.Events: %w", err)
 		}
-		for _, event := range foundEvents {
-			foundEventIDs[event.EventID()] = true
+		for _, knownEvent := range alreadyKnown {
+			knownEventIDs[knownEvent.EventID()] = true
+			addsStateEvents = append(addsStateEvents, knownEvent)
```
kegsay, line 186: This still isn't done for the initial event if there are no missing events.

kegsay: I would not add incrementally to `addStateEvents` like this. We know that the events we want to add all have keys in the unhelpfully named `knownEventIDs`, we just need to loop that and append rather than do it twice (and forget as in this case).

PR #2441, opened 2022-05-09
Federation consumer changes

neilalexander:
Like #2437 but for the federation API. A bit simpler because it doesn't ask the local database.

### DIFF
--- a/federationapi/consumers/roomserver.go
+++ b/federationapi/consumers/roomserver.go
@@ -146,28 +146,30 @@ func (s *OutputRoomEventConsumer) processInboundPeek(orp api.OutputNewInboundPee
 // processMessage updates the list of currently joined hosts in the room
 // and then sends the event to the hosts that were joined before the event.
 func (s *OutputRoomEventConsumer) processMessage(ore api.OutputNewRoomEvent) error {
-	eventsRes := &api.QueryEventsByIDResponse{}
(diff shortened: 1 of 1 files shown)


federationapi/consumers/roomserver.go
```diff
@@ -146,28 +146,33 @@ func (s *OutputRoomEventConsumer) processInboundPeek(orp api.OutputNewInboundPee
 // processMessage updates the list of currently joined hosts in the room
 // and then sends the event to the hosts that were joined before the event.
 func (s *OutputRoomEventConsumer) processMessage(ore api.OutputNewRoomEvent) error {
-	eventsRes := &api.QueryEventsByIDResponse{}
-	if len(ore.AddsStateEventIDs) > 0 {
+	addsStateEvents := make([]*gomatrixserverlib.HeaderedEvent, 0, len(ore.AddsStateEventIDs))
+	missingEventIDs := make([]string, 0, len(ore.AddsStateEventIDs))
+	for _, eventID := range ore.AddsStateEventIDs {
```
kegsay, line 151: This is basically identical to syncapi. It would be nicer if `OutputNewRoomEvent` had a function on it to return `addsStateEvents, missingEventIDs` perhaps, then both fedapi and syncapi can just call that and then hit roomserver/local db.

kegsay (approved):

PR #2448, opened 2022-05-10
Fix OTK upload spam

S7evinK:
This should stop clients, e.g. gomuks or go-neb, from uploading OTKs if it is not needed.
We now return `device_one_time_keys_count` if the client gave up or the timeout was reached.


syncapi/sync/requestpool.go
```diff
@@ -248,7 +248,15 @@ func (rp *RequestPool) OnIncomingSyncRequest(req *http.Request, device *userapi.
 		defer userStreamListener.Close()
 
 		giveup := func() util.JSONResponse {
+			syncReq.Log.Debugln("Responding to sync since client gave up or timeout was reached")
 			syncReq.Response.NextBatch = syncReq.Since
+			// NOTSPEC: We should always include OTKs in sync responses, otherwise clients might upload keys
+			// even if that's not required. See also:
+			// https://github.com/matrix-org/synapse/blob/29f06704b8871a44926f7c99e73cf4a978fb8e81/synapse/rest/client/sync.py#L276-L281
```
turt2live, line 255: Note that there is actually spec for this:

> Optional. For each key algorithm, the number of unclaimed one-time keys currently held on the server for this device. If an algorithm is unlisted, the count for that algorithm is assumed to be zero. If this entire parameter is missing, the count for all algorithms is assumed to be zero.

https://spec.matrix.org/v1.2/client-server-api/#extensions-to-sync-1

or in other words: by *not* specifying it on the response, the server is saying there's zero keys, thus upload attempts.

S7evinK: Right, but clients don't always seem to handle this (or not correctly), as seen in https://github.com/vector-im/element-android/issues/3725 or now with gomuks.

turt2live: I may have misunderstood the concern, but it originally sounded like clients were spamming OTK uploads, which would be expected if the server excluded the field.

If clients are just outright not uploading keys when they should, that's a client bug and not a server issue.

neilalexander (approved):

syncapi/sync/requestpool.go
turt2live: have opened a spec clarification issue for this: https://github.com/matrix-org/matrix-spec/issues/1057

kegsay (approved):
LGTM

PR #2454, opened 2022-05-12
Test_UserStatistics Fix expected results to match observed results

brianathere:
### Pull Request Checklist

### DIFF
--- a/userapi/storage/tables/stats_table_test.go
+++ b/userapi/storage/tables/stats_table_test.go
@@ -300,10 +300,10 @@ func Test_UserStatistics(t *testing.T) {
 				},
 				R30UsersV2: map[string]int64{
 					"ios":      0,
-					"android":  1,
-					"web":      1,
+					"android":  0,
+					"web":      0,
 					"electron": 0,
-					"all":      2,
+					"all":      0,
 				},
 				AllUsers:        6,
 				NonBridgedUsers: 5,


S7evinK:
The intention of this test is to actually test the `R30UsersV2` output - so if this test fails, there's something wrong anywhere.
I just did a `go test -count=100 -failfast  ./...` for the test on `main` and didn't hit any issues with, on x86_64 (besides a weird SQLite DB corruption).

Therefore some questions:
- Does it fail for SQLite or Postgres (or both?)
- iirc you are using an M1 chip?

brianathere:
> The intention of this test is to actually test the `R30UsersV2` output - so if this test fails, there's something wrong anywhere. I just did a `go test -count=100 -failfast ./...` for the test on `main` and didn't hit any issues with it, on x86_64 (besides a weird SQLite DB corruption).
> 
> Therefore some questions:
> 
> * Does it fail for SQLite or Postgres (or both?)
> * iirc you are using an M1 chip?
> * Any special build flags in `go env`?

It is failing on both SQLite and Postgres
This is running on an M1 Mac

I'm running it via 
```
./build/scripts/build-test-lint.sh
```

No special build flags

I'll take a deeper look today, I'm curious why it's working in one environment, not another.

S7evinK:
This has been failing on (newly added) [scheduled](https://github.com/matrix-org/dendrite/actions/runs/3365826433/jobs/5584225429#step:6:270) runs as well.
Turns out that adding 1h to the last updated time causes issues, when the test is run short past midnight. Should be fixed with https://github.com/matrix-org/dendrite/commit/869bf4d0ac985a34f83d24a74de855089f182b18, therefore closing this.
If this is still happening, please let me know.

PR #2460, opened 2022-05-13
Update NATS Server to version 2.8.2

neilalexander:
This updates the built-in NATS Server.


S7evinK (approved):
Version 2.8.3 :)

neilalexander:
> Version 2.8.3 :)

It's actually 2.8.2, it's just that `go.mod` does weird things when the commit comes after the tagged commit.

PR #2466, opened 2022-05-16
bugfix: E2EE device keys could sometimes not be sent to remote servers

kegsay:
This PR fixes a flakey sytest 'Local device key changes get to remote servers'.

The root cause of this bug is as follows:
 - User A joins a federated room.
 - clientapi talks to roomserver to `PerformJoin`.
 - roomserver talks to federationapi to `PerformJoin`.
 - federationapi does the federated request, checks the response then injects the input room events into the roomserver synchronously (via use of `Asynchronous: false`).
 - roomserver processes the `send_join` events and emits output room events asynchronously.
 - roomserver returns, unblocking federationapi, which returns, which unblocks roomserver's `PerformJoin`, which unblocks clientapi, which returns 200 OK.

At this point, the client thinks they are joined to the room and so may do other actions which depend on that, such as changing device keys and expect those updates to go to servers in the newly joined room. However, there are no guarantees any downstream components are aware of this room yet. Anything which consumes output room events may be lagging. This isn't a problem so much for `syncapi` as it just means the room won't be in the `join` section of the response for a bit. This a problem however for the `federationapi`, which relies on the output room events to update the "joined hosts" table which controls which events get sent where. This can result in this race condition:

- User A gets 200 OK from /join.
- User A uploads new device keys.
- `keyserver` emits key change events.
- `federationapi` consumes the key change event.
- `federationapi` asks `roomserver` for all joined rooms for this user, which returns the newly joined room.
- `federationapi` checks its database for joined hosts in this room, finds nothing, so does not send this event to other servers.
- Eventually, `federationapi` consumes the output room event from `roomserver`, which then updates its joined hosts table correctly, but by this point we've already lost data.

This PR implements a temporary hacky fix by updating the joined hosts table immediately after verifying the `/send_join` response, and before passing these events to the `roomserver`. It's hacky because it only works for `/send_join`: this problem can happen any time a server joins or leaves a room: we may work on stale data.


S7evinK (approved):

S7evinK (approved):
Assuming CI is happy :)

PR #2472, opened 2022-05-18
Really SKIP_NODB

S7evinK:
It's not enough to check the error returned from `sql.Open` as:
> Open may just validate its arguments without creating a connection
to the database. To verify that the data source name is valid, call
Ping.

### DIFF
--- a/test/db.go
+++ b/test/db.go
@@ -55,7 +55,7 @@ func createLocalDB(t *testing.T, dbName string) {
 	}
 	err := createDB.Run()
 	if err != nil && !Quiet {
-		fmt.Println(t, "createLocalDB returned error: %s", err)
+		fmt.Println("createLocalDB returned error:", err)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


test/db.go
```diff
@@ -54,7 +54,7 @@ func createLocalDB(t *testing.T, dbName string) {
 	}
 	err := createDB.Run()
 	if err != nil && !Quiet {
-		fmt.Println("createLocalDB returned error:", err)
+		fatalError(t, "createLocalDB returned error: %s", err)
```
kegsay, line 57: Please don't. Createdb can fail if the database already exists locally which is fine.

test/db.go
```diff
@@ -44,8 +44,9 @@ func fatalError(t *testing.T, format string, args ...interface{}) {
 }
 
 func createLocalDB(t *testing.T, dbName string) {
-	if !Quiet {
-		t.Log("Note: tests require a postgres install accessible to the current user")
+	if _, err := exec.LookPath("createdb"); err != nil && !Quiet {
```
kegsay, line 47: Remove `!Quiet` here else we won't panic when it doesn't exist.

kegsay (approved):

PR #2478, opened 2022-05-23
Restricted join support on `/make_join`, `/send_join`

neilalexander:
This PR does the following things:
* Checks restricted joins and populates `join_authorised_via_users_server` correctly for allowed restricted joins
* Kick back restricted joins that don't match a rule if a new join comes in via `/make_join`
* Signs membership events in `/send_join` so that the signature for the `join_authorised_via_users_server` user is present

### DIFF
--- a/federationapi/internal/perform.go
+++ b/federationapi/internal/perform.go
@@ -213,7 +213,9 @@ func (r *FederationInternalAPI) performJoinUsingServer(
 	// If the remote server returned an event in the "event" key of
 	// the send_join request then we should use that instead. It may
 	// contain signatures that we don't know about.
-	if respSendJoin.Event != nil {
+	if respSendJoin.Event != nil && isWellFormedMembershipEvent(
+		respSendJoin.Event, roomID, userID, r.cfg.Matrix.ServerName,
+	) {
 		event = respSendJoin.Event
 	}
 
@@ -275,6 +277,26 @@ func (r *FederationInternalAPI) performJoinUsingServer(
 	}
 
 	return nil
+}
+
+// isWellFormedMembershipEvent returns true if the event looks like a legitimate
+// membership event.
+func isWellFormedMembershipEvent(event *gomatrixserverlib.Event, roomID, userID string, origin gomatrixserverlib.ServerName) bool {
+	if membership, err := event.Membership(); err != nil {
+		return false
+	} else if membership != gomatrixserverlib.Join {
+		return false
+	}
+	if event.RoomID() != roomID {
+		return false
+	}
+	if event.Origin() != origin {
+		return false
+	}
+	if !event.StateKeyEquals(userID) {
+		return false
+	}
+	return true
 }
 
 // PerformOutboundPeekRequest implements api.FederationInternalAPI
--- a/federationapi/internal/perform.go
+++ b/federationapi/internal/perform.go
@@ -209,10 +209,16 @@ func (r *FederationInternalAPI) performJoinUsingServer(
 	}
 	r.statistics.ForServer(serverName).Success()
 
-	authEvents := respSendJoin.AuthEvents.UntrustedEvents(respMakeJoin.RoomVersion)
+	// If the remote server returned an event in the "event" key of
+	// the send_join request then we should use that instead. It may
+	// contain signatures that we don't know about.
+	if respSendJoin.Event != nil {
+		event = respSendJoin.Event
+	}
 
 	// Sanity-check the join response to ensure that it has a create
 	// event, that the room version is known, etc.
+	authEvents := respSendJoin.AuthEvents.UntrustedEvents(respMakeJoin.RoomVersion)
 	if err = sanityCheckAuthChain(authEvents); err != nil {
 		return fmt.Errorf("sanityCheckAuthChain: %w", err)
 	}
--- a/federationapi/routing/join.go
+++ b/federationapi/routing/join.go
@@ -327,6 +327,14 @@ func SendJoin(
 		}
 	}
 
+	// Sign the membership event. This is required for restricted joins to work
+	// in the case that the authorised via user is one of our own users.
+	signed := event.Sign(
+		string(cfg.Matrix.ServerName),
+		cfg.Matrix.KeyID,
+		cfg.Matrix.PrivateKey,
+	)
+
 	// Send the events to the room server.
 	// We are responsible for notifying other servers that the user has joined
 	// the room, so set SendAsServer to cfg.Matrix.ServerName
@@ -336,7 +344,7 @@ func SendJoin(
 			InputRoomEvents: []api.InputRoomEvent{
 				{
 					Kind:          api.KindNew,
-					Event:         event.Headered(stateAndAuthChainResponse.RoomVersion),
+					Event:         signed.Headered(stateAndAuthChainResponse.RoomVersion),
 					SendAsServer:  string(cfg.Matrix.ServerName),
 					TransactionID: nil,
 				},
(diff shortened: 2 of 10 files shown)


federationapi/internal/perform.go
```diff
@@ -210,10 +210,16 @@ func (r *FederationInternalAPI) performJoinUsingServer(
 	}
 	r.statistics.ForServer(serverName).Success()
 
-	authEvents := respSendJoin.AuthEvents.UntrustedEvents(respMakeJoin.RoomVersion)
+	// If the remote server returned an event in the "event" key of
+	// the send_join request then we should use that instead. It may
+	// contain signatures that we don't know about.
+	if respSendJoin.Event != nil {
+		event = respSendJoin.Event
```
kegsay, line 217: Is this safe to do in the general case? Or can this be used to get homeservers to arbitrarily sign things?

federationapi/routing/join.go
```diff
@@ -314,6 +329,40 @@ func SendJoin(
...
+				Code: http.StatusBadRequest,
+				JSON: jsonerror.BadJSON(fmt.Sprintf("The authorising username %q does not belong to this server.", memberContent.AuthorisedVia)),
+			}
+		}
+	}
+
+	// Sign the membership event. This is required for restricted joins to work
+	// in the case that the authorised via user is one of our own users. It also
+	// doesn't hurt to do it even if it isn't a restricted join.
+	signed := event.Sign(
```
kegsay, line 360: Again, is it safe to be signing arbitrary bytes and then sending it back? Surely we should do more checks?

federationapi/routing/join.go
```diff
@@ -354,10 +403,66 @@ func SendJoin(
...
 		},
 	}
 }
 
+func checkRestrictedJoin(
+	httpReq *http.Request,
+	rsAPI api.FederationRoomserverAPI,
+	roomVersion gomatrixserverlib.RoomVersion,
+	roomID, userID string,
+) (*util.JSONResponse, string, error) {
```
kegsay, line 416: Needs docs on what the return args are.

roomserver/api/api.go
```diff
@@ -134,6 +134,7 @@ type ClientRoomserverAPI interface {
 	QueryRoomVersionForRoom(ctx context.Context, req *QueryRoomVersionForRoomRequest, res *QueryRoomVersionForRoomResponse) error
 	QueryPublishedRooms(ctx context.Context, req *QueryPublishedRoomsRequest, res *QueryPublishedRoomsResponse) error
 	QueryRoomVersionCapabilities(ctx context.Context, req *QueryRoomVersionCapabilitiesRequest, res *QueryRoomVersionCapabilitiesResponse) error
+	QueryRestrictedJoinAllowed(ctx context.Context, req *QueryRestrictedJoinAllowedRequest, res *QueryRestrictedJoinAllowedResponse) error
```
kegsay, line 137: This is wrong? The client API doesn't need to call this function.

roomserver/api/query.go
```diff
@@ -348,6 +348,18 @@ type QueryServerBannedFromRoomResponse struct {
...
 
+type QueryRestrictedJoinAllowedRequest struct {
+	UserID string `json:"user_id"`
+	RoomID string `json:"room_id"`
+}
+
+type QueryRestrictedJoinAllowedResponse struct {
+	Restricted    bool   `json:"restricted"`               // Is the room membership restricted?
+	Resident      bool   `json:"resident"`                 // Is our homeserver in the relevant rooms?
+	Allowed       bool   `json:"allowed"`                  // Is the join allowed by the rules?
```
kegsay, line 359: All of these flags need additional comments. They aren't common terms in use so it's unclear what they mean. Please explain in simple terms.

roomserver/internal/query/query.go
```diff
@@ -757,3 +758,129 @@ func (r *Queryer) QueryAuthChain(ctx context.Context, req *api.QueryAuthChainReq
...
+	if roomInfo == nil || roomInfo.IsStub {
+		return nil // fmt.Errorf("room %q doesn't exist or is stub room", req.RoomID)
+	}
+	// If the room version doesn't allow restricted joins then don't
+	// try to process any further.
+	allowRestrictedJoins, err := roomInfo.RoomVersion.AllowRestrictedJoinsInEventAuth()
+	if err != nil {
+		return fmt.Errorf("roomInfo.RoomVersion.AllowRestrictedJoinsInEventAuth: %w", err)
+	} else if !allowRestrictedJoins {
+		return nil
```
kegsay, line 779: I'm surprised that this query endpoint makes no distinction between "you gave me a bogus room ID" and "your room ID is valid but the room doesn't allow restricted joins"?

roomserver/internal/query/query.go
```diff
@@ -757,3 +758,129 @@ func (r *Queryer) QueryAuthChain(ctx context.Context, req *api.QueryAuthChainReq
...
+	if pending, _, _, err = helpers.IsInvitePending(ctx, r.DB, req.RoomID, req.UserID); err != nil {
+		return fmt.Errorf("helpers.IsInvitePending: %w", err)
+	} else if pending {
+		res.Allowed = true
+		return nil
+	}
+	// We need to get the power levels content so that we can determine which
+	// users in the room are entitled to issue invites. We need to use one of
+	// these users as the authorising user.
+	// Get the join rules to work out if the join rule is "restricted".
```
kegsay, line 812: Comments lie. You get the PL event.

roomserver/internal/query/query.go
```diff
@@ -757,3 +758,129 @@ func (r *Queryer) QueryAuthChain(ctx context.Context, req *api.QueryAuthChainReq
...
+		if err != nil || len(joinNIDs) == 0 {
+			// There should always be more than one join NID at this point
+			// because we are gated behind GetLocalServerInRoom, but y'know,
+			// sometimes strange things happen.
+			continue
+		}
+		// For each of the joined users, let's see if we can get a valid
+		// membership event.
+		for _, joinNID := range joinNIDs {
+			events, err := r.DB.Events(ctx, []types.EventNID{joinNID})
```
kegsay, line 867: Wouldn't it be more efficient to just query all `joinNIDs` in one go? Then loop the result?

kegsay (changes requested):
Needs tests - I'm surprised no additional sytests pass?

neilalexander:
@kegsay The newly passing tests are Complement tests, not Sytests.

I'll pick up your review comments tomorrow!

federationapi/internal/perform.go
neilalexander: I'll add some checks here to ensure the membership event is well-formed and what we expect, but if it is, then yes it should be safe.

federationapi/routing/join.go
neilalexander: In this particular case yes, because we're already gated behind checks that ensure that the event is a membership event, the room ID and state key matches and that the event origin matches the actual request origin.

roomserver/internal/query/query.go
neilalexander: It doesn't really need to. If the room doesn't exist or the room version is unknown to us, we still can't work out if a restricted join is allowed or not, so the net effect is the same.

roomserver/internal/query/query.go
neilalexander: The list of join NIDs has the potential to be quite large in a big room and we may be satisfied with a very early result, so not sure what the best answer here is.

roomserver/internal/query/query.go
neilalexander: Oops, copy-pasta.

kegsay (approved):

PR #2480, opened 2022-05-23
Fulltext implementation incl. config

S7evinK:
This adds the main component of the fulltext search.
This PR doesn't do anything yet, besides creating an empty fulltextindex folder if enabled. Indexing events is done in a separate PR.


setup/base/base.go
```diff
@@ -146,6 +148,15 @@ func NewBaseDendrite(cfg *config.Dendrite, componentName string, options ...Base
 		logrus.WithError(err).Panicf("failed to start opentracing")
 	}
 
+	var fts *fulltext.Search
+	isSyncOrMonolith := componentName == "syncapi" || isMonolith
```
neilalexander, line 152: This just makes me think we should probably define some consts for the various component names somewhere, so that we don't accidentally change them in one place and not another later on. Happy for that to be in a separate PR.

neilalexander (approved):
Looks good!

PR #2484, opened 2022-05-24
Use `/admin/v1/register` in `create-account`

S7evinK:
The `create-account` utility is updated to use the existing `/_synapse/admin/v1/register` endpoint, similar to `register_new_matrix_user` on Synapse. This ensures all needed functions are called to propagate e.g. account data.
Unfortunately, this means "passwordless" accounts can't be created anymore.
Resetting a password still uses a direct database connection.


neilalexander:
Things like this make me wonder whether we should turn user creation into an admin endpoint instead of having `create-account` talk to the database directly. I don't really like the fact that we have to create workarounds like this when we should just update the accounts & account data using the built-in APIs and have the stream positions advanced properly.

genofire:
@neilalexander , i und erstand you position.

So we should use the same function (of creating accounts) for cmd and URL-Endpoint, without going over the Endpoint ... to prevent missing to create data in the database (for notification or something else).

McSinyx:
How do I fix accounts created with `create-account`?

PR #2511, opened 2022-06-02
Implement history visibility on `/messages`, `/context`, `/sync`

S7evinK:
This implements history visibility on `/messages`, `/context` and `/sync`.
A test for the SyncAPI is added, similar to [Sytest](https://github.com/matrix-org/sytest/blob/441ec0157f476617a72e65119764d62dfd6d9c30/tests/30rooms/30history-visibility.pl#L37), to verify we only get the expected events.

The test infra gains the ability to set the user account type and the room history visibility.

There's some noise in `syncapi/storage/shared/syncserver.go`, as the IDE was complaining about different receiver names and other things.


syncapi/storage/sqlite3/deltas/2022061412000000_history_visibility_column.go
```diff
@@ -51,6 +54,41 @@ func UpAddHistoryVisibilityColumnCurrentRoomState(ctx context.Context, tx *sql.T
 	if err != nil {
 		return fmt.Errorf("failed to execute upgrade: %w", err)
 	}
+
+	// get the current room history visibility
+	rows, err := tx.Query(`SELECT DISTINCT room_id, headered_event_json FROM syncapi_current_room_state
```
kegsay, line 59: Why are you adding this to an _existing_ migration? Won't it not be applied?

syncapi/streams/stream_pdu.go
```diff
@@ -296,13 +299,18 @@ func (p *PDUStreamProvider) addRoomDeltaToResponse(
...
+		}
 		jr := types.NewJoinResponse()
 		if hasMembershipChange {
 			p.addRoomSummary(ctx, jr, delta.RoomID, device.UserID, latestPosition)
 		}
 		jr.Timeline.PrevBatch = &prevBatch
-		jr.Timeline.Events = gomatrixserverlib.HeaderedToClientEvents(recentEvents, gomatrixserverlib.FormatSync)
-		jr.Timeline.Limited = limited
+		jr.Timeline.Events = gomatrixserverlib.HeaderedToClientEvents(events, gomatrixserverlib.FormatSync)
+		jr.Timeline.Limited = limited && len(events) == len(recentEvents)
```
kegsay, line 313: Why isn't `limited` alone good enough? Needs comments.

syncapi/streams/stream_pdu.go
```diff
@@ -462,6 +475,17 @@ func (p *PDUStreamProvider) getJoinResponseForCompleteSync(
 	recentEvents := p.DB.StreamEventsToEvents(device, recentStreamEvents)
 	stateEvents = removeDuplicates(stateEvents, recentEvents)
 
+	events := recentEvents
+	// Only apply history visibility checks if the response is for joined rooms
+	if !isPeek {
+		events, err = applyHistoryVisibilityFilter(ctx, p.DB, p.rsAPI, roomID, device.UserID, recentEvents)
```
kegsay, line 481: Does this mean peek rooms will always see everything regardless of the history visibility? Seems unsafe!

syncapi/streams/stream_pdu.go
```diff
@@ -331,6 +339,37 @@ func (p *PDUStreamProvider) addRoomDeltaToResponse(
...
+// sure we always return the required events in the timeline.
+func applyHistoryVisibilityFilter(
+	ctx context.Context,
+	db storage.Database,
+	rsAPI roomserverAPI.SyncRoomserverAPI,
+	roomID, userID string,
+	recentEvents []*gomatrixserverlib.HeaderedEvent,
+) ([]*gomatrixserverlib.HeaderedEvent, error) {
+	// We need to make sure we always include the latest states events, if they are in the timeline
+	stateEvents, err := db.CurrentState(ctx, roomID, &gomatrixserverlib.StateFilter{Limit: 1000}, nil)
```
kegsay, line 352: 1000 value is arbitrary and not justified in comments.

syncapi/streams/stream_pdu.go
```diff
@@ -331,6 +339,37 @@ func (p *PDUStreamProvider) addRoomDeltaToResponse(
...
+	if err != nil {
+		logrus.WithError(err).Warnf("failed to get current state")
+	}
```
kegsay, line 355: Fallthrough - missing return? No justification for fallthrough.

syncapi/syncapi_test.go
```diff
@@ -311,7 +323,180 @@ func testSyncAPIUpdatePresenceImmediately(t *testing.T, dbType test.DBType) {
...
+	}
+}
+
+func verifyEventVisible(t *testing.T, wantVisible bool, wantVisibleEvent *gomatrixserverlib.HeaderedEvent, chunk []gomatrixserverlib.ClientEvent) {
+	t.Helper()
+	if wantVisible {
+		found := false
+		for _, ev := range chunk {
+			if ev.EventID == wantVisibleEvent.EventID() {
+				found = true
```
kegsay, line 483: Would be a bit cleaner to just `return` here no? Then instead of `if !found` just have `t.Fatalf` since that line will only be reached if there is no match.

syncapi/syncapi_test.go
```diff
@@ -311,7 +323,180 @@ func testSyncAPIUpdatePresenceImmediately(t *testing.T, dbType test.DBType) {
...
+
+		bobDev.AccountType = accType
+		userType := "guest"
+		if accType == userapi.AccountTypeUser {
+			userType = "real user"
+		}
+
+		base, close := testrig.CreateBaseDendrite(t, dbType)
+		defer close()
+		_ = close
```
kegsay, line 400: ?

syncapi/syncapi_test.go
```diff
@@ -311,7 +323,180 @@ func testSyncAPIUpdatePresenceImmediately(t *testing.T, dbType test.DBType) {
 
 }
 
-func toNATSMsgs(t *testing.T, base *base.BaseDendrite, input []*gomatrixserverlib.HeaderedEvent) []*nats.Msg {
+// This is mainly what Sytest is doing in "test_history_visibility"
+func TestMessageHistoryVisibility(t *testing.T) {
```
kegsay, line 327: Excellent test overall, reads really well.

syncapi/storage/sqlite3/deltas/2022061412000000_history_visibility_column.go
```diff
@@ -51,6 +54,41 @@ func UpAddHistoryVisibilityColumnCurrentRoomState(ctx context.Context, tx *sql.T
...
+`)
+	if err != nil {
+		return fmt.Errorf("failed to query current room state: %w", err)
+	}
+	var eventBytes []byte
+	var roomID string
+	var event gomatrixserverlib.HeaderedEvent
+	var hisVis gomatrixserverlib.HistoryVisibility
+	historyVisibilities := make(map[string]gomatrixserverlib.HistoryVisibility)
+	for rows.Next() {
```
kegsay, line 70: Missing `defer rows.Close()`

syncapi/storage/sqlite3/deltas/2022061412000000_history_visibility_column.go
```diff
@@ -51,6 +54,41 @@ func UpAddHistoryVisibilityColumnCurrentRoomState(ctx context.Context, tx *sql.T
...
+			return fmt.Errorf("failed to unmarshal event: %w", err)
+		}
+		historyVisibilities[roomID] = gomatrixserverlib.HistoryVisibilityShared
+		if hisVis, err = event.HistoryVisibility(); err == nil {
+			historyVisibilities[roomID] = hisVis
+		}
+	}
+
+	// update the history visibility
+	for roomID, hisVis = range historyVisibilities {
```
kegsay, line 84: So this migration is a bit wrong because aiui it will load the _current_ history visibility value then apply this for _all_ events? Probaby good enough, but definitely needs comments.

syncapi/storage/postgres/deltas/2022061412000000_history_visibility_column.go
```diff
@@ -39,6 +42,41 @@ func UpAddHistoryVisibilityColumnCurrentRoomState(ctx context.Context, tx *sql.T
...
+`)
+	if err != nil {
+		return fmt.Errorf("failed to query current room state: %w", err)
+	}
+	var eventBytes []byte
+	var roomID string
+	var event gomatrixserverlib.HeaderedEvent
+	var hisVis gomatrixserverlib.HistoryVisibility
+	historyVisibilities := make(map[string]gomatrixserverlib.HistoryVisibility)
+	for rows.Next() {
```
kegsay, line 58: Same comments as sqlite migration.

kegsay (approved):
By and large this is fantastic, thank you! Most of my comments are minor.

syncapi/internal/history_visibility.go
```diff
@@ -0,0 +1,204 @@
...
+		EventIDs: eventIDs,
+		UserID:   userID,
+	}, resp)
+	if err != nil {
+		return res, err
+	}
+
+	// Create a map from eventID -> eventVisibility
+	for _, eventID := range eventIDs {
+		vis := eventVisibility{membershipAtEvent: gomatrixserverlib.Leave}
```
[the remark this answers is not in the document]
S7evinK: Updated.
Also, it should never be 0 (besides if we manually update the visibility of an event), since the `sql.Scanner`/`sql.Valuer` from https://github.com/matrix-org/gomatrixserverlib/pull/326/files default to `shared`.

syncapi/internal/history_visibility.go
```diff
@@ -0,0 +1,204 @@
...
+			continue
+		}
+		// Handle history visibility changes
+		if hisVis, err := ev.HistoryVisibility(); err == nil {
+			prevHisVis := gjson.GetBytes(ev.Unsigned(), "prev_content.history_visibility").String()
+			if oldPrio, ok := historyVisibilityPriority[gomatrixserverlib.HistoryVisibility(prevHisVis)]; ok {
+				// no OK check, since this should have been validated when setting the value
+				newPrio := historyVisibilityPriority[hisVis]
+				if oldPrio < newPrio {
+					d.visibility = gomatrixserverlib.HistoryVisibility(prevHisVis)
```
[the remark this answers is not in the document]
S7evinK: Added a comment and a link to the [spec](https://spec.matrix.org/v1.3/client-server-api/#server-behaviour-5)

syncapi/storage/sqlite3/deltas/2022061412000000_history_visibility_column.go
S7evinK: Fixed.

syncapi/streams/stream_pdu.go
S7evinK: This should be something for a different PR, otherwise this PR will get even larger.

roomserver/api/api.go
```diff
@@ -97,6 +97,12 @@ type SyncRoomserverAPI interface {
 		req *PerformBackfillRequest,
 		res *PerformBackfillResponse,
 	) error
+
+	QueryMembershipAtEvent(
```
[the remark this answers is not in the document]
S7evinK: `QueryStateAfterEvents` returns a slice of headered events, whereas `QueryMembershipAtEvent` returns a map from `eventID` to headered events for easier lookup of the membership event.

PR #2519, opened 2022-06-07
Rate limiting changes

### LINKED ISSUE #2518: Option to disable rate limiting for a specific user
Synapse has this https://github.com/matrix-org/synapse/pull/9648 and it would be useful for mjolnir.

Related to #1280

neilalexander:
This makes the following changes:

* For logged in users, the rate limiting now applies to the device session rather than the remote IP address;
* For non-logged in users, the rate limiting continues to apply to remote address as it does today;
* It is now possible to add user IDs to the `exempt_user_ids` option under `rate_limiting` to exclude specific accounts (i.e. Mjolnirs) from rate limiting;
* Admin and appservice users are now exempt from rate limiting by default.

Closes #2518.


bones-was-here:
Might be good to put " " around @user:domain.com in the rate limiting example config, dendrite doesn't start without them.

neilalexander:
@bones-was-here Done, thanks for the heads-up.

PR #2523, opened 2022-06-09
Add a `servername` column to `account_profiles`

S7evinK:
This adds a `servername` column to the `account_profiles` table, so it is possible to store federated profiles.
The `MakeAuthAPI` function now also adds the server name to the device struct, so it can be used on different endpoints.

Storing the profile of federated servers avoids possible round trips over federation, when e.g. querying the user directory.


userapi/storage/postgres/deltas/2022060815233800_userprofile_servername.go
```diff
@@ -0,0 +1,36 @@
...
+
+var serverName gomatrixserverlib.ServerName
+
+func LoadProfilePrimaryKey(m *sqlutil.Migrations, s gomatrixserverlib.ServerName) {
+	serverName = s
+	m.AddMigration(UpProfilePrimaryKey, DownProfilePrimaryKey)
+}
+
+func UpProfilePrimaryKey(tx *sql.Tx) error {
+	_, err := tx.Exec(fmt.Sprintf(`ALTER TABLE account_profiles ADD COLUMN IF NOT EXISTS servername TEXT NOT NULL DEFAULT '%s';
```
neilalexander, line 19: This is going to sound nitpicky but, in the interests of consistency across the project, can we make this `server_name` instead?

neilalexander (commented):
As far as storing remote profiles goes, what mechanism would we have for invalidating our local copy of the profile? I don't think we get notified about remote profile updates (but rather just the updated membership events in rooms which can be different per-room).

S7evinK:
> As far as storing remote profiles goes, what mechanism would we have for invalidating our local copy of the profile? I don't think we get notified about remote profile updates (but rather just the updated membership events in rooms which can be different per-room).

Membership events might be a good pointer to when we should invalidate our local copy. If there's a change, we remove the entry from the database and fetch a fresh profile when `/profile` or `/user_directory/search` is hit.

neilalexander (approved):

federationapi/routing/send.go
```diff
@@ -275,6 +280,13 @@ func (t *txnReq) processTransaction(ctx context.Context) (*gomatrixserverlib.Res
 			continue
 		}
 
+		// Clear our local user profile cache, if this is a membership event
```
neilalexander, line 283: It would actually be better for the user API to have a consumer for the output room events topic and follow that, rather than hooking into the federation API. The roomserver is the single source of truth for the member state.

neilalexander (changes requested):

neilalexander (approved):

S7evinK:
Closing this, as it's more likely we're going to use `user_id` instead of an extra column. (Which is also more consistent with other tables)

PR #2530, opened 2022-06-11
mv sample config to docs/dendrite

ealib:
Move

* `/dendrite-sample.monolith.yaml` to `/docs/dendrite/dendrite-sample.monolith.yaml`
* `/dendrite-sample.polylith.yaml` to `/docs/dendrite/dendrite-sample.polylith.yaml`

This PR does not require testing because it simply moves two sample files to another directory (in which all other example files reside).

Signed-off-by: `Emanuele Aliberti <dev@mtka.eu>`


neilalexander:
Not clear if we want to do this. The sample configs being at the top-level is good for visibility for people who check out the repo.

neilalexander:
Thanks, but we will close for now and may revisit where we put documentation and configs in the future.

PR #2532, opened 2022-06-13
Handle state before, send history visibility in output

neilalexander:
This PR auths events against the state after the prev events and enables us to work out what the history visibility was at the time of the event, which is also now sent in the roomserver output events.

### DIFF
--- a/roomserver/internal/input/input_events.go
+++ b/roomserver/internal/input/input_events.go
@@ -296,58 +296,18 @@ func (r *Inputer) processRoomEvent(
 	}
 
 	// Get the state before the event so that we can work out if the event was
-	// allowed at the time. Don't bother doing this if the event is already
-	// rejected since it's just wasting CPU time.
+	// allowed at the time, and also to get the history visibility. We won't
+	// bother doing this if the event was already rejected as it just ends up
+	// burning CPU time.
+	historyVisibility := "joined" // Default to restrictive.
 	if rejectionErr == nil {
-		var stateBeforeEvent []*gomatrixserverlib.Event
-		if input.HasState {
-			// If we're overriding the state then we need to go and retrieve
-			// them from the database. It's a hard error if they are missing.
-			stateEvents, err := r.DB.EventsFromIDs(ctx, input.StateEventIDs)
-			if err != nil {
-				return fmt.Errorf("r.DB.EventsFromIDs: %w", err)
-			}
-			stateBeforeEvent = make([]*gomatrixserverlib.Event, 0, len(stateEvents))
-			for _, entry := range stateEvents {
-				stateBeforeEvent = append(stateBeforeEvent, entry.Event)
-			}
-		} else if event.Type() != gomatrixserverlib.MRoomCreate && len(event.PrevEventIDs()) > 0 {
-			// For all non-create events, there must be prev events, so we'll
-			// ask the query API for the relevant tuples needed for auth.
-			tuplesNeeded := gomatrixserverlib.StateNeededForAuth([]*gomatrixserverlib.Event{event}).Tuples()
-			stateBeforeReq := &api.QueryStateAfterEventsRequest{
-				RoomID:       event.RoomID(),
-				PrevEventIDs: event.PrevEventIDs(),
-				StateToFetch: tuplesNeeded,
-			}
-			stateBeforeRes := &api.QueryStateAfterEventsResponse{}
-			if err := r.Queryer.QueryStateAfterEvents(ctx, stateBeforeReq, stateBeforeRes); err != nil {
-				return fmt.Errorf("r.Queryer.QueryStateAfterEvents: %w", err)
-			}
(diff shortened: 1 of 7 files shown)


roomserver/internal/input/input_latest_events.go
```diff
@@ -69,15 +70,16 @@ func (r *Inputer) updateLatestEvents(
...
+		ctx:               ctx,
+		api:               r,
+		updater:           updater,
+		roomInfo:          roomInfo,
+		stateAtEvent:      stateAtEvent,
+		event:             event,
+		sendAsServer:      sendAsServer,
+		transactionID:     transactionID,
+		rewritesState:     rewritesState,
+		historyVisibility: historyVisibility,
```
kegsay, line 82: Unclear when this is set:
 - When the latest event is a history visibility event
 - For all events, what the current (before the event) history visibility is.

Needs comments or better yet, a better variable name.

roomserver/internal/input/input_events.go
```diff
@@ -402,6 +420,94 @@ func (r *Inputer) processRoomEvent(
 	return nil
 }
 
+// processStateBefore works out what the state is before the event and
+// then checks the event auths against the state at the time. It also
+// tries to determine what the history visibility was of the event at
+// the time, so that it can be sent in the output event to downstream
+// components.
+// nolint:nakedret
```
kegsay, line 428: I'm with the linter on this one. This function is huge, and having naked returns make it hard to figure out _what_ is returned.

roomserver/internal/input/input_events.go
```diff
@@ -402,6 +420,94 @@ func (r *Inputer) processRoomEvent(
...
+		stateBeforeEvent = make([]*gomatrixserverlib.Event, 0, len(stateEvents))
+		for _, entry := range stateEvents {
+			stateBeforeEvent = append(stateBeforeEvent, entry.Event)
+		}
+	} else if missingPrev {
+		// If we're missing prev events and still failed to fetch them
+		// before then we're stuck.
+		rejectionErr = fmt.Errorf("event %q should have prev events", event.EventID())
+		return
+	} else if event.Type() != gomatrixserverlib.MRoomCreate && len(event.PrevEventIDs()) > 0 {
```
kegsay, line 453: You need to check the state key as well as the event type.

roomserver/internal/input/input_events.go
```diff
@@ -402,6 +420,94 @@ func (r *Inputer) processRoomEvent(
...
+		default:
+			stateBeforeEvent = gomatrixserverlib.UnwrapEventHeaders(stateBeforeRes.StateEvents)
+		}
+	} else if event.Type() != gomatrixserverlib.MRoomCreate && len(event.PrevEventIDs()) == 0 {
+		// Non-create events that don't have any prev event IDs are
+		// impossible in theory, so reject them.
+		rejectionErr = fmt.Errorf("event %q should have prev events", event.EventID())
+		return
+	}
+	if rejectionErr == nil {
```
kegsay, line 489: Bad control flow here. This will never ever be false because every time this is set we immediately return.

The point of this part of the function is to load up the state events and reject otherwise. This code could and should be refactored to make there be fewer branching paths: it would reduce the quality of the errors messages but greatly improve readability of this function, e.g:
```go
if input.HasState {
    // load from the input event, return error otherwise
} else {
    // load from the Queryer, return error otherwise
}

// at this point, `stateBeforeEvent` is set.
```

Currently you have a bunch of if elseif blocks and it's mentally hard to parse if you've caught every single case.

roomserver/internal/input/input_events.go
```diff
@@ -402,6 +420,94 @@ func (r *Inputer) processRoomEvent(
...
+			return
+		}
+	}
+	// Work out what the history visibility was at the time of the
+	// event.
+	for _, event := range stateBeforeEvent {
+		if event.Type() != gomatrixserverlib.MRoomHistoryVisibility || !event.StateKeyEquals("") {
+			continue
+		}
+		if hisVis, err := event.HistoryVisibility(); err == nil {
```
kegsay, line 503: What if there is no history visibility event? It's not clear if you intend to return the empty string. Only at the call-site does it seem you set a default, perhaps that should be here instead so you can guarantee you always return the history visibility?

kegsay (changes requested):

roomserver/internal/input/input_events.go
neilalexander: The default is actually set at the call site (because we don't know if we'll hit `processStateBefore` or not) *and* at the start of `processStateBefore` so that it always returns *something*.

roomserver/internal/input/input_latest_events.go
neilalexander: This is *always* the history visibility before the event in question, not after it, even if it's a history visibility event. I've added a comment to that effect.

PR #2536, opened 2022-06-15
Add `InputDeviceListUpdate` to the keyserver, remove old input API

neilalexander:
This queues up device list updates from federation into the key server using a NATS stream. This ensures that:
1. We don't head-of-line block waiting for the key server to do things before responding to `/send`
2. Updates are guaranteed to be processed

### DIFF
--- a/keyserver/consumers/devicelistupdate.go
+++ b/keyserver/consumers/devicelistupdate.go
@@ -1,4 +1,4 @@
-// Copyright 2020 The Matrix.org Foundation C.I.C.
+// Copyright 2022 The Matrix.org Foundation C.I.C.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/keyserver/consumers/devicelistupdate.go
+++ b/keyserver/consumers/devicelistupdate.go
@@ -1,0 +1,77 @@
+// Copyright 2020 The Matrix.org Foundation C.I.C.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


keyserver/consumers/devicelistupdate.go
```diff
@@ -0,0 +1,77 @@
...
+func NewDeviceListUpdateConsumer(
+	process *process.ProcessContext,
+	cfg *config.KeyServer,
+	js nats.JetStreamContext,
+	updater *internal.DeviceListUpdater,
+) *DeviceListUpdateConsumer {
+	return &DeviceListUpdateConsumer{
+		ctx:       process.Context(),
+		jetstream: js,
+		durable:   cfg.Matrix.JetStream.Prefixed("KeyServerInputDeviceListConsumer"),
```
kegsay, line 49: Should this be made into a constant?

keyserver/consumers/devicelistupdate.go
```diff
@@ -0,0 +1,77 @@
...
+// key change events topic from the key server.
+func (t *DeviceListUpdateConsumer) onMessage(ctx context.Context, msg *nats.Msg) bool {
+	var m gomatrixserverlib.DeviceListUpdateEvent
+	if err := json.Unmarshal(msg.Data, &m); err != nil {
+		logrus.WithError(err).Errorf("Failed to read from device list update input topic")
+		return true
+	}
+	err := t.updater.Update(ctx, m)
+	if err != nil {
+		logrus.WithError(err).Errorf("Failed to update device list")
```
kegsay, line 73: Log info about `m` maybe?

kegsay (commented):
By making this async, we will likely risk upsetting tests which assume that when /send 200 OKs that the EDU has been processed. Is there another way those tests can assert when the update has been processed?

neilalexander:
> By making this async, we will likely risk upsetting tests which assume that when /send 200 OKs that the EDU has been processed.

This isn't a guarantee anywhere, nor is it necessarily true in Synapse because of worker mode. The only right thing to do is to wait for `/sync`, which the tests must do because they pass.

keyserver/consumers/devicelistupdate.go
neilalexander: At the moment the consumer names aren't constants anywhere in the codebase, so just following the pattern here.

kegsay (approved):

PR #2540, opened 2022-06-26
Fix nats.go commit

flacks:
Corrects nats.go commit hash to latest available from neilalexander/nats.go, since currently-set version doesn't exist.


neilalexander (approved):
Thanks for this. I imagine this happened when rebasing the PR for upstream.

Issue #2387, opened 2022-04-27
History not fetched from other servers when (re)joining room


bones-was-here:
### Background information

- **Dendrite version or git SHA**: 54ff4cf690918886c7e7a59a65ccff970c3aa1fc
- **Monolith or Polylith?**: monolith
- **SQLite3 or Postgres?**: postgres
- **Running in Docker?**: no
- **`go version`**: 1.17.9
- **Client used (if applicable)**: element web

### Description

 - **What** is the problem: if a homeserver has had no users in a room for some time and re-joins it, the missing events are not fetched from remote servers
 - **Who** is affected: potentially any user of a homserver that has been out of a room it previously was in
 - **How** is this bug manifesting: when the user re-joins the room, no warns or errors are logged, and the fetched events are all retrieved locally (so in Element the user can see only very old messages, from when the server previously participated in the room)
 - **When** did this first appear: may have always been like this

After re-joining the room, if a user on a different homeserver posts a message, the re-joined user does not see it.  

If the re-joined user posts a message in a problematic room, this can trigger sudden fetching of events - not just in the room where the message was posted, but also in other rooms that are in the same state of missing recent events.  Not all history is loaded however, and it logs:
`level=warning msg="Stored rejected event" context=missing error="missingState.processEventWithMissingState: t.lookupState (backwards extremity): t.lookupStateAfterEvent: t.lookupStateBeforeEvent: contents=`
`level=warning msg="Roomserver failed to process async event" error="missingState.processEventWithMissingState: t.lookupState (backwards extremity): t.lookupStateAfterEvent: t.lookupStateBeforeEvent: contents=`
`level=warning msg="failed to get state after ... context=missing error="storage: event IDs missing from the database (0 != 1)"`

### Steps to reproduce

- create a room, invite user(s) from a different homeserver
- leave the room with all user(s) on the homeserver that created the room
- rejoin with the user that originally created the room, after "some time" and many messages

<!--
Describe how what happens differs from what you expected.

If you can identify any relevant log snippets from server logs, please include
those (please be careful to remove any personal or private data). Please surround them with
``` (three backticks, on a line on their own), so that they are formatted legibly.

Alternatively, please send logs to @kegan:matrix.org or @neilalexander:matrix.org
with a link to the respective Github issue, thanks!
-->

bones-was-here:
This seems related to the broken state I see for #dendrite:matrix.org

[...]

PR #2548, opened 2022-07-01
Try to fix backfilling

S7evinK:
With this PR we're ignoring the ([to be](https://github.com/matrix-org/gomatrixserverlib/pull/319)) added `SignatureErr` in GMSL, which should help in getting events where e.g. the signing key was changed, but the key ID stayed the same.

We're now also trying to get the events from all servers in the defined list and not only from one.

It also adds `start`/`end` to the `/context` endpoint, which could cause (e.g.) Element Web to use an empty `from` on `/messages` after hitting `/context`.

Should hopefully fix https://github.com/matrix-org/dendrite/issues/2387


neilalexander (approved):
Please update the `go.mod` to point to the new squashed commit for #318 before merge:
```
go get github.com/matrix-org/gomatrixserverlib@da53994
go mod tidy

PR #1988, opened 2021-08-25
Added .well-known/matrix/server endpoint

twentybit:
### Pull Request Checklist

* [ ] I have added any new tests that need to pass to `sytest-whitelist` as specified in [docs/sytest.md](https://github.com/matrix-org/dendrite/blob/master/docs/sytest.md)
* [x] Pull request includes a [sign off](https://github.com/matrix-org/dendrite/blob/master/docs/CONTRIBUTING.md#sign-off)

Signed-off-by: `Ryan Whittington <twentybitdev@gmail.com>`

PR #2551, opened 2022-07-03
Add .well-known/matrix/client to clientapi

Jonnobrow:
### Overview
Added the well-known route for clients, since the server route already exists.

Built on work done in #1988 

### Pull Request Checklist

Signed-off-by: `Jonathan Bartlett <jonathan@jonnobrow.co.uk>`


genofire:
Does we want to set `m.identity_server` also? (And maybe custom.properties) like in specification: https://spec.matrix.org/v1.2/client-server-api/#getwell-knownmatrixclient

Jonnobrow:
> Does we want to set `m.identity_server` also? (And maybe custom.properties) like in specification: https://spec.matrix.org/v1.2/client-server-api/#getwell-knownmatrixclient

Adding identity server should be fairly straightforward. 

Custom properties could be slightly more complex, but I will take a look.

neilalexander (approved):

neilalexander:
Thanks for this!

PR #2563, opened 2022-07-07
Ristretto cache

neilalexander:
This places the built-in LRU caches with a new [Ristretto](https://github.com/dgraph-io/ristretto) cache. 

The discipline of this cache is rather different to before. Whereas the LRU cache would have admitted any new entry and evicted the least used one, Ristretto is much more conservative — it won't admit a cache entry on the first try and will only admit new entries if they are actually an improvement on the ones that are already in the cache. On the whole, this means that the cache quality should overall be much better and maintain items that are more useful to us over time.

In addition to this, because the entries are now costed (based on an in-memory estimate of their size), we can now *reasonably* cap the size of the cache. It's not a hard limit but it should be good enough to prevent unbounded cache growth on systems with limited memory.

Age is bounded too (although configurable) because this gives the cache a better opportunity to refresh itself over time. A much bigger cache will benefit from a longer age but a smaller cache may benefit from refreshing more often.

Finally, the other notable change is the use of generics here, so this effectively lifts our baseline supported Go version from 1.16 to 1.18. The upside is that managing the different cache partitions is now considerably simpler as they don't require runtime type assertions anymore.

A future piece of work could be to remove some of the outward-facing interfaces but there's very little overhead there to worry about for now.

### DIFF
--- a/dendrite-sample.monolith.yaml
+++ b/dendrite-sample.monolith.yaml
@@ -44,10 +44,18 @@ global:
   # Configuration for in-process caches, to help speed up processing transactions
   # and reducing load on the database.
   cache:
-    # The estimated maximum size for the global cache in bytes. Note that this is
-    # not a hard limit, nor is it a memory limit for the entire process. A cache that
-    # is too small may end up yielding little benefit.
-    max_bytes_est: 1073741824
+    # The estimated maximum size for the global cache in bytes, or in terabytes,
+    # gigabytes, megabytes or kilobytes when the appropriate 'tb', 'gb', 'mb' or
+    # 'kb' suffix is specified. Note that this is not a hard limit, nor is it a
+    # memory limit for the entire process. A cache that is too small may ultimately
+    # provide little or no benefit.
+    max_bytes_est: 1gb
+
+    # The maximum amount of time that a cache entry can live for in memory before
+    # it will be evicted and/or refreshed from the database. Lower values result in
+    # easier admission of new cache entries but may also increase database load in
+    # comparison to higher values, so adjust conservatively.
+    max_age: 1h
 
   # The server name to delegate server-server communications to, with optional port
   # e.g. localhost:443
--- a/dendrite-sample.monolith.yaml
+++ b/dendrite-sample.monolith.yaml
@@ -41,15 +41,16 @@ global:
     max_idle_conns: 5
     conn_max_lifetime: -1
 
-  # Configuration for in-process caches, to help speed up processing transactions
-  # and reducing load on the database.
+  # Configuration for in-memory caches. Caches can often improve performance by
+  # keeping frequently accessed items (like events, identifiers etc.) in memory
+  # rather than having to read them from the database.
   cache:
     # The estimated maximum size for the global cache in bytes, or in terabytes,
     # gigabytes, megabytes or kilobytes when the appropriate 'tb', 'gb', 'mb' or
     # 'kb' suffix is specified. Note that this is not a hard limit, nor is it a
     # memory limit for the entire process. A cache that is too small may ultimately
     # provide little or no benefit.
-    max_bytes_est: 1gb
+    max_size_est: 1gb
 
     # The maximum amount of time that a cache entry can live for in memory before
     # it will be evicted and/or refreshed from the database. Lower values result in
--- a/dendrite-sample.monolith.yaml
+++ b/dendrite-sample.monolith.yaml
@@ -55,7 +55,9 @@ global:
     # The maximum amount of time that a cache entry can live for in memory before
     # it will be evicted and/or refreshed from the database. Lower values result in
     # easier admission of new cache entries but may also increase database load in
-    # comparison to higher values, so adjust conservatively.
+    # comparison to higher values, so adjust conservatively. Higher values may make
+    # it harder for new items to make it into the cache, e.g. if new rooms suddenly
+    # become popular.
     max_age: 1h
 
   # The server name to delegate server-server communications to, with optional port
--- a/dendrite-sample.monolith.yaml
+++ b/dendrite-sample.monolith.yaml
@@ -41,6 +41,14 @@ global:
     max_idle_conns: 5
     conn_max_lifetime: -1
 
+  # Configuration for in-process caches, to help speed up processing transactions
+  # and reducing load on the database.
+  cache:
+    # The estimated maximum size for the global cache in bytes. Note that this is
+    # not a hard limit, nor is it a memory limit for the entire process. A cache that
+    # is too small may end up yielding little benefit.
+    max_bytes_est: 1073741824
+
   # The server name to delegate server-server communications to, with optional port
   # e.g. localhost:443
   well_known_server_name: ""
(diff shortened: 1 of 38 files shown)


internal/caching/impl_ristretto.go
```diff
@@ -0,0 +1,185 @@
+package caching
```
S7evinK, line 1: Copyright missing. :) (as it was before with LRU)

setup/config/config_test.go
```diff
@@ -268,3 +270,22 @@ n0Xq64k7fc42HXJpF8CGBkSaIhtlzcruO+vqR80B9r62+D0V7VmHOnP135MT6noU
...
 -----END CERTIFICATE-----
 `
+
+func TestUnmarshalDataUnit(t *testing.T) {
+	target := struct {
+		Got DataUnit `yaml:"value"`
+	}{}
+	for input, expect := range map[string]DataUnit{
+		"value: 0.6tb": 659706976665,
+		"value: 1.2gb": 1288490188,
```
S7evinK, line 280: Maybe also test something like `1,2gb`, `1.2 gb` or something, as this currently fails.

S7evinK (changes requested):
Mostly minor things, otherwise looks OK to me.

dendrite-sample.monolith.yaml
```diff
@@ -41,6 +41,23 @@ global:
...
+    # 'kb' suffix is specified. Note that this is not a hard limit, nor is it a
+    # memory limit for the entire process. A cache that is too small may ultimately
+    # provide little or no benefit.
+    max_size_est: 1gb
+
+    # The maximum amount of time that a cache entry can live for in memory before
+    # it will be evicted and/or refreshed from the database. Lower values result in
+    # easier admission of new cache entries but may also increase database load in
+    # comparison to higher values, so adjust conservatively.
+    max_age: 1h
```
kegsay, line 59: You don't fully explain why high values are bad. You imply that it is harder to admit them into the cache with a higher age, but surely that would also result in increased database load then..?

roomserver/storage/shared/storage.go
```diff
@@ -439,8 +439,18 @@ func (d *Database) Events(
 }
 
 func (d *Database) events(
-	ctx context.Context, txn *sql.Tx, eventNIDs []types.EventNID,
+	ctx context.Context, txn *sql.Tx, inputEventNIDs types.EventNIDs,
 ) ([]types.Event, error) {
+	sort.Sort(inputEventNIDs)
```
kegsay, line 444: Why sort?

internal/caching/cache_roomevents.go
```diff
@@ -0,0 +1,21 @@
+package caching
+
+import (
+	"github.com/matrix-org/dendrite/roomserver/types"
+	"github.com/matrix-org/gomatrixserverlib"
+)
+
+// RoomServerNIDsCache contains the subset of functions needed for
+// a roomserver NID cache.
+type RoomServerEventsCache interface {
```
kegsay, line 10: Docs are wrong.

internal/caching/cache_roomevents.go
```diff
@@ -0,0 +1,21 @@
...
+	GetRoomServerEvent(eventNID types.EventNID) (*gomatrixserverlib.Event, bool)
+	StoreRoomServerEvent(eventNID types.EventNID, event *gomatrixserverlib.Event)
+}
+
+func (c Caches) GetRoomServerEvent(eventNID types.EventNID) (*gomatrixserverlib.Event, bool) {
+	return c.RoomServerEvents.Get(int64(eventNID))
+}
+
+func (c Caches) StoreRoomServerEvent(eventNID types.EventNID, event *gomatrixserverlib.Event) {
+	c.RoomServerEvents.Set(int64(eventNID), event)
```
kegsay, line 20: Events are mutable (they can be redacted, etc). How does this cache work with that?

internal/caching/impl_ristretto.go
```diff
@@ -0,0 +1,199 @@
...
+	"github.com/matrix-org/gomatrixserverlib"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+func mustCreateCache(maxCost config.DataUnit, enablePrometheus bool) *ristretto.Cache {
+	cache, err := ristretto.NewCache(&ristretto.Config{
+		NumCounters: 1e5,
+		MaxCost:     int64(maxCost),
+		BufferItems: 64,
```
kegsay, line 36: Need comments / justification on why 64 and why 1e5.

internal/caching/impl_ristretto.go
```diff
@@ -0,0 +1,199 @@
...
+	federationPDUsCache
+	federationEDUsCache
+	spaceSummaryRoomsCache
+	lazyLoadingCache
+)
+
+func NewRistrettoCache(maxCost config.DataUnit, maxAge time.Duration, enablePrometheus bool) (*Caches, error) {
+	cache := mustCreateCache(maxCost, enablePrometheus)
+	return &Caches{
+		RoomVersions: &RistrettoCachePartition[string, gomatrixserverlib.RoomVersion]{
```
kegsay, line 79: It would be nice to just give a hint to the reader what `string` is (room ID in this case), likewise throughout this struct. e.g

```go
RoomVersions: &RistrettoCachePartition[string, gomatrixserverlib.RoomVersion]{ // room ID -> version
```

internal/caching/caches.go
```diff
@@ -1,28 +1,52 @@
...
-	RoomServerRoomNIDs Cache // RoomServerNIDsCache
-	RoomServerRoomIDs  Cache // RoomServerNIDsCache
-	RoomInfos          Cache // RoomInfoCache
-	FederationEvents   Cache // FederationEventsCache
-	SpaceSummaryRooms  Cache // SpaceSummaryRoomsCache
-	LazyLoading        Cache // LazyLoadCache
+	RoomVersions       Cache[string, gomatrixserverlib.RoomVersion]
+	ServerKeys         Cache[string, gomatrixserverlib.PublicKeyLookupResult]
+	RoomServerRoomNIDs Cache[string, types.RoomNID]
+	RoomServerRoomIDs  Cache[int64, string]
```
kegsay, line 29: Having one-liner comments to explain what these are would be nice.

kegsay (approved):
Overall LGTM!

roomserver/storage/shared/storage.go
neilalexander: You know, that's a good question. The sort was there before (albeit in a different place in the function) and I now don't remember why.

internal/caching/cache_roomevents.go
neilalexander: The `applyRedactions` function in the storage package likely side-effects onto the cache here, so there's no change from the old LRU cache. We probably want to investigate that further later.

internal/caching/impl_ristretto.go
neilalexander: I've added some comments. The 64 is a sane default recommended by the ristretto godoc. The `1e5` is possibly not the optimum value — making it larger increases the size of the bloom filter and making it smaller makes the chance of collisions higher, but this value just *seemed* to be OK after some experimenting. We may want to change it in future.

PR #2567, opened 2022-07-12
Configuration tweaks

neilalexander:
This makes the following changes:

* The various `Defaults` functions are now responsible for setting sane defaults if `generate` is specified, rather than hiding them in `generate-config`
* Some configuration options have been marked as `omitempty` so that they don't appear in generated configs unnecessarily (monolith-specific vs. polylith-specific options)
* A new option `-polylith` has been added to `generate-config` to create a config that makes sense for polylith deployments (i.e. including the internal/external API listeners and per-component database sections)
* A new option `-normalise` has been added to `generate-config` to take an existing file and add any missing options and/or defaults


setup/config/config.go
```diff
@@ -293,19 +293,31 @@ func (config *Dendrite) Derive() error {
 }
 
 // SetDefaults sets default config values if they are not explicitly set.
-func (c *Dendrite) Defaults(generate bool) {
+func (c *Dendrite) Defaults(generate bool, isMonolith bool) {
```
kegsay, line 296: I'd prefer to see an opts struct like:
```go
type DefaultOpts {
    IsMonolith bool
    Generate bool
}
```
which is then used here. It makes it clearer at call-sites to know what `Defaults(true, true)` mean.

cmd/generate-config/main.go
```diff
@@ -11,74 +11,42 @@ import (
 )
 
 func main() {
-	defaultsForCI := flag.Bool("ci", false, "sane defaults for CI testing")
+	defaultsForCI := flag.Bool("ci", false, "Populate the configuration with sane defaults for use in CI")
 	serverName := flag.String("server", "", "The domain name of the server if not 'localhost'")
 	dbURI := flag.String("db", "", "The DB URI to use for all components if not SQLite files")
+	normalise := flag.String("normalise", "", "Normalise a given configuration file")
```
kegsay, line 17: Unclear what this is supposed to do from the docstring alone.

kegsay (approved):
Otherwise LGTM

cmd/generate-config/main.go
spaetz: May your humble guinea pig suggest that what you want to do is to: "update" an existing config file to the latest and greatest, adding missing defaults, rather than "normalizing" things?
But then, any term is good.

neilalexander: I specifically didn't want to use words like "update" because it isn't really what it does — it won't migrate an old config version to a new one for example, and it won't correct the formatting of old options vs new ones at present.

PR #2572, opened 2022-07-15
Use sync API database in `filterSharedUsers`

neilalexander:
This reduces the amount of calls we make to the roomserver by using the sync API current room state directly in `filterSharedUsers` when building sync responses containing device list updates, as the roomserver had quite a heavy way of calculating this, while we can do this on the `syncapi_current_room_state` table using index scans only.

This should help to speed up building sync responses, although attention is also needed in the rest of `TrackChangedUsers` to get rid of all the RS calls altogether.

### DIFF
--- a/syncapi/internal/keychange.go
+++ b/syncapi/internal/keychange.go
@@ -21,6 +21,7 @@ import (
 	keyapi "github.com/matrix-org/dendrite/keyserver/api"
 	keytypes "github.com/matrix-org/dendrite/keyserver/types"
 	roomserverAPI "github.com/matrix-org/dendrite/roomserver/api"
+	"github.com/matrix-org/dendrite/syncapi/storage"
 	"github.com/matrix-org/dendrite/syncapi/types"
 	"github.com/matrix-org/gomatrixserverlib"
 	"github.com/matrix-org/util"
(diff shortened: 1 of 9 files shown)


syncapi/internal/keychange.go
```diff
@@ -215,15 +216,15 @@ func TrackChangedUsers(
...
-	var result []string
-	var sharedUsersRes roomserverAPI.QuerySharedUsersResponse
-	err := rsAPI.QuerySharedUsers(ctx, &roomserverAPI.QuerySharedUsersRequest{
-		UserID:       userID,
-		OtherUserIDs: usersWithChangedKeys,
-	}, &sharedUsersRes)
+	sharedUsersRes := &roomserverAPI.QuerySharedUsersResponse{
+		UserIDsToCount: map[string]int{},
+	}
+	sharedUsers, err := db.SharedUsers(ctx, userID, usersWithChangedKeys)
```
kegsay, line 227: These users aren't in the map above anymore. Is this okay?

kegsay (approved):

syncapi/internal/keychange.go
neilalexander: I think you've highlighted a bug in the map handling there which I've now fixed.

PR #2577, opened 2022-07-19
Explain how SRV works in Matrix and discourage using it

tulir:
Signed-off-by: `Tulir Asokan <tulir@maunium.net>`


neilalexander (approved):
Looks good, thanks!

Issue #2504, opened 2022-05-30
Spaces Summary not working properly over federation


boogewooge:
### Background information

- **Dendrite version**: 0.8.4
- **Monolith or Polylith?**: Monolith
- **SQLite3 or Postgres?**: Postgres
- **Running in Docker?**: No
- **`go version`**: go1.17.10 linux/amd64
- **Client used (if applicable)**: Element Web/Desktop

### Description

 - **What** is the problem: Over federation, users can only see rooms they're already in in the space summary, instead of the full list. This is with the non-listed rooms being public.
 - **Who** is affected: Users on external homeservers
 - **How** is this bug manifesting:
From the logs on my Synapse `1.59.1` instance trying to fetch the room summary on my dendrite instance:
```
Request failed: GET matrix://my.dendrite.instance/_matrix/federation/v1/hierarchy/%21[REDACTED]%3Amy.dendrite.instance?suggested_only=true: HttpResponseException('404: Not Found')
```
 - **When** did this first appear: The first time a user on a synapse instance tried to see a space summary based on my dendrite instance in element.

### Steps to reproduce

1. Create a space on a dendrite instance and add public rooms to it.
2. Invite a user on a synapse instance to the space and optionally one of the rooms in the space
3. The synapse-based user will only be able to see any rooms they're already in in the space summary, instead of all (public) rooms in it.

<!--
Describe how what happens differs from what you expected.

If you can identify any relevant log snippets from server logs, please include
those (please be careful to remove any personal or private data). Please surround them with
``` (three backticks, on a line on their own), so that they are formatted legibly.

Alternatively, please send logs to @kegan:matrix.org or @neilalexander:matrix.org
with a link to the respective Github issue, thanks!
-->

bones-was-here:
I also consistently see this issue on 2 instances of 0.8.6, I think it has existed for as long as spaces summary has been working.

When a dendrite has no users in a room it tries to get summary information from the other dendrite (the one that owns the namespace and is still in the room) which fails with a similar error to the synapse one above:  
[...]

PR #2578, opened 2022-07-19
/hierarchy - return public and knockable rooms for authed users

texuf:
When requesting the room hierarchy with an authenticated user, return public and knockable rooms.

According to the spec, https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/2946-spaces-summary.md

```
Any child room that the user is joined or is potentially joinable is included in the response.
```

This is currently not the case. See discussion here: https://matrix.to/#/!NasysSDfxKxZBzJJoE:matrix.org/$t2Csj-6y1PVsn8GOnFZfXzeQW13NfqvrFCxB-XI_uhA?via=matrix.org&via=libera.chat&via=element.io and here: https://matrix.to/#/!NasysSDfxKxZBzJJoE:matrix.org/$EHp1x1DY7tnYZtx_PVEb-sKB9lmJajqHx2uGlhrRh6k?via=matrix.org&via=libera.chat&via=element.io

Test Plan:
create and register clients bob and alice
have bob create a public space
have bob create a public room parented to the space
have alice join the space(room)
have alice sync the space using /rooms/$roomId/hierarchy
expect alice to see two rooms in the space hierarchy, the space and the child room

### DIFF
--- a/setup/mscs/msc2946/msc2946.go
+++ b/setup/mscs/msc2946/msc2946.go
@@ -479,7 +479,7 @@ func (w *walker) authorised(roomID, parentRoomID string) (authed, isJoinedOrInvi
 	return w.authorisedServer(roomID), false
 }
 
-// authorisedServer returns true iff the server is joined this room or the room is world_readable
+// authorisedServer returns true iff the server is joined this room or the room is world_readable, public, or knockable
 func (w *walker) authorisedServer(roomID string) bool {
 	// Check history visibility / join rules first
 	hisVisTuple := gomatrixserverlib.StateKeyTuple{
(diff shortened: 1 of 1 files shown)


bones-was-here:
#2504

texuf:
The checks failed, but I don't believe this is my fault, is it possible to restart those stages and see if they work a second time around?

setup/mscs/msc2946/msc2946.go
```diff
@@ -45,6 +45,9 @@ const (
 	ConstCreateEventContentValueSpace = "m.space"
 	ConstSpaceChildEventType          = "m.space.child"
 	ConstSpaceParentEventType         = "m.space.parent"
+	ConstJoinRulePublic               = "public"
```
neilalexander, line 48: We have constants defined in gomatrixserverlib already: `gomatrixserverlib.Public`, `gomatrixserverlib.Knock` and `gomatrixserverlib.Restricted`. Please use those instead of defining additional ones.

neilalexander (changes requested):

setup/mscs/msc2946/msc2946.go
texuf: Awesome suggestion! I made those changes, thanks for the heads up.

clokep:
> It would be nice to add a new test to ensure that public join rule (but NOT world readable rooms) show up in the list, along with knockable rooms.

I'd suggest augmenting the `TestClientSpacesSummaryJoinRules` test case, that seems to be missing these cases.

clokep:
matrix-org/matrix-spec#1186 is likely relevant to this, by the way.

PR #2587, opened 2022-07-22
Add race testing to tests, and fix a few small race conditions in the tests

brianathere:
### Pull Request Checklist

In our CI pipeline the unit tests are non-deterministically failing. The first step in addressing this I propose is to make sure they don't have race conditions that are detected by the race checker.


neilalexander:
Can you please update your sign-off line?

neilalexander (changes requested):
Although I do agree that the mutexes are good to have here for avoiding races in these tests, I'm not mad on bringing in more dependencies (like `github.com/sasha-s/go-deadlock`) to achieve it.

syncapi/sync/requestpool_test.go
```diff
@@ -125,11 +128,19 @@ func TestRequestPool_updatePresence(t *testing.T) {
 	go rp.cleanPresence(db, time.Millisecond*50)
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			beforeCount := publisher.count
+			beforeCount := func() int {
```
S7evinK, line 131: Probably easier to `Lock`/`Unlock` before/after setting `beforeCount`, for readability.

internal/sqlutil/migrate.go
```diff
@@ -50,6 +50,7 @@ type Migration struct {
 
 // Migrator
 type Migrator struct {
+	gooseMutex      sync.Mutex
```
S7evinK, line 53: Do you see any races without this mutex?
There's already a mutex on this struct, which is protecting `knownMigrations`, the other maps shouldn't be accessed concurrently.

brianathere: Yes, the race is coming from accessing goose in two separate goroutines, I'll add a comment clarifying the need for this lock.

S7evinK: Oh, I just saw that this PR was created before #2264 was merged. So I _think_ we may not need this mutex at all.

brianathere: Ah, nice, thank you, I missed this. I removed the mutex.

neilalexander (approved):

PR #2592, opened 2022-07-25
Make the User API responsible for sending account data output events

neilalexander:
The client API was doing this for some reason when the User API is actually the thing that owns the account data.

### DIFF
--- a/userapi/producers/syncapi.go
+++ b/userapi/producers/syncapi.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/matrix-org/dendrite/internal/eventutil"
 	"github.com/matrix-org/dendrite/setup/jetstream"
+	"github.com/matrix-org/dendrite/syncapi/types"
 	"github.com/matrix-org/dendrite/userapi/storage"
 	"github.com/matrix-org/gomatrixserverlib"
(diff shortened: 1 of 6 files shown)


userapi/producers/syncapi.go
```diff
@@ -34,7 +35,7 @@ func NewSyncAPI(db storage.Database, js JetStreamPublisher, clientDataTopic stri
 }
 
 // SendAccountData sends account data to the Sync API server.
-func (p *SyncAPI) SendAccountData(userID string, roomID string, dataType string) error {
+func (p *SyncAPI) SendAccountData(userID string, roomID string, dataType string, readMarker *eventutil.ReadMarkerJSON, ignoredUsers *types.IgnoredUsers) error {
 	m := &nats.Msg{
 		Subject: p.clientDataTopic,
```
kegsay, line 40: Is this the same as `Subject: p.TopicClientData,` from clientapi?

kegsay (approved):

userapi/producers/syncapi.go
neilalexander: Yes, derived from `cfg.Matrix.JetStream.Prefixed(jetstream.OutputClientData)`.

PR #2593, opened 2022-07-25
Fix: Handle joining room not previously synced

texuf:
Mostly making this PR to see if I can flush out the intended behavior. Seems like the intended method would be to first `peek` the room, but peek is currently unimplemented in dendrite. What do you guys think?

Test Plan:
register new account bob
bob creates public, shared history room
bob sends a “hello world” message to the room
register new account jane
jane joins bobs room via room id

ISSUE:
Jane only sees her new membership event, doesn’t sync the name of the room, or bobs “hello world” message

NOTE: this happens any time jane syncs something after bob sends "hello world"
 then joins the room. All that's needed is a more recent FROM offset for Jane
to miss important information about the room

FIX
If a join event happened during a current sync delta, reset the to parameter to sync the entire room to Jane

### Pull Request Checklist

* [ ] I have added added tests for PR _or_ I have justified why this PR doesn't need tests.
* [ ] Pull request includes a [sign off](https://github.com/matrix-org/dendrite/blob/main/docs/CONTRIBUTING.md#sign-off)

### DIFF
--- a/syncapi/storage/shared/syncserver.go
+++ b/syncapi/storage/shared/syncserver.go
@@ -765,6 +765,7 @@ func (d *Database) GetStateDeltas(
 		}
 	}
 
+	var membershipPositions = make(map[string]types.StreamPosition)
 	// handle newly joined rooms and non-joined rooms
 	for roomID, stateStreamEvents := range state {
 		for _, ev := range stateStreamEvents {
@@ -784,6 +785,7 @@ func (d *Database) GetStateDeltas(
 						}
 						return nil, nil, err
 					}
+					membershipPositions[roomID] = ev.StreamPosition
 					state[roomID] = s
 					continue // we'll add this room in when we do joined rooms
 				}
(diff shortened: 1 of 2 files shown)


texuf:
Looks like I can call /state to get this info https://spec.matrix.org/v1.2/client-server-api/#getting-events-for-a-room
I was confused by the js sdk which just had `roomState(...)` and not `fetchRoomState` or `getRoomState` which are the other conventions used for making requests 
However, the documentation implies that sync should return all the info about rooms that I care about. Would love to hear other opinions about this.

texuf:
> Looks like I can call /state to get this info https://spec.matrix.org/v1.2/client-server-api/#getting-events-for-a-room I was confused by the js sdk which just had `roomState(...)` and not `fetchRoomState` or `getRoomState` which are the other conventions used for making requests However, the documentation implies that sync should return all the info about rooms that I care about. Would love to hear other opinions about this.

I take this back, trying to fetch events in the matrix-js-sdk is proving to be a disaster. I can fetch them, but getting them to play nice with the existing room / event architecture is proving to be a disaster.

texuf:
> > Looks like I can call /state to get this info https://spec.matrix.org/v1.2/client-server-api/#getting-events-for-a-room I was confused by the js sdk which just had `roomState(...)` and not `fetchRoomState` or `getRoomState` which are the other conventions used for making requests However, the documentation implies that sync should return all the info about rooms that I care about. Would love to hear other opinions about this.
> 
> I take this back, trying to fetch events in the matrix-js-sdk is not as simple as I had hoped. I can fetch them, but getting them to play nice with the existing room / event architecture is proving to be a disaster.

Figure it out, dendrite returns a `prev_batch` token, which allows me to call `await client.scrollback(room)` which syncs the room, so I'm good here.

texuf:
> Figure it out, dendrite returns a `prev_batch` token, which allows me to call `await client.scrollback(room)` which syncs the room, so I'm good here.

However, I don't believe this will give me all the state events... If anyone wants to chime in with a suggestion I'm all ears.

PR #2596, opened 2022-07-27
Optimise checking other servers allowed to see events

neilalexander:
This is done quite a lot for federation requests, i.e. `/event`, and it's expensive enough today to warrant its own optimised query.

The query filters out all events from the state snapshots that aren't relevant, leaving only the history visibility event and the membership events for the domain name in question.

### DIFF
--- a/roomserver/storage/postgres/state_snapshot_table.go
+++ b/roomserver/storage/postgres/state_snapshot_table.go
@@ -72,9 +72,32 @@ const bulkSelectStateBlockNIDsSQL = "" +
 	"SELECT state_snapshot_nid, state_block_nids FROM roomserver_state_snapshots" +
 	" WHERE state_snapshot_nid = ANY($1) ORDER BY state_snapshot_nid ASC"
 
+// Looks up both the history visibility event and relevant membership events from
+// a given domain name from a given state snapshot. This is used to optimise the
+// helpers.CheckServerAllowedToSeeEvent function.
+// TODO: There's a sequence scan here because of the hash join strategy, which is
+// probably O(n) on state key entries, so there must be a way to avoid that somehow.
+const bulkSelectStateForHistoryVisibilitySQL = `
+SELECT event_nid FROM (
+	SELECT event_nid, event_type_nid, event_state_key_nid FROM roomserver_events
+	WHERE (event_type_nid = 5 OR event_type_nid = 7)
+	AND event_nid = ANY(
+		SELECT UNNEST(event_nids) FROM roomserver_state_block
+		WHERE state_block_nid = ANY(
+			SELECT UNNEST(state_block_nids) FROM roomserver_state_snapshots
+			WHERE state_snapshot_nid = $1
+		)
+	)
+) AS roomserver_events
+INNER JOIN roomserver_event_state_keys
+	ON roomserver_events.event_state_key_nid = roomserver_event_state_keys.event_state_key_nid
+	AND (event_type_nid = 7 OR event_state_key LIKE '%:$2');
+`
+
 type stateSnapshotStatements struct {
-	insertStateStmt              *sql.Stmt
-	bulkSelectStateBlockNIDsStmt *sql.Stmt
+	insertStateStmt                         *sql.Stmt
+	bulkSelectStateBlockNIDsStmt            *sql.Stmt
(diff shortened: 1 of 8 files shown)


roomserver/storage/shared/storage.go
```diff
@@ -988,6 +988,38 @@ func (d *Database) loadEvent(ctx context.Context, eventID string) *types.Event {
 	return &evs[0]
 }
 
+func (d *Database) GetHistoryVisibilityState(ctx context.Context, roomInfo *types.RoomInfo, eventID string, domain string) ([]*gomatrixserverlib.Event, error) {
+	eventStates, err := d.EventsTable.BulkSelectStateAtEventByID(ctx, nil, []string{eventID})
+	if err != nil {
+		return nil, err
+	}
+	stateSnapshotNID := eventStates[0].BeforeStateSnapshotNID
```
kegsay, line 996: Are we guaranteed to have entries if there is no error?

roomserver/storage/postgres/state_snapshot_table.go
```diff
@@ -136,3 +159,26 @@ func (s *stateSnapshotStatements) BulkSelectStateBlockNIDs(
...
+	ctx context.Context, txn *sql.Tx, stateSnapshotNID types.StateSnapshotNID, domain string,
+) ([]types.EventNID, error) {
+	stmt := sqlutil.TxStmt(txn, s.bulkSelectStateForHistoryVisibilityStmt)
+	rows, err := stmt.QueryContext(ctx, stateSnapshotNID, domain)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close() // nolint: errcheck
+	results := make([]types.EventNID, 0, 16)
+	for i := 0; rows.Next(); i++ {
```
kegsay, line 173: Why use `i`?

roomserver/storage/postgres/state_snapshot_table.go
```diff
@@ -136,3 +159,26 @@ func (s *stateSnapshotStatements) BulkSelectStateBlockNIDs(
...
+		var eventNID types.EventNID
+		if err = rows.Scan(&eventNID); err != nil {
+			return nil, err
+		}
+		results = append(results, eventNID)
+	}
+	if err = rows.Err(); err != nil {
+		return nil, err
+	}
+	return results, nil
```
kegsay, line 183: Can just be `retun results, rows.Err()`

roomserver/storage/postgres/state_snapshot_table.go
```diff
@@ -72,9 +72,31 @@ const bulkSelectStateBlockNIDsSQL = "" +
 	"SELECT state_snapshot_nid, state_block_nids FROM roomserver_state_snapshots" +
 	" WHERE state_snapshot_nid = ANY($1) ORDER BY state_snapshot_nid ASC"
 
+// Looks up both the history visibility event and relevant membership events from
+// a given domain name from a given state snapshot. This is used to optimise the
+// helpers.CheckServerAllowedToSeeEvent function.
+// TODO: There's a sequence scan here because of the hash join strategy, which is
+// probably O(n) on state key entries, so there must be a way to avoid that somehow.
+const bulkSelectStateForHistoryVisibilitySQL = "" +
```
kegsay, line 80: I would just use \` personally so you don't have to use `"` and `+` everywhere, makes it a bit easier to read at the cost of a bit of whitespace.

roomserver/storage/postgres/state_snapshot_table.go
```diff
@@ -72,9 +72,31 @@ const bulkSelectStateBlockNIDsSQL = "" +
...
+	"SELECT event_nid FROM (" +
+	" SELECT event_nid, event_type_nid, event_state_key_nid FROM roomserver_events" +
+	" WHERE (event_type_nid = 5 OR event_type_nid = 7)" +
```
kegsay, line 83: Explain what 5 and 7 are, with xrefing to code please.

kegsay (approved):

roomserver/storage/shared/storage.go
neilalexander: Yes, if there's no error then the returned list is always the same size as the input list.

PR #2600, opened 2022-08-01
De-race `types.RoomInfo`

neilalexander:
Since the `types.RoomInfo`s can live in the cache, they can be mutated by the latest events updater.

Therefore this should eliminate potential data races that can occur between `SetLatestEvents()` and other places.


neilalexander:
@kegsay The reason we do it like this is because if we use `types.RoomInfo` in the cache instead of `*types.RoomInfo`, then we can't guarantee that any future `Set`s containing changes will be admitted (because of how Ristretto's admission policy works). By using a single instance in memory, we can guarantee that if we are getting a reference from the cache, we can always update it.

PR #2608, opened 2022-08-02
Delete correct Send-to-Device messages

S7evinK:
This should make sure we delete the correct send-to-device messages, once the client sends us an updated sync token.

### DIFF
--- a/syncapi/syncapi_test.go
+++ b/syncapi/syncapi_test.go
@@ -370,26 +370,26 @@ func testSendToDevice(t *testing.T, dbType test.DBType) {
 		},
 		{
 			name:  "incremental sync, one message", // this deletes message 1, as we advanced the since token
-			since: "s0_0_0_1_0_0_0_0_0",
+			since: types.StreamingToken{SendToDevicePosition: 1}.String(),
 			want: []string{
 				"message 2",
 			},
 		},
 		{
 			name:  "failed incremental sync, one message", // didn't advance since, so still the same message
-			since: "s0_0_0_1_0_0_0_0_0",
+			since: types.StreamingToken{SendToDevicePosition: 1}.String(),
 			want: []string{
 				"message 2",
 			},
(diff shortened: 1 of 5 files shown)


syncapi/syncapi_test.go
```diff
@@ -311,6 +314,138 @@ func testSyncAPIUpdatePresenceImmediately(t *testing.T, dbType test.DBType) {
...
+			name:              "initial sync, two new messages", // we didn't advance the since token, so we'll receive two messages
+			sendMessagesCount: 1,
+			want: []string{
+				"message 1",
+				"message 2",
+			},
+		},
+		{
+			name:  "incremental sync, one message", // this deletes message 1, as we advanced the since token
+			since: "s0_0_0_1_0_0_0_0_0",
```
neilalexander, line 373: Please use something like `types.StreamingToken{SendToDevicePosition: 1}.String()`, so that we don't break the tests if we change the format of the streaming tokens in the future.

syncapi/syncapi_test.go
```diff
@@ -311,6 +314,138 @@ func testSyncAPIUpdatePresenceImmediately(t *testing.T, dbType test.DBType) {
...
+		}
+		time.Sleep((time.Millisecond * 15) * time.Duration(tc.sendMessagesCount)) // wait a bit, so the messages can be processed
+		// sync
+		w := httptest.NewRecorder()
+		base.PublicClientAPIMux.ServeHTTP(w, test.NewRequest(t, "GET", "/_matrix/client/v3/sync", test.WithQueryParams(map[string]string{
+			"access_token": alice.AccessToken,
+			"since":        tc.since,
+		})))
+
+		// verify
```
neilalexander, line 435: Please expand on the comments for this section some more, since this code isn't immediately obvious to the untrained eye.

neilalexander (approved):

PR #2610, opened 2022-08-02
Try more servers when calling `/state_ids`

neilalexander:
... otherwise if we hit a single server that times out, we seemingly give up and reject the events.


roomserver/internal/input/input_missing.go
```diff
@@ -662,9 +665,20 @@ func (t *missingStateReq) lookupMissingStateViaStateIDs(ctx context.Context, roo
 
 	util.GetLogger(ctx).WithField("room_id", roomID).Infof("lookupMissingStateViaStateIDs %s", eventID)
 	// fetch the state event IDs at the time of the event
-	stateIDs, err := t.federation.LookupStateIDs(ctx, t.origin, roomID, eventID)
+	var stateIDs gomatrixserverlib.RespStateIDs
+	var err error
+	count := 0
+	for _, serverName := range t.servers {
```
kegsay, line 671: What order is this in? Previously we hit the origin of the /send txn which feels like the right thing to do, as they really should know the state.

kegsay (approved):

roomserver/internal/input/input_missing.go
```diff
@@ -662,9 +665,20 @@ func (t *missingStateReq) lookupMissingStateViaStateIDs(ctx context.Context, roo
...
+		reqctx, cancel := context.WithTimeout(ctx, time.Second*30)
```
kegsay, line 672: This should maybe be a global timer not per-server else HQ will take aaaaaages

roomserver/internal/input/input_missing.go
neilalexander: In this order: Origin server of the federation request, origin server of the event (if it's different for some reason) and then everything else.

PR #2612, opened 2022-08-02
Add Cache-Control header to media endpoints

### LINKED ISSUE #2589: Rate limiting causes decryption errors in Element
### Background information

- **Dendrite version or git SHA**: 0.8.9
- **Running in Docker?**: No
- **`go version`**: 1.18.4
- **Client used (if applicable)**: Element Web/Desktop

### Description

 - **What** is the problem: 

When the `rate_limiting` config block is left at default, viewing an encrypted room with a lot of media in Element tends to show some of the media as "Error decrypting image/video". 

Switching rooms resets which of the media are decrypted correctly. 

After clearing the cache, all media are decrypted correctly, switching rooms (sometimes >1 times) makes the issue resurface.

Original event sources always stay the same for all media. Decrypted event sources change depending on how the media event in question is currently behaving.

This does not happen on nheko or fluffychat.

The only errors in the dendrite logs were rate-limit triggers.

Disabling `rate_limiting` fixes the issue.

 - **Who** is affected: All our users we tested this with.
 
 - **How** is this bug manifesting: By viewing a room with a bunch of media in the recent timeline in Element Web/Desktop.
 
 - **When** did this first appear: Immediately on a fresh dendrite install. 

### Steps to reproduce

- (Optional: Fresh dendrite install)
- Leave `rate_limiting` config block on these defaults: https://github.com/matrix-org/dendrite/blob/main/dendrite-sample.monolith.yaml#L184=
- Create room
- Send some 10+ media files (images and videos) into the room
- In element (with `Show previews/thumbnails for images` turned on): switch to another room and then to the one where the media were posted (and are still in recent timeline). Maybe do this a couple times, but only rarely more than one switch was needed in our case.

<!--

``` (three backticks, on a line on their own), so that they are formatted legibly.

-->

S7evinK:
This PR
- sets a `Cache-Control` (taken from [Synapse](https://github.com/matrix-org/synapse/blob/e10a2fe0c28ec9206c0e2275df492f61ff5025f2/synapse/rest/media/v1/_base.py#L181-L186)) header on media download endpoints, making sure clients with the ability (e.g. browsers) can cache already downloaded media files. This saves bandwidth and helps with user experience. 
- raises `rate_limiting.threshold` to 20, this should fix #2589

### DIFF
--- a/dendrite-sample.monolith.yaml
+++ b/dendrite-sample.monolith.yaml
@@ -192,7 +192,7 @@ client_api:
   # and appservice users are exempt from rate limiting by default.
   rate_limiting:
     enabled: true
-    threshold: 5
+    threshold: 20
     cooloff_ms: 500
     exempt_user_ids:
     #  - "@user:domain.com"
--- a/dendrite-sample.polylith.yaml
+++ b/dendrite-sample.polylith.yaml
@@ -195,7 +195,7 @@ client_api:
   # and appservice users are exempt from rate limiting by default.
   rate_limiting:
     enabled: true
-    threshold: 5
+    threshold: 20
     cooloff_ms: 500
     exempt_user_ids:
     #  - "@user:domain.com"
--- a/internal/httputil/rate_limiting.go
+++ b/internal/httputil/rate_limiting.go
@@ -1,7 +1,6 @@
 package httputil
 
 import (
-	"fmt"
 	"net/http"
 	"sync"
 	"time"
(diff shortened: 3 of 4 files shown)


neilalexander (changes requested):
The `Cache-Control` bit makes sense, but I don't think changing the rate limiting to blocking behaviour is a good idea.

As well as curbing abuse, rate limiting is also about preventing single users/devices from consuming too many server resources. Every HTTP request to Dendrite is a socket connection, a file descriptor and a request goroutine.

Today Dendrite will simply close the connection immediately with an `M_LIMIT_EXCEEDED` which frees up those resources immediately, but this change will allow the abuser to hold onto those resources for potentially a long time while they "wait", which feels bad.

neilalexander (approved):

PR #2614, opened 2022-08-03
Fix syncapi shared users query & device lists

S7evinK:
This PR
- fixes a bug with the postgres shared users query (missing `pq.Array()`)
- fixes wrongly returned device lists (we were adding users we don't share a room with to `device_list.changed`)
- early returns `QueryKeys` if the context is done/canceled

### DIFF
--- a/syncapi/storage/postgres/current_room_state_table.go
+++ b/syncapi/storage/postgres/current_room_state_table.go
@@ -112,7 +112,7 @@ const selectEventsWithEventIDsSQL = "" +
 const selectSharedUsersSQL = "" +
 	"SELECT state_key FROM syncapi_current_room_state WHERE room_id = ANY(" +
 	"	SELECT room_id FROM syncapi_current_room_state WHERE state_key = $1 AND membership='join'" +
-	") AND state_key = ANY($2) AND membership='join';"
+	") AND state_key = ANY($2) AND membership IN ('join', 'invite');"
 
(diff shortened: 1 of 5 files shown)


neilalexander (approved):

syncapi/storage/postgres/current_room_state_table.go
```diff
@@ -112,7 +112,7 @@ const selectEventsWithEventIDsSQL = "" +
 const selectSharedUsersSQL = "" +
 	"SELECT state_key FROM syncapi_current_room_state WHERE room_id = ANY(" +
 	"	SELECT room_id FROM syncapi_current_room_state WHERE state_key = $1 AND membership='join'" +
-	") AND state_key = ANY($2) AND membership='join';"
+	") AND state_key = ANY($2) AND membership IN ('join', 'invite');"
```
neilalexander, line 115: Shouldn't this be `membership = ANY('join', 'invite')` for PostgreSQL?

neilalexander (approved):

syncapi/storage/postgres/current_room_state_table.go
S7evinK: `membership = ANY(ARRAY['join', 'invite'])`, I think `IN` is easier to read.

neilalexander: If it works then ship it!

PR #2620, opened 2022-08-04
Only create a new destinationQueue if we don't have one

S7evinK:
From the looks of it, the federationapi seems a bit more calm with this change.


neilalexander (approved):
Good catch.

PR #2623, opened 2022-08-05
Invalidate lazyLoadCache if we're doing an initial sync

S7evinK:
Otherwise we might end up with a timeline without displaynames and/or avatars.
Also fixes a minor bug withing `/messages` were we missed the `room_id` in state events.


internal/caching/cache_lazy_load_members.go
```diff
@@ -14,6 +14,7 @@ type lazyLoadingCacheKey struct {
 type LazyLoadCache interface {
 	StoreLazyLoadedUser(device *userapi.Device, roomID, userID, eventID string)
 	IsLazyLoadedUserCached(device *userapi.Device, roomID, userID string) (string, bool)
+	Invalidate(device *userapi.Device, roomID, userID string)
```
neilalexander, line 17: Please name this `InvalidateLazyLoadedUser` so that the interfaces can overlap if we need to define another invalidate function on another cache later.

neilalexander (approved):

PR #2626, opened 2022-08-05
Generic-based internal HTTP API

neilalexander:
Tested out so far on a few endpoints within the federation API, to centralise the machinery about how the internal endpoints work in full HTTP mode.

### DIFF
--- a/internal/httputil/internalapi.go
+++ b/internal/httputil/internalapi.go
@@ -63,8 +63,5 @@ func CallInternalProxyAPI[req, res any, errtype error](name, url string, client 
 	defer span.Finish()
 
 	var response res
-	if err := PostJSON[req, res, errtype](ctx, span, client, url, request, &response); err != nil {
-		return response, err
-	}
-	return response, nil
+	return response, PostJSON[req, res, errtype](ctx, span, client, url, request, &response)
 }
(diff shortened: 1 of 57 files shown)


internal/httputil/internalapi.go
```diff
@@ -0,0 +1,93 @@
...
+			}
+		}
+		return util.JSONResponse{
+			Code: http.StatusOK,
+			JSON: response,
+		}
+	})
+}
+
+func CallInternalRPCAPI[reqtype, restype any](name, url string, client *http.Client, ctx context.Context, request *reqtype, response *restype) error {
```
S7evinK, line 80: Having the context somewhere in the middle feels unnatural.

internal/httputil/internalapi.go
```diff
@@ -0,0 +1,93 @@
...
+	span, ctx := opentracing.StartSpanFromContext(ctx, name)
+	defer span.Finish()
+
+	return PostJSON[reqtype, restype, InternalAPIError](ctx, span, client, url, request, response)
+}
+
+func CallInternalProxyAPI[reqtype, restype any, errtype error](name, url string, client *http.Client, ctx context.Context, request *reqtype) (restype, error) {
```
S7evinK, line 87: Same here.

S7evinK (approved):

internal/httputil/internalapi.go
neilalexander: It's grouped with the `request` and `response` parameters because those are the ones that we pass through.

PR #2633, opened 2022-08-11
Add SQLite WAL and busy_timeout

sfPlayer1:
SQLite has rather unfortunate defaults for concurrent accesses, leading to database locked errors. https://github.com/matrix-org/dendrite/pull/1290 appears to try to fix some of this, but doesn't appear to cover a concurrent read+write and the errors still happened in large quantities when I joined the dendrite room.

This PR enables WAL for concurrent read+write and much better performance. It also sets busy_timeout to 10 seconds to not immediately fail operations while the database is locked, providing a DB level solution to PR 1290's objective with additional coverage.

I chose this particular location to add the two statements as they should apply to every SQLite DB connection.

From a quick test it looks as if the db locked errors are gone with this change.

### Pull Request Checklist

* [ ] I have added added tests for PR _or_ I have justified why this PR doesn't need tests.
* [ ] Pull request includes a [sign off](https://github.com/matrix-org/dendrite/blob/main/docs/CONTRIBUTING.md#sign-off)


neilalexander (commented):
The reason we use locking writers is because in some cases (i.e. in WebAssembly with sql.js) we can't rely on WAL being present. If there are `database is locked` errors then that sounds like there is probably a bug somewhere with something not using a writer when it should be and enabling WAL may just be covering up the actual cause.

Also be aware that we don't recommend the use of SQLite for production federating homeservers.

neilalexander:
Oh, additionally, if we are to accept this PR, you need to [update the sign-off line](https://github.com/matrix-org/dendrite/blob/main/docs/CONTRIBUTING.md#sign-off) as per the PR template.

internal/sqlutil/sqlutil.go
```diff
@@ -46,6 +46,9 @@ func Open(dbProperties *config.DatabaseOptions, writer Writer) (*sql.DB, error)
 		db.SetMaxOpenConns(dbProperties.MaxOpenConns())
 		db.SetMaxIdleConns(dbProperties.MaxIdleConns())
 		db.SetConnMaxLifetime(dbProperties.ConnMaxLifetime())
+	} else {
+		db.Exec("PRAGMA busy_timeout = 10000;")
```
neilalexander, line 50: The linter unhappiness here is because it thinks you might have ignored the return value by accident — I suggest dropping the error explicitly in case it isn't supported on a given underlying platform (like WASM), so in that case you need to do `_ = db.Exec(...`.

kegsay:
We recently updated our [contributing guidelines](https://github.com/matrix-org/dendrite/blob/main/docs/CONTRIBUTING.md#contribution-types). This PR is being closed because it isn't a feature we want to maintain going forwards. Supporting WAL is nice but adds extra risks with the extra concurrency it allows. We currently have a mechanism to limit concurrency on SQLite which we are happy with overall. We feel that landing this PR would increase the overall maintenance burden of SQLite mode due to the additional race conditions exposed with extra concurrency.

When we have more bandwidth as a team, we would be interested in adding support for this.

PR #2647, opened 2022-08-16
Reprocess rejected input events

neilalexander:
... otherwise they stay perma-rejected, even if we've since learned about something that might allow us to reprocess it successfully later.


S7evinK (approved):

kegsay (changes requested):
So previously we would reprocess events if they first came in as outliers and then later as a different Kind. This PR breaks that: its mere presence in the events table is enough for us to skip reprocessing. I worry that means outliers will never get a place in the dag as they won't ever be reprocessed. Outliers don't get rejected in the same way as live events, so I expect is_rejected to be false.

Issue #2121, opened 2022-01-28
DM rooms are displayed as normal rooms


kegsay:
> New 1:1 chat initiate by my synapse server to dentrite shows up under rooms not under person on the dendrite element clienrt

> I see the same. Invite from dendrite homeserver to matrix.org sows up as person, but from matrix.org to homeserver shows up in rooms on the homeserver, both element clients

Could be a few things:
 - Account data isn't sent to clients
 - `is_direct` hints aren't making it across federation
 
Ref: https://spec.matrix.org/v1.1/client-server-api/#direct-messaging

zeeZ:
That second quote is me, running v0.5.1.

kegsay:
We now have complement tests to assert that:
 - Account data IS sent to clients
 - `is_direct` hints ARE making it across federation

Without further logs I cannot help more on this one. Additional Qs:
 - Does this consistently happen with all new DM rooms?
 - Does it happen when you DM users on the same homeserver?

S7evinK:
As far as my testing goes, this only happens when a user is invited over federation. As soon as the invite is accepted, the DM appears as a normal room in Element Web (and probably other clients)

PR #2648, opened 2022-08-17
Try to reconstuct membership events if it's a federated invite

S7evinK:
This should fix https://github.com/matrix-org/dendrite/issues/2121

The issue here is that `/rooms/{roomID}/members` was failing with "You aren't a member of the room and weren't previously a member of the room.", as we don't have a `membershipNID` for invites coming over federation, and we don't know anything about the room yet. This tries to reconstruct the membership events for invites coming in over federation, making Element Web a bit more happy.

### DIFF
--- a/roomserver/internal/perform/perform_invite.go
+++ b/roomserver/internal/perform/perform_invite.go
@@ -18,6 +18,10 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/matrix-org/gomatrixserverlib"
+	"github.com/matrix-org/util"
+	log "github.com/sirupsen/logrus"
+
 	federationAPI "github.com/matrix-org/dendrite/federationapi/api"
 	"github.com/matrix-org/dendrite/roomserver/api"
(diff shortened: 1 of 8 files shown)


neilalexander:
Is Element Web *really* trying to hit `/members` for an invite that hasn't been accepted yet? If so I would consider that bad behaviour.

S7evinK:
Yep, as soon as the room/invite is selected.

neilalexander:
I'm not very comfortable with this.

This feels like a hack to fix something that Element Web shouldn't be doing to begin with. I think first of all we should be raising an issue and talking to the EW guys about why they need to hit `/members` to work out if the room is a DM or not when the invite stripped state should already contain the `is_direct` field. 

When a federated invite comes in, the stripped state is literally all we know about the room — we don't have any of the full auth events or anything useful enough to construct an actual "room" in the roomserver. We also don't even *really* know what the event IDs are, which is a required field in the `/members` response, and we have no way to calculate them. I don't really want to get into the realm of Dendrite needing to fabricate events/responses in this way.

PR #2652, opened 2022-08-18
State handling tweaks

neilalexander:
This tweaks how rejected events are handled in room state and also to not apply checks we can't complete to outliers.

### DIFF
--- a/roomserver/internal/helpers/auth.go
+++ b/roomserver/internal/helpers/auth.go
@@ -39,7 +39,7 @@ func CheckForSoftFail(
 	var authStateEntries []types.StateEntry
 	var err error
 	if rewritesState {
-		authStateEntries, err = db.StateEntriesForEventIDs(ctx, stateEventIDs, false)
+		authStateEntries, err = db.StateEntriesForEventIDs(ctx, stateEventIDs, true)
 		if err != nil {
 			return true, fmt.Errorf("StateEntriesForEventIDs failed: %w", err)
 		}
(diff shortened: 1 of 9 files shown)


roomserver/storage/interface.go
```diff
@@ -79,7 +79,7 @@ type Database interface {
 	// Look up the state entries for a list of string event IDs
 	// Returns an error if the there is an error talking to the database
 	// Returns a types.MissingEventError if the event IDs aren't in the database.
```
kegsay, line 81: Need to comment how `excludeRejected` interacts with this. If all `eventIDs` are excluded, does it return `[]` or a `types.MissingEventError`?

roomserver/storage/sqlite3/events_table.go
```diff
@@ -221,17 +237,18 @@ func (s *eventStatements) BulkSelectStateEventByID(
...
 		if err = rows.Scan(
 			&result.EventTypeNID,
 			&result.EventStateKeyNID,
 			&result.EventNID,
 		); err != nil {
 			return nil, err
 		}
+		results = append(results, result)
 	}
 	if i != len(eventIDs) {
```
kegsay, line 253: Missing `excludeRejected` check.

roomserver/internal/helpers/auth.go
```diff
@@ -39,7 +39,7 @@ func CheckForSoftFail(
 	var authStateEntries []types.StateEntry
 	var err error
 	if rewritesState {
-		authStateEntries, err = db.StateEntriesForEventIDs(ctx, stateEventIDs)
+		authStateEntries, err = db.StateEntriesForEventIDs(ctx, stateEventIDs, false)
```
kegsay, line 42: Why do we include rejected events here? Seems dangerous?

roomserver/internal/helpers/auth.go
```diff
@@ -97,7 +97,7 @@ func CheckAuthEvents(
 	authEventIDs []string,
 ) ([]types.EventNID, error) {
 	// Grab the numeric IDs for the supplied auth state events from the database.
-	authStateEntries, err := db.StateEntriesForEventIDs(ctx, authEventIDs)
+	authStateEntries, err := db.StateEntriesForEventIDs(ctx, authEventIDs, false)
```
kegsay, line 100: Why do we include rejected events here? Seems dangerous?

kegsay (changes requested):

roomserver/storage/interface.go
neilalexander: Updated the comment, it returns `[]`.

PR #2657, opened 2022-08-19
Fetch more data for newly joined rooms in an incremental sync

S7evinK:
If we've joined a new room in an incremental sync, try fetching more data.
This deflakes the complement server notices test (at least in my tests).

### DIFF
--- a/syncapi/streams/stream_pdu.go
+++ b/syncapi/streams/stream_pdu.go
@@ -178,24 +178,24 @@ func (p *PDUStreamProvider) IncrementalSync(
 
 	var err error
 	var stateDeltas []types.StateDelta
-	var joinedRooms []string
+	var syncJoinedRooms []string
 
 	stateFilter := req.Filter.Room.State
 	eventFilter := req.Filter.Room.Timeline
 
 	if req.WantFullState {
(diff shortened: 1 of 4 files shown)


syncapi/streams/stream_pdu.go
```diff
@@ -222,6 +222,32 @@ func (p *PDUStreamProvider) IncrementalSync(
...
 
+	// If we joined a new room in this sync, make sure we add enough information about it.
+	// This does an "initial sync" for the newly joined rooms
+	newlyJoinedRooms := joinedRooms(req.Response, req.Device.UserID)
+	if len(newlyJoinedRooms) > 0 {
+		// remove already added rooms, as we're doing an "initial sync"
+		for _, x := range newlyJoinedRooms {
+			delete(req.Response.Rooms.Join, x)
+		}
+		r = types.Range{
```
neilalexander, line 233: You may want this backwards:
```
	r := types.Range{
		From:      to,
		To:        0,
		Backwards: true,
	}
```
... so that you get the most recent timeline events, not the earliest.

neilalexander (approved):

PR #2662, opened 2022-08-22
Add `/_dendrite/admin/purgeRoom/{roomID}` (PostgreSQL only)

neilalexander:
This adds a new admin endpoint `/_dendrite/admin/purgeRoom/{roomID}`. It completely erases all database entries for a given room ID.

The roomserver will start by clearing all data for that room and then will generate an output event to notify downstream components (i.e. the sync API and federation API) to do the same.

It does not currently clear media and it is currently not implemented for SQLite since it relies on SQL array operations right now.


roomserver/internal/perform/perform_admin.go
```diff
@@ -231,3 +232,37 @@ func (r *Admin) PerformAdminEvacuateUser(
 	}
 	return nil
 }
+
+func (r *Admin) PerformAdminPurgeRoom(
+	ctx context.Context,
+	req *api.PerformAdminPurgeRoomRequest,
+	res *api.PerformAdminPurgeRoomResponse,
+) error {
+	if _, _, err := gomatrixserverlib.SplitID('!', req.RoomID); err != nil {
```
kegsay, line 241: Would be nice to comment this. It's just basically checking that the input starts with `!`, but it's a nice validation check on the off chance this API is abused.

roomserver/internal/perform/perform_admin.go
```diff
@@ -231,3 +232,37 @@ func (r *Admin) PerformAdminEvacuateUser(
...
+		res.Error = &api.PerformError{
+			Code: api.PerformErrorBadRequest,
+			Msg:  fmt.Sprintf("Malformed room ID: %s", err),
+		}
+		return nil
+	}
+
+	logrus.WithField("room_id", req.RoomID).Warn("Purging room from roomserver")
+	if err := r.DB.PurgeRoom(ctx, req.RoomID); err != nil {
+		logrus.WithField("room_id", req.RoomID).WithError(err).Warn("Failed to purge room from roomserver")
```
kegsay, line 251: We still fallthrough to notifying downstream consumers here, which could cause problems as they will now be out of sync if you were to purge -> join the same room from fresh again. We should probably not notify downstream components when we fail to purge room data like this.

roomserver/storage/tables/interface.go
```diff
@@ -171,6 +172,20 @@ type Redactions interface {
 	MarkRedactionValidated(ctx context.Context, txn *sql.Tx, redactionEventID string, validated bool) error
 }
 
+type Purge interface {
```
kegsay, line 175: This is confusing at it isn't a table.

roomserver/storage/shared/storage.go
```diff
@@ -1352,6 +1353,57 @@ func (d *Database) ForgetRoom(ctx context.Context, userID, roomID string, forget
...
+			if err == sql.ErrNoRows {
+				return nil
+			}
+			return fmt.Errorf("failed to lock the room: %w", err)
+		}
+		if err := d.Purge.PurgeStateBlocks(ctx, txn, roomNID); err != nil {
+			return fmt.Errorf("failed to purge state blocks: %w", err)
+		}
+		if err := d.Purge.PurgeStateSnapshots(ctx, txn, roomNID); err != nil {
+			return fmt.Errorf("failed to purge state blocks: %w", err)
```
kegsay, line 1374:
```suggestion
			return fmt.Errorf("failed to purge state snapshots: %w", err)
```

roomserver/storage/postgres/purge_statements.go
```diff
@@ -0,0 +1,159 @@
...
+	"context"
+	"database/sql"
+
+	"github.com/matrix-org/dendrite/internal/sqlutil"
+	"github.com/matrix-org/dendrite/roomserver/types"
+)
+
+const purgeEventJSONSQL = "" +
+	"DELETE FROM roomserver_event_json WHERE event_nid = ANY(" +
+	"	SELECT event_nid FROM roomserver_events WHERE room_nid = $1" +
```
kegsay, line 13: Would a JOIN be a bit more efficient than this? i.e join on the event_nid then delete where the room_nid matches $1.

kegsay (changes requested):
Needs tests just to exercise the DELETE statements to ensure there are no typos in them / stop them from bitrotting when we update column names but inevitably forget to update the names in `purge_statements.go`.

syncapi/storage/shared/syncserver.go
```diff
@@ -1078,3 +1078,36 @@ func (d *Database) MaxStreamPositionForPresence(ctx context.Context) (types.Stre
 func (d *Database) SelectMembershipForUser(ctx context.Context, roomID, userID string, pos int64) (membership string, topologicalPos int, err error) {
 	return d.Memberships.SelectMembershipForUser(ctx, nil, roomID, userID, pos)
 }
+
+func (d *Database) PurgeRoom(ctx context.Context, roomID string) error {
+	return d.Writer.Do(d.DB, nil, func(txn *sql.Tx) error {
+		if err := d.BackwardExtremities.PurgeBackwardExtremities(ctx, txn, roomID); err != nil {
+			return fmt.Errorf("failed to purge backward extremities: %w", err)
+		}
+		if err := d.CurrentRoomState.DeleteRoomStateForRoom(ctx, txn, roomID); err != nil {
```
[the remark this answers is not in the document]
neilalexander: Yes, every trace of the room.

clientapi/admin_test.go
```diff
@@ -132,3 +135,99 @@ func TestAdminResetPassword(t *testing.T) {
...
+			wantOK bool
+		}{
+			{name: "Can purge existing room", wantOK: true, roomID: room.ID},
+			{name: "Can not purge non-existent room", wantOK: false, roomID: "!doesnotexist:localhost"},
+			{name: "rejects invalid room ID", wantOK: false, roomID: "@doesnotexist:localhost"},
+		}
+
+		for _, tc := range testCases {
+			t.Run(tc.name, func(t *testing.T) {
+				req := test.NewRequest(t, http.MethodGet, "/_dendrite/admin/purgeRoom/"+tc.roomID)
```
kegsay, line 219: I'm surprised this works with iteration vs loop variables, but I guess because `t.Run` blocks this is fine. Might be worth doing `tc := tc` regardless.

clientapi/routing/routing.go
```diff
@@ -165,6 +165,12 @@ func Setup(
 		}),
 	).Methods(http.MethodGet, http.MethodOptions)
 
+	dendriteAdminRouter.Handle("/admin/purgeRoom/{roomID}",
+		httputil.MakeAdminAPI("admin_purge_room", userAPI, func(req *http.Request, device *userapi.Device) util.JSONResponse {
+			return AdminPurgeRoom(req, cfg, device, rsAPI)
+		}),
+	).Methods(http.MethodGet, http.MethodOptions)
```
kegsay, line 172: Very strange this is a GET request given it is destructive. POST would be a more suitable verb.

internal/sqlutil/sql.go
```diff
@@ -153,6 +158,22 @@ func RunLimitedVariablesQuery(ctx context.Context, query string, qp QueryProvide
 	return nil
 }
 
+// RunLimitedVariablesExec split up a query with more variables than the used database can handle in multiple queries.
+func RunLimitedVariablesExec(ctx context.Context, query string, qp ExecProvider, variables []interface{}, limit uint) error {
```
kegsay, line 162: This absolutely needs tests. For sliding sync I use a different API - [Chunkify](https://github.com/matrix-org/sliding-sync/blob/main/sqlutil/sql.go#L49) but it ends up doing the same thing.

roomserver/storage/shared/storage.go
```diff
@@ -1445,6 +1446,54 @@ func (d *Database) ForgetRoom(ctx context.Context, userID, roomID string, forget
...
+		if err = d.Purge.PurgeStateBlocks(ctx, txn, roomNID); err != nil {
+			return fmt.Errorf("failed to purge state blocks: %w", err)
+		}
+		if err = d.Purge.PurgeStateSnapshots(ctx, txn, roomNID); err != nil {
+			return fmt.Errorf("failed to purge state snapshots: %w", err)
+		}
+		if err = d.Purge.PurgeInvites(ctx, txn, roomNID); err != nil {
+			return fmt.Errorf("failed to purge invites: %w", err)
+		}
+		if err = d.Purge.PurgeMemberships(ctx, txn, roomNID); err != nil {
```
kegsay, line 1469: I don't understand why this isn't just `d.Purge.PurgeRoom(ctx, txn, roomNID)`. We've added a bazillion functions here which end up being easy to forget one, and hard to verify they are all being called. Just combine all these SQL statements into a single function?

roomserver/storage/tables/interface.go
```diff
@@ -173,6 +174,20 @@ type Redactions interface {
...
+	PurgeEvents(ctx context.Context, txn *sql.Tx, roomNID types.RoomNID) error
+	PurgeRoom(ctx context.Context, txn *sql.Tx, roomNID types.RoomNID) error
+	PurgeStateSnapshots(ctx context.Context, txn *sql.Tx, roomNID types.RoomNID) error
+	PurgeStateBlocks(ctx context.Context, txn *sql.Tx, roomNID types.RoomNID) error
+	PurgePreviousEvents(ctx context.Context, txn *sql.Tx, roomNID types.RoomNID) error
+	PurgeInvites(ctx context.Context, txn *sql.Tx, roomNID types.RoomNID) error
+	PurgeMemberships(ctx context.Context, txn *sql.Tx, roomNID types.RoomNID) error
+	PurgePublished(ctx context.Context, txn *sql.Tx, roomID string) error
+	PurgeRedactions(ctx context.Context, txn *sql.Tx, roomNID types.RoomNID) error
+	PurgeRoomAliases(ctx context.Context, txn *sql.Tx, roomID string) error
```
kegsay, line 188: As previously mentioned, this seems crazy to have 11 functions which need to be called sequentially else it isn't purging the room. Just do it all in a single function?!

syncapi/consumers/roomserver.go
```diff
@@ -127,6 +128,11 @@ func (s *OutputRoomEventConsumer) onMessage(ctx context.Context, msgs []*nats.Ms
 		s.onRetirePeek(s.ctx, *output.RetirePeek)
 	case api.OutputTypeRedactedEvent:
 		err = s.onRedactEvent(s.ctx, *output.RedactedEvent)
+	case api.OutputTypePurgeRoom:
+		err = s.onPurgeRoom(s.ctx, *output.PurgeRoom)
+		if err != nil {
+			return true
```
kegsay, line 134:
```suggestion
			return true // non-fatal if we don't purge
```

kegsay (approved):
LGTM!

primalmotion:
I had an issue when I tried to join the Arch Linux room (which is pretty fat room). It ended joining it, but it was a mess. When I tried to leave, it was impossible. I tried many things to no avail, until I tried this PR to clean up everything. That worked great for me.

PR #2632, opened 2022-08-10
Rewrite Startup for using cobra (and koanf)

genofire:
improvements are:
- one binary also for service like
  - [x] create-account
  - [x] generete-config
  - [x] generate-keys
  - [x] monolith
  - [x] polylith (with there components)
  - [x] version
- config reading with:
  - toml
  - yaml
  - json
  - env:
      - [ ] split database uri into components (e.g. for set password on another way then host - k8s secrets vs. configmaps)
      - works level of `struct` are `__` (two underscore - so e.g. `server_name` is the same on yaml): 
          - e.g. set `DENDRITE_GLOBAL__SERVER_NAME`
      - e.g for `create-account` without config-file:
       ```bash
       export DENDRITE_GLOBAL__PRIVATE_KEY="matrix_key.pem" DENDRITE_CLIENT_API__REGISTRATION_SHARED_SECRET="a"; ./dendrite create-account --username a
       ```

- config writing (by `generete-config` with env and cli-flags configurated) into yaml

fix #2591, #1958, #393.

PS: viper has some problems with parsing env and is not possible to disable file reading

---

that would be the new structure:
if you like it, i will reimplement the rest - please feedback @neilalexander 

---
### CMD Helps
```
Dendrite is a second-generation Matrix homeserver written in Go.
It intends to provide an efficient, reliable and scalable alternative to Synapse.

Usage:
  dendrite [command]

Available Commands:
  account     Manage Account on this server
  completion  Generate the autocompletion script for the specified shell
  config      Configuration tools
  help        Help about any command
  keys        key tools
  server      run the server

Flags:
  -h, --help      help for dendrite
  -v, --version   version for dendrite

Use "dendrite [command] --help" for more information about a command.
```
<details>
  <summary>account</summary>

```
Manage Account on this server

Usage:
  dendrite account [command]

Available Commands:
  create      Creates a new user account on the homeserver.

Flags:
  -h, --help   help for account

Use "dendrite account [command] --help" for more information about a command.
```
</details>
<details>
  <summary>account create</summary>

```
Creates a new user account on the homeserver.

Usage:
  dendrite account create [flags]

Flags:
      --admin                 Create an admin account
  -c, --config string         config file
  -h, --help                  help for create
      --password string       The password to associate with the account
      --passwordfile string   The file to use for the password (e.g. for automated account creation)
      --passwordstdin         Reads the password from stdin
      --reset-password        Deprecated
      --timeout duration      Timeout for the http client when connecting to the server (default 30s)
[...]

PR #2664, opened 2022-08-22
Add helm chart

samip5:
The original origin of this chart is from the deprecated k8s-at-home [repo](https://github.com/k8s-at-home/charts/tree/master/charts/incubator/dendrite), I hope that it would get some more love being an official one.

This requires some tests / Github workflow that I cannot really do, but examples of how that works are at the deprecated repo mentioned above.

### Pull Request Checklist


charts/dendrite/values.yaml
```diff
@@ -0,0 +1,599 @@
...
+#
+---
+image:
+  # -- image repository
+  repository: ghcr.io/matrix-org/dendrite-monolith
+  # -- image tag
+  # @default -- chart.appVersion
+  tag:
+  # -- image pull policy
+  pullPolicy: IfNotPresent
```
genofire, line 16: missing `PullSecrets` for own mirrors and own image registries

charts/dendrite/values.yaml
```diff
@@ -0,0 +1,599 @@
...
+        enabled: true
+        port: 8448
+        protocol: HTTPS
+
+ingress:
+  # -- (Monolith Only) Enable and configure ingress settings for the chart under
+  # this key.
+  # @default -- See values.yaml
+  main:
+    enabled: false
```
genofire, line 44: - missing `annotations` for `certmanager` and other parts tooling in a k8s
- missing `tls` for store `certs`
- missing `ingressClassName` for multiple ingress setups in k8s

charts/dendrite/values.yaml
```diff
@@ -0,0 +1,599 @@
...
+    mountPath: &mediaPath /var/dendrite/media
+    accessMode: ReadWriteOnce
+    size: 5Gi
+  # -- Configure Jetsream persistence. This is highly recommended in production.
+  # @default -- See values.yaml
+  jetstream:
+    enabled: false
+    mountPath: /var/dendrite/jetstream
+    accessMode: ReadWriteOnce
+    size: 1Gi
```
genofire, line 60: - missing `storageClass`

genofire (changes requested):
thank you for your helm chat, this projects really need one.

Missing a lot of default configuration options for a default helmChart.
`resource` on every container e.g.

and much more.


--- 
optional i would wish this that a helmchart deploy also CRDs for monitoring, e.g.:
- ServiceMonitor https://prometheus-operator.dev/docs/operator/api/#monitoring.coreos.com/v1.ServiceMonitor for fetch metrics
- PrometheusRules https://prometheus-operator.dev/docs/operator/api/#monitoring.coreos.com/v1.PrometheusRule for alerting

charts/dendrite/values.yaml
genofire: I believe you have some of them for a poly setup (but not documented in values.yaml)
And for a monolith setup, i do not see an ingress

charts/dendrite/templates/dendrite-config.yaml
```diff
@@ -0,0 +1,208 @@
+{{- $componentSpecificDatabaseConfig := or .Values.dendrite.polylithEnabled -}}
```
genofire, line 1: i hope, to get it unnessary with #2632 using env.

genofire (commented):
I do not find any "Deployment", i like to search for an label with hash of the configuration -> if it change, that it restart ...

Does i unseen it or you miss to commit it?

charts/dendrite/values.yaml
samip5: This is quite complicated chart, so please see the dependencies as quite a lot of those missing things come from there, unless you mean that a user wouldn't need to specify them separately?

The ingress is currently ONLY for a monolith deployment due to the complexities of the polylith one.

I'm currently utilizing this here (using my own chart repo and modified chart): https://github.com/samip5/k8s-cluster/blob/main/cluster/apps/default/dendrite/helm-release.yaml

samip5:
> I do not find any "Deployment", i like to search for an label with hash of the configuration -> if it change, that it restart ...
> 
> Does i unseen it or you miss to commit it?

There are dependencies, which have an controller which specifies that.

charts/dendrite/values.yaml
samip5: > missing `PullSecrets` for own mirrors and own image registries

Nope, https://github.com/k8s-at-home/library-charts/blob/main/charts/stable/common/values.yaml#L49

charts/dendrite/values.yaml
samip5: Better to not specify one by default, as it would use global default if one exists in a cluster without it being set.

neilalexander:
It may be best to maintain this in a separate repository. My concern is that by putting this into the main Dendrite repository, it is giving the illusion of it being official and supported. I don't think we're in a position right now to maintain it.

samip5:
> It may be best to maintain this in a separate repository. My concern is that by putting this into the main Dendrite repository, it is giving the illusion of it being official and supported. I don't think we're in a position right now to maintain it.

Regarding this, what would be preferred approach? I don't really have the time to keep it up-to-date, and it would probably make more sense to have under matrix-org to be maintained under the org.

S7evinK:
Closing this in favor of https://github.com/matrix-org/dendrite/pull/2905.
This adds too much complexity around CI (setting up k8s-at-home CI stuff etc.)

PR #2667, opened 2022-08-23
Detect `types.MissingStateError` in `CheckServerAllowedToSeeEvent`

neilalexander:
This will hopefully stop some 500 errors on `/event` where there is no state-before known.


S7evinK (approved):
LGTM, if CI is happy.

PR #2675, opened 2022-08-26
Fulltext implementation using Bleve

S7evinK:
Based on #2480

This actually indexes events based on their event type. They are removed from the index if we receive a `m.room.redaction` event on the `OutputRoomEvent` stream.
An admin endpoint is added to reindex all existing events.

Todo:
[ ] Pagination
[ ] Fix grouping (iirc)


neilalexander (approved):
Overall this look OK to me!

PR #2678, opened 2022-08-26
Fix race condition in monitor API

brianathere:
Signed-off-by: Brian Meek <brian@hntlabs.com>

Fix race condition in monitor API


S7evinK:
Should be fixed by #2679 now. If you're still seeing a race, feel free to re-open.
(Also: Didn't know you can build/install with `-race` enabled, so thanks :) )

PR #2687, opened 2022-08-31
Refactor appservices component

neilalexander:
This PR refactors the app services component. It makes the following changes:

* Each appservice now gets its own NATS JetStream consumer
* The appservice database is now removed entirely, since we just use JetStream as a data source instead
* The entire component is now much simpler and we deleted lots of lines of code 💅

The result is that it should be much lighter and hopefully much more performant.


appservice/consumers/roomserver.go
```diff
@@ -33,178 +37,192 @@ import (
...
+}
+
+// backoff pauses the calling goroutine for a 2^some backoff exponent seconds
+func (s *appserviceState) backoffAndPause(err error) error {
+	if s.backoff < 6 {
+		s.backoff++
+	}
+	duration := time.Second * time.Duration(math.Pow(2, float64(s.backoff)))
+	log.WithField("appservice", s.ID).WithError(err).Errorf("Unable to send transaction to appservice, backing off for %s", duration.String())
+	time.Sleep(duration)
```
S7evinK, line 211: Can we actually sleep here?

appservice/consumers/roomserver.go
```diff
@@ -33,178 +37,192 @@ import (
...
-		log.WithFields(log.Fields{
-			"room_id": event.RoomID(),
-		}).WithError(err).Errorf("Unable to get membership for room")
+	// If the response was fine then we can clear any backoffs in place and
+	// report that everything was OK. Otherwise, back off for a while.
+	switch resp.StatusCode {
+	case http.StatusOK:
+		state.backoff = 0
+	default:
+		_ = state.backoffAndPause(err)
```
S7evinK, line 199: We should return a new error here, as this will fall through and return nil. Resulting in "losing" those events, if I see that correctly. (line 157 would be `true`)

S7evinK (approved):
Besides the comments, LGTM!

appservice/consumers/roomserver.go
neilalexander: Yes, because each appservice has its own consumer and therefore worker goroutine, so this effectively just blocks the worker for that single appservice for a while.

PR #2688, opened 2022-08-31
Refactor notifications

S7evinK:
This PR changes the handling of notifications
- removes the `StreamEvent` and `ReadUpdate` stream
- listens on the `OutputRoomEvent` stream in the UserAPI to inform the SyncAPI about unread notifications
- listens on the `OutputReceiptEvent` stream in the UserAPI to set receipts/update notifications
- sets the `read_markers` directly from within the internal UserAPI

### DIFF
--- a/userapi/storage/postgres/pusher_table.go
+++ b/userapi/storage/postgres/pusher_table.go
@@ -19,11 +19,12 @@ import (
 	"database/sql"
 	"encoding/json"
 
+	"github.com/sirupsen/logrus"
+
 	"github.com/matrix-org/dendrite/internal"
 	"github.com/matrix-org/dendrite/internal/sqlutil"
 	"github.com/matrix-org/dendrite/userapi/api"
 	"github.com/matrix-org/dendrite/userapi/storage/tables"
-	"github.com/sirupsen/logrus"
 )
 
 // See https://matrix.org/docs/spec/client_server/r0.6.1#get-matrix-client-r0-pushers
@@ -136,7 +137,7 @@ func (s *pushersStatements) SelectPushers(
 		pushers = append(pushers, pusher)
 	}
 
-	logrus.Debugf("Database returned %d pushers", len(pushers))
+	logrus.Tracef("Database returned %d pushers", len(pushers))
 	return pushers, rows.Err()
 }
 
--- a/userapi/storage/sqlite3/pusher_table.go
+++ b/userapi/storage/sqlite3/pusher_table.go
@@ -19,11 +19,12 @@ import (
 	"database/sql"
 	"encoding/json"
 
+	"github.com/sirupsen/logrus"
+
 	"github.com/matrix-org/dendrite/internal"
 	"github.com/matrix-org/dendrite/internal/sqlutil"
 	"github.com/matrix-org/dendrite/userapi/api"
 	"github.com/matrix-org/dendrite/userapi/storage/tables"
-	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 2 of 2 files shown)


texuf:
What is this refactor trying to solve or achieve? I don't see a linked issue. I personally see two problems with the sync
- the notifications and rooms have different stream positions, but we only send notifications to the client when there happen to be updates to both (almost never because of the way the threads line up) and we end up sending 0, 0 notification numbers when a different number exists in the sync db
- you can't clear old notifications because we clean up the user notification db, but rely on the existence of a specific eventId in that db to forward on a message to the sync db where we store the aggregate.

I fixed both here, https://github.com/matrix-org/dendrite/pull/2701/files, but I'm lacking a lot of the bigger picture. I would love to hear your feedback on these two issues and understand what else you're trying to fix.

S7evinK:
It's basically trying to solve the same issue you are trying to, that the notification counts are not getting updated "in time"/correctly.
The reason for this still being a draft is, that there's still something odd with clearing notifications, like you've mentioned. Good catch with the `updated` check, will see if that works better here.

With this PR the different stream positions would "disappear", there would only be one in the SyncAPI.

syncapi/storage/sqlite3/notification_data_table.go
```diff
@@ -81,20 +82,26 @@ func (r *notificationDataStatements) UpsertRoomUnreadCounts(ctx context.Context,
 	return
 }
 
-func (r *notificationDataStatements) SelectUserUnreadCounts(ctx context.Context, txn *sql.Tx, userID string, fromExcl, toIncl types.StreamPosition) (map[string]*eventutil.NotificationData, error) {
-	rows, err := sqlutil.TxStmt(txn, r.selectUserUnreadCounts).QueryContext(ctx, userID, fromExcl, toIncl)
+func (r *notificationDataStatements) SelectUnserUnreadCountsForRooms(
```
neilalexander, line 85: Please correct the spelling `Unser` -> `User`.

userapi/consumers/roomserver.go
```diff
@@ -129,7 +137,7 @@ func (s *OutputStreamEventConsumer) processMessage(ctx context.Context, event *g
 		"room_id":     event.RoomID(),
 		"num_members": len(members),
 		"room_size":   roomSize,
-	}).Tracef("Notifying members")
+	}).Debugf("Notifying members")
```
neilalexander, line 140: Can we pop this back to `Tracef` please?

userapi/consumers/roomserver.go
```diff
@@ -447,7 +455,7 @@ func (s *OutputStreamEventConsumer) evaluatePushRules(ctx context.Context, event
 		"room_id":   event.RoomID(),
 		"localpart": mem.Localpart,
 		"rule_id":   rule.RuleID,
-	}).Tracef("Matched a push rule")
+	}).Debugf("Matched a push rule")
```
neilalexander, line 458: Same here, as well as some other places.

neilalexander (approved):
I think this looks OK aside from one or two minor cosmetic things.

texuf:
I'm seeing tons of `sql: converting argument $1 type: unsupported type []interface {}, a slice of interface` after merging this pr, 
```
level=error msg="GetUserUnreadNotificationCountsForRooms failed" device_id=JVIH5pVr error="sql: converting argument $1 type: unsupported type []interface {}, a slice of interface" limit=20"
```

Is this something I've messed up locally?

texuf:
Wow this is really broken, I'll put up a PR.

Issue #2564, opened 2022-07-08
Considerations for Go 1.19 update


0x1a8510f2:
Reading through the [Go 1.19rc1 release notes](https://tip.golang.org/doc/go1.19), I've noticed some things which may have an impact on Dendrite and may be worth considering when Go 1.19+ becomes the minimum version supported:

- The file descriptor limit increase code added in #2527 can be removed as the Go runtime will automatically increase the file descriptor limit in binaries importing the `os` package:
>On Unix operating systems, Go programs that import package [os](https://tip.golang.org/pkg/os/) now automatically increase the open file limit (RLIMIT_NOFILE) to the maximum allowed value; that is, they change the soft limit to match the hard limit. This corrects artificially low limits set on some systems for compatibility with very old C programs using the [select](https://en.wikipedia.org/wiki/Select_(Unix)) system call. Go programs are not helped by that limit, and instead even simple programs like gofmt often ran out of file descriptors on such systems when processing many files in parallel. One impact of this change is that Go programs that in turn execute very old C programs in child processes may run those programs with too high a limit. This can be corrected by setting the hard limit before invoking the Go program.
- Files with a `unix` build constraint, such as the one with the code mentioned above, can be simplified:
>The build constraint unix is now recognized in //go:build lines. The constraint is satisfied if the target operating system, also known as GOOS, is a Unix or Unix-like system. For the 1.19 release it is satisfied if GOOS is one of aix, android, darwin, dragonfly, freebsd, hurd, illumos, ios, linux, netbsd, openbsd, or solaris. In future releases the unix constraint may match additional newly supported operating systems.
- As per [arewep2pyet](https://arewep2pyet.com/), Dendrite must
>Ensure memory usage of embedded instances is bounded e.g cache sizes.

This means that the new soft memory limit may come in handy:

[...]

PR #2711, opened 2022-09-10
Switch to Go 1.19 as minimum

0x1a8510f2:
## Reasoning

[Go 1.19](https://tip.golang.org/doc/go1.19) has been out for over a month now (02/08/2022) and features some changes which are likely to be quite beneficial here (see https://github.com/matrix-org/dendrite/issues/2564). Perhaps most importantly, the release allows one to set a soft memory limit on the runtime which helps a lot in the P2P usecase.

Other noteworthy changes include:
- Performance improvements (especially on RISC-V)
- Some security improvements
- More stdlib atomic types (no need for `go.uber.org/atomic` anymore)
- Automatic file descriptor limit increase on unix-like systems

Hence, I think it's worth incrementing the minimum supported version to benefit from the above.

As an alternative, I could revert backwards-incompatible changes and just use go 1.19 in CI to benefit from most of this but still allow building with 1.18. That said, I personally think the code removed because of the upgrade warrants incrementing the minimum.

No tests needed beyond what already exists as this doesn't really change the code in any meaningful way.

## Changes Summary

- Replace all usage of Go 1.18 with Go 1.19
- Use debian bullseye over debian stretch
- Remove now-obsolete file descriptor limit increase code (the commit says memory limit because it's 2am and I'm tired)
- Simplify unix build constraints
- Replace usage of `go.uber.org/atomic` with `sync/atomic` and move the former into the indirect deps section in `go.mod`
- Fix a minor typo in log

## Side Note

https://github.com/matrix-org/dendrite/blob/af9a204cc0b997bf0b7fd6ea434a7c04157bdf28/syncapi/storage/postgres/filtering.go#L41-L50

I came across this comment while searching for `1.18`. I assume the comment means the function should be merged with the duplicate below it using generics? Unfortunately, Go generics don't allow for accessing properties, only methods. Should I remove the comment?

## Pull Request Checklist

Signed-off-by: `0x1a8510f2 <admin@0x1a8510f2.space>`


0x1a8510f2:
My git-fu failed me. Gonna re-open

bones-was-here:
Debian stable backports doesn't have 1.19 yet.

0x1a8510f2:
@bones-was-here do you mean, as a reason not to upgrade to 1.19 yet or are you referring to the commit where I switch to using bullseye for building? If the latter, I'm using the container (https://hub.docker.com/_/golang) which does have 1.19.

bones-was-here:
dendrite 0.9.8 builds and works on go 1.18, which is in backports

PR #2713, opened 2022-09-11
Send-to-device consumer/producer tweaks

S7evinK:
Some tweaks for the send-to-device consumers/producers:
- use `json.RawMessage` without marshalling it first
- try further devices (if available) if we failed to `PublishMsg` in the producers
- some logging changes (to better debug E2EE issues)


clientapi/producers/syncapi.go
```diff
@@ -115,15 +111,17 @@ func (p *SyncAPIProducer) SendToDevice(
...
+		m := nats.NewMsg(p.TopicSendToDeviceEvent)
+		m.Data = eventJSON
 		m.Header.Set("sender", sender)
 		m.Header.Set(jetstream.UserID, userID)
+
 		if _, err = p.JetStream.PublishMsg(m, nats.Context(ctx)); err != nil {
-			log.WithError(err).Error("sendToDevice failed t.Producer.SendMessage")
+			if i < len(devices)-1 {
+				log.WithError(err).Warn("sendToDevice failed to PublishMsg, trying further devices")
+				continue
```
neilalexander, line 122: Do we want to do this? It seems like a transient failure in NATS could result in messages being dropped by skipping over them?

S7evinK: Right now they are ALL definitely dropped, if only one device fails to send. If it's the first, that's bad, since we don't try other devices, like with this change.

neilalexander: Fair enough!

PR #2716, opened 2022-09-13
Tweak `LoadMembershipAtEvent` behaviour when state not known

neilalexander:
Previously `LoadMembershipAtEvent` would fail if the state before one of the events was not known, i.e. because it was an outlier. This modifies it so that it gracefully handles not knowing the state and returns no memberships instead, so that history visibility doesn't freak out and kill `/sync` requests dead.

### DIFF
--- a/roomserver/internal/query/query.go
+++ b/roomserver/internal/query/query.go
@@ -237,7 +237,11 @@ func (r *Queryer) QueryMembershipAtEvent(
 	}
 
 	for _, eventID := range request.EventIDs {
-		stateEntry := stateEntries[eventID]
+		stateEntry, ok := stateEntries[eventID]
+		if !ok {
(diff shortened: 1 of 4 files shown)


roomserver/internal/query/query.go
```diff
@@ -237,7 +237,11 @@ func (r *Queryer) QueryMembershipAtEvent(
 	}
 
 	for _, eventID := range request.EventIDs {
-		stateEntry := stateEntries[eventID]
+		stateEntry, ok := stateEntries[eventID]
+		if !ok {
+			response.Memberships[eventID] = []*gomatrixserverlib.HeaderedEvent{}
```
kegsay, line 242: Please update the docs for `QueryMembershipAtEvent` to say that sending back `[]` is possible.

PR #2719, opened 2022-09-14
Bug fix #2718 appservice txnid should be different for each batch of events

### LINKED ISSUE #2718: Appservice incorrectly handling a batch of events with the same transaction ID
### Background information

- **Dendrite version or git SHA**: 0ea948c70548875601a544c57a588271af3adce4 (latest main on 9/14)
- **Running in Docker?**: No
- **`go version`**: go version go1.19 darwin/amd64

### Description
 - **What** is the problem:

I am seeing a regression with appservice. I am working on an appservice that uses the matrix-bot-sdk. My appservice was working prior to this PR that was merged: [Refactor appservices component](https://github.com/matrix-org/dendrite/pull/2687)

Before the PR was merged, my appservice is able to see all events associated with a new space/ room creation. When a space is created, it was able to see all the related events - room.create, room.join, etc. After pulling the latest main, the appservice is no longer able to see the events when a space is created.

I traced it to this logic in appservice/consumers/roomserver.go. In line 178 of the func sendEvents( ... ):

`txnID := events[0].Event.OriginServerTS()`

What's happening is that the batch of events associated with space / room creation has the same OriginServerTS(). Dendrite sends the first event to the appservice. The appservice processes the event, and completes the transaction. The sendEvents func is invoked again with the rest of the events. The txnID is exactly the same as before. Consequently, the appservice thinks that this transaction has been processed, and ignores the events.

 - **Who** is affected:
I am building on matrix-bot-sdk. This regression could affect any projects using that sdk. Didn't try other appservice sdk. 

Configure an appservice (based on matrix-bot-sdk). Create a space or a room. Appservice no longer gets all the associated events with creating the room.

After this PR was merged: [Refactor appservices component](https://github.com/matrix-org/dendrite/pull/2687)

I will create a PR for fixing it. Will gather feedback in the PR.

tak-hntlabs:
See issue: [#2718](https://github.com/matrix-org/dendrite/issues/2718) for more details.

The fix assumes that if the number of transaction items are different, then the txnid should be different. 

txnid := OriginalServerTS()_len(transactions) 

The case that it doesn't address is if the txnid generated this way is the same for 2 different batches of events which have the same OriginalServerTS and the same array length.

Another option:

txnid := OriginalServerTS()_hash(transactions)

Would love to hear other ideas and ways to fix this.

### Pull Request Checklist

* [x ] I have added added tests for PR _or_ I have justified why this PR doesn't need tests.
* [x ] Pull request includes a [sign off](https://github.com/matrix-org/dendrite/blob/main/docs/CONTRIBUTING.md#sign-off)

### DIFF
--- a/appservice/consumers/roomserver.go
+++ b/appservice/consumers/roomserver.go
@@ -174,9 +174,9 @@ func (s *OutputRoomEventConsumer) sendEvents(
 	}
 
 	// If the number of items in the array is different,
-	// then this should be a different transaction. Incorporate the length
+	// then this should be treated as a different transaction. Incorporate the length
 	// of events into the transaction ID.
-	txnID := events[0].Event.OriginServerTS().Time().Add(time.Duration(len(events)))
+	txnID := events[0].Event.OriginServerTS().Time().Nanosecond() + len(transaction)
 
 	// Send the transaction to the appservice.
 	// https://matrix.org/docs/spec/application_service/r0.1.2#put-matrix-app-v1-transactions-txnid
--- a/appservice/consumers/roomserver.go
+++ b/appservice/consumers/roomserver.go
@@ -22,6 +22,7 @@ import (
 	"math"
 	"net/http"
 	"net/url"
+	"strconv"
 	"time"
 
 	"github.com/matrix-org/gomatrixserverlib"
@@ -151,15 +152,25 @@ func (s *OutputRoomEventConsumer) onMessage(
 		return true
 	}
 
+	txnID := ""
+	// Try to get the message metadata, if we're able to, use the timestamp as the txnID
+	if len(msgs) > 0 {
+		metadata, err := msgs[0].Metadata()
+		if err == nil {
+			txnID = strconv.Itoa(int(metadata.Timestamp.UnixNano()))
+		}
+	}
+
 	// Send event to any relevant application services. If we hit
 	// an error here, return false, so that we negatively ack.
 	log.WithField("appservice", state.ID).Debugf("Appservice worker sending %d events(s) from roomserver", len(events))
-	return s.sendEvents(ctx, state, events) == nil
+	return s.sendEvents(txnID, ctx, state, events) == nil
 }
 
 // sendEvents passes events to the appservice by using the transactions
 // endpoint. It will block for the backoff period if necessary.
 func (s *OutputRoomEventConsumer) sendEvents(
+	txnID string,
 	ctx context.Context, state *appserviceState,
 	events []*gomatrixserverlib.HeaderedEvent,
 ) error {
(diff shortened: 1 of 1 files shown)


S7evinK:
As the `TODO` mentions, I'd probably also go for the NATS timestamp, something like 

```go
// If the number of items in the array is different,
// then this should be treated as a different transaction. Incorporate the length
// of events into the transaction ID.
txnID := fmt.Sprintf("%d_%d", events[0].Event.OriginServerTS(), len(events))
// Try to get the message metadata, if we're able to, use the timestamp as the txnID
metadata, err := msgs[0].Metadata()
if err == nil {
	txnID = strconv.Itoa(int(metadata.Timestamp.Unix()))
}
// Send event to any relevant application services. If we hit
// an error here, return false, so that we negatively ack.
return s.sendEvents(ctx, state, events, txnID) == nil
```
in `onMessage`

tak-hntlabs:
> As the `TODO` mentions, I'd probably also go for the NATS timestamp, something like
> 
> ```go
> // If the number of items in the array is different,
> // then this should be treated as a different transaction. Incorporate the length
> // of events into the transaction ID.
> txnID := fmt.Sprintf("%d_%d", events[0].Event.OriginServerTS(), len(events))
> // Try to get the message metadata, if we're able to, use the timestamp as the txnID
> metadata, err := msgs[0].Metadata()
> if err == nil {
> 	txnID = strconv.Itoa(int(metadata.Timestamp.Unix()))
> }
> // Send event to any relevant application services. If we hit
> // an error here, return false, so that we negatively ack.
> return s.sendEvents(ctx, state, events, txnID) == nil
> ```
> 
> in `onMessage`

what is msgs[0].Metadata() ? I don't see anything like that in the sendEvents function.

S7evinK:
> what is msgs[0].Metadata() ? I don't see anything like that in the sendEvents function.

That's in `onMessage`, the metadata of a NATS message. With my example `sendEvents` would get an additional parameter `txnID`, and the end of `onMessage` (after the `if len(events) == 0`-check) would become the example.

tak-hntlabs:
> That's in `onMessage`, the metadata of a NATS message. With my example `sendEvents` would get an additional parameter `txnID`, and the end of `onMessage` (after the `if len(events) == 0`-check) would become the example.

that worked! Thanks!

appservice/consumers/roomserver.go
```diff
@@ -151,15 +152,25 @@ func (s *OutputRoomEventConsumer) onMessage(
 		return true
 	}
 
+	txnID := ""
```
S7evinK, line 155: `msgs` can't be 0 at this point, we already iterated over them and did a `len(events) == 0` check before. So this can be removed.
I'd prefer having
```go
txnID := fmt.Sprintf("%d_%d", events[0].Event.OriginServerTS(), len(transaction))
```
from `sendEvents` here (instead of `txnID := ""`), so it's easier to read.

appservice/consumers/roomserver.go
```diff
@@ -151,15 +152,25 @@ func (s *OutputRoomEventConsumer) onMessage(
...
 	// an error here, return false, so that we negatively ack.
 	log.WithField("appservice", state.ID).Debugf("Appservice worker sending %d events(s) from roomserver", len(events))
-	return s.sendEvents(ctx, state, events) == nil
+	return s.sendEvents(txnID, ctx, state, events) == nil
 }
 
 // sendEvents passes events to the appservice by using the transactions
 // endpoint. It will block for the backoff period if necessary.
 func (s *OutputRoomEventConsumer) sendEvents(
+	txnID string,
```
S7evinK, line 173: Can you move the new parameter to the end?
Having `context.Context` as the first parameter is more idiomatic Go.

S7evinK (changes requested):
Besides the comments, I'm happy with this. :)

appservice/consumers/roomserver.go
tak-hntlabs: done.

appservice/consumers/roomserver.go
S7evinK: Fair enough.

S7evinK (approved):

S7evinK:
Thanks for this!

PR #2720, opened 2022-09-15
Tweak `InsertMigration` to avoid logging

S7evinK:
We'd still produce logs in Postgres when trying to insert a migration we already ran. This should stop us from creating those log entries.

### DIFF
--- a/internal/sqlutil/migrate.go
+++ b/internal/sqlutil/migrate.go
@@ -55,7 +55,7 @@ type Migrator struct {
 	migrations      []Migration
 	knownMigrations map[string]struct{}
 	mutex           *sync.Mutex
-	provider        execProvider
+	insertStmt      *sql.Stmt
 }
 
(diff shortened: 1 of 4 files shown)


internal/sqlutil/migrate.go
```diff
@@ -82,42 +84,46 @@ func (m *Migrator) AddMigrations(migrations ...Migration) {
...
-		dendriteVersion = internal.VersionString()
-	)
 	// ensure there is a table for known migrations
 	executedMigrations, err := m.ExecutedMigrations(ctx)
 	if err != nil {
 		return fmt.Errorf("unable to create/get migrations: %w", err)
 	}
 
 	return WithTransaction(m.db, func(txn *sql.Tx) error {
+		m.provider = txn
```
neilalexander, line 94: This seems like it would leak?

neilalexander (approved):

PR #2722, opened 2022-09-15
Verify `msgtype` is set on `m.room.message` events

S7evinK:
https://spec.matrix.org/v1.3/client-server-api/#mroommessage
"Discovered" because I forgot to set the `msgtype` in Complement, Dendrite was happy, Synapse failed the tests because of the missing `msgtype`

### DIFF
--- a/clientapi/routing/sendevent.go
+++ b/clientapi/routing/sendevent.go
@@ -120,6 +120,16 @@ func SendEvent(
 		delete(r, "join_authorised_via_users_server")
 	}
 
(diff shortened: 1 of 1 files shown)


tulir:
If you want to do unnecessary validation, shouldn't you also check that `body` exists?

S7evinK:
In which way unnecessary?

neilalexander:
Closing because we don't think it's necessary for Dendrite to enforce this.

PR #2725, opened 2022-09-16
Ensure that all state event IDs are included in the `added` section when rewriting state

neilalexander:
This should hopefully fix an entire class of problems where components downstream from the roomserver (i.e. the sync API) could just lose a whole bunch of state after a rewrite operation like a federated join.

The root of the bug is that we set `RewritesState` in the output event which instructs downstream components to purge their copy of any room state, but then didn't send the entire state snapshot in `adds_state_event_ids` so the downstream state ends up being incomplete as a result.


kegsay (approved):
Whoops, can't believe we missed this one. Can we add a test for it please?

neilalexander:
Will PR a test separately.

PR #2740, opened 2022-09-27
Return `M_UNRECOGNIZED` for unknown CS API endpoints/actions

### LINKED ISSUE #2739: Unrecognized APIs yield a non-JSON error response
### Description

 - **What** is the problem: Dendrite returns `404 page not found` error instead of a Matrix JSON `M_UNRECOGNIZED` errcode
 - **Who** is affected: Anyone using features which Dendrite does not yet support
 - **How** is this bug manifesting: 404 response with non-JSON response
 - **When** did this first appear: Presumably day 0

### Steps to reproduce

- Spin up a dendrite without msc2946 enabled
- `curl 'https://$DENDRITE/_matrix/client/v1/rooms/roomId/hierarchy'`

Synapse in comparison returns `{"errcode":"M_UNRECOGNIZED","error":"Unrecognized request"}` for unknown APIs

neilalexander:
Fixes #2739.

### DIFF
--- a/setup/base/base.go
+++ b/setup/base/base.go
@@ -392,17 +392,26 @@ func (b *BaseDendrite) configureHTTPErrors() {
 		_, _ = w.Write([]byte(fmt.Sprintf("405 %s not allowed on this endpoint", r.Method)))
 	}
 
+	clientNotFoundHandler := func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusNotFound)
+		w.Header().Set("Content-Type", "application/json")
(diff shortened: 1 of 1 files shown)


S7evinK (approved):

richvdh:
Was doing this only for the CS API, and not the Federation API, a deliberate decision?

neilalexander:
> Was doing this only for the CS API, and not the Federation API, a deliberate decision?

Not especially, just that it wasn't clear whether it makes any difference over federation whereas it obviously does seem to matter to clients. We could just as easily do it for federation too.

richvdh:
It would definitely be helpful for federation; we have the same hacks there.

Might want to see where MSC3743 goes though.

PR #2743, opened 2022-09-27
Fix for `sql: converting argument $1 type: unsupported type []interfa…

texuf:
…ce {}, a slice of interface` in new notifications select

The sqlite3 version was just not working, original pr here: https://github.com/matrix-org/dendrite/pull/2688

signed off by: austin ellis <austin@hntlabs.com>

This doesn't fix the notification counts, they still only work about 1 out of every 5 times in my tests. I will stick with my other fix locally for reliable notification delivery: https://github.com/matrix-org/dendrite/pull/2701


S7evinK (approved):
Oh, good catch! Thanks!
Really wondering why its not working for you. Is syncing not working (e.g. the counts are not shown) or the push delivery to mobile devices? Maybe good to create an issue for that.

PR #2744, opened 2022-09-28
Consistent `*sql.Tx` usage across sync API

neilalexander:
This tidies up the `storage` package so that everything takes a transaction parameter instead of something things that do and some that don't.

### DIFF
--- a/syncapi/storage/sqlite3/notification_data_table.go
+++ b/syncapi/storage/sqlite3/notification_data_table.go
@@ -90,8 +90,13 @@ func (r *notificationDataStatements) SelectUserUnreadCountsForRooms(
 	for i := range roomIDs {
 		params[i+1] = roomIDs[i]
 	}
-	sql := strings.Replace(selectUserUnreadNotificationsForRooms, "($2)", sqlutil.QueryVariadicOffset(len(roomIDs), 1), 1)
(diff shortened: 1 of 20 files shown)


syncapi/storage/sqlite3/notification_data_table.go
```diff
@@ -90,8 +90,13 @@ func (r *notificationDataStatements) SelectUserUnreadCountsForRooms(
 	for i := range roomIDs {
 		params[i+1] = roomIDs[i]
 	}
-	sql := strings.Replace(selectUserUnreadNotificationsForRooms, "($2)", sqlutil.QueryVariadicOffset(len(roomIDs), 1), 1)
-	rows, err := r.db.QueryContext(ctx, sql, params...)
+	sql := strings.Replace(selectUserUnreadNotificationsForRooms, "($1)", sqlutil.QueryVariadic(len(params)), 1)
```
S7evinK, line 93: `$2` is the variadic one here, so we need `QueryVariadicOffset` as well.

S7evinK (approved):

syncapi/storage/sqlite3/notification_data_table.go
neilalexander: Ah, looks like this got reverted on my branch somehow.

PR #2745, opened 2022-09-28
Transactional isolation for `/sync`

neilalexander:
This should transactional snapshot isolation for `/sync` etc requests.

For now we don't use repeatable read due to some odd test failures with invites.


syncapi/routing/messages.go
```diff
@@ -70,6 +72,13 @@ func OnIncomingMessagesRequest(
 ) util.JSONResponse {
 	var err error
 
+	snapshot, err := db.NewDatabaseTransaction(req.Context())
```
kegsay, line 75: Why is this using `NewDatabaseTransaction` but /context uses `NewDatabaseSnapshot`?

syncapi/storage/shared/syncserver.go
```diff
@@ -1036,17 +538,6 @@ func (d *Database) UpsertRoomUnreadNotificationCounts(ctx context.Context, userI
 	return
 }
 
-func (d *Database) GetUserUnreadNotificationCountsForRooms(ctx context.Context, userID string, rooms map[string]string) (map[string]*eventutil.NotificationData, error) {
```
kegsay, line 1039: So am I right in saying that all DB functions which are used by the `/sync` side are now in `transaction.go` and all DB functions used by the roomserver consumer remain here?

kegsay (approved):
Overall looks like a good improvement. I worry that we will leak transactions now we leave them hanging after making a snapshot. It would be nice to make it impossible to forget to `defer Rollback()` e.g similar to `WithTransaction`.

syncapi/routing/messages.go
neilalexander: I'll add a comment about this, but the main idea is that `NewDatabaseSnapshot` is supposed to be read-only and repeatable read, whereas `NewDatabaseTransaction` isn't. Since `/messages` can backfill and store events directly in the sync API database, it needs to be able to write.

syncapi/storage/shared/syncserver.go
neilalexander: That's pretty much it, yes.

PR #2746, opened 2022-09-28
Add `/_dendrite/admin/refreshDevices/{userID}`

S7evinK:
Allows to immediately query `/devices/{userID}` over federation to (hopefully) resolve E2EE issues.

### DIFF
--- a/clientapi/routing/admin.go
+++ b/clientapi/routing/admin.go
@@ -2,20 +2,23 @@ package routing
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"time"
 
 	"github.com/gorilla/mux"
-	"github.com/matrix-org/dendrite/clientapi/jsonerror"
-	"github.com/matrix-org/dendrite/internal/httputil"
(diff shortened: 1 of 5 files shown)


neilalexander (commented):
I realise internally we refer to it as “marking as stale” but it might be better for this endpoint to be named something that explains what it does, like “refresh device list”.

S7evinK:
Right, had the same thought after having closed my notebook.

neilalexander (approved):

Mic92:
Do I need to put in my own userid or the other persons userid in the http endpoint?

S7evinK:
> Do I need to put in my own userid or the other persons userid in the http endpoint?

Of the other user, the userID you want to refresh. If you try to use a local userID, you should get an error message.

Mic92:
Ok. If I have a room, should refreshing keys for specified user also help with an encrypted room work or is room encryption completely disconnected from this? Because I tried this endpoint and the other person was still not able to read my messages.

PR #2749, opened 2022-09-29
Allow more time for device list updates

neilalexander:
This updates the device list updater so that it has a context per-request, rather than a global 30 seconds for the entire server. This could mean that talking to a slow remote server or requesting a lot of user IDs was pretty much guaranteed to fail.

It also uses the process context to allow correct cancellation when Dendrite wants to shut down cleanly.


S7evinK (approved):
LGTM!

PR #2755, opened 2022-10-02
Remove empty fields from `/sync` response

S7evinK:
First attempt at removing empty fields from `/sync` responses. Needs https://github.com/matrix-org/sytest/pull/1298 to keep Sytest happy.


syncapi/streams/stream_typing.go
```diff
@@ -35,9 +36,9 @@ func (p *TypingStreamProvider) IncrementalSync(
 			continue
 		}
 
-		jr := *types.NewJoinResponse()
-		if existing, ok := req.Response.Rooms.Join[roomID]; ok {
-			jr = existing
+		jr, ok := req.Response.Rooms.Join[roomID]
+		if !ok {
+			jr = *types.NewJoinResponse()
```
neilalexander, line 41: Dereferencing in these places feels weird (I realise it's not a thing you added, just that it's been that way for a while and feels strange).

Maybe we can just update `RoomsResponse` so they are pointers instead and remove the dereferencing/copying?

neilalexander (approved):
LGTM if CI happy.

PR #2757, opened 2022-10-02
Fix fragility of selectEventsWithEventIDsSQL queries

### LINKED ISSUE #601: Fix database errors
Currently, we have two occurrences of `sql: expected x destination arguments in Scan, not y` runtime errors popping up:

- [x] one when querying the CS API's `GET /devices` endpoint
- [ ] one in the syncapi's `rowsToStreamEvents` function

Both should be quite trivial to fix as the former should be pretty straightforward and the latter on require a small bit of refactoring.

fantashley:
This fixes a temporary workaround with the `selectEventsWithEventIDsSQL` queries where fields need to be artificially added to the queries so the row results match the format of the `syncapi_output_room_events` table. I made similar functions that accept row results from the `syncapi_current_room_state` table and convert them into StreamEvents without the fields that are specific to output room events.

There is also a unit test in the first commit to ensure the resulting behavior doesn't change from the modified queries and functions.

Fixes #601.

### Pull Request Checklist

* [x] I have added tests for PR _or_ I have justified why this PR doesn't need tests.

Signed-off-by: `Ashley Nelson <fant@shley.email>`

### DIFF
--- a/syncapi/storage/postgres/current_room_state_table.go
+++ b/syncapi/storage/postgres/current_room_state_table.go
@@ -104,12 +104,7 @@ const selectStateEventSQL = "" +
 	"SELECT headered_event_json FROM syncapi_current_room_state WHERE room_id = $1 AND type = $2 AND state_key = $3"
 
 const selectEventsWithEventIDsSQL = "" +
-	// TODO: The session_id and transaction_id blanks are here because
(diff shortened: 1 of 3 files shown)


neilalexander (approved):
LGTM, thanks!

PR #2759, opened 2022-10-04
Add utility command to generate schema for config

balki:
Usage:
``` sh
❯ ./bin/generate-jsonschema > schema.yaml                 
❯ ./bin/generate-config --schema ./schema.yaml > config.yaml
```
Now `config.yaml` has IDE¹ support with validation and documentation² built in
image
image

¹ IDEs with language server support. Tested with vscode(with YAML plugin) and VIM (vim-lsc plugin with [yaml-language-server](https://github.com/redhat-developer/yaml-language-server))
² Documentation is picked up from go comments for the struct member

Signed-off-by: `Balakrishnan Balasubramanian <git.user@balki.me>`


balki:
@neilalexander @S7evinK Do you like this idea to add schema? I have found this helpful to understand and write config easily. Let me know if you have any feedback and if you are going forward with this PR.

balki:
Looks like it does not work if not run from the source code (to pick up the comments). May be can be used in CI to generate schema automatically if interested. Closing now.

PR #2763, opened 2022-10-04
Side effect import bleve analyzer languages

### LINKED ISSUE #2761: Fulltext search doesn't find analyzer for any other lang than en
### Background information

- **Dendrite version or git SHA**:  0.10.1+b17b111.loginsso
- **Monolith or Polylith?**:  Monolith
- **SQLite3 or Postgres?**:  Postgres
- **Running in Docker?**:  Kubernetes
- **`go version`**: 1.19

### Description

 - **What** is the problem: Using the new fulltext search with language set as "fi" results in an error.
 - **Who** is affected: My whole Dendrite server
 - **How** is this bug manifesting: The server panics on startup
 - **When** did this first appear: After enabling the feature

### Steps to reproduce

- Install Dendrite 0.10.1
- Set it up with search using any other language than en
- Have it panic on boot

<!--

``` (three backticks, on a line on their own), so that they are formatted legibly.

-->

S7evinK:
... to actually allow different languages. Fixes #2761 
Binary size increases by ~1MB.


neilalexander (approved):

neilalexander:
I wonder if this increases the binary size significantly?

PR #2765, opened 2022-10-05
Add `RoomEventType` nats.Header to avoid unneeded unmarshalling

S7evinK:


syncapi/consumers/roomserver.go
```diff
@@ -93,6 +93,23 @@ func (s *OutputRoomEventConsumer) Start() error {
 // sync stream position may race and be incorrectly calculated.
 func (s *OutputRoomEventConsumer) onMessage(ctx context.Context, msgs []*nats.Msg) bool {
 	msg := msgs[0] // Guaranteed to exist if onMessage is called
+
+	receivedType := api.OutputType(msg.Header.Get(jetstream.RoomEventType))
+	ok := false
+	// Only handle events we care about
+	for _, eventType := range api.OutputTypes {
```
neilalexander, line 100: I'd be surprised if we were receiving completely unknown output event types, so in that case `api.OutputTypes` and this loop are probably redundant.

S7evinK: The SyncAPI should receive all events, right.

roomserver/api/output.go
```diff
@@ -21,6 +21,18 @@ import (
 // An OutputType is a type of roomserver output.
 type OutputType string
 
+// OutputTypes contains all possible output types from below
+var OutputTypes = []OutputType{
```
neilalexander, line 25: This is now dead code, can we remove it until we have a need for it?

neilalexander (approved):

PR #2767, opened 2022-10-05
Don't allow inviting users that don't exist

neilalexander:
... otherwise some new users might be in for a surprise when they register a name that has previously been invited.


S7evinK (approved):

kegsay:
This won't work for appservices or 3PID invites where the user doesn't yet exist. It also lacks tests.

PR #2779, opened 2022-10-07
Add Monolith config options for HTTP(s) Bind Addresses

BtbN:
It so far wasn't possible to specify the monoliths bind address(es) via the config file, only via command line options.
Pretty straight forward implementation. If the CLI option is set, it takes precedence over the config one in all cases.
Behaviour should be unchanged, even if the options are unset, given that the defaults match up with the old cli option defaults.

### DIFF
--- a/cmd/dendrite-monolith-server/main.go
+++ b/cmd/dendrite-monolith-server/main.go
@@ -46,6 +46,9 @@ func main() {
 	cfg := setup.ParseFlags(true)
 	httpAddr := cfg.Global.Monolith.HTTPBindAddr
 	httpsAddr := cfg.Global.Monolith.HTTPBindAddr
+	certPath := string(cfg.Global.Monolith.TlsCertificatePath)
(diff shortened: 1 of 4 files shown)


spaetz:
Changes look good to this non-dev. My only question is: What happens if the key is not existing in the config file? Currently there is no fallback to the default ports, right? Would it balk out?

I think, we should still use the default ports if nothing is configured...

BtbN:
My understanding is that the defaults from config_monolith.go propagate through in that case.
Though I'm not 100% familiar with the config mechanism used there, that was how it behaved when I tested it.

BtbN:
Only the internal API listener is left now, but that can probably safely stay a CLI-Only option.

BtbN:
Excuse me?
This is a quite central configuration option, and not exposing it in the config file is a major flaw.
It forces me to fork the systemd service and put configuration in there, splitting the config into two different places and making config management extremely annoying and inconsistent.

The configuration already exists. It's just in a very annoying place.

PR #2782, opened 2022-10-10
Move /event to the SyncAPI

S7evinK:
This allows us to apply history visibility without having to recalculate it in the roomserver.
Unblocks https://github.com/matrix-org/complement/pull/495, fix missing part of https://github.com/matrix-org/dendrite/issues/617


neilalexander (approved):
Please also update the `docs/nginx/`, `docs/hiawatha/` and `docs/caddy/` sample configs.

PR #2783, opened 2022-10-10
Repeatable reads for `/sync`

neilalexander:
This puts repeatable reads into all sync streams.


S7evinK (approved):

kegsay:
LGTM assuming CI passes.

PR #2791, opened 2022-10-12
Fix `outliers whose auth_events are in a different room are correctly rejected`

S7evinK:
Fixes `outliers whose auth_events are in a different room are correctly rejected`, by validating that auth events are all from the same room and not using rejected events for event auth.

### DIFF
--- a/roomserver/internal/helpers/auth.go
+++ b/roomserver/internal/helpers/auth.go
@@ -19,10 +19,11 @@ import (
 	"fmt"
 	"sort"
 
+	"github.com/matrix-org/gomatrixserverlib"
+
 	"github.com/matrix-org/dendrite/roomserver/state"
 	"github.com/matrix-org/dendrite/roomserver/storage"
 	"github.com/matrix-org/dendrite/roomserver/types"
-	"github.com/matrix-org/gomatrixserverlib"
 )
 
 // CheckForSoftFail returns true if the event should be soft-failed
@@ -129,6 +130,14 @@ type authEvents struct {
 	stateKeyNIDMap map[string]types.EventStateKeyNID
 	state          stateEntryMap
 	events         EventMap
+}
+
+func (ae *authEvents) Valid() bool {
+	roomIDs := make(map[string]struct{})
+	for _, ev := range ae.events {
+		roomIDs[ev.RoomID()] = struct{}{}
+	}
+	return len(roomIDs) <= 1
 }
 
 // Create implements gomatrixserverlib.AuthEventProvider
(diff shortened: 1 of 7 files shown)


roomserver/internal/helpers/auth.go
```diff
@@ -131,6 +132,14 @@ type authEvents struct {
 	events         EventMap
 }
 
+func (ae *authEvents) Valid() bool {
+	roomIDs := make(map[string]struct{})
+	for _, ev := range ae.events {
+		roomIDs[ev.RoomID()] = struct{}{}
+	}
+	return len(roomIDs) <= 1
```
kegsay, line 140: This is inefficient as we'll loop through all events regardless. Just remember the first room id and return false if any event isn't that.

roomserver/internal/input/input_events_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"testing"
+
+	"github.com/matrix-org/gomatrixserverlib"
+
+	"github.com/matrix-org/dendrite/test"
+)
+
+func Test_1(t *testing.T) {
```
kegsay, line 11: No name?

roomserver/internal/input/input_events_test.go
```diff
@@ -0,0 +1,63 @@
...
+			authEvents = append(authEvents, x.Event)
+		}
+		if x.Type() == gomatrixserverlib.MRoomJoinRules {
+			authEventIDs = append(authEventIDs, x.EventID())
+			authEvents = append(authEvents, x.Event)
+		}
+	}
+
+	// Add the illegal auth event from room1
+	for _, x := range room1.Events() {
```
kegsay, line 39: You fail to explain why it's illegal (different room)

roomserver/internal/helpers/auth.go
neilalexander: Where is this function called from, and do we know the room ID at that point?

neilalexander (approved):
Let's see how it goes.

roomserver/internal/helpers/auth.go
```diff
@@ -229,6 +226,16 @@ func loadAuthEvents(
 	if result.events, err = db.Events(ctx, eventNIDs); err != nil {
 		return
 	}
+	roomID := ""
+	for _, ev := range result.events {
+		if ev.RoomID() != "" {
```
neilalexander, line 231: Presumably this is meant to be `if roomID == "" { ... }` so that you only set `roomID` once?

Issue #2771, opened 2022-10-06
Dendrite advertises non-existent version "v1.0"


turt2live:
v1.0 is symbolic and doesn't actually exist - it'll be some combination of the `r0.X` series instead.

PR #2797, opened 2022-10-15
Advertize spec 1.3 & 1.4 compatibility

spaetz:
Removing the nonexisting 1.0 at the same time. I would not know what to test or how to test things though.
This is needed to support the private read markers by clients.

Fixes: https://github.com/matrix-org/dendrite/issues/2771

Signed-off-by: Sebastian Spaeth <Sebastian@SSpaeth.de>


kegsay:
I don't think Dendrite does everything for v1.4, `m.rule.room.server_acl` doesn't exist, no CORP header on mediaapi, etc https://spec.matrix.org/v1.5/changelog/#v14

Do we even support private read markers? This would need routing changes to only send that kind of read marker to that user ID and no one else.

kegsay (changes requested):
I'm okay with removing `v1.0` but the others should not be added, we are lying otherwise.

spaetz:
Given that this PR is just about removing 1.0 from the list of advertized spec versions, closing. It can be folded into a PR when we are really changing the advertized specs.

PR #2806, opened 2022-10-17
Namespace user API tables

neilalexander:
This migrates all the various user API tables, indices and sequences to be `userapi_`-namespaced, rather than the mess they are all now.


S7evinK (approved):
LGTM!

PR #2823, opened 2022-10-23
Add support for config "auto_join_rooms"

Neboer:
Add support for config "auto_join_rooms". Now new accounts can join the rooms in config file automatically.

### Pull Request Checklist

* [x] Pull request includes a [sign off below using a legally identifiable name](https://matrix-org.github.io/dendrite/development/contributing#sign-off)

Signed-off-by: `Rubin Poster <rubinposter@gmail.com>`

### DIFF
--- a/clientapi/routing/register.go
+++ b/clientapi/routing/register.go
@@ -699,7 +699,8 @@ func handleGuestRegistration(
 // handleRegistrationFlow will direct and complete registration flow stages
 // that the client has requested.
 // nolint: gocyclo
-func handleRegistrationFlow(req *http.Request,
+func handleRegistrationFlow(
+	req *http.Request,
 	r registerRequest,
 	sessionID string,
 	cfg *config.ClientAPI,
@@ -774,7 +775,8 @@ func handleRegistrationFlow(req *http.Request,
 	// Check if the user's registration flow has been completed successfully
 	// A response with current registration flow and remaining available methods
 	// will be returned if a flow has not been successfully completed yet
-	return checkAndCompleteFlow(sessions.getCompletedStages(sessionID), req, r, sessionID, cfg, userAPI)
+	return checkAndCompleteFlow(sessions.getCompletedStages(sessionID),
+		req, r, sessionID, cfg, userAPI)
 }
 
 // handleApplicationServiceRegistration handles the registration of an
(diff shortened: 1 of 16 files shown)


dendrite-sample.monolith.yaml
```diff
@@ -181,6 +181,9 @@ client_api:
   recaptcha_bypass_secret: ""
   recaptcha_siteverify_api: ""
 
+#  auto_join_rooms:
+#    - "#main:matrix.org"
+
```
S7evinK, line 186: Please add some comments as to what is happening with this config. Also, this should be added to the polylith example config.
It should probably be noted which rooms can be added here, as invite only rooms are probably going to cause issues.

clientapi/routing/register.go
```diff
@@ -825,15 +827,25 @@ func handleApplicationServiceRegistration(
...
+	clientRsApi rsAPI.ClientRoomserverAPI) util.JSONResponse {
+
 	if checkFlowCompleted(flow, cfg.Derived.Registration.Flows) {
+		// POST register behavior: add user to room
+		for room := range cfg.AutoJoinRooms {
+			err := addUserToRoom(req.Context(), clientRsApi, cfg.AutoJoinRooms[room], r.Username,
+				userutil.MakeUserID(r.Username, cfg.Matrix.ServerName))
+			if err != nil {
+				log.Fatal(err)
+				return util.JSONResponse{Code: http.StatusInternalServerError,
```
S7evinK, line 845: Please don't `log.Fatal`, `log.WithError(err).Error("failed to auto-join room")` instead (possibly with information about  which room and user?).
Also I don't think we should fail the registration, only because we couldn't auto-join a room.

S7evinK (changes requested):
In its current form, this is not going to work as expected.
`addUserToRoom` is executed before the user account is even created (happens in `completeRegistration`), so it should be deferred until after that call.

As mentioned in the comments, it should be noted which kind of rooms can be used for auto-joining, as invite only rooms are probably going to cause issues.

Neboer:
I've updated my code, Please check my new commit.

clientapi/routing/register.go
```diff
@@ -945,6 +973,17 @@ func completeRegistration(
 	}
 	sessions.addCompletedRegistration(sessionID, result)
 
+	defer func() {
```
S7evinK, line 976: I think we can remove the `defer` at this point (even though the defer makes the `post register behavior` clearer).
We should check that `accType` is not `AccountTypeAppService` and/or that `appserviceID` is empty, not sure how appservices react if they are automatically joined to rooms on registration.

Neboer: So what I need is to remove the defer func wrapper and add a check of `accType` and `appserviceID` ?
```
// POST register behavior: check if the user is a normal user.
// If the user is a normal user, add user to room specified in the configuration "auto_join_rooms".
if accType != userapi.AccountTypeAppService || appserviceID == "" {
	for room := range cfg.AutoJoinRooms {
        	...
	}
}
```

S7evinK: Yep, thinking about it, it's probably better to use `&&`, but that looks fine otherwise.

Neboer:
I've completed it. Please check :)

S7evinK (approved):
LGTM! Thanks for this!

PR #2829, opened 2022-10-25
Initial support for multiple server names

neilalexander:
This PR is the first step towards virtual hosting by laying the groundwork for multiple server names being configured.


clientapi/routing/profile.go
```diff
@@ -261,6 +263,12 @@ func updateProfile(
 		return jsonerror.InternalServerError(), err
 	}
 
+	_, domain, err := gomatrixserverlib.SplitID('@', userID)
```
kegsay, line 266: This is untrusted, it isn't from a Device. Need to check if the domain is valid.

federationapi/federationapi_test.go
```diff
@@ -164,6 +164,7 @@ func TestFederationAPIJoinThenKeyUpdate(t *testing.T) {
 func testFederationAPIJoinThenKeyUpdate(t *testing.T, dbType test.DBType) {
 	base, close := testrig.CreateBaseDendrite(t, dbType)
 	base.Cfg.FederationAPI.PreferDirectFetch = true
+	base.Cfg.FederationAPI.KeyPerspectives = nil
```
kegsay, line 167: Why?

federationapi/routing/routing.go
```diff
@@ -479,14 +479,14 @@ func Setup(
...
-		"federation_keys_claim", cfg.Matrix.ServerName, keys, wakeup,
+		"federation_keys_claim", cfg.Matrix.ServerName, cfg.Matrix.IsLocalServerName, keys, wakeup,
 		func(httpReq *http.Request, request *gomatrixserverlib.FederationRequest, vars map[string]string) util.JSONResponse {
 			return ClaimOneTimeKeys(httpReq, request, keyAPI, cfg.Matrix.ServerName)
 		},
 	)).Methods(http.MethodPost)
 
 	v1fedmux.Handle("/user/keys/query", MakeFedAPI(
-		"federation_keys_query", cfg.Matrix.ServerName, keys, wakeup,
+		"federation_keys_query", cfg.Matrix.ServerName, cfg.Matrix.IsLocalServerName, keys, wakeup,
```
kegsay, line 489: How does this work given we are passing in a single server name for the destination?

userapi/internal/api.go
```diff
@@ -131,6 +131,11 @@ func (a *UserInternalAPI) setFullyRead(ctx context.Context, req *api.InputAccoun
 }
 
 func (a *UserInternalAPI) PerformAccountCreation(ctx context.Context, req *api.PerformAccountCreationRequest, res *api.PerformAccountCreationResponse) error {
+	serverName := req.ServerName
```
kegsay, line 134: We should probably validate this with the config as a sanity check at internal API boundaries like this.

userapi/internal/api.go
```diff
@@ -530,8 +544,13 @@ func (a *UserInternalAPI) queryAppServiceToken(ctx context.Context, token, appSe
 
 // PerformAccountDeactivation deactivates the user's account, removing all ability for the user to login again.
 func (a *UserInternalAPI) PerformAccountDeactivation(ctx context.Context, req *api.PerformAccountDeactivationRequest, res *api.PerformAccountDeactivationResponse) error {
+	serverName := req.ServerName
+	if serverName == "" {
+		serverName = a.Config.Matrix.ServerName
```
kegsay, line 549: Likewise, sanity check please.

kegsay (approved):

federationapi/federationapi_test.go
neilalexander: Mostly because I noticed the various key tests were configuring perspective servers, which just seemed surprising to me.

federationapi/routing/routing.go
neilalexander: We pass in both the *default* server name and the check function into `MakeFedAPI`. If the `X-Matrix` header contains a `destination=` field then we will match it using `cfg.Matrix.IsLocalServerName`. If there is no `destination=` then we use the given server name as the default value.

userapi/internal/api.go
neilalexander: Yeah, these little bits in the user API are just stubs to keep things working until we've got columns in the tables to store server names in.

PR #2831, opened 2022-10-26
Implement `/thirdparty` endpoints

S7evinK:
Implements the following endpoints 
```
GET /_matrix/client/v3/thirdparty/protocols
GET /_matrix/client/v3/thirdparty/protocols/{protocol}
GET /_matrix/client/v3/thirdparty/location
GET /_matrix/client/v3/thirdparty/location/{protocol}
GET /_matrix/client/v3/thirdparty/user
GET /_matrix/client/v3/thirdparty/user/{protocol}
```

### DIFF
--- a/appservice/api/query.go
+++ b/appservice/api/query.go
@@ -19,11 +19,13 @@ package api
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
+
+	"github.com/matrix-org/gomatrixserverlib"
 
 	"github.com/matrix-org/dendrite/clientapi/auth/authtypes"
 	userapi "github.com/matrix-org/dendrite/userapi/api"
-	"github.com/matrix-org/gomatrixserverlib"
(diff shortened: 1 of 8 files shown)


neilalexander (approved):

neilalexander:
Looks like there are newly passing tests for `sytest-whitelist`.

S7evinK:
Yep, but I'm not yet happy with the cache, as it is only refreshed when calling `/protocol` without a protocol.

S7evinK:
Oh, and this would need https://github.com/matrix-org/sytest/pull/1310, which currently breaks the tests for Synapse.
Even with the fallback implemented, the tests fail due to `Destroying unresponded HTTP request to /appservs/1/_matrix/app/v1/thirdparty/protocol/ymca`

neilalexander (approved):

PR #2819, opened 2022-10-20
Fix moderncsqlite `database is locked` errors

0x1a8510f2:
As [promised](https://matrix.to/#/!RcWPWcZrMeBxOGaalX:matrix.org/$YT4PMrOLwWNU7_CoDBmMYxWUlVAd2yATAELyZyXmKEk?via=0x1a8510f2.space&via=matrix.org&via=envs.net) :D

Turns out the fix was simple enough, based on https://gitlab.com/cznic/sqlite/-/issues/106#note_1058094993.

~~I've also updated all dependencies (except `gomatrixserverlib` as it was incompatible) before coming across this solution, hoping that maybe that would help. It didn't, but I left the commit there since it seemed to work fine and up-to-date deps are always a good thing. That said, I'm happy to roll that back if you prefer, since it's not really relevant to this PR.~~

### Pull Request Checklist

* [ ] I have added tests for PR _or_ I have justified why this PR doesn't need tests.
* [x] Pull request includes a [sign off below using a legally identifiable name](https://matrix-org.github.io/dendrite/development/contributing#sign-off) _or_ I have already signed off privately

(signed off privately)

PR #2832, opened 2022-10-26
Fix `moderncsqlite` errors and rebase onto `main`

0x1a8510f2:
This is #2819 but rebased on latest `main`. This PR is against main too as opposed to the `moderncsqlite` branch.

The main change here is simply:

```go
// add query parameters to the dsn
if strings.Contains(dsn, "?") {
	dsn += "&"
} else {
	dsn += "?"
}

// wait some time before erroring if the db is locked
// https://gitlab.com/cznic/sqlite/-/issues/106#note_1058094993
dsn += "_pragma=busy_timeout%3d10000"
```

### Pull Request Checklist

* [x] I have added tests for PR _or_ I have justified why this PR doesn't need tests.

Signed off privately.


neilalexander:
I'm going to try producing a couple of P2P Demo builds with this branch. If they work fine then I'm happy to merge!

0x1a8510f2:
Awesome!

>One thing I'm aware of is that we might break OpenBSD etc users, since I don't think modernc.org/sqlite will support those. I wonder if we can build-tag this somehow.

I was gonna say, this should be simple enough and I could take a look at it over the weekend, but it looks like it was so simple you did it before I got a chance to reply :P

So it looks like all that's left to do is to set `CGO_ENABLED=false` in all CI builds except OpenBSD.

neilalexander:
OK, I've updated the branch to use `modernc.org/sqlite` if `CGO_ENABLED=0` and to continue using `github.com/mattn/go-sqlite3` if `CGO_ENABLED=1`. I've also updated GHA to test both with Sytest and Complement. Things look OK.

PR #2833, opened 2022-10-26
Associate events in db before queueing them to send

devonh:
Fixes a race condition between sending federation events and having them fully associated in the database.

### DIFF
--- a/federationapi/queue/queue.go
+++ b/federationapi/queue/queue.go
@@ -247,7 +247,7 @@ func (oqs *OutgoingQueues) SendEvent(
 		return fmt.Errorf("sendevent: oqs.db.StoreJSON: %w", err)
 	}
 
-	destQueues := []*destinationQueue{}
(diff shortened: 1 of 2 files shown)


federationapi/queue/queue.go
```diff
@@ -247,9 +247,10 @@ func (oqs *OutgoingQueues) SendEvent(
 		return fmt.Errorf("sendevent: oqs.db.StoreJSON: %w", err)
 	}
 
+	destQueues := []*destinationQueue{}
```
neilalexander, line 250: `destQueues := make([]*destinationQueue, 0, len(destmap))` is a good way to prevent reallocations here.

neilalexander (approved):

federationapi/queue/queue.go
neilalexander: Same in `SendEDU` too.

devonh: Good catch

Issue #2838, opened 2022-10-27
bob loses membership to room when he accepts a new invite for a new room


tak-hntlabs:
### Background information

- **Dendrite version or git SHA**: a2706e6498287a5b052ef47413175bf7551b36b1
- **Monolith or Polylith?**: Monolith
- **SQLite3 or Postgres?**: Postgres
- **Running in Docker?**: No
- **`go version`**: 1.19.2
- **Client used (if applicable)**:

### Description

- **What** is the problem: 

bob is invited to multiple spaces. Bob accepted the invites. Found that his membership was changed to 'Leave' in the earlier spaces that he had just joined.

- **Who** is affected:

Users who receive multiple invites, and accepted the invites to join the spaces/ rooms.

- **How** is this bug manifesting:

Users' membership changed to 'Leave' in the room that they have just accepted. In order for the bug to manifest, the users must accept another invite.

- **When** did this first appear:
Not sure. Likely https://github.com/matrix-org/dendrite/pull/2817

### Steps to reproduce

- Alice creates roomA and roomB
- Alice sends both invites to bob
- Bob starts accepting the invites
- Bob loses access to roomA when he accepted the invite for roomB

PR with debug notes: https://github.com/matrix-org/dendrite/pull/2839

tak-hntlabs:
fix merge to main.

PR #2839, opened 2022-10-27
Fix issue where a member is forced to leave a room when the invite is marked deleted

tak-hntlabs:
Proposed fix for issue: https://github.com/matrix-org/dendrite/issues/2838

Suppose bob received invites to spaceA and spaceB.
When Bob joins spaceA, we add an OutputEvent event to retire the invite. This sets the invite to "deleted" in the database. This makes sense. 

The bug is in stream_invites.go. Triggered when bob received a new invite for spaceB, and does a client sync.

In the block (line 76)
`for roomID := range retiredInvites

   if _, ok := req.Response.Rooms.Invite[roomID]; ok {
	continue
   }

   if _, ok := req.Response.Rooms.Join[roomID]; ok {
	continue
  }
...
` 

Bob is not in either maps even though he had just accepted the invite for spaceA. Consequently, the spaceA invite is treated as a retired invite, and a membership Leave event is generated. What bob sees is that after accepting the invite to spaceB, he lose access to spaceA.

### Pull Request Checklist

* [ ] I have added tests for PR _or_ I have justified why this PR doesn't need tests.
* [x ] Pull request includes a [sign off below using a legally identifiable name](https://matrix-org.github.io/dendrite/development/contributing#sign-off) _or_ I have already signed off privately


syncapi/streams/stream_invite.go
```diff
@@ -74,27 +74,43 @@ func (p *InviteStreamProvider) IncrementalSync(
...
-			OriginServerTS: gomatrixserverlib.AsTimestamp(time.Now()),
-			RoomID:         roomID,
-			Sender:         req.Device.UserID,
-			StateKey:       &req.Device.UserID,
-			Type:           "m.room.member",
-			Content:        gomatrixserverlib.RawJSON(`{"membership":"leave"}`),
-		})
-		req.Response.Rooms.Leave[roomID] = lr
 
+		joinedUsers, err := snapshot.AllJoinedUsersInRoom(ctx, []string{roomID})
```
neilalexander, line 84: This is a rather expensive way of approaching this problem as it results in us pulling out the entire membership state for a room, only to discard the vast majority of it. The membership table already has `SelectMembershipForUser` so it seems like we could just add some wrapper for that into `storage_sync.go`.

tak-hntlabs: thanks, this is the kind of feedback I'm hoping to get. I can work on a better fix in storage_sync.  I'm investigating a second regression that doesn't sync space room state properly after a user accepts an invite and joins a room. All happening from the commits in the last week - all related to invite-join flow.  Missing test coverage?

tak-hntlabs: @neilalexander Fixed to use SelectMembershipForUser. Please re-review

neilalexander (approved):

PR #2843, opened 2022-10-28
Fix #2842 - room state incorrectly filtered away for newly joined member

### LINKED ISSUE #2842: Client not getting m.room.create state when doing incremental sync
### Background information

- **Dendrite version or git SHA**: c62ac3d6ad5c60f5f28a0f50bba50f7cbc2436ce
- **Running in Docker?**: no
- **`go version`**: 1.19.2

### Description

- **What** is the problem:

matrix-js-sdk room.isSpace() has an error because the m.room.create event is not sync'ed from the server. This causes client issues which depends on the function to determine if the room is a space or not.

- **Who** is affected:

client app that uses the sdk to determine room state such as isSpaceRoom.

Affecting client apps that uses matrix-js-sdk room function that require m.room.create event to be fetched.

Changes in the past 1 week. This wasn't a problem before.

### Steps to reproduce

- alice creates a space and a child room
- alice sends 20 messages in the space
- alice invites bob to the room
- bob joins the room, but unable to get proper room states

Root cause of issue is in stream_pdu.go in the addRoomDeltaToResponse function. Will include debug notes in my PR.

tak-hntlabs:
Issue: #2842 

repro:
- alice creates a space and a child room
- alice sends 20 messages in the space
- alice invites bob to the room
- bob joins the room, but unable to get proper room states

In stream_pdu.go in the func addRoomDeltaToResponse:55 the stateEvents are removed.

`delta.StateEvents = removeDuplicates(delta.StateEvents, recentEvents) `

This is normally fine, but because recentEvents contains many other events (> 20), when historyVisibility filtering is applied, the original room events (like m.room.create) is filtered out because the filter limit is set.

This is a problem for newly joined members like bob who needs the room state such as the m.room.create event to determine room vs space types among other states.

### DIFF
--- a/syncapi/streams/stream_pdu.go
+++ b/syncapi/streams/stream_pdu.go
@@ -252,7 +252,13 @@ func (p *PDUStreamProvider) addRoomDeltaToResponse(
 		return r.From, fmt.Errorf("p.DB.RecentEvents: %w", err)
 	}
 	recentEvents := snapshot.StreamEventsToEvents(device, recentStreamEvents)
-	delta.StateEvents = removeDuplicates(delta.StateEvents, recentEvents) // roll back
+	// Don't remove the state events if this is a newly joined room. The member
+	// needs the full room state. For example, to figure out if the room is a space
+	// or not, it needs the m.room.create event which may be chronologically outside
+	// the filter limit.
+	if !delta.NewlyJoined {
+		delta.StateEvents = removeDuplicates(delta.StateEvents, recentEvents) // roll back
+	}
 	prevBatch, err := snapshot.GetBackwardTopologyPos(ctx, recentStreamEvents)
 	if err != nil {
 		return r.From, fmt.Errorf("p.DB.GetBackwardTopologyPos: %w", err)
--- a/syncapi/streams/stream_pdu.go
+++ b/syncapi/streams/stream_pdu.go
@@ -257,7 +257,7 @@ func (p *PDUStreamProvider) addRoomDeltaToResponse(
 	// or not, it needs the m.room.create event which may be chronologically outside
 	// the filter limit.
 	if !delta.NewlyJoined {
-		delta.StateEvents = removeDuplicates(delta.StateEvents, recentEvents) // roll back
+		delta.StateEvents = removeDuplicates(delta.StateEvents, recentEvents)
 	}
 	prevBatch, err := snapshot.GetBackwardTopologyPos(ctx, recentStreamEvents)
 	if err != nil {


syncapi/streams/stream_pdu.go
```diff
@@ -252,7 +252,13 @@ func (p *PDUStreamProvider) addRoomDeltaToResponse(
 		return r.From, fmt.Errorf("p.DB.RecentEvents: %w", err)
 	}
 	recentEvents := snapshot.StreamEventsToEvents(device, recentStreamEvents)
-	delta.StateEvents = removeDuplicates(delta.StateEvents, recentEvents) // roll back
+	// Don't remove the state events if this is a newly joined room. The member
+	// needs the full room state. For example, to figure out if the room is a space
+	// or not, it needs the m.room.create event which may be chronologically outside
+	// the filter limit.
+	if !delta.NewlyJoined {
```
neilalexander, line 259: This isn't the right way to approach this I don't think. The idea here is that an event should appear in *either* `"timeline"` or `"state"` but not both.

I suspect this is just a case of us calling `removeDuplicates` at the wrong time, i.e. before filtering rather than afterwards.

neilalexander:
So I think the actual root of this issue is in #2844, which is that the `"state"` section shouldn't have been limited at all. That's now fixed on `main`.

Can you please check to see if that improves things for you?

FYI the idea behind `removeDuplicates` is that an event should appear in *either* `"timeline"` or `"state"` but not both.

tak-hntlabs:
I had tested it 2 days ago after it was merged to main. Still saw the issue
unfortunately. I can retest with an older commit that had worked, and
observe the "state" and "timeline".

tak-hntlabs:
ah no - not yet. I'll test with latest main and let you know.

tak-hntlabs:
It fails. Looks like stateEvents need to be included to be recognized as
the room state. Can the recentEvents be pruned instead?

neilalexander:
Just to confirm the exact problem here: are you seeing the state events in `"timeline"` or are you not seeing them at all in either section?

Normally we deduplicate things from the state section if they appear in the timeline, otherwise clients often won't render those state transitions (i.e. joins, name changes etc.) to users. Clients should handle a state event regardless of which section it appears in, but they aren't guaranteed to render something that only appears in `"state"` (since the ordering is indeterminate).

tak-hntlabs:
If the states are in timeline only it does not work. i.e. duplicates
removed from stateEvents.

If in stateEvents and in timeline it works. I didn't try removing dups from
timeline

tak-hntlabs:
Any idea about this bug: https://github.com/matrix-org/dendrite/issues/2847
?

tak-hntlabs:
fixed by PR https://github.com/matrix-org/dendrite/pull/2856

PR #2844, opened 2022-10-31
Tweak `"state"` and `"timeline"` filtering

neilalexander:
This should stop state events disappearing down a gap where we'd try to separate out the sections *before* applying history visibility instead of after.


tak-hntlabs:
Still seeing issue #2842 

It works if I add this condition in line 318:

	if !delta.NewlyJoined {
		delta.StateEvents = gomatrixserverlib.HeaderedReverseTopologicalOrdering(
			removeDuplicates(delta.StateEvents, recentEvents),
			gomatrixserverlib.TopologicalOrderByAuthEvents,
		)
	}

Issue #2847, opened 2022-10-31
regression in scrollback function to get previous events


tak-hntlabs:
### Background information

- **Dendrite version or git SHA**: 8c7b274e4e223cbb888c1eb892767e46a25381aa
- **Monolith or Polylith?**: monolith
- **SQLite3 or Postgres?**: posgres
- **Running in Docker?**: no
- **`go version`**: 1.19.2
- **Client used (if applicable)**:

### Description

- **What** is the problem:

Using matrix-js-sdk to scroll back message history. 

matrixClient.scrollback(....)

This was working, but now has regressed.

- **Who** is affected:

client calling the scrollback function to go backwards in message history.

- **How** is this bug manifesting:

Unable to scroll back past events.

- **When** did this first appear:

Not sure.

### Steps to reproduce

- bob creates a public room. m.room.history_visibility is set to "[shared](https://spec.matrix.org/v1.4/client-server-api/#room-history-visibility)". 
- bob sends 25 messages
- alice joins the room
- alice calls scrollback(30) to get the previous 30 events. 

The matrix-js-sdk scrollback function hits the endpoint /rooms/{roomID}/messages in syncapi/routing.go

What used to happen is that the client gets back at least 30 events. But now it's only getting back 1 event.

Debugging the `OnIncomingMessagesRequest` function, it looks like the `fromQuery` parameter is wrong. 

Expected: something like 't14_19952' which translates to maxDepth = 14, maxStreamPosforMaxDepth = 19952

Actual: 't1_19445' which translates to maxDepth = 1, maxStreamPosforMaxDepth = 19445 

Because the maxDepth is only 1, it didn't return the correct number of events.

It looks like the bug happen earlier. After the client had sync'ed, it did not get back the correct pagination token. Consequently, when the client makes a request to this endpoint, it is passing the wrong 'fromQuery' parameter.

neilalexander:
Can you try the `neilalexander/prevbatch` branch?

tak-hntlabs:
It works!

neilalexander:
OK, fixed by #2856.

PR #2856, opened 2022-11-03
Move `prev_batch` calculation

neilalexander:
This might help #2847.


neilalexander:
I'm not actually sure this does the right thing, since it's possible that it causes us to make endless requests for periods of the timeline that have already been filtered out by history visibility.

neilalexander:
I've done some testing with this and it seems OK so will merge.

PR #2857, opened 2022-11-04
Cleanup stale device lists for users we don't share a room with anymore

S7evinK:
The stale device lists table might contain entries for users we don't share a room with anymore. This now asks the roomserver about left users and removes those entries from the table.


keyserver/storage/tables/stale_device_lists_test.go
```diff
@@ -0,0 +1,94 @@
+package tables_test
```
S7evinK, line 1: Do we want to use `package table` or `package tables_test` in future table tests?

kegsay: Not `table`. If you meant `tables` vs `tables_test` then it depends. The former exposes private functions, the latter doesn't. Given they are unit tests, am happy to expose private functions.

roomserver/storage/shared/storage.go
```diff
@@ -1365,6 +1365,36 @@ func (d *Database) JoinedUsersSetInRooms(ctx context.Context, roomIDs, userIDs [
...
+	}
+
+	userNIDs := make([]types.EventStateKeyNID, 0, len(stateKeyNIDMap))
+	userNIDtoUserID := make(map[types.EventStateKeyNID]string, len(stateKeyNIDMap))
+	for userID, nid := range stateKeyNIDMap {
+		userNIDs = append(userNIDs, nid)
+		userNIDtoUserID[nid] = userID
+	}
+
+	stillJoinedUsersNIDs, err := d.MembershipTable.SelectJoinedUsers(ctx, nil, userNIDs)
```
kegsay, line 1381: I don't grok this, we aren't even using room IDs anywhere..? How can we possibly know if we're joined to anything?

kegsay (changes requested):
Needs more comments. Possible race conditions.

roomserver/storage/shared/storage.go
S7evinK: Will add comments.

roomserver/storage/postgres/membership_table.go
```diff
@@ -157,6 +157,12 @@ const selectServerInRoomSQL = "" +
 	" JOIN roomserver_event_state_keys ON roomserver_membership.target_nid = roomserver_event_state_keys.event_state_key_nid" +
 	" WHERE membership_nid = $1 AND room_nid = $2 AND event_state_key LIKE '%:' || $3 LIMIT 1"
 
+const selectJoinedUsersSQL = `
+SELECT DISTINCT target_nid
+FROM roomserver_membership m
+WHERE membership_nid > $1 AND target_nid = ANY($2)
```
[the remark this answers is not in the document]
S7evinK: We're passing `tables.MembershipStateLeaveOrBan` (1) to it, so we're getting all users which are either joined, invited or have a knocking state.

keyserver/internal/device_list_update.go
```diff
@@ -154,11 +159,16 @@ func NewDeviceListUpdater(
...
 		userIDToChan:   make(map[string]chan bool),
 		userIDToChanMu: &sync.Mutex{},
+		rsAPI:          rsAPI,
 	}
 }
 
 // Start the device list updater, which will try to refresh any stale device lists.
 func (u *DeviceListUpdater) Start() error {
+	if err := u.cleanUp(); err != nil {
+		return fmt.Errorf("failed to cleanup stale device lists: %w", err)
```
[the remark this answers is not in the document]
S7evinK: We would have panicked on a failure, but changed that now anyway.

keyserver/internal/device_list_update.go
```diff
@@ -186,6 +196,36 @@ func (u *DeviceListUpdater) Start() error {
...
+		return nil
+	}
+	staleUsers, err := u.db.StaleDeviceLists(u.process.Context(), []gomatrixserverlib.ServerName{})
+	if err != nil {
+		return err
+	}
+
+	maxRetries := 3
+
+	// In polylith mode, the roomserver api might not be up yet, so we try again
```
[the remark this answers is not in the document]
S7evinK: It is, indeed. Something I was thinking about - we could implement some retry mechanism in our internal API calls, but this would increase the complexity around them.

kegsay (approved):

PR #2871, opened 2022-11-09
Add test code coverage reporting

devonh:


devonh:
With this change it looks like the github runner for unit tests now takes rougly 5 - 5.5 minutes.
I bumped the timeout up to 10 minutes to allow for variations in runner speeds.

Ideally we could pull this down, but due to the way dendrite's tests are written we rely on coverage metrics from cross package testing. That means, in order to get a full test coverage picture, we need to perform a more expensive coverage calculation than the default go settings provide.

PR #2872, opened 2022-11-11
Handle guest access [1/2?]

S7evinK:
Needs https://github.com/matrix-org/sytest/pull/1315, as otherwise the membership events aren't persisted yet when hitting `/state` after kicking guest users.

Makes the following tests pass:
```
Guest users denied access over federation if guest access prohibited
Guest users are kicked from guest_access rooms on revocation of guest_access
Guest users are kicked from guest_access rooms on revocation of guest_access over federation
```

Todo (in a follow up PR):
- Restrict access to CS API Endpoints as per https://spec.matrix.org/v1.4/client-server-api/#client-behaviour-14


roomserver/internal/perform/perform_join.go
```diff
@@ -271,6 +271,25 @@ func (r *Joiner) performJoinRoomByID(
 		}
 	}
 
+	// If a guest is trying to join a room, check that the room has a m.room.guest_access event
+	if req.IsGuest {
+		var guestAccessEvent *gomatrixserverlib.HeaderedEvent
+		guestAccess := "forbidden"
+		guestAccessEvent, err = r.DB.GetStateEvent(ctx, req.RoomIDOrAlias, gomatrixserverlib.MRoomGuestAccess, "")
```
neilalexander, line 278: Looks like this `err` is dropped?

neilalexander (commented):
Looks overall fine!

roomserver/internal/perform/perform_join.go
S7evinK: The error is ignored because `guestAccess` is set to forbidden above. We only use the result of `GetStateEvent` if there's no error and we actually have a `guestAccessEvent`.

neilalexander: The error could potentially be a failure contacting the database, so at the very least that should make it into the logs. Agreed we can ignore things like `sql.ErrNoRows` etc though.

kegsay (changes requested):
Just a few things.

roomserver/internal/input/input_events.go
```diff
@@ -729,3 +735,104 @@ func (r *Inputer) calculateAndSetState(
...
+
+// kickGuests kicks guests users from m.room.guest_access rooms, if guest access is now prohibited.
+func (r *Inputer) kickGuests(ctx context.Context, event *gomatrixserverlib.Event, roomInfo *types.RoomInfo) error {
+	if event.Type() != gomatrixserverlib.MRoomGuestAccess {
+		return nil
+	}
+	if gjson.GetBytes(event.Content(), "guest_access").String() == "can_join" {
+		return nil
+	}
+	membershipNIDs, err := r.DB.GetMembershipEventNIDsForRoom(ctx, roomInfo.RoomNID, true, true)
```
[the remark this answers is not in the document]
S7evinK: Can do in a different PR.

roomserver/internal/input/input_events.go
```diff
@@ -442,8 +443,11 @@ func (r *Inputer) processRoomEvent(
 		}
 	}
 
-	if err = r.kickGuests(ctx, event, roomInfo); err != nil {
-		logrus.WithError(err).Error("failed to kick guest users on m.room.guest_access revocation")
+	// If guest_access changed and is not can_join, kick all guest users.
+	if event.Type() == gomatrixserverlib.MRoomGuestAccess && gjson.GetBytes(event.Content(), "guest_access").String() != "can_join" {
```
kegsay, line 447: Please do not use `.String()` as that type-coerces the value which is not what we want: https://github.com/tidwall/gjson/blob/v1.14.4/gjson.go#L72 - use `.Str` instead which is only populated if the value is in fact a string.

kegsay (approved):

PR #2876, opened 2022-11-11
Virtual hosting schema and logic changes

neilalexander:
Note that virtual users cannot federate correctly yet.

### DIFF
--- a/userapi/storage/postgres/deltas/2022110411000001_server_names.go
+++ b/userapi/storage/postgres/deltas/2022110411000001_server_names.go
@@ -1,0 +1,28 @@
+package deltas
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+
+	"github.com/lib/pq"
+	"github.com/matrix-org/gomatrixserverlib"
+)
+
+// I know what you're thinking: you're wondering "why doesn't this use $1
+// and pass variadic parameters to ExecContext?" — the answer is because
+// PostgreSQL doesn't expect the table name to be specified as a substituted
+// argument in that way so it results in a syntax error in the query.
+
+func UpServerNamesPopulate(ctx context.Context, tx *sql.Tx, serverName gomatrixserverlib.ServerName) error {
+	for _, table := range serverNamesTables {
+		q := fmt.Sprintf(
+			"UPDATE %s SET server_name = %s WHERE server_name = '';",
+			pq.QuoteIdentifier(table), pq.QuoteLiteral(string(serverName)),
+		)
+		if _, err := tx.ExecContext(ctx, q); err != nil {
+			return fmt.Errorf("write server names to %q error: %w", table, err)
+		}
+	}
+	return nil
+}
--- a/userapi/storage/postgres/threepid_table.go
+++ b/userapi/storage/postgres/threepid_table.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/matrix-org/dendrite/internal/sqlutil"
 	"github.com/matrix-org/dendrite/userapi/storage/tables"
+	"github.com/matrix-org/gomatrixserverlib"
 
 	"github.com/matrix-org/dendrite/clientapi/auth/authtypes"
 )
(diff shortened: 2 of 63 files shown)


userapi/storage/postgres/threepid_table.go
```diff
@@ -33,21 +34,22 @@ CREATE TABLE IF NOT EXISTS userapi_threepids (
...
 	PRIMARY KEY(threepid, medium)
 );
 
-CREATE INDEX IF NOT EXISTS userapi_threepid_idx ON userapi_threepids(localpart);
+CREATE INDEX IF NOT EXISTS userapi_threepid_idx ON userapi_threepids(localpart, server_name);
 `
 
 const selectLocalpartForThreePIDSQL = "" +
-	"SELECT localpart FROM userapi_threepids WHERE threepid = $1 AND medium = $2"
+	"SELECT localpart, server_name FROM userapi_threepids WHERE threepid = $1 AND medium = $2"
```
S7evinK, line 46: I think we need to update this to also filter on the `server_name` column, as otherwise this might return more than one result.

userapi/storage/postgres/threepid_table.go
```diff
@@ -33,21 +34,22 @@ CREATE TABLE IF NOT EXISTS userapi_threepids (
...
 const selectThreePIDsForLocalpartSQL = "" +
-	"SELECT threepid, medium FROM userapi_threepids WHERE localpart = $1"
+	"SELECT threepid, medium FROM userapi_threepids WHERE localpart = $1 AND server_name = $2"
 
 const insertThreePIDSQL = "" +
-	"INSERT INTO userapi_threepids (threepid, medium, localpart) VALUES ($1, $2, $3)"
+	"INSERT INTO userapi_threepids (threepid, medium, localpart, server_name) VALUES ($1, $2, $3, $4)"
 
 const deleteThreePIDSQL = "" +
 	"DELETE FROM userapi_threepids WHERE threepid = $1 AND medium = $2"
```
S7evinK, line 55: Probably also needs filtering on `server_name`.

userapi/storage/tables/stats_table_test.go
```diff
@@ -79,6 +79,7 @@ func mustMakeAccountAndDevice(
 	accDB tables.AccountsTable,
 	devDB tables.DevicesTable,
 	localpart string,
+	serverName gomatrixserverlib.ServerName, // nolint:unparam
```
S7evinK, line 82: Since we're using `serverName`, `nolint` can be removed?

S7evinK (changes requested):
Overall, looks fine. Just a few places where we're missing `IsLocalServerName` checks or missing `ServerName` in the parameter to API requests. 
I'm wondering if we can avoid internal API roundtrips by already checking possible `ServerName` fields in requests on the client (mostly in Polylith scenarios)?

Oh, and the Postgres comments apply to SQLite as well.

userapi/storage/tables/stats_table_test.go
neilalexander: The `nolint` is because it always receives the same value and the linter is picky :-)

userapi/storage/postgres/deltas/2022110411000001_server_names.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/lib/pq"
+	"github.com/matrix-org/gomatrixserverlib"
+)
+
+// I know what you're thinking: you're wondering "why doesn't this use $1
+// and pass variadic parameters to ExecContext?" — the answer is because
+// PostgreSQL doesn't expect the table name to be specified as a substituted
+// argument in that way so it results in a syntax error in the query.
+
+func UpServerNamesPopulate(ctx context.Context, tx *sql.Tx, serverName gomatrixserverlib.ServerName) error {
```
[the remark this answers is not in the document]
neilalexander: Mostly kept this separate to make it very exceptionally obvious that it runs at a different stage to the other migration.

PR #2885, opened 2022-11-18
Added Landing Page

LukasLJL:
I have added/copied a landing page like Synpase does.
Recently I have installed Dendrite and was wondering why it´s not working. After some troubleshooting I figured out there is no landing page like synpase has, so the Server was running just fine. 

Hopefuly this PR can fix this problem and may help other users who run into this issue.
I have not written any unit tests, because it´s just a simple landing page with a redirect to a static site.

### Pull Request Checklist

Signed-off-by: `Lukas Huida<lukas@leucali.net>`


kegsay:
This needs tests to ensure the routing is correct.

kegsay (changes requested):
Needs tests to make sure routing is working correctly in monolith mode at the very least. Otherwise LGTM!

LukasLJL:
Thanks for the reply, could you give me a hint where I can add the test? Are there any example?

kegsay:
There aren't as there hasn't been anything like this before. You need a `setup.Monolith` and a `base.Dendrite` and to glue them together similar to https://github.com/matrix-org/dendrite/blob/main/cmd/dendrite-monolith-server/main.go - then you can make a fake request with [httptest](https://pkg.go.dev/net/http/httptest#NewRequest) and pass it to the handler to ensure you get back the HTML page.

S7evinK:
We should probably also embed the html file and/or make it a template to include the actual Dendrite version.
Something like
```html
    <h1>It works! Dendrite {{ .dendriteVersion }} is running</h1>
```
would be great.

setup/base/base.go
```diff
@@ -478,13 +486,30 @@ func (b *BaseDendrite) SetupAndServeHTTP(
 
 	b.configureHTTPErrors()
 
+	//Redirect for Landing Page
+	externalRouter.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		http.Redirect(w, r, httputil.PublicStaticPath, http.StatusSeeOther)
```
kegsay, line 491: `http.StatusFound` surely?

setup/base/base.go
```diff
@@ -478,13 +486,30 @@ func (b *BaseDendrite) SetupAndServeHTTP(
...
 		internalRouter.Handle("/metrics", httputil.WrapHandlerInBasicAuth(promhttp.Handler(), b.Cfg.Global.Metrics.BasicAuth))
 	}
 
 	b.ConfigureAdminEndpoints()
 
+	tmpl := template.Must(template.ParseFS(staticContent, "static/*.gotmpl"))
+
+	b.PublicStaticMux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		if err := tmpl.ExecuteTemplate(w, "index.gotmpl", map[string]string{
+			"Version": internal.VersionString(),
```
kegsay, line 505: It seems inefficient to template this every time the client asks. Just do it once at startup and return those bytes?

kegsay (approved):
Otherwise LGTM

PR #2891, opened 2022-11-27
Don't omit `pattern` key if empty.

### LINKED ISSUE #2882: Potential Spec violation with push rules / incompatible with FluffyChat
### Background information

Homeserver: dendrite.matrix.org

neilalexander accepted opening a tracking issue also at Dendrite in case it's Dendrite and not Famedly SDK (most notably used in FluffyChat) at fault. Their issue: https://gitlab.com/famedly/company/frontend/famedlysdk/-/issues/349

I apologise for not knowing enough to open a more detailed issue.

### Description

- **What** is the problem: FluffyChat cannot sync with Dendrite.
- **Who** is affected: FluffyChat users
- **How** is this bug manifesting: FluffyChat never finishes syncing
- **When** did this first appear: FluffyChat development buids for version that were stabilized today.

### Steps to reproduce

1. Login to Dendrite homeserver using FlufyChat.
2. Observe it not syncing.

webzwo0i:
This is needed to match state events with an empty string value. Fixes #2882

Feel free to dismiss this PR in case you think that clients should interpret a missing pattern key in state event conditions as if the pattern would be `*`. However, for a lot of state events, this would also match events that have an arbitrary value, not only those that have `pattern: ''`` (empty string). So maybe clients should interpret a missing pattern key as if `pattern: ''` was given.

### DIFF
--- a/internal/pushrules/condition.go
+++ b/internal/pushrules/condition.go
@@ -14,7 +14,7 @@ type Condition struct {
 
 	// Pattern indicates the value pattern that must match. Required
 	// for EventMatchCondition.
-	Pattern string `json:"pattern,omitempty"`
+	Pattern string `json:"pattern"`
 
 	// Is indicates the condition that must be fulfilled. Required for
 	// RoomMemberCountCondition.


neilalexander:
Looking at [the spec](https://spec.matrix.org/v1.5/client-server-api/#push-rules-events), `pattern` is not a "Required" field, which makes me think Fluffychat should be able to cope with it not being present and do the right thing.

spaetz:
The spec continues to state:

`pattern | string | Required for event_match conditions. The glob-style pattern to match against.`

So, pattern is required in the case of event_match conditions.

The dendrite code base recognizes that on line 15 of conditions.go:

```
	// Pattern indicates the value pattern that must match. Required
	// for EventMatchCondition.
	Pattern string `json:"pattern"`
```

So, we need to make sure that "pattern" is included ONLY when we have a case of
 `EventMatchCondition ConditionKind = "event_match"`.

kegsay:
What does Synapse do here?

kegsay (changes requested):
Needs unit tests, and confirmation that this is right behaviour (what does Synapse do?) Also, we require sign-off.

S7evinK:
Closing in favor of https://github.com/matrix-org/dendrite/pull/2918

PR #2901, opened 2022-12-02
Add possibility to run complement with coverage enabled

S7evinK:
This adds the possibility to run Complement with coverage enabled.
In combination with https://github.com/matrix-org/complement/pull/566 we should then be able to extract the coverage logs, combine them with https://github.com/wadey/gocovmerge (or similar) and upload them to Codecov (with different flags, depending on SQLite, HTTP etc.)


Dockerfile
```diff
@@ -82,11 +86,12 @@ RUN /usr/bin/generate-keys --private-key matrix_key.pem && \
...
+ENV COVER=0
 EXPOSE 8008 8448
 
 # At runtime, generate TLS cert based on the CA now mounted at /ca
 # At runtime, replace the SERVER_NAME with what we are told
 CMD /usr/bin/generate-keys --server $SERVER_NAME --tls-cert server.crt --tls-key server.key --tls-authority-cert /ca/ca.crt --tls-authority-key /ca/ca.key && \
     /usr/bin/generate-config -server $SERVER_NAME --ci > dendrite.yaml && \
     cp /ca/ca.crt /usr/local/share/ca-certificates/ && update-ca-certificates && \
-    /usr/bin/dendrite-monolith-server --really-enable-open-registration --tls-cert server.crt --tls-key server.key --config dendrite.yaml -api=${API:-0}
+    exec /complement-cmd.sh
```
kegsay, line 97: Running "complement" anything in this Dockerfile feels very very wrong. This is the base image for the entire server?

Dockerfile
```diff
@@ -21,7 +21,9 @@ RUN --mount=target=. \
     GOARCH="$TARGETARCH" \
     GOOS="linux" \
     CGO_ENABLED=$([ "$TARGETARCH" = "$USERARCH" ] && echo "1" || echo "0") \
-    go build -v -ldflags="${FLAGS}" -trimpath -o /out/ ./cmd/...
+    go build -v -ldflags="${FLAGS}" -trimpath -o /out/ ./cmd/... && \
+    go test -c -cover -covermode=atomic -o /out/dendrite-monolith-server-cover -coverpkg "github.com/matrix-org/..." ./cmd/dendrite-monolith-server && \
+    cp build/scripts/complement-cmd.sh /out/
```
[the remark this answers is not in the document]
S7evinK: The "main" Dockerfile is multi staged and that's the "build" stage.
The binaries are later only copied to the stages which need them.

Dockerfile
S7evinK: Same as above, multi staged Dockerfile, this is the "complement" build target.

S7evinK: Ripped it out, since the `complement-cmd.sh` would not have worked anyway.

kegsay (approved):
Assuming it works, LGTM

Issue #2878, opened 2022-11-13
hcaptcha fails with "Invalid site key"


greenship24:
### Background information

- **Dendrite version or git SHA**:
- **Monolith or Polylith?**:
- **SQLite3 or Postgres?**:
- **Running in Docker?**:
- **`go version`**:
- **Client used (if applicable)**:

- Latest dendrite as of 2022-11-12
- Monolith
- postgres
- docker
- built-in to image
- Element, schilidchat, element on the web, syphon, etc.

### Description

- **What** is the problem:
hcaptcha fails to work - 
- **Who** is affected:
Anyone using hcaptcha in lieu of recaptcha
- **How** is this bug manifesting:
Response is "ERROR for site owner: Invalid site key" upon trying to register for the server
- **When** did this first appear:

### Steps to reproduce

- Place site key in the recaptcha_public_key
- Place private key in recaptcha_private_key
- Uncomment all fields under # To use hcaptcha.com..... and use the defaults (as in the dendrite.yaml template)

<!--
Describe how what happens differs from what you expected.

If you can identify any relevant log snippets from server logs, please include
those (please be careful to remove any personal or private data). Please surround them with
``` (three backticks, on a line on their own), so that they are formatted legibly.

Alternatively, please send logs to @kegan:matrix.org or @neilalexander:matrix.org
with a link to the respective Github issue, thanks!
-->

### Additional
I think it *might* have to do with whatever v2 and v3 recaptcha uses which I presume hcaptcha uses v3 now? Just a guess. 
I've verified all settings and have done simple tests otherwise where the site key works fine.

**Comment by twiguard** [comment]
I just want to add, that is seems reCaptcha is returning the error, not hCaptcha as configured.

**Comment by greenship24** [comment]
```
  # Settings for ReCAPTCHA.
  recaptcha_public_key: "<key here>"
  recaptcha_private_key: "<private key here>"
  recaptcha_bypass_secret: ""

  # To use hcaptcha.com instead of ReCAPTCHA, set the following parameters, otherwise just keep them empty.
  recaptcha_siteverify_api: "https://hcaptcha.com/siteverify"
  recaptcha_api_js_url: "https://js.hcaptcha.com/1/api.js"
  recaptcha_form_field: "h-captcha-response"
  recaptcha_sitekey_class: "h-captcha"
```

This is what I have in dendrite.yaml (with my hcaptcha keys obviously in the key here portions). Am I missing something?

wait-what:
[...]

Issue #2493, opened 2022-05-25
Allow other captcha Providers


mpldr:
**Description:**

I'd rather not use reCaptcha for privacy reasons. Is there a way to configure other providers such as hCaptcha or – even better – serve a custom captcha.

neilalexander:
#2415 does mention [patching in hCaptcha support](https://github.com/matrix-org/dendrite/issues/2415#issuecomment-1122058084), since seemingly the site-verify API is quite similar. I absolutely think this is a thing we can reasonably support.

array-in-a-matrix:
hCaptcha is an option as of [Dendrite version 0.10.5](https://github.com/matrix-org/dendrite/releases/tag/v0.10.5). Having other options would be nice.

kegsay:
We have no plans to add more captcha providers.

PR #2916, opened 2022-12-19
Add clientapi tests

S7evinK:
This PR
- adds several tests for the clientapi, mostly around `/register` and auth fallback.
- removes the now deprecated `homeserver` field from responses to `/register`
- slightly refactors auth fallback handling

### DIFF
--- a/clientapi/routing/auth_fallback.go
+++ b/clientapi/routing/auth_fallback.go
@@ -15,6 +15,7 @@
 package routing
 
 import (
+	"fmt"
 	"html/template"
 	"net/http"
 
@@ -102,6 +103,22 @@ func AuthFallback(
 	w http.ResponseWriter, req *http.Request, authType string,
 	cfg *config.ClientAPI,
 ) *util.JSONResponse {
+	// We currently only support reCaptcha, so fail early if that's not requested
+	if authType == authtypes.LoginTypeRecaptcha {
+		if !cfg.RecaptchaEnabled {
+			return writeHTTPMessage(w, req,
+				"Recaptcha login is disabled on this Homeserver",
+				http.StatusBadRequest,
+			)
+		}
+	} else {
+		_ = writeHTTPMessage(w, req, fmt.Sprintf("Unknown authtype %q", authType), http.StatusNotImplemented)
+		return &util.JSONResponse{
+			Code: http.StatusNotFound,
+			JSON: jsonerror.NotFound("Unknown auth stage type"),
+		}
+	}
+
 	sessionID := req.URL.Query().Get("session")
 
 	if sessionID == "" {
@@ -130,70 +147,35 @@ func AuthFallback(
 
 	if req.Method == http.MethodGet {
 		// Handle Recaptcha
-		if authType == authtypes.LoginTypeRecaptcha {
-			if err := checkRecaptchaEnabled(cfg, w, req); err != nil {
-				return err
-			}
-
-			serveRecaptcha()
-			return nil
-		}
-		return &util.JSONResponse{
-			Code: http.StatusNotFound,
-			JSON: jsonerror.NotFound("Unknown auth stage type"),
-		}
+		serveRecaptcha()
+		return nil
 	} else if req.Method == http.MethodPost {
 		// Handle Recaptcha
-		if authType == authtypes.LoginTypeRecaptcha {
-			if err := checkRecaptchaEnabled(cfg, w, req); err != nil {
-				return err
-			}
-
-			clientIP := req.RemoteAddr
-			err := req.ParseForm()
-			if err != nil {
-				util.GetLogger(req.Context()).WithError(err).Error("req.ParseForm failed")
-				res := jsonerror.InternalServerError()
-				return &res
-			}
-
-			response := req.Form.Get(cfg.RecaptchaFormField)
-			if err := validateRecaptcha(cfg, response, clientIP); err != nil {
-				util.GetLogger(req.Context()).Error(err)
-				return err
-			}
-
-			// Success. Add recaptcha as a completed login flow
-			sessions.addCompletedSessionStage(sessionID, authtypes.LoginTypeRecaptcha)
-
-			serveSuccess()
-			return nil
+		clientIP := req.RemoteAddr
+		err := req.ParseForm()
+		if err != nil {
+			util.GetLogger(req.Context()).WithError(err).Error("req.ParseForm failed")
+			res := jsonerror.InternalServerError()
+			return &res
 		}
 
-		return &util.JSONResponse{
-			Code: http.StatusNotFound,
-			JSON: jsonerror.NotFound("Unknown auth stage type"),
+		response := req.Form.Get(cfg.RecaptchaFormField)
+		if err := validateRecaptcha(cfg, response, clientIP); err != nil {
+			util.GetLogger(req.Context()).Error(err)
+			return err
 		}
+
+		// Success. Add recaptcha as a completed login flow
+		sessions.addCompletedSessionStage(sessionID, authtypes.LoginTypeRecaptcha)
+
+		serveSuccess()
+		return nil
 	}
+	_ = writeHTTPMessage(w, req, "Bad method", http.StatusMethodNotAllowed)
 	return &util.JSONResponse{
 		Code: http.StatusMethodNotAllowed,
 		JSON: jsonerror.NotFound("Bad method"),
 	}
-}
-
-// checkRecaptchaEnabled creates an error response if recaptcha is not usable on homeserver.
-func checkRecaptchaEnabled(
-	cfg *config.ClientAPI,
-	w http.ResponseWriter,
-	req *http.Request,
-) *util.JSONResponse {
-	if !cfg.RecaptchaEnabled {
-		return writeHTTPMessage(w, req,
-			"Recaptcha login is disabled on this Homeserver",
-			http.StatusBadRequest,
-		)
-	}
-	return nil
 }
 
 // writeHTTPMessage writes the given header and message to the HTTP response writer.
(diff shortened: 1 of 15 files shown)


S7evinK:
Needs https://github.com/matrix-org/complement/pull/573 for Complement to pass, as this also removes the deprecated (since r0.4.0) `home_server` field from responses to `/register` and `/login`.

setup/config/config_clientapi.go
```diff
@@ -93,6 +90,10 @@ func (c *ClientAPI) Verify(configErrs *ConfigErrors, isMonolith bool) {
 		if c.RecaptchaSitekeyClass == "" {
 			c.RecaptchaSitekeyClass = "g-recaptcha-response"
 		}
+		checkNotEmpty(configErrs, "client_api.recaptcha_public_key", c.RecaptchaPublicKey)
```
kegsay, line 93: Why shift the order of checks here?

clientapi/routing/auth_fallback.go
```diff
@@ -102,6 +103,22 @@ func AuthFallback(
 	w http.ResponseWriter, req *http.Request, authType string,
 	cfg *config.ClientAPI,
 ) *util.JSONResponse {
+	// We currently only support reCaptcha, so fail early if that's not requested
+	if authType == authtypes.LoginTypeRecaptcha {
```
kegsay, line 107: Is this true? I thought we supported hCaptcha as well? But unsure it works? https://github.com/matrix-org/dendrite/issues/2878 https://github.com/matrix-org/dendrite/issues/2493

clientapi/routing/auth_fallback.go
```diff
@@ -102,6 +103,22 @@ func AuthFallback(
...
+			return writeHTTPMessage(w, req,
+				"Recaptcha login is disabled on this Homeserver",
+				http.StatusBadRequest,
+			)
+		}
+	} else {
+		_ = writeHTTPMessage(w, req, fmt.Sprintf("Unknown authtype %q", authType), http.StatusNotImplemented)
+		return &util.JSONResponse{
+			Code: http.StatusNotFound,
+			JSON: jsonerror.NotFound("Unknown auth stage type"),
```
kegsay, line 118: These JSON responses are presumably just logged? Given we write back text on `:115`? Needs clarification / removal of `util.JSONResponse` here.

clientapi/routing/auth_fallback.go
```diff
@@ -130,72 +147,39 @@ func AuthFallback(
...
-				util.GetLogger(req.Context()).Error(err)
-				return err
-			}
-
-			// Success. Add recaptcha as a completed login flow
-			sessions.addCompletedSessionStage(sessionID, authtypes.LoginTypeRecaptcha)
-
-			serveSuccess()
-			return nil
+		clientIP := req.RemoteAddr
```
kegsay, line 154: This needs to honour the real IP header in case of reverse proxies. It's a config option iirc

clientapi/routing/auth_fallback.go
```diff
@@ -130,72 +147,39 @@ func AuthFallback(
...
+		err := req.ParseForm()
+		if err != nil {
+			util.GetLogger(req.Context()).WithError(err).Error("req.ParseForm failed")
+			res := jsonerror.InternalServerError()
```
kegsay, line 158: Most certainly not a 500, surely 4xx?

clientapi/routing/auth_fallback.go
kegsay: x100 on this. It's currently really unclear when it is right to send back text, and when to send back JSON, and the whole interplay between the two. Like here, it's an error and we write back text. But if we fail to parse the form values we send back JSON. Why?

clientapi/routing/auth_fallback_test.go
```diff
@@ -0,0 +1,131 @@
...
+				t.Run(fmt.Sprintf("useHCaptcha(%v) - recaptchaEnabled(%v) - wantErr(%v)", useHCaptcha, recaptchaEnabled, wantErr), func(t *testing.T) {
+					// Set the defaults for each test
+					base.Cfg.ClientAPI.Defaults(config.DefaultOpts{Generate: true, Monolithic: true})
+					base.Cfg.ClientAPI.RecaptchaEnabled = recaptchaEnabled
+					base.Cfg.ClientAPI.RecaptchaPublicKey = "pub"
+					base.Cfg.ClientAPI.RecaptchaPrivateKey = "priv"
+					if useHCaptcha {
+						base.Cfg.ClientAPI.RecaptchaSiteVerifyAPI = "https://hcaptcha.com/siteverify"
+						base.Cfg.ClientAPI.RecaptchaApiJsUrl = "https://js.hcaptcha.com/1/api.js"
+						base.Cfg.ClientAPI.RecaptchaFormField = "h-captcha-response"
```
kegsay, line 32: After saying how we only support recaptcha in comments, it's amusing to see the tests referring to hcaptcha :D

clientapi/routing/register_test.go
```diff
@@ -264,3 +279,318 @@ func TestSessionCleanUp(t *testing.T) {
...
+			domain:    "localhost",
+			want: &util.JSONResponse{
+				Code: http.StatusBadRequest,
+				JSON: jsonerror.InvalidUsername("Username cannot start with a '_'"),
+			},
+		},
+		{
+			name:      "valid username",
+			localpart: "valid",
+			domain:    "localhost",
```
kegsay, line 330: Surprised we don't ever test bad domains.

clientapi/routing/register_test.go
```diff
@@ -264,3 +279,318 @@ func TestSessionCleanUp(t *testing.T) {
...
+			name:      "empty username",
+			localpart: "",
+			domain:    "localhost",
+			want: &util.JSONResponse{
+				Code: http.StatusBadRequest,
+				JSON: jsonerror.InvalidUsername("Username can only contain characters a-z, 0-9, or '_-./='"),
+			},
+		},
+		{
+			name:      "invalid username",
```
kegsay, line 301: Say why (caps).

Also, it would be good to add tests for:
 - unicode characters e.g emoji
 - ASCII symbols e.g $
 - a complex but valid username beyond the existing test for "valid" e.g `f00_bar-baz.=40/` is valid I think? Test it.

clientapi/routing/register_test.go
```diff
@@ -264,3 +279,318 @@ func TestSessionCleanUp(t *testing.T) {
...
+		},
+		{
+			name:      "valid username",
+			localpart: "valid",
+			domain:    "localhost",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := validateUsername(tt.localpart, tt.domain); !reflect.DeepEqual(got, tt.want) {
```
kegsay, line 335: Whilst we are here... please move `validateUsername` to `internal` or some other package and call it from `cmd/create-account` to prevent people registering usernames with caps or other things like that.

clientapi/routing/register_test.go
```diff
@@ -264,3 +279,318 @@ func TestSessionCleanUp(t *testing.T) {
...
+				t.Errorf("validateUsername() = %v, want %v", got, tt.want)
+			}
+			if got := validateApplicationServiceUsername(tt.localpart, tt.domain); !reflect.DeepEqual(got, tt.want) {
+				if got != nil && got.JSON != jsonerror.InvalidUsername("Username cannot start with a '_'") {
```
kegsay, line 339: Needs more comments to explain why you're doing this check. I assume because appservices _are_ allowed to use `_`.

clientapi/routing/register_test.go
```diff
@@ -264,3 +279,318 @@ func TestSessionCleanUp(t *testing.T) {
...
+		name     string
+		password string
+		want     *util.JSONResponse
+	}{
+		{
+			name:     "password too short",
+			password: "shortpw",
+			want: &util.JSONResponse{
+				Code: http.StatusBadRequest,
+				JSON: jsonerror.WeakPassword(fmt.Sprintf("password too weak: min %d chars", minPasswordLength)),
```
kegsay, line 358: I generally dislike doing direct checks on error strings, as they aren't really part of the public API. We should be checking the `errcode`. Write a helper function and `gjson` out the code? This applies throughout these tests.

kegsay (changes requested):
Mostly small things, the main code here is good, thanks!

clientapi/routing/auth_fallback.go
S7evinK: We do, but it is also using the `authtypes.LoginTypeRecaptcha`, given it's just using a different backend for validation.

clientapi/routing/auth_fallback.go
S7evinK: TBH, I also don't like all the JSON and normal HTML stuff combined. Since we're just using the `MakeHTMLAPI` for auth fallback right now, I'd be totally fine to completely remove JSON here.

clientapi/routing/register_test.go
S7evinK: I also had that added, just to see that we either use the `server_name` from the configuration OR one from a virtual host.
And yea, something like `http://localhost` would produce a valid username

clientapi/routing/register_test.go
S7evinK: Like above, we should just remove the JSON here and actually return error types. (Same for the other `validate*` functions)

clientapi/routing/auth_fallback.go
S7evinK: It is, but unfortunately just for the sync api :/

setup/config/config_clientapi.go
S7evinK: Could remove it completely, given we either have something set or default to reCaptcha (if nothing is set).

kegsay (approved):
LGTM!

PR #2918, opened 2022-12-20
Pushrule tweaks, make `pattern` non-optional on `EventMatchCondition`

S7evinK:
This should fix https://github.com/matrix-org/dendrite/issues/2882 (Tested with FluffyChat 1.7.1)
Also adds tests that the predefined push rules (as per the spec) is what we have in Dendrite.

### DIFF
--- a/internal/pushrules/condition.go
+++ b/internal/pushrules/condition.go
@@ -14,7 +14,7 @@ type Condition struct {
 
 	// Pattern indicates the value pattern that must match. Required
 	// for EventMatchCondition.
-	Pattern string `json:"pattern,omitempty"`
+	Pattern *string `json:"pattern,omitempty"`
 
 	// Is indicates the condition that must be fulfilled. Required for
 	// RoomMemberCountCondition.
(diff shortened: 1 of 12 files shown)


Undef-a:
Just tested this one (pulled your branch and built/installed) against Fluffychat 1.7.2 (arm64 flatpak). I still get the old error `Error during processing events - PushCondition is missing pattern`.

S7evinK:
@Undef-a: In case it's possible, can you provide your pushrules (`SELECT * FROM userapi_account_datas WHERE localpart = 'YourLocalpart' and type = 'm.push_rules'`)?

Might be related to already stored pushrules in the database. If you don't have configured anything special or are OK with configuring that again: Can you remove your pushrules from the database? (`DELETE FROM userapi_account_datas WHERE localpart = 'YourLocalpart' and type = 'm.push_rules'`)
After deleting you can recreate them by clearing cache in Element Web/doing an initial sync.

Undef-a:
I did have a few, but sadly deleting them didn't fix the issue either.

Undef-a:
Looks like the issue above is on the Fluffychat side. If you've already logged in upgrading to the new version will still fail to start even with these patches. 

Logging out of Fluffychat and back in fixes it.

kegsay (approved):
Need coverage up, but LGTM

Issue #2914, opened 2022-12-14
Signing in with uppercase username breaks device list updates


nico-famedly:
### Background information

- **Dendrite version or git SHA**: 0.10.8+ed497aa
- **Monolith or Polylith?**: Monolith
- **SQLite3 or Postgres?**: sqlite
- **Running in Docker?**: yes
- **`go version`**:
- **Client used (if applicable)**: famedlysdk, but shouldn't matter

### Description

- **What** is the problem: If I sign in using `Bob` instead of `bob`, I get a device list update for `@Bob:localhost` instead of `@bob:localhost` with only one entry
- **Who** is affected: Uhhh, everyone who uses an uppercase username to sign in?
- **How** is this bug manifesting: Wrong device lists, that break e2ee
- **When** did this first appear:

### Steps to reproduce

- curl -fS -XPOST -d '{"username":"bob", "password":"Something", "inhibit_login":true, "auth": {"type":"m.login.dummy"}}' "http://$IP_ADDRESS/_matrix/client/r0/register"
- then login using Bob as the identifier for the user
- login succeeds
- If you share an encrypted room with bob, you will get a device list update for `@Bob:servername` instead of `@bob:servername`

<!--
Describe how what happens differs from what you expected.

If you can identify any relevant log snippets from server logs, please include
those (please be careful to remove any personal or private data). Please surround them with
``` (three backticks, on a line on their own), so that they are formatted legibly.

Alternatively, please send logs to @kegan:matrix.org or @neilalexander:matrix.org
with a link to the respective Github issue, thanks!
-->
I would expect device lists to always use the exact mxid instead of the casing from /login.

Logs:

```
# Normal login using "bob"

Updating device keys for {@bob:famedlysdk.test: []}

Response {@bob:famedlysdk.test: {pnv71Ijt: Instance of 'MatrixDeviceKeys'}}

# Login using "Bob"

[Matrix] Successfully connected as Bob with http://127.0.0.1

Updating device keys for {@Bob:famedlysdk.test: []}

Response {@Bob:famedlysdk.test: {2Cfqnc6M: Instance of 'MatrixDeviceKeys'}}
```

Notice how the latter does not include the devices form "bob" and uses the wrong mxid in the response.

bones-was-here:
How are you able to log in with an uppercase char?

I still patch the source to support very old accounts https://github.com/matrix-org/dendrite/issues/2073#issuecomment-1047934258

deepbluev7:
[...]

Issue #2073, opened 2021-12-16
Usernames with uppercase chars can no longer log in


bones-was-here:
### Background information

- **Dendrite version or git SHA**: 002c3e0a5ff18c0b36c4c8e6cebea61547c9ac51
- **Monolith or Polylith?**: Monolith
- **SQLite3 or Postgres?**: Postgres
- **Running in Docker?**: no
- **`go version`**: 1.17.5

### Description

If a user has created an account with `create-account` and their username contains a capital letter, they can no longer log in.

### Steps to reproduce

- `create-account --username "Test" --password "SomePassword"`
- try to log in with version 388d7a197446bce5493c52d36affecf0cf706296 or later

We have had such an account for a long time now and were unaware it was not compliant with the spec.  Is this going to cause federation problems for the rooms?  How can we recover from the situation?

kegsay:
It shouldn't cause federation problems as there are existing matrix user IDs on matrix.org which also have upper-case letters.

hoernschen:
I would like to get into developing for this project and wanted to start with this issue. 
But I am not completely sure what the expected behaviour should be here. Should the login also support old uppercase usernames or be completely not case sensitive?

kegsay:
It should be completely not case-sensitive.

bones-was-here:
"Fixed" by adding a line to the build script...
`sed -i 's/username := strings.ToLower(r.Username())/username := r.Username()/' clientapi/auth/password.go`


> It should be completely not case-sensitive.

Forcing to lower case is not the same as case insensitivity...

avdb13:
> Forcing to lower case is not the same as case insensitivity...
This is indeed true but if you want to have case insensitivity you'd have to change it all the way down to the database level which might break compatibility with a lot of other parts of the code. As for as I know SQLite supports case-insensitive comparisons for ASCII characters but for PostgreSQL the `citext` extension would be required.

bones-was-here:
> > Forcing to lower case is not the same as case insensitivity...
> 
> This is indeed true but if you want to have case insensitivity you'd have to change it all the way down to the database level which might break compatibility with a lot of other parts of the code. As for as I know SQLite supports case-insensitive comparisons for ASCII characters but for PostgreSQL the `citext` extension would be required.

No, how it's supposed to work (at least according to MSC threads) is:
1. "under the hood" (in the DB, in federation traffic, in JSON) the usernames are fully case sensitive and case is always preserved
[...]

PR #2922, opened 2022-12-23
Fix `/login` issue causing wrong device list updates

S7evinK:
Fixes https://github.com/matrix-org/dendrite/issues/2914 and possibly https://github.com/matrix-org/dendrite/issues/2073?

### DIFF
--- a/clientapi/auth/login_test.go
+++ b/clientapi/auth/login_test.go
@@ -22,6 +22,7 @@ import (
 	"testing"
 
 	"github.com/matrix-org/dendrite/clientapi/jsonerror"
+	"github.com/matrix-org/dendrite/clientapi/userutil"
 	"github.com/matrix-org/dendrite/setup/config"
 	uapi "github.com/matrix-org/dendrite/userapi/api"
 	"github.com/matrix-org/gomatrixserverlib"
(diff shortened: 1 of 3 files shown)


kegsay (changes requested):
Make `PasswordRequest.Username()` return the lower-cased correct username by default please. Then add something like `.RawUsername()` for matching when there is no matching user.

S7evinK:
> Make `PasswordRequest.Username()` return the lower-cased correct username by default please. Then add something like `.RawUsername()` for matching when there is no matching user.

I'm currently not seeing the benefit of that. With that we'd need to actually take care of those cases where we use `.Username()` right now. e.g. when deactivating an account, we would have to try both cases (raw and lower-cased) instead of setting it correctly once on login.

kegsay:
The benefit is that you don't need to set the field (or forget to set the field..), and it makes it clear what is processed/validated input vs untrusted/potentially invalid input. When verifying the account exists, we are forced unfortunately to check both, but then once this check has been done then code which just wants to know the username can just call `.Username()`?

kegsay (approved):
After talking it through, LGTM

PR #2930, opened 2023-01-05
Fix room summary returning wrong heroes

### LINKED ISSUE #2910: DM room summary can have incorrect entries for m.heroes when switching accounts
### Background information
- Dendrite built locally from git, tag v0.10.8
- Monolith
- Postgres: psql (13.8 (Debian 13.8-0+deb11u1))
- Docker: no, systemd
- Go: 1.19.1
- Element Web version: 1.11.16 / Olm version: 3.2.12

### Description

I have some pre-existing DMs created using my matrix.org user. I have invited my new Dendrite user, and then left the room with my matrix.org user. At this point, the Element Web UI displays the room name as "$myuser and 0 others"

image

Discussion in #dendrite:matrix.org pointed me to the /sync summary, and it seems the response contains multiple hero entries:

```
> sync$rooms$join$`!(redacted-roomid):matrix.org`$summary
$m.heroes
$m.heroes[[1]]
[1] "@gwmngilfen:matrix.org"

$m.heroes[[2]]
[1] "@myfriend:anotherserver.foo"

$m.joined_member_count
[1] 2

$m.invited_member_count
[1] 0
```

So there appears to be an entry for my matrix.org ID, even though I left the room with that account. This only seems to happen in Element Web - SchildiChat displays the DM name correctly. I've been noticing this for a while, so it's hard to say when it started.

### Steps to reproduce
- Create a DM using a matrix.org account (or presumably any other server, but that is untested) 
- Invite a Dendrite user
- Leave the room
- Observe the room name from Element Web, logged in as the Dendrite user
- Check the sync object for the Dendrite user

S7evinK:
This should fix #2910.
Probably makes Sytest/Complement a bit upset, since this not using `sort.Strings` anymore.


syncapi/storage/shared/storage_sync.go
```diff
@@ -92,8 +93,61 @@ func (d *DatabaseTransaction) MembershipCount(ctx context.Context, roomID, membe
...
+		return summary, err
+	}
+
+	for _, ev := range evs {
+		switch ev.Type() {
+		case gomatrixserverlib.MRoomName:
+			if gjson.GetBytes(ev.Content(), "name").Str != "" {
+				return summary, nil
+			}
+		case gomatrixserverlib.MRoomCanonicalAlias:
```
kegsay, line 128: Why do we bail early if there is a name/canonical alias? I appreciate they take precedence over the calculated hero name, but surely we need to return the heroes regardless?

kegsay (approved):
It's unclear why we moved from using the memberships table to using the current state table though. Why?

S7evinK:
> It's unclear why we moved from using the memberships table to using the current state table though. Why?

If a user joins a room and leaves afterwards, the join membership still exists in the memberships table, resulting in wrong heroes returned (#2910).

PR #2939, opened 2023-01-14
Switch the default config option values for `recaptcha_sitekey_class` and `recaptcha_form_field`

anoadragon453:
Attempting to use the [web auth fallback mechanism](https://spec.matrix.org/v1.5/client-server-api/#fallback) for Google ReCAPTCHA with the default setting for `client_api.recaptcha_sitekey_class` of "g-recaptcha-response" results in no captcha being rendered:

image

I cross-checked the captcha code between [dendrite.matrix.org's fallback page](https://dendrite.matrix.org/_matrix/client/r0/auth/m.login.recaptcha/fallback/web?session=asdhjaksd) and [matrix-client.matrix.org's one](https://matrix-client.matrix.org/_matrix/client/r0/auth/m.login.recaptcha/fallback/web?session=asdhjaksd) (which both use the same captcha public key) and noticed a discrepancy in the `class` attribute of the div that renders the captcha. [ReCAPTCHA's docs state](https://developers.google.com/recaptcha/docs/v3#automatically_bind_the_challenge_to_a_button) to use "g-recaptcha" as the class for the submit button.

I noticed this when user `@parappanon:parappa.party` reported that they were also seeing no captcha being rendered on their Dendrite instance. Changing `client_api.recaptcha_sitekey_class` to "g-recaptcha" caused their captcha to render properly as well.

There may have been a change in the class name from ReCAPTCHA v2 to v3? The [docs for v2](https://developers.google.com/recaptcha/docs/display#auto_render) also request one uses "g-recaptcha" though.

Thus I propose changing the default setting to unbreak people's recaptcha auth fallback pages. Should fix dendrite.matrix.org as well.

### DIFF
--- a/setup/config/config_clientapi.go
+++ b/setup/config/config_clientapi.go
@@ -85,7 +85,7 @@ func (c *ClientAPI) Verify(configErrs *ConfigErrors, isMonolith bool) {
 			c.RecaptchaApiJsUrl = "https://www.google.com/recaptcha/api.js"
 		}
 		if c.RecaptchaFormField == "" {
-			c.RecaptchaFormField = "g-recaptcha"
+			c.RecaptchaFormField = "g-recaptcha-response"
 		}
 		if c.RecaptchaSitekeyClass == "" {
 			c.RecaptchaSitekeyClass = "g-recaptcha"
--- a/setup/config/config_clientapi.go
+++ b/setup/config/config_clientapi.go
@@ -88,7 +88,7 @@ func (c *ClientAPI) Verify(configErrs *ConfigErrors, isMonolith bool) {
 			c.RecaptchaFormField = "g-recaptcha"
 		}
 		if c.RecaptchaSitekeyClass == "" {
-			c.RecaptchaSitekeyClass = "g-recaptcha-response"
+			c.RecaptchaSitekeyClass = "g-recaptcha"
 		}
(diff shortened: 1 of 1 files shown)


S7evinK (approved):

setup/config/config_clientapi.go
```diff
@@ -88,7 +88,7 @@ func (c *ClientAPI) Verify(configErrs *ConfigErrors, isMonolith bool) {
 			c.RecaptchaFormField = "g-recaptcha"
```
anoadragon453, line 88: This should be `g-recaptcha-response` instead - it's what recaptcha sends to the server upon a successful form submission.

```suggestion
			c.RecaptchaFormField = "g-recaptcha-response"
```

anoadragon453:
@S7evinK I've updated this PR to also change the default of the `client_api.recaptcha_form_field` option. My theory is that the default values for the `recaptcha_sitekey_class` and `recaptcha_form_field` options were accidentally swapped.

S7evinK (approved):

anoadragon453:
@S7evinK CI failures appear unrelated to the PR. Also heads up I don't have perms to merge this PR, if you're waiting on me to do so :)

S7evinK:
Yep, they are related to https://github.com/matrix-org/complement/commit/8550c224815cbcc0e15caa94c9a169df1eb82a30 I think (at least for Complement)

PR #2940, opened 2023-01-16
Omit state field from `/messages` response if empty

handlerug:
The field type is `[ClientEvent]` in the [spec](https://spec.matrix.org/v1.5/client-server-api/#get_matrixclientv3roomsroomidmessages), but right now `null` can also be returned. Omit the field completely if it's empty. Some clients (rightfully) assume it's either not present at all or it's of the right type (see https://github.com/matrix-org/matrix-react-sdk/pull/9913).

### Pull Request Checklist

    * The PR is a simple struct tag fix

Signed-off-by: `Umar Getagazov <umar@handlerug.me>`


S7evinK:
Thanks!

PR #2948, opened 2023-01-20
Make tests more reliable

S7evinK:
When using `testrig.CreateBase` and then using that base for other `NewInternalAPI` calls, we never actually shutdown the components.
`testrig.CreateBase` returns a `close` function, which only removes the database, so still running components have issues connecting to the database, since we ripped it out underneath it - which can result in "Disk I/O" or "pq deadlock detected" issues.

### DIFF
--- a/federationapi/federationapi_test.go
+++ b/federationapi/federationapi_test.go
@@ -273,7 +273,7 @@ func TestRoomsV3URLEscapeDoNot404(t *testing.T) {
 	cfg.Global.ServerName = gomatrixserverlib.ServerName("localhost")
 	cfg.Global.PrivateKey = privKey
 	cfg.Global.JetStream.InMemory = true
-	base := base.NewBaseDendrite(cfg, "Monolith")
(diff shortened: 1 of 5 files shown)


federationapi/federationapi_test.go
```diff
@@ -273,7 +273,7 @@ func TestRoomsV3URLEscapeDoNot404(t *testing.T) {
 	cfg.Global.ServerName = gomatrixserverlib.ServerName("localhost")
 	cfg.Global.PrivateKey = privKey
 	cfg.Global.JetStream.InMemory = true
-	base := base.NewBaseDendrite(cfg, "Monolith")
+	base := base.NewBaseDendrite(cfg, "Monolith", base.DisableMetrics)
```
kegsay, line 276: This is terribly confusing because there is also `base.EnableMetrics`, however that is a struct field whereas this I guess is a static const? Can we not shadow the package name `base` with the variable `base` please!

kegsay (approved):

PR #2953, opened 2023-01-26
Docs restructure

S7evinK:
Needs to be merged into `gh-pages` later on.


clientapi/admin_test.go
```diff
@@ -133,7 +133,11 @@ func TestPurgeRoom(t *testing.T) {
 		cfg, processCtx, close := testrig.CreateConfig(t, dbType)
 		caches := caching.NewRistrettoCache(128*1024*1024, time.Hour, caching.DisableMetrics)
 		natsInstance := jetstream.NATSInstance{}
-		defer close()
+		defer func() {
+			// give components the time to process purge requests
+			time.Sleep(time.Millisecond * 50)
```
devonh, line 138: Seems like this will still be quite flakey. Is there a better way to handle this in the test?

docs/INSTALL.md
```diff
@@ -8,6 +8,6 @@ or alternatively, in the [installation](installation/) folder:
 2. [Setting up the domain](installation/2_domainname.md)
 3. [Preparing database storage](installation/3_database.md)
 4. [Generating signing keys](installation/4_signingkey.md)
-5. [Installing as a monolith](installation/5_install_monolith.md)
+5. [Installing Dendrite](installation/5_install_monolith.md)
 6. [Populate the configuration](installation/7_configuration.md)
-7. [Starting the monolith](installation/8_starting_monolith.md)
+7. [Starting Dendrite](installation/8_starting_monolith.md)
```
devonh, line 13: The filename still refers to the old monolith.md

docs/installation/manual/3_configuration.md
```diff
@@ -79,22 +79,17 @@ You do not need to configure the `storage_path` when using a standalone NATS Ser
...
+If you want to use a single connection pool to a single PostgreSQL database, 
+then you must uncomment and configure the `database` section within the `global` section:
 
 ```yaml
 global:
   # ...
   database:
     connection_string: postgres://user:pass@hostname/database?sslmode=disable
-    max_open_conns: 100
+    max_open_conns: 90
```
devonh, line 92: out of curiosity, why the change in number here? :)

S7evinK: As mentioned, the default configuration for Postgres is set to 100, but has reserved 3 slots for administrative usage. So having 90 conns for Dendrite sounds like a fair trade. (Even though we probably could go further down)

PR #2960, opened 2023-01-31
Roomserver published pkey migration

devonh:
Adds a missed migration to update the primary key on the roomserver_published table in postgres.
Primary key was changed in #2836.


S7evinK (approved):
Looks good to me.

PR #2972, opened 2023-02-15
Merge keyserver & userapi

S7evinK:
As discussed yesterday, a first draft of merging the keyserver and the userapi.


kegsay (approved):
Changes seem good overall, let's land this before it gets too big. Please ensure upgrades work in terms of it preserving the data in keyserver DB (and keep that DB separate in SQLite mode).

Issue #2924, opened 2022-12-25
Support listening on Unix socket


Saklad5:
Dendrite currently only supports listening on TCP ports, one for HTTP and one for HTTPS. Since it is already designed with reverse proxies in mind, it would be much more helpful to support listening on a Unix domain socket.

Unix domain sockets are much simpler and efficient, particularly with regards to access control: configuring them to only be usable by desired users (e.g. nginx) allows connections to remain secure from other users on the server without requiring TLS or other complications like listening on the loopback interface does.

This also has the obvious benefit of not having to keep track of arbitrary ports and addresses.

When implementing this, it is vital that permissions for the socket are configurable: if anyone can access it, it is only marginally better than a loopback address.

S7evinK:
Fixed with #2974

PR #2974, opened 2023-02-16
unix socket support

cyberb:
### Pull Request Checklist

Signed-off-by: `Boris Rybalkin <ribalkin@gmail.com>`

I need this for Syncloud project (https://github.com/syncloud/platform) where I run multiple apps behind an nginx on the same RPi like device so unix socket is very convenient to not have port conflicts between apps.
Also someone opened this Issue: https://github.com/matrix-org/dendrite/issues/2924


cyberb:
I think I fixed most of lint errors from the first run.
Is it possible to run locally the checks?

cmd/dendrite/main.go
```diff
@@ -30,16 +31,33 @@ import (
 )
 
 var (
-	httpBindAddr  = flag.String("http-bind-address", ":8008", "The HTTP listening port for the server")
-	httpsBindAddr = flag.String("https-bind-address", ":8448", "The HTTPS listening port for the server")
-	certFile      = flag.String("tls-cert", "", "The PEM formatted X509 certificate to use for TLS")
-	keyFile       = flag.String("tls-key", "", "The PEM private key to use for TLS")
+	unixSocket           = flag.String("unix-socket", "", "The HTTP listening unix socket for the server")
+	unixSocketPermission = flag.Int("unix-socket-permission", 0755, "The HTTP listening unix socket permission for the server")
```
kegsay, line 35: Please mark these flags as experimental and not stable, as we will likely move these around over the coming months. Just put `EXPERIMENTAL(unstable): ` in each description line.

cmd/dendrite/main.go
```diff
@@ -30,16 +31,33 @@ import (
...
+			logrus.WithError(err).Fatalf("Failed to parse http address")
+		}
+		httpAddr = http
+		https, err := config.HttpAddress("https://" + *httpsBindAddr)
+		if err != nil {
+			logrus.WithError(err).Fatalf("Failed to parse https address")
+		}
+		httpsAddr = https
+	} else {
+		httpAddr = config.UnixSocketAddress(*unixSocket, fs.FileMode(*unixSocketPermission))
```
kegsay, line 58: The flag description doesn't make it clear that this is either/or. I'm happy for it to be either/or, but you need to document it to the end user.

setup/config/config_address.go
```diff
@@ -0,0 +1,45 @@
...
+	} else {
+		return NetworkTcp
+	}
+}
+
+func UnixSocketAddress(path string, perm fs.FileMode) ServerAddress {
+	return ServerAddress{Address: path, Scheme: NetworkUnix, UnixSocketPermission: perm}
+}
+
+func HttpAddress(urlAddress string) (ServerAddress, error) {
```
kegsay, line 39: Idiomatic Go would have this be `func HTTPAddress`.

setup/config/config_address.go
```diff
@@ -0,0 +1,45 @@
+package config
+
+import (
+	"io/fs"
+	"net/url"
+)
+
+const (
+	NetworkTcp  = "tcp"
```
kegsay, line 9: Idiomatic Go would have this be `NetworkTCP`.

setup/base/base.go
```diff
@@ -437,9 +434,27 @@ func (b *BaseDendrite) SetupAndServeHTTP(
 					}
 				}
 			} else {
-				if err := externalServ.ListenAndServe(); err != nil {
-					if err != http.ErrServerClosed {
-						logrus.WithError(err).Fatal("failed to serve HTTP")
+				if externalHTTPAddr.IsUnixSocket() {
+					_ = os.Remove(externalHTTPAddr.Address)
```
kegsay, line 438: Needs docs to explain why you're removing this and why you're ignoring the return variable.

setup/base/base_test.go
```diff
@@ -4,11 +4,15 @@ import (
...
 	"embed"
 	"html/template"
+	"net"
 	"net/http"
 	"net/http/httptest"
+	"path"
 	"testing"
 	"time"
 
+	"golang.org/x/net/context"
```
kegsay, line 14: Surely `"context"`?

kegsay (changes requested):
Broadly LGTM - a few small things, needs more documentation.

This is a rather esoteric feature, and we debated whether or not we even wanted to maintain this. However, it does have its uses in embedded contexts (if Dendrite is on the same device as the client, using unix sockets would be ideal) which is why this is being accepted into mainline.

cmd/dendrite/main.go
cyberb: fixed

cmd/dendrite/main.go
cyberb: added description

setup/config/config_address.go
cyberb: fixed

setup/base/base.go
cyberb: fixed the ignore.
I think there is a general consensus about removing socket files on start, I think it is useful, but I have strong opinion.

setup/base/base_test.go
cyberb: fixed

dvob:
I only now stumbled over this issue so I'm probably a little late to chime in.

@cyberb I also think it would be nice to move all settings into the config struct to unify the configuration a little bit.

Concerning the new options for listening on a unix socket I ask my self if we really need two additional flags/options for this? Other servers usually solve this with a single configuration called listen address or bind address. See for example:
* [Nginx](https://nginx.org/en/docs/http/ngx_http_core_module.html#listen) `127.0.0.1:8000` or `unix:/path/to/socket`
* [Caddy](https://caddyserver.com/docs/conventions#network-addresses) `127.0.0.1:8000` or `unix//path/to/socket`
* [Docker](https://docs.docker.com/engine/reference/commandline/dockerd/#daemon-socket-option) `127.0.0.1:8000` or `unix:///path/to/socket`

Recently when I added support for Let's Encrypt (see https://github.com/dvob/dendrite/commit/f120c42e8d4f3ee061c9523c362f6db9295f129f) in a private branch I also asked my self if there are really both flags needed: `-https-bind-address`and `-http-bind-address`? Do we want to support Dendrite running on multiple interfaces?

I suggest to always only listen on one interface an hence only have one flag `--bind-address` which uses a similar format to an existing tool.
To enable TLS we either use a separate flag like `--tls-enable` or it could be enabled implicitly if a certificate is configured (`--tls-cert`).

Examples:
```
# listen on 0.0.0.0:8008
./dendrite --bind-address :8008

# listen on 192.168.1.5:443 with tls
./dendrite --bind-address 192.168.1.5:443  --tls-cert cert.pem --tls-key key.pem

# listen on unix socket
./dendrite --bind-address unix//path/to/socket
```

cyberb:
[dvob](https://github.com/dvob) I can do whatever is needed just need code reviewers to tell me what should I do to get this merged

cyberb:
@kegsay could you review the fixes to your comments?

setup/base/base.go
```diff
@@ -435,18 +435,21 @@ func (b *BaseDendrite) SetupAndServeHTTP(
 				}
 			} else {
 				if externalHTTPAddr.IsUnixSocket() {
-					_ = os.Remove(externalHTTPAddr.Address)
+					err := os.RemoveAll(externalHTTPAddr.Address)
```
kegsay, line 438: Why did you change this to `RemoveAll`? The address should be pointing to a file, not a directory.

kegsay: If you're worried about panicing on a "not found error" then you need to check it via `errors.Is(err, fs.ErrNotExist)`

cyberb: Just a nice function to do a cleanup which gives you only errors you care about, I can change it if you want.
Do you not use removeall for files?

kegsay: No because if you point this at a directory it'll remove that and all files within it.

cyberb: fixed

kegsay (approved):

cyberb:
@kegsay sorry was this merged or not?
PR mentions merge commit but I cannot find it on main branch and PR is still open.

S7evinK:
Now it is merged :)
Thanks!

bones-was-here:
It looks like there's a minor issue where if you specify `--unix-socket-permission=666` the resulting permissions are a broken mess, but if the leading zero is included `--unix-socket-permission=0666` then it works properly.

cyberb:
Right, I do not really know what is right way of doing this, looks like pg allows with and without 0:

```
The parameter value is expected to be a numeric mode specified in the format accepted by the chmod and umask system calls. (To use the customary octal format the number must start with a 0 (zero).)

The default permissions are 0777, meaning anyone can connect. Reasonable alternatives are 0770
```

https://www.postgresql.org/docs/current/runtime-config-connection.html

Should I create a PR which will convert nnn to 0nnn in this case?

bones-was-here:
I would, but I'm pedantic about robustness :)

S7evinK:
Indeed, a new pull request sounds good.

cyberb:
sure, here you are: https://github.com/matrix-org/dendrite/pull/3014

PR #2976, opened 2023-02-16
Unify logging by using logrus for jetstream logs

dvob:
I guess tests for the logging is rather unusual so I omitted tests for this change.

Signed-off-by: `David Schneider <dsbrng25b@gmail.com>`


setup/jetstream/nats.go
```diff
@@ -53,7 +53,7 @@ func (s *NATSInstance) Prepare(process *process.ProcessContext, cfg *config.JetS
 		if err != nil {
 			panic(err)
 		}
-		s.ConfigureLogger()
+		s.SetLogger(NewLogAdapter(), false, false)
```
S7evinK, line 56: Maybe worth to explain that `false, false` is `debugFlag` and `traceFlag`

setup/jetstream/log.go
```diff
@@ -0,0 +1,42 @@
...
+		entry: logrus.StandardLogger().WithField("component", "jetstream"),
+	}
+}
+
+func (l *LogAdapter) Noticef(format string, v ...interface{}) {
+	l.entry.Infof(format, v...)
+}
+
+func (l *LogAdapter) Warnf(format string, v ...interface{}) {
+	l.entry.Errorf(format, v...)
```
S7evinK, line 25: Looks like there are a few copy and paste mistakes? e.g. `Warnf` logging on `Errorf` and `Fatalf` only on `Errorf`?

S7evinK (changes requested):

setup/jetstream/nats.go
dvob: Changed this to use the values from the NATS options which mimics the exact behavior of ConfigureLogger which was used before.

dvob:
@S7evinK Thank you for the review. 

Yes these there were some copy/paste mistakes which I fixed.

I also changed the call to SetLogger to use the values from the NATS options. This should make the call more clear and behave exactly as ConfigureLogger would.

S7evinK (approved):
Looks good! Thanks!

PR #2977, opened 2023-02-17
Fix SQLite `session_id`

S7evinK:
This fixes an issue with device_id/session_ids.
If a `device_id` is reused, we would reuse the same `session_id`, since we delete one device and insert a new one directly, resulting in the query to get a new `session_id` to return the previous session_id. (`SELECT count(access_token)`)


hughns:
For future reference, the complement test ended up in https://github.com/matrix-org/complement/pull/622

Issue #2980, opened 2023-02-19
Device keys are lost when logging in with existing device ID


tulir:
### Background information
- **Dendrite version or git SHA**: v0.11.1
- **SQLite3 or Postgres?**: Postgres
- **Running in Docker?**: Yes
- **`go version`**: 1.19
- **Client used (if applicable)**: mautrix-go

### Description

- **What** is the problem: device e2ee keys are lost when 
- **Who** is affected: clients that relogin instead of persisting access token
- **How** is this bug manifesting: either encryption is broken, or the client refuses to start because it validates that the user or server didn't mess up
- **When** did this first appear: after restarting the client

### Steps to reproduce
- log in with an account and set up encryption keys normally
- restart the program and log in with the same device id
- query keys for device

Expected result: query returns the keys uploaded in step 1
Actual result: query returns no keys

https://github.com/mautrix/go/tree/master/example can be used for testing. The program will panic on the second run because it validates that the keys are on the server on startup and is sad when it can't find the keys.

tulir:
Maybe caused by https://github.com/matrix-org/dendrite/blob/v0.11.1/userapi/internal/api.go#L269-L270 not checking if it's an existing device?

PR #2982, opened 2023-02-20
Fix issue where device keys are removed if a device ID is reused

S7evinK:
Fixes https://github.com/matrix-org/dendrite/issues/2980


kegsay (approved):
Please write tests.

PR #2989, opened 2023-02-28
Refactor `StoreEvent`, add `MaybeRedactEvent`, create an `EventDatabase`

S7evinK:
This PR changes the following:
- `StoreEvent` now only stores an event (and possibly prev event), instead of also doing redactions
- Adds a `MaybeRedactEvent` (pulled out from `StoreEvent`), which should be called after storing events
- a few other things


federationapi/routing/events.go
```diff
@@ -38,7 +38,7 @@ func GetEvent(
 	if err != nil {
 		return *err
 	}
-	event, err := fetchEvent(ctx, rsAPI, eventID)
```
kegsay, line 41: Needs comments.

roomserver/api/query.go
```diff
@@ -86,6 +86,7 @@ type QueryStateAfterEventsResponse struct {
 
 // QueryEventsByIDRequest is a request to QueryEventsByID
 type QueryEventsByIDRequest struct {
+	RoomID string `json:"room_id"`
```
kegsay, line 89: Needs comments on what happens if this is empty.

roomserver/internal/helpers/helpers_test.go
```diff
@@ -38,7 +38,7 @@ func TestIsInvitePendingWithoutNID(t *testing.T) {
 		var authNIDs []types.EventNID
 		for _, x := range room.Events() {
 
-			roomNID, err := db.GetOrCreateRoomNID(context.Background(), x.Unwrap())
+			roomNID, roomInfo, err := db.GetOrCreateRoomNID(context.Background(), x.Unwrap())
```
kegsay, line 41: If we are returning a `roomInfo` here surely we can remove the `roomNID` as well as the info will have it?

roomserver/internal/query/query.go
```diff
@@ -138,18 +138,39 @@ func (r *Queryer) QueryEventsByID(
 	request *api.QueryEventsByIDRequest,
 	response *api.QueryEventsByIDResponse,
```
kegsay, line 139: This API needs to make it very clear that you cannot pull events for multiple rooms in a single request.

roomserver/storage/shared/storage.go
```diff
@@ -28,24 +27,48 @@ import (
...
-	InvitesTable        tables.Invites
-	MembershipTable     tables.Membership
-	PublishedTable      tables.Published
 	RedactionsTable     tables.Redactions
-	Purge               tables.Purge
-	GetRoomUpdaterFn    func(ctx context.Context, roomInfo *types.RoomInfo) (*RoomUpdater, error)
+}
+
+// GetEventDatabase returns an EventDatabase to work with events only.
+func (d *Database) GetEventDatabase() *EventDatabase {
```
kegsay, line 59: Why do we need a getter? Constructors are making the embedded `EventDatabase` anyway, so callers can just do `db.EventDatabase.whatever`?

PR #2995, opened 2023-03-02
Make redaction check easier to read

S7evinK:
We need to check the redaction PL in Dendrite, if we do it in GMSL, we end up not sending the event to the output stream because it will be rejected.


roomserver/internal/input/input_events.go
```diff
@@ -402,6 +384,23 @@ func (r *Inputer) processRoomEvent(
 		}
 	}
 
+	// if storing this event results in it being redacted then do so.
```
kegsay, line 387:
```suggestion
	// if storing this event results in it being redacted then do so.
	// we do this after calculating state for this event as we may need to get power levels
```

kegsay (approved):

Issue #644, opened 2019-01-08
Implement missing registration features


babolivier:
- [ ] Save/reuse session parameters
- [ ] Remove old sessions. Need to do so on a session-specific timeout.
- [ ] Use device ID in request to create device if present
- [x] "Enable registration" config flag
- [ ] Guest account upgrading

anstue:
There is already a `registration_disabled` flag which is used accordingly in code. I guess we can check that box.

Kegsay:
This issue has too many disparate bits in it, closing for more concrete issues.

PR #2996, opened 2023-03-03
Add basic runtime tracing

S7evinK:
This allows us in almost all places to use regions to further trace down long running tasks.
Also removes an unused function.

### DIFF
--- a/internal/httputil/httpapi.go
+++ b/internal/httputil/httpapi.go
@@ -21,12 +21,12 @@ import (
 	"net/http/httptest"
 	"net/http/httputil"
 	"os"
+	"runtime/trace"
 	"strings"
 
 	"github.com/getsentry/sentry-go"
 	"github.com/matrix-org/util"
 	"github.com/opentracing/opentracing-go"
-	"github.com/opentracing/opentracing-go/ext"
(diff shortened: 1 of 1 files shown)


kegsay:
` FAILURE: #644: New federated private chats get full presence information (SYN-115)` is failing, probably flakey.

kegsay (changes requested):

kegsay:
Can you also please give a screenshot of what this looks like in the runtime/trace web viewer?

roomserver/storage/shared/storage.go
```diff
@@ -1020,7 +1020,7 @@ func (d *EventDatabase) MaybeRedactEvent(
 			return fmt.Errorf("d.RedactionsTable.MarkRedactionValidated: %w", err)
 		}
 
-		d.Cache.StoreRoomServerEvent(redactedEvent.EventNID, redactedEvent.Event)
+		d.Cache.InvalidateRoomServerEvent(redactedEvent.EventNID)
```
kegsay, line 1023: Why? Just because the event is redacted doesn't mean it should be removed from the cache?

PR #2997, opened 2023-03-04
Update admin-promotion instructions; clarify ID for evacuation

timmc:
Table name has changed since instructions were written.

There's probably a better way to describe how to get the internal room ID than I've attempted here, so feel free to adjust as needed. (It may even be good to show an example of what an internal room ID looks like, e.g. `!nc93825:example.com`)

### Pull Request Checklist

  * Doc-only change
* [x] Pull request includes a [sign off below using a legally identifiable name](https://matrix-org.github.io/dendrite/development/contributing#sign-off)

Signed-off-by: `Tim McCormack <cortex@brainonfire.net>`

### DIFF
--- a/docs/administration/4_adminapi.md
+++ b/docs/administration/4_adminapi.md
@@ -22,12 +22,12 @@ curl --header "Authorization: Bearer <access_token>" -X <POST|GET|PUT> <Endpoint
 An `access_token` can be obtained through most Element-based matrix clients by going to `Settings` -> `Help & About` -> `Advanced` -> `Access Token`.
 Be aware that an `access_token` allows a client to perform actions as an user and should be kept **secret**.
 
-The user must be an administrator in the `account_accounts` table in order to use these endpoints.
+The user must be an administrator in the `userapi_accounts` table in order to use these endpoints.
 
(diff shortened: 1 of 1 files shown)


docs/administration/4_adminapi.md
```diff
@@ -38,6 +38,10 @@ This endpoint will instruct Dendrite to part all local users from the given `roo
 in the URL. It may take some time to complete. A JSON body will be returned containing
 the user IDs of all affected users.
 
+If the room has a published address set, the room's ID will not be in the URL, but may
+be found as the room's "internal ID" (e.g. under Advanced within the room settings in
+Element.)
```
S7evinK, line 43: Does this sound OK?
```suggestion
If the room has an alias set (e.g. is published), the room's ID will not be visible in the URL, but it can
be found as the room's "internal ID" in Element Web (Settings -> Advanced)
```

S7evinK (approved):

docs/administration/4_adminapi.md
timmc: Yes, that looks better, thank you!

S7evinK:
Thanks!

PR #3008, opened 2023-03-10
Update ACLs when received as outliers

### LINKED ISSUE #3004: Dendrite does not respect m.room.server_acl on room join
### Background information

- **Dendrite version or git SHA**: 7d83f8b6332cc41f0ac721d014aaad535069282f
- **Running in Docker?**: false
- **`go version`**: `go version go1.18.1 linux/amd64`

### Description

- **What** is the problem: Dendrite does not respect `m.room.server_acl`
- **Who** is affected: Dendrite servers joined the room **after** this rule was set
- **How** is this bug manifesting: Dendrite will show all events in the room, including events from servers denied with [m.room.server_acl](https://spec.matrix.org/v1.4/client-server-api/#mroomserver_acl)
- **When** did this first appear: This is relevant at least since [Dendrite 0.10.7](https://github.com/matrix-org/dendrite/releases/tag/v0.10.7)

### Steps to reproduce

1. Prepare 3 or more servers. 1st will be room host, 2nd will be blocked with [m.room.server_acl](https://spec.matrix.org/v1.4/client-server-api/#mroomserver_acl) and 3rd server will be affected by this issue. 1st and 2nd servers don't have to be powered by dendrite.
2. Create room on 1st server
3. Join this room from 2nd server
4. Add `deny` rule, to block 2nd server from participating
5. Join from 3rd (dendrite) server.
6. Any events sent by 2nd server will not appear for users from 1st server, but will be visible for users from 3rd server
7. You can set [m.room.server_acl](https://spec.matrix.org/v1.4/client-server-api/#mroomserver_acl) again, to "fix" this.

Alternatively, you can join #russian:kde.org using dendrite server. This room has t2bot.io in deny list. So from any dendrite server you will be able to see messages, hidden for everyone else.

<!--

``` (three backticks, on a line on their own), so that they are formatted legibly.

-->

### PR DESCRIPTION
**Pull request by S7evinK**
This should fix #3004 by making sure we also update our in-memory ACLs after joining a new room.
Also makes use of more caching in `GetStateEvent`

Bonus: Adds some tests, as I was about to use `GetBulkStateContent`, but turns out that `GetStateEvent` is basically doing the same, just that it only gets the `eventTypeNID`/`eventStateKeyNID` once and not for every call.

### CODE REVIEW

File: roomserver/internal/input/input_events.go [hunk 1]
```diff
@@ -576,15 +582,17 @@ func (r *Inputer) processStateBefore(
 	if rejectionErr = gomatrixserverlib.Allowed(event, &stateBeforeAuth); rejectionErr != nil {
 		return
 	}
-	// Work out what the history visibility was at the time of the
+	// Work out what the history visibility/ACLs was at the time of the
 	// event.
 	for _, event := range stateBeforeEvent {
+		if event.Type() == "m.room.server_acl" && event.StateKeyEquals("") {
+			r.ACLs.OnServerACLUpdate(event)
```
kegsay, line 589: It feels very wrong to side-effect and call this in a function called `processStateBefore`. What happens if we call `processStateBefore` with an older event?

roomserver/storage/shared/storage.go
```diff
@@ -1154,22 +1154,24 @@ func (d *Database) GetStateEvent(ctx context.Context, roomID, evType, stateKey s
...
 		// No rooms have an event of this type, otherwise we'd have an event type NID
 		return nil, nil
 	}
 	if err != nil {
 		return nil, err
 	}
-	stateKeyNID, err := d.EventStateKeysTable.SelectEventStateKeyNID(ctx, nil, stateKey)
+	eventTypeNID := eventTypeNIDMap[evType]
+	stateKeyNIDMap, err := d.eventStateKeyNIDs(ctx, nil, []string{stateKey})
 	if err == sql.ErrNoRows {
```
kegsay, line 1167: Does `eventStateKeyNIDs()` really return `sql.ErrNoRows`? Likewise for `eventTypeNIDs()`

roomserver/internal/input/input_events.go
```diff
@@ -576,15 +582,17 @@ func (r *Inputer) processStateBefore(
 	if rejectionErr = gomatrixserverlib.Allowed(event, &stateBeforeAuth); rejectionErr != nil {
 		return
 	}
-	// Work out what the history visibility was at the time of the
+	// Work out what the history visibility/ACLs was at the time of the
 	// event.
 	for _, event := range stateBeforeEvent {
+		if event.Type() == "m.room.server_acl" && event.StateKeyEquals("") {
```
kegsay, line 588: Please parameterise this into a constant. Even just somewhere in Dendrite is better than having to check each occurance for typos.

kegsay (changes requested):

roomserver/storage/shared/storage.go
S7evinK: `eventStateKeyNIDs` possibly yes, `eventTypeNIDs` not.

kegsay (approved):

kegsay (approved):

PR #3011, opened 2023-03-14
Update docs and sample config for the relay_api

akirk:
This adds an empty `relay_api` section to the sample configuration. For SQLite environments, or others where a `database.connection_string` is needed for each section, there should be an entry in the configuration sample as a basis.

This PR also changes the "Configuring Dendrite" documentation in that respect.

The requirement was introduced in #2917. When upgrading dendrite, it will complain about `relay_api.database.connection_string` not being configured.

### Pull Request Checklist

Signed-off-by: `Alex Kirk <akirk@users.noreply.github.com>`

### DIFF
--- a/dendrite-sample.yaml
+++ b/dendrite-sample.yaml
@@ -273,9 +273,6 @@ media_api:
       height: 480
       method: scale
 
-# Configuration for the Relay API.
-relay_api:
-
 # Configuration for enabling experimental MSCs on this homeserver.
 mscs:
   mscs:
--- a/dendrite-sample.yaml
+++ b/dendrite-sample.yaml
@@ -273,6 +273,9 @@ media_api:
       height: 480
       method: scale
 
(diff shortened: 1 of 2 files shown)


dendrite-sample.yaml
```diff
@@ -273,6 +273,9 @@ media_api:
       height: 480
       method: scale
 
+# Configuration for the Relay API.
```
S7evinK, line 276: I don't think we should add this, the sample is more for usage with a global connection pool.
May be worth to add a sample file for SQLite instead, where we definitely need this section including the `database.connection_string`

akirk: I see. Happy to remove this change here if you intend to create a SQLite sample file in another PR.

S7evinK (approved):

S7evinK:
Thanks!

PR #3016, opened 2023-03-16
Preparations for removing `BaseDendrite`

### LINKED ISSUE #2903: Dendrite returns improper error codes for unknown endpoints
Dendrite returns improper responses for unknown endpoints per MSC3743, see matrix-org/complement#565 for tests I'm adding for this.

Dendrite is:

* Returning a 404 instead of a 405 for an unknown methods on Client-Server APIs.
* Returning a text response, instead of JSON, error on `/media`, `/key`, `/federation` and unknown prefixes.

S7evinK:
Preparations to actually remove/replace `BaseDendrite`.
Quite a few changes:
- SyncAPI accepts an `fulltext.Indexer` interface (fulltext is removed from `BaseDendrite`)
- Caches are removed from `BaseDendrite`
- Introduces a `Router` struct (likely to change)
  - also fixes #2903
- Introduces a `sqlutil.ConnectionManager`, which should remove `base.DatabaseConnection` later on
- probably more

### DIFF
--- a/appservice/appservice_test.go
+++ b/appservice/appservice_test.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/matrix-org/dendrite/appservice"
 	"github.com/matrix-org/dendrite/appservice/api"
+	"github.com/matrix-org/dendrite/internal/caching"
 	"github.com/matrix-org/dendrite/roomserver"
 	"github.com/matrix-org/dendrite/setup/config"
 	"github.com/matrix-org/dendrite/test"
@@ -123,8 +124,9 @@ func TestAppserviceInternalAPI(t *testing.T) {
 			},
 		}
 
+		caches := caching.NewRistrettoCache(base.Cfg.Global.Cache.EstimatedMaxSize, base.Cfg.Global.Cache.MaxAge, caching.DisableMetrics)
 		// Create required internal APIs
-		rsAPI := roomserver.NewInternalAPI(base)
+		rsAPI := roomserver.NewInternalAPI(base, caches)
 		usrAPI := userapi.NewInternalAPI(base, rsAPI, nil)
 		asAPI := appservice.NewInternalAPI(base, usrAPI, rsAPI)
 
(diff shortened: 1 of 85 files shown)


internal/httputil/routing.go
```diff
@@ -15,7 +15,10 @@
 package httputil
 
 import (
+	"net/http"
 	"net/url"
+
+	"github.com/gorilla/mux"
```
kegsay, line 21: As a side note, do we have an issue to replace mux? We need to as it has been archived and isn't maintained anymore.

internal/sqlutil/connection_manager.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"database/sql"
+	"fmt"
+
+	"github.com/matrix-org/dendrite/setup/config"
+)
+
+type ConnectionManager interface {
+	Connection(dbProperties *config.DatabaseOptions) (*sql.DB, Writer, error)
```
kegsay, line 25: Not sure I see the value in interfacing this? Any concrete impl would need to return a DB which is pretty difficult to mock out!

kegsay (approved):
Great improvement so far! Let's land it.

appservice/appservice_test.go
```diff
@@ -123,8 +124,9 @@ func TestAppserviceInternalAPI(t *testing.T) {
 			},
 		}
 
+		caches := caching.NewRistrettoCache(base.Cfg.Global.Cache.EstimatedMaxSize, base.Cfg.Global.Cache.MaxAge, caching.DisableMetrics)
 		// Create required internal APIs
-		rsAPI := roomserver.NewInternalAPI(base)
+		rsAPI := roomserver.NewInternalAPI(base, caches)
```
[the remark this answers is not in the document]
S7evinK: Indeed, that is the idea. Gradually remove the need for `base`.

internal/httputil/routing.go
S7evinK: Now we have https://github.com/matrix-org/dendrite/issues/3017

PR #3023, opened 2023-03-21
Remove `BaseDendrite`

S7evinK:
Removes `BaseDendrite` to, hopefully, make testing and composing of components easier in the future.

### DIFF
--- a/appservice/appservice.go
+++ b/appservice/appservice.go
@@ -21,6 +21,7 @@ import (
 	"sync"
 	"time"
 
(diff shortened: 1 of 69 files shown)


kegsay (approved):
This is an improvement on the status quo because we now have each component only accepting what they actually _need_ rather than `BaseDendrite` everywhere. The component constructors are still very complex at present, but we can make composable intermediate structs in another PR. LGTM

PR #3025, opened 2023-03-23
Add `/search` tests

S7evinK:


syncapi/routing/search_test.go
```diff
@@ -0,0 +1,244 @@
...
+			name:      "with alice ID",
+			wantOK:    true,
+			searchReq: SearchRequest{},
+			device:    &aliceDevice,
+		},
+		{
+			name:   "searchTerm specified",
+			wantOK: true,
+			searchReq: SearchRequest{
+				SearchCategories: SearchCategories{RoomEvents: RoomEvents{SearchTerm: "hello"}},
```
kegsay, line 59: Can it do partial matches like `hell`? Can we also check that it isn't just doing prefix matching by searching for `world3`? Does it strip punctuation?

kegsay (changes requested):
Needs a bit more but otherwise LGTM

kegsay (approved):

PR #3027, opened 2023-03-24
Improve proxy support

paigeadelethompson:
### Pull Request Checklist

* [ ] I have added Go unit tests or [Complement integration tests](https://github.com/matrix-org/complement) for this PR _or_ I have justified why this PR doesn't need tests
* [ ] Pull request includes a [sign off below using a legally identifiable name](https://matrix-org.github.io/dendrite/development/contributing#sign-off) _or_ I have already signed off privately

Signed-off-by: `Paige Thompson <paige@paige.bio>`

### DIFF
--- a/appservice/appservice.go
+++ b/appservice/appservice.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"crypto/tls"
 	"net/http"
+	"net/url"
 	"sync"
 	"time"
 
@@ -51,7 +52,9 @@ func NewInternalAPI(
 			TLSClientConfig: &tls.Config{
 				InsecureSkipVerify: cfg.AppServiceAPI.DisableTLSValidation,
 			},
-			Proxy: http.ProxyFromEnvironment,
+			Proxy: func(req *http.Request) (*url.URL, error) {
+				return cfg.AppServiceAPI.Proxy.GetApplicableProxy(req, &cfg.Global.Proxy)
+			},
 		},
 	}
 	// Create appserivce query API with an HTTP client that will be used for all
--- a/internal/pushgateway/client.go
+++ b/internal/pushgateway/client.go
@@ -7,9 +7,11 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"net/url"
 	"time"
 
 	"github.com/matrix-org/dendrite/internal"
+	"github.com/matrix-org/dendrite/setup/config"
 )
 
 type httpClient struct {
@@ -17,15 +19,17 @@ type httpClient struct {
 }
 
 // NewHTTPClient creates a new Push Gateway client.
-func NewHTTPClient(disableTLSValidation bool) Client {
+func NewHTTPClient(cfg *config.Dendrite) Client {
 	hc := &http.Client{
 		Timeout: 30 * time.Second,
 		Transport: &http.Transport{
 			DisableKeepAlives: true,
 			TLSClientConfig: &tls.Config{
-				InsecureSkipVerify: disableTLSValidation,
+				InsecureSkipVerify: cfg.UserAPI.PushGatewayDisableTLSValidation,
 			},
-			Proxy: http.ProxyFromEnvironment,
+			Proxy: func(req *http.Request) (*url.URL, error) {
+				return cfg.UserAPI.Proxy.GetApplicableProxy(req, &cfg.Global.Proxy)
+			},
 		},
 	}
 	return &httpClient{hc: hc}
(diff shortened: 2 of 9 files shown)


paigeadelethompson:
Hey so I realize this probably a subject that nobody wants to talk about. Currently I have observed some problems with relying solely on the environment to specify `HTTP_PROXY` due to the fact that there's no way to specify an exclusion. And so it will attempt to proxy requests to app services thru the specified proxy. Ideally you would be able to specify an exclusion for a network or a single IP address to avoid proxying requests to addresses or networks that can't be reached via the specified proxy. 

I want to make sure I have some idea of what's expected for a pull request. In the past someone began working on this and then reverted their work to rely solely on the HTTP_PROXY environment variables for configuring this. It's not a bad way of doing it, but in and of itself it's very rudimentary and without the ability to exclude some destination networks from being proxied introduces a problem in which the proxy itself must determine the correct destination which may require a significant change to somebody's network when there doesn't really need to be.

Synapse never was perfect about this; the distinction between excluding addresses which just simply should never be reached by the server and excluding addresses just from being proxied. I think both are pretty important especially with regards to fetching media and other ways that clients can cause the server to hit sensitive internal addresses. They should both be distinctly configurable I think.

It's kind of late and I'm tired but I would be glad to continue working on this later. There are several areas of Dendrite in which http requests are made that I'm aware of, what I have in mind is something like: 

- HTTP_PROXY and HTTPS_PROXY take highest precedence
- Global proxy_outbound takes second highest precedence 
- and then for every section 1 level beneath global proxy_outbound can be specified (where applicable) and takes the lowest precedence

Regarding having networks or addresses that the server should never try to send requests to regardless of whether through a proxy or not I'm happy to leave that out of this scope and revisit it in another pull request. For this scope I would focus on a way to just exclude addresses and ranges from being proxied.

paigeadelethompson:
I'll leave this PR in draft until its actually ready or until I get some feedback and then finish whichever comes first

S7evinK:
Looking at the implementation of `ProxyFromEnvironment`, Go doesn't just use the `HTTP_PROXY`, it is totally aware of other env variables
https://cs.opensource.google/go/x/net/+/refs/tags/v0.8.0:http/httpproxy/proxy.go;l=89-96

```go
func FromEnvironment() *Config {
	return &Config{
		HTTPProxy:  getEnvAny("HTTP_PROXY", "http_proxy"),
		HTTPSProxy: getEnvAny("HTTPS_PROXY", "https_proxy"),
		NoProxy:    getEnvAny("NO_PROXY", "no_proxy"),
		CGI:        os.Getenv("REQUEST_METHOD") != "",
	}
}
```

paigeadelethompson:
yeah but none for making exceptions for networks or addresses that you don't want to use a proxy for. I had another idea that completely escaped me when I started writing this but it's possible to do interpolation of environment variables in yaml with go: https://mtyurt.net/post/go-using-environment-variables-in-configuration-files.html

You can also use setenv https://gobyexample.com/environment-variables

I almost said that the easiest way to accomplish this would be to unset the environment variable if the destination address of a request matches a list of addresses not intended to be proxied but the problem with that is that this is all threaded and unless there's a way to isolate coroutines to their own environment in golang, am I understanding correctly that coroutine environments are isolated by default? If that is the case then a coroutine that needs to make an exception for proxying to a specific destination you would just https://pkg.go.dev/os#Unsetenv before creating a Transport or getting the default. 

Unfortunately I'm not seeing anything that tells me definitively whether or not coro environments are isolated or shared. I saw something on stack but that's not reliable. I'll test this at some point if I can't find a definitive answer in a minute.

S7evinK:
https://cs.opensource.google/go/x/net/+/refs/tags/v0.8.0:http/httpproxy/proxy.go;l=38-51

Correct me if I am wrong or misunderstood, but that's what you'd like? So no special Config is needed for that.

paigeadelethompson:
No, not at all. The problem isn't whether or not PROXY_HTTP works it with the fact that it's all-on or all-off. There are networks that it needs to take exception to, such as 192.168.0.0/16 for example. I have much more specific prefixes myself, but this is a common thing to have with any application that supports proxying.

paigeadelethompson:
Hey! that might actually work! I will check in a sec

kegsay:
This feature falls outside the remit of Dendrite, and isn't really what we want as [per the contributing guidelines](https://github.com/matrix-org/dendrite/blob/main/docs/development/CONTRIBUTING.md#contribution-types). I see the value in allowing exclusions, but I don't feel putting that in Dendrite is the right thing to do. It should ideally be something your proxy of choice can manage. It looks like Go's `NO_PROXY` env var may do what you want.

paigeadelethompson:
Hey guys, thanks again, 

I think @S7evinK's recommendation will work ok

PR #3029, opened 2023-03-24
Add CS API device tests

S7evinK:
Adds tests for
- `/devices`
- `/delete_devices` (also adds UIA)

### DIFF
--- a/clientapi/clientapi_test.go
+++ b/clientapi/clientapi_test.go
@@ -37,42 +37,39 @@ func TestGetPutDevices(t *testing.T) {
 
 	test.WithAllDatabases(t, func(t *testing.T, dbType test.DBType) {
 		testCases := []struct {
-			name         string
-			requestUser  *test.User
-			deviceUser   *test.User
-			request      *http.Request
-			withDeviceID bool
-			wantOK       bool
-			validateFunc func(t *testing.T, device userDevice, routers httputil.Routers)
+			name           string
+			requestUser    *test.User
+			deviceUser     *test.User
+			request        *http.Request
+			wantStatusCode int
+			validateFunc   func(t *testing.T, device userDevice, routers httputil.Routers)
 		}{
 			{
-				name:        "can get all devices",
-				requestUser: alice,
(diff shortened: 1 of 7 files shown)


userapi/internal/user_api.go
```diff
@@ -386,11 +386,6 @@ func (a *UserInternalAPI) PerformDeviceUpdate(ctx context.Context, req *api.Perf
 	}
 	res.DeviceExists = true
 
-	if dev.UserID != req.RequestingUserID {
```
S7evinK, line 389: Removed because we should never ever hit this case, because we are querying the database for the `dev.UserID`. (except the database is lying to us?)

clientapi/clientapi_test.go
```diff
@@ -0,0 +1,376 @@
...
+					}
+					gotDisplayName := gjson.GetBytes(rec.Body.Bytes(), "display_name").Str
+					if gotDisplayName != "my new displayname" {
+						t.Fatalf("expected displayname '%s', got '%s'", "my new displayname", gotDisplayName)
+					}
+				},
+			},
+			{
+				// this should return "device does not exist"
+				name:         "can not update device for different user",
```
kegsay, line 92: AKA a 404? Maybe swap `wantOK` with `wantStatusCode: 200` and then you can assert the HTTP status code for errors?

clientapi/clientapi_test.go
```diff
@@ -0,0 +1,376 @@
...
+				requestUser:  alice,
+				deviceUser:   bob,
+				request:      httptest.NewRequest(http.MethodPut, "/_matrix/client/v3/devices/", strings.NewReader(`{"display_name":"my new displayname"}`)),
+				withDeviceID: true,
```
kegsay, line 96: This bool is pointless. It is always true when `deviceUser` is set. And always false when `deviceUser` is unset. Just use the presence of `deviceUser` instead.

userapi/internal/user_api.go
S7evinK: We need to look up the device, as one device can update another of the same user.

PR #3031, opened 2023-03-27
Don't fail config validation if existingSecret is provided but .Vaues.signing_key.create is false (use existing key)

Arkaniad:
This change allows Helm templating to proceed when using an existing key mounted from a Kubernetes Secret. Prior to this, templating would fail because `.Values.signing_key.create` is `false` as we don't want to create a new key, but this blocked templating even if we were mounting a secret and configuring it properly.

### Pull Request Checklist

  - Unneeded - Fix for Helm Chart deployment

Signed-off-by: Rhea Danzey <rdanzey@element.io>


Arkaniad:
Closing, consolidating in [#3033](https://github.com/matrix-org/dendrite/pull/3033)

PR #3034, opened 2023-03-28
Add tests for CSAPI membership changes

S7evinK:
Adds some more checks in regards to power levels, uses a less heavy way to get the membership of a user, avoids asking the database for the room version, since it will be queried later.


clientapi/routing/membership.go
```diff
@@ -250,20 +255,10 @@ func sendInvite(
...
-	if err == errMissingUserID {
-		return util.JSONResponse{
-			Code: http.StatusBadRequest,
-			JSON: jsonerror.BadJSON(err.Error()),
-		}, err
-	} else if err == eventutil.ErrRoomNoExists {
-		return util.JSONResponse{
-			Code: http.StatusNotFound,
-			JSON: jsonerror.NotFound(err.Error()),
-		}, err
```
S7evinK, line 265: Removed because `errMissingUserID` is never returned and `checkMemberInRoom` already checks if the room exists and returns `IsInRoom=false`, resulting in an error returned earlier.

clientapi/routing/membership.go
```diff
@@ -357,19 +352,7 @@ func loadProfile(
...
-	verReq := roomserverAPI.QueryRoomVersionForRoomRequest{RoomID: roomID}
-	verRes := roomserverAPI.QueryRoomVersionForRoomResponse{}
-	if err := rsAPI.QueryRoomVersionForRoom(req.Context(), &verReq, &verRes); err != nil {
-		resErr = &util.JSONResponse{
-			Code: http.StatusBadRequest,
-			JSON: jsonerror.UnsupportedRoomVersion(err.Error()),
-		}
-		return
-	}
-	roomVer = verRes.RoomVersion
```
S7evinK, line 372: When sending membership events, we are getting the create event, which also has the `roomVersion`.

clientapi/routing/membership.go
```diff
@@ -432,34 +415,17 @@ func checkAndProcessThreepid(
...
-	tuple := gomatrixserverlib.StateKeyTuple{
-		EventType: gomatrixserverlib.MRoomMember,
-		StateKey:  userID,
-	}
-	var membershipRes roomserverAPI.QueryCurrentStateResponse
-	err := rsAPI.QueryCurrentState(ctx, &roomserverAPI.QueryCurrentStateRequest{
-		RoomID:      roomID,
-		StateTuples: []gomatrixserverlib.StateKeyTuple{tuple},
+	var membershipRes roomserverAPI.QueryMembershipForUserResponse
+	err := rsAPI.QueryMembershipForUser(ctx, &roomserverAPI.QueryMembershipForUserRequest{
```
S7evinK, line 419: Only hits the database 3 times instead of 7 times.

kegsay (approved):

clientapi/routing/membership.go
S7evinK: `checkMemberInRoom` returns `IsInRoom=false`, either if the room doesn't exist or the user isn't in the room, that's what I mean.

PR #3040, opened 2023-03-30
Fix potential state reset when trying to join a room

S7evinK:
When trying to join a room in short sequence, it is possible that a state reset occurs.


kegsay (changes requested):
Needs comments.

PR #3064, opened 2023-04-21
Use IRoomVersion

kegsay:
This is a step towards allowing arbitrary room version impls.


federationapi/routing/invite.go
```diff
@@ -78,7 +77,7 @@ func InviteV1(
 ) util.JSONResponse {
 	roomVer := gomatrixserverlib.RoomVersionV1
 	body := request.Content()
-	event, err := roomVer.NewEventFromTrustedJSON(body, false)
+	event, err := gomatrixserverlib.MustGetRoomVersion(roomVer).NewEventFromTrustedJSON(body, false)
```
devonh, line 80: Should we add a comment here explaining why this MustGetRoomVersion is acceptable here? In case we move the hardcoded room version above later on

devonh (approved):

federationapi/routing/invite.go
kegsay: Sure!

PR #3065, opened 2023-04-24
Move high level room joining logic to GMSL

devonh:
GMSL PR: https://github.com/matrix-org/gomatrixserverlib/pull/372

### DIFF
--- a/federationapi/internal/federationclient.go
+++ b/federationapi/internal/federationclient.go
@@ -19,14 +19,11 @@ func (a *FederationInternalAPI) MakeJoin(
 ) (res gomatrixserverlib.MakeJoinResponse, err error) {
 	ctx, cancel := context.WithTimeout(ctx, defaultTimeout)
 	defer cancel()
-	ires, err := a.doRequestIfNotBlacklisted(s, func() (interface{}, error) {
-		return a.federation.MakeJoin(ctx, origin, s, roomID, userID)
-	})
+	ires, err := a.federation.MakeJoin(ctx, origin, s, roomID, userID)
 	if err != nil {
 		return &fclient.RespMakeJoin{}, err
 	}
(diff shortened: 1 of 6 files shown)


kegsay (approved):
Pending GMSL changes.

federationapi/internal/federationclient.go
```diff
@@ -9,14 +9,46 @@ import (
...
+	r := ires.(fclient.RespMakeJoin)
+	return &r, nil
+}
+
+func (a *FederationInternalAPI) SendJoin(
+	ctx context.Context, origin, s spec.ServerName, event *gomatrixserverlib.Event,
+) (res gomatrixserverlib.SendJoinResponse, err error) {
+	ctx, cancel := context.WithTimeout(ctx, defaultTimeout)
+	defer cancel()
+	ires, err := a.doRequestIfNotBlacklisted(s, func() (interface{}, error) {
```
kegsay, line 37: I like this _but_ the behaviour in the following situation feels wrong:
 - Server X is blacklisted.
 - Server X comes back online but Dendrite doesn't know it yet.
 - User wants to join a room via Server X.
 - Blacklist says no.
 - Dendrite marks that as a failure (Reachable=false), reinforcing the blacklist.

I don't know the full implications of this. If reinforcing the blacklist is a no-op then fine, but if it resets the timer or something then I can see this being a problem. The specification is silent on how best to handle blacklists, because after all we don't want to waste time sending requests to dead servers. However, the spec is particularly unclear on which endpoints should be considered "unblacklistable" - i.e always try them to see if the server is now up. We will unblacklist when the remote server sends us something (I hope hope hope) but that may be at some indeterminate point in the future.

The safest option would be to preserve the original implementation and don't check the blacklist. Longer term, we need to take a holistic view of this and figure out a way forward.

kegsay (approved):

PR #3066, opened 2023-04-25
Remove `PerformError`

S7evinK:
This removes `PerformError`, which was needed when we still had polylith. 

This removes quite a bunch of
```go
if err != nil {
	return err
}
if err := res.Error; err != nil {
	return err.JSONResponse()
}
```

Hopefully can be read commit by commit.


roomserver/internal/perform/perform_join.go
```diff
@@ -342,16 +314,10 @@ func (r *Joiner) performJoinRoomByID(
 			}
 			inputRes := rsAPI.InputRoomEventsResponse{}
 			if err = r.Inputer.InputRoomEvents(ctx, &inputReq, &inputRes); err != nil {
-				return "", "", &rsAPI.PerformError{
-					Code: rsAPI.PerformErrorNoOperation,
-					Msg:  fmt.Sprintf("InputRoomEvents failed: %s", err),
-				}
+				return "", "", rsAPI.ErrNotAllowed{Err: err}
```
kegsay, line 317: This probably isn't a "not allowed" error? It used to be err no op.

kegsay (approved):
Clearer API overall, LGTM!

PR #3071, opened 2023-04-28
Add key backup tests

S7evinK:
Also slightly refactors the functions and methods to rely less on the req/res pattern we had for polylith.

Returns `M_WRONG_ROOM_KEYS_VERSION` for some endpoints as per the spec

### DIFF
--- a/userapi/internal/user_api.go
+++ b/userapi/internal/user_api.go
@@ -683,22 +683,11 @@ func (a *UserInternalAPI) QueryOpenIDToken(ctx context.Context, req *api.QueryOp
 	return nil
 }
 
+func (a *UserInternalAPI) DeleteKeyBackup(ctx context.Context, userID, version string) (bool, error) {
+	return a.DB.DeleteKeyBackup(ctx, userID, version)
+}
+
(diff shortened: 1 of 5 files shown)


userapi/internal/user_api.go
```diff
@@ -683,62 +684,43 @@ func (a *UserInternalAPI) QueryOpenIDToken(ctx context.Context, req *api.QueryOp
...
-		exists, err := a.DB.DeleteKeyBackup(ctx, req.UserID, req.Version)
-		if err != nil {
-			res.Error = fmt.Sprintf("failed to delete backup: %s", err)
-		}
-		res.Exists = exists
-		res.Version = req.Version
-		return nil
-	}
+func (a *UserInternalAPI) DeleteKeyBackup(ctx context.Context, userID, version string) (bool, error) {
+	return a.DB.DeleteKeyBackup(ctx, userID, version)
```
kegsay, line 688: We've lost validation here. Previous code checked `version` was not empty.

kegsay (approved):

userapi/internal/user_api.go
S7evinK: This is already handled at the router level, with a missing `{version}` variable the functions isn't called. Added a test anyway.

PR #3074, opened 2023-05-03
Use PDU in even more places

kegsay:
- No longer rely on *Event returning from NewEventFrom... functions
 
Requires https://github.com/matrix-org/gomatrixserverlib/pull/377


S7evinK (approved):
LGTM!

PR #3079, opened 2023-05-08
Only require room version instead of room info for db.Events()

devonh:
This reduces the API requirements for the Events database to align with what is actually required.


roomserver/state/state.go
```diff
@@ -1134,7 +1137,11 @@ func (v *StateResolution) loadStateEvents(
 			eventNIDs = append(eventNIDs, entry.EventNID)
 		}
 	}
-	events, err := v.db.Events(ctx, v.roomInfo, eventNIDs)
+
+	if v.roomInfo == nil {
+		return nil, nil, fmt.Errorf("cannot get events without room info")
```
S7evinK, line 1142: Maybe we should make this an error we can reuse?

PR #3081, opened 2023-05-11
Move MakeJoin logic to GMSL

devonh:


roomserver/internal/query/query.go
```diff
@@ -858,6 +858,52 @@ func (r *Queryer) QueryAuthChain(ctx context.Context, req *api.QueryAuthChainReq
...
+		if err != nil || len(events) != 1 {
+			continue
+		}
+		event := events[0]
+		if event.Type() != spec.MRoomMember || event.StateKey() == nil {
+			continue // shouldn't happen
+		}
+
+		joinedUsers = append(joinedUsers, event)
+	}
```
S7evinK, line 902: Thiiis is going to hurt performance. Why not just query the events all at once?

federationapi/routing/routing.go
```diff
@@ -328,6 +326,24 @@ func Setup(
...
+			}
+			roomID, err := spec.NewRoomID(vars["roomID"])
+			if err != nil {
+				return util.JSONResponse{
+					Code: http.StatusBadRequest,
+					JSON: spec.BadJSON("Invalid RoomID"),
+				}
+			}
+
+			logrus.Warnf("Processing make_join for user %s, room %s", userID.String(), roomID.String())
```
S7evinK, line 345: Likely more `Debugf` than `Warnf`

roomserver/internal/query/query.go
devonh: This is the current behaviour. I'll update it since I'm in here anyway

federationapi/routing/join.go
```diff
@@ -33,153 +34,144 @@ import (
...
-		return util.JSONResponse{
-			Code: http.StatusForbidden,
-			JSON: spec.Forbidden("The join must be sent by the server of the user"),
+		event, err := eventutil.QueryAndBuildEvent(httpReq.Context(), proto, cfg.Matrix, identity, time.Now(), rsAPI, &queryRes)
+		if err == eventutil.ErrRoomNoExists {
+			return nil, nil, &util.JSONResponse{
+				Code: http.StatusNotFound,
+				JSON: spec.NotFound("Room does not exist"),
+			}
+		} else if e, ok := err.(gomatrixserverlib.BadJSONError); ok {
```
kegsay, line 126: Super unfortunate that we both use constant errors _and_ error types here. Can we just use error types as they are superior due to being able to add context to the error messages and remove `eventutil.ErrRoomNoExists`? This means we can then type switch

federationapi/routing/join.go
```diff
@@ -33,153 +34,144 @@ import (
...
-	content := gomatrixserverlib.MemberContent{
-		Membership:    spec.Join,
-		AuthorisedVia: authorisedVia,
-	}
-	if err = proto.SetContent(content); err != nil {
-		util.GetLogger(httpReq.Context()).WithError(err).Error("builder.SetContent failed")
-		return spec.InternalServerError()
+	response, internalErr := gomatrixserverlib.HandleMakeJoin(input)
+	if internalErr != nil {
+		return *internalErr
```
kegsay, line 162: It's not ideal that this returns a `JSONResponse`. I would rather we defined GMSL error types and then map them to JSON responses, as that:
 - allows us to inspect the error more easily (type switch) vs interpreting JSON errors
 - decouples "something went wrong joining this room" with "returning an HTTP response".

federationapi/routing/join.go
```diff
@@ -33,153 +34,144 @@ import (
...
+		BuildEventTemplate: createJoinTemplate,
 	}
-	content := gomatrixserverlib.MemberContent{
-		Membership:    spec.Join,
-		AuthorisedVia: authorisedVia,
-	}
-	if err = proto.SetContent(content); err != nil {
-		util.GetLogger(httpReq.Context()).WithError(err).Error("builder.SetContent failed")
-		return spec.InternalServerError()
+	response, internalErr := gomatrixserverlib.HandleMakeJoin(input)
```
kegsay, line 160: I'd rather this function wasn't so implementation specific (it heavily implies the /make_join endpoint, but also presumably does /send_join despite it not saying so?) Can we just call it `JoinRoom(input)`?

federationapi/routing/join.go
```diff
@@ -33,153 +34,144 @@ import (
...
-		}
-	} else if e, ok := err.(gomatrixserverlib.BadJSONError); ok {
-		return util.JSONResponse{
-			Code: http.StatusBadRequest,
-			JSON: spec.BadJSON(e.Error()),
-		}
-	} else if err != nil {
-		util.GetLogger(httpReq.Context()).WithError(err).Error("eventutil.BuildEvent failed")
+	if response == nil {
+		util.GetLogger(httpReq.Context()).Error("gmsl.HandleMakeJoin returned invalid response")
```
kegsay, line 166: I strongly dislike this: we shouldn't be returning nil responses AND nil errors. If I see this log line in the logs it gives me no clue what went wrong here. Please return an error instead with contextual information. Returning `nil, nil` makes sense only when A) the data being `nil` isn't an error and B) the `nil` indicates absence rather than as a result of an `OR` operation (error OR data).

roomserver/internal/query/query.go
```diff
@@ -858,6 +858,52 @@ func (r *Queryer) QueryAuthChain(ctx context.Context, req *api.QueryAuthChainReq
...
+	return isIn, err
+}
+
+func (r *Queryer) GetLocallyJoinedUsers(ctx context.Context, roomVersion gomatrixserverlib.RoomVersion, roomNID types.RoomNID) ([]gomatrixserverlib.PDU, error) {
+	joinNIDs, err := r.DB.GetMembershipEventNIDsForRoom(ctx, roomNID, true, true)
+	if err != nil || len(joinNIDs) == 0 {
+		// There should always be more than one join NID at this point
+		// because we are gated behind GetLocalServerInRoom, but y'know,
+		// sometimes strange things happen.
+		return nil, err
```
kegsay, line 885: This means it is possible for `GetLocallyJoinedUsers` to return `nil, nil` in the case where `err == nil` and `len(joinNIDs) == 0`. Is this intentional? If so, please comment. I think it probably is intentional as if there are no joined users then we want to return an empty slice with no error?

roomserver/internal/query/query.go
kegsay: Yeah please bulk fetch the nids.

kegsay (changes requested):

federationapi/routing/join.go
devonh: This is on the receiving side and only deals with handling the /make_join endpoint.
The /send_join handling is completely separate.

roomserver/internal/query/query.go
devonh: I think the logic should change to remove the `len(joinNIDs) == 0` check altogether, and return an empty slice later on in that case. And this comment appears to be obsolete as well so it should be removed.
That should clean up the logic.

federationapi/routing/join.go
devonh: We don't ever actually return `nil, nil`, but since it returns a pointer to the result it technically could have been nil.
The result is a pointer since it returns a full `ProtoEvent` struct. I could change that to be behind a pointer, but that could also technically be nil too. 
Which we could just not check, and just let the server panic assuming it was a programmer error that led to this case. I really don't like leaving in intentional panics across a library boundary. It's one thing to have them for internal programming errors, but relying on an external library api behaviour is dangerous.

federationapi/routing/join.go
devonh: Agreed, that approach is much cleaner. Doing that touches a lot of code, but it's nothing complicated so I'll include the changes in this PR.

kegsay (approved):
Are we sure we are returning the correct HTTP response codes with this refactor? Otherwise LGTM

PR #3083, opened 2023-05-16
Drop `reference_sha` column

S7evinK:
Companion PR to https://github.com/matrix-org/gomatrixserverlib/pull/383

### DIFF
--- a/roomserver/storage/postgres/deltas/20230516154000_drop_reference_sha.go
+++ b/roomserver/storage/postgres/deltas/20230516154000_drop_reference_sha.go
@@ -38,12 +38,15 @@ func UpDropEventReferenceSHAEvents(ctx context.Context, tx *sql.Tx) error {
 }
 
 func UpDropEventReferenceSHAPrevEvents(ctx context.Context, tx *sql.Tx) error {
-	_, err := tx.ExecContext(ctx, `ALTER TABLE roomserver_previous_events DROP COLUMN IF EXISTS previous_reference_sha256;`)
+	_, err := tx.ExecContext(ctx, "ALTER TABLE roomserver_previous_events DROP CONSTRAINT roomserver_previous_event_id_unique;")
 	if err != nil {
 		return fmt.Errorf("failed to execute upgrade: %w", err)
 	}
-	// re-add constraint, as it seems to be dropped as well
-	_, err = tx.ExecContext(ctx, `CREATE UNIQUE INDEX IF NOT EXISTS roomserver_previous_event_id_unique ON roomserver_previous_events (previous_event_id);`)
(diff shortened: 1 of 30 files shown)


roomserver/storage/postgres/deltas/20230516154000_drop_reference_sha.go
```diff
@@ -0,0 +1,38 @@
...
+
+func UpDropEventReferenceSHA(ctx context.Context, tx *sql.Tx) error {
+	var count int
+	err := tx.QueryRowContext(ctx, `SELECT count(*) FROM roomserver_events GROUP BY event_id HAVING count(event_id) > 1`).
+		Scan(&count)
+	if err != nil && err != sql.ErrNoRows {
+		return fmt.Errorf("failed to query duplicate event ids")
+	}
+	if count > 0 {
+		return fmt.Errorf("unable to drop column, as there are duplicate event ids")
```
kegsay, line 31: This is unfortunate as it would prevent people from upgrading Dendrite. Can we instead mix together the event nids for the affected rows? E.g given:
```
event_id | sha256 | nids
$foo | abc123 | [2,4]
$foo | def456 | [2,3]
```
replace it with:
```
$foo | [2,3,4]
```
?

roomserver/storage/postgres/previous_events_table.go
```diff
@@ -82,22 +82,21 @@ func (s *previousEventStatements) InsertPreviousEvent(
 	ctx context.Context,
 	txn *sql.Tx,
 	previousEventID string,
-	previousEventReferenceSHA256 []byte,
 	eventNID types.EventNID,
 ) error {
 	stmt := sqlutil.TxStmt(txn, s.insertPreviousEventStmt)
 	_, err := stmt.ExecContext(
-		ctx, previousEventID, previousEventReferenceSHA256, int64(eventNID),
+		ctx, previousEventID, []byte(""), int64(eventNID),
```
kegsay, line 89: Why not remove `[]byte("")` entirely? I guess removing `previous_reference_sha256` is for another PR? Actually it looks like you just forgot about this, given sqlite does remove this.

kegsay (approved):

roomserver/storage/sqlite3/previous_events_table.go
```diff
@@ -72,7 +74,34 @@ type previousEventStatements struct {
...
+		return err
+	}
+	m := sqlutil.NewMigrator(db)
+	m.AddMigrations([]sqlutil.Migration{
+		{
+			Version: migrationName,
+			Up:      deltas.UpDropEventReferenceSHAPrevEvents,
+		},
+	}...)
+	logrus.Infof("cName: %s", cName)
```
kegsay, line 100: Debug logging?

kegsay (approved):

PR #3085, opened 2023-05-23
Move `MakeLeave` to GMSL

S7evinK:
Basically the same API shape as for `/make_join`
https://github.com/matrix-org/gomatrixserverlib/pull/385

### DIFF
--- a/federationapi/routing/leave.go
+++ b/federationapi/routing/leave.go
@@ -34,108 +34,115 @@ func MakeLeave(
 	request *fclient.FederationRequest,
 	cfg *config.FederationAPI,
 	rsAPI api.FederationRoomserverAPI,
-	roomID, userID string,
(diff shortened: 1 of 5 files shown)


federationapi/routing/leave.go
```diff
@@ -34,108 +34,115 @@ func MakeLeave(
...
-		if mem, merr := state.Membership(); merr == nil && mem == spec.Leave {
-			return util.JSONResponse{
-				Code: http.StatusOK,
-				JSON: map[string]interface{}{
-					"room_version": event.Version(),
-					"event":        state,
-				},
-			}
+	default:
+		util.GetLogger(httpReq.Context()).WithError(internalErr)
```
kegsay, line 126: This doesn't actually log though?

kegsay (approved):

PR #3093, opened 2023-05-31
Move CreateRoom logic to Roomserver

devonh:
Move create room logic over to roomserver.


S7evinK (approved):
Didn't see anything obviously wrong, from the looks of it, a 1:1 move from the CS API.

PR #3098, opened 2023-06-02
Add initial support for storing user room keys

S7evinK:


roomserver/storage/postgres/user_room_keys_table.go
```diff
@@ -0,0 +1,82 @@
...
+const userRoomKeysSchema = `
+CREATE TABLE IF NOT EXISTS roomserver_user_room_keys (     
+    user_nid    INTEGER NOT NULL,
+    room_nid    INTEGER NOT NULL,
+    pseudo_id_key BYTEA NOT NULL,
+    CONSTRAINT roomserver_user_room_keys_pk PRIMARY KEY (user_nid, room_nid)
+);
+`
+
+const insertUserRoomKeySQL = `INSERT INTO roomserver_user_room_keys (user_nid, room_nid, pseudo_id_key) VALUES ($1, $2, $3)`
```
kegsay, line 37: Without an ON CONFLICT we'll get errors if we try to insert the same key twice surely?

roomserver/storage/shared/storage.go
```diff
@@ -1589,6 +1591,29 @@ func (d *Database) UpgradeRoom(ctx context.Context, oldRoomID, newRoomID, eventS
 	})
 }
 
+// InsertUserRoomKey inserts a new user room key for the given user and room.
+// Returns an error if a database error occurred, also if the primary constraint was violated.
+func (d *Database) InsertUserRoomKey(ctx context.Context, userNID types.EventStateKeyNID, roomNID types.RoomNID, key ed25519.PrivateKey) error {
```
kegsay, line 1596: This API shape isn't the most helpful and is racey.

The main scenario we are designing for is "I want to join a pseudo ID room, please ensure there is a key for me". If you concurrently call `InsertUserRoomKey`, the slower call will fail with an error, when in reality you just want to return the key that was previously inserted. I would probably change this to return the private key to use and make use of `ON CONFLICT ... RETURNING` like we do elsewhere in Dendrite.

roomserver/storage/shared/storage.go
```diff
@@ -1589,6 +1591,29 @@ func (d *Database) UpgradeRoom(ctx context.Context, oldRoomID, newRoomID, eventS
...
+	return d.Writer.Do(d.DB, nil, func(txn *sql.Tx) error {
+		return d.UserRoomKeyTable.InsertUserRoomKey(ctx, txn, userNID, roomNID, key)
+	})
+}
+
+// SelectUserRoomKey queries the user room key for a given user.
+// Returns the key and an error.
+// TODO: should we handle absent keys (sql.ErrNoRows) as non-fatal?
+func (d *Database) SelectUserRoomKey(ctx context.Context, userNID types.EventStateKeyNID, roomNID types.RoomNID) (key ed25519.PrivateKey, err error) {
```
kegsay, line 1605: It's not entirely clear that this is for selecting the private key i.e because Alice wants to send a message and now you need to sign an event, as opposed to "here's a set of events you're about to return to the client, please pull out the user IDs for these sender keys". It's not clear because of comments like:

> queries the user room key for a given user

It's not clear if this is the public part or private part.

> TODO: should we handle absent keys (sql.ErrNoRows) as non-fatal?

Yes, and let the caller decide what to do if the user has no key (maybe it can make one).

kegsay (changes requested):
General shape is good but API could be made a bit kinder to the caller.

roomserver/storage/shared/storage.go
S7evinK: In theory it shouldn't race when called through `/join`, since there can only be one join in progress.

roomserver/storage/interface.go
```diff
@@ -194,8 +194,10 @@ type Database interface {
 		ctx context.Context, roomInfo *types.RoomInfo, eventNID types.EventNID, event gomatrixserverlib.PDU, plResolver state.PowerLevelResolver,
 	) (gomatrixserverlib.PDU, gomatrixserverlib.PDU, error)
 
-	InsertUserRoomKey(ctx context.Context, userNID types.EventStateKeyNID, roomNID types.RoomNID, key ed25519.PrivateKey) error
-	SelectUserRoomKey(ctx context.Context, userNID types.EventStateKeyNID, roomNID types.RoomNID) (key ed25519.PrivateKey, err error)
+	InsertUserRoomPrivateKey(ctx context.Context, userID spec.UserID, roomID spec.RoomID, key ed25519.PrivateKey) (result ed25519.PrivateKey, err error)
+	SelectUserRoomPrivateKey(ctx context.Context, userID spec.UserID, roomID spec.RoomID) (key ed25519.PrivateKey, err error)
+	InsertUserRoomPublicKey(ctx context.Context, userID spec.UserID, roomID spec.RoomID, key ed25519.PublicKey) (result ed25519.PublicKey, err error)
```
kegsay, line 199: We almost certainly want to warn against doing something like this for local users:
 - InsertUserRoomPrivateKey
 - InsertUserRoomPublicKey

Because these functions don't use a txn, you could race and end up with the wrong priv/pub key combo.

roomserver/storage/postgres/user_room_keys_table.go
```diff
@@ -29,17 +31,33 @@ const userRoomKeysSchema = `
...
+
+const insertUserRoomPublicKeySQL = `
+	INSERT INTO roomserver_user_room_keys (user_nid, room_nid, pseudo_id_pub_key) VALUES ($1, $2, $3)
+	ON CONFLICT ON CONSTRAINT roomserver_user_room_keys_pk DO UPDATE SET pseudo_id_pub_key = roomserver_user_room_keys.pseudo_id_pub_key
+	RETURNING (pseudo_id_pub_key)
+`
+
 const selectUserRoomKeySQL = `SELECT pseudo_id_key FROM roomserver_user_room_keys WHERE user_nid = $1 AND room_nid = $2`
 
+const selectUserNIDsSQL = `SELECT user_nid, pseudo_id_pub_key  FROM roomserver_user_room_keys WHERE pseudo_id_pub_key = ANY($1)`
```
kegsay, line 54: This isn't safe to do because it lacks a `WHERE room_nid` clause. The protocol does not say that pseudo ID keys need to be globally unique.

kegsay (changes requested):
A few tweaks required, but this is looking good.

roomserver/storage/interface.go
kegsay: This comment doesn't make sense anymore because `InsertUserRoomPrivateKey` actually also inserts the public part too, see further down comments.

roomserver/storage/postgres/user_room_keys_table.go
```diff
@@ -29,17 +31,33 @@ const userRoomKeysSchema = `
...
+    pseudo_id_key BYTEA NULL, -- may be null for users not local to the server
+    pseudo_id_pub_key BYTEA NOT NULL,
     CONSTRAINT roomserver_user_room_keys_pk PRIMARY KEY (user_nid, room_nid)
 );
 `
 
-const insertUserRoomKeySQL = `INSERT INTO roomserver_user_room_keys (user_nid, room_nid, pseudo_id_key) VALUES ($1, $2, $3)`
+const insertUserRoomPrivateKeySQL = `
+	INSERT INTO roomserver_user_room_keys (user_nid, room_nid, pseudo_id_key, pseudo_id_pub_key) VALUES ($1, $2, $3, $4)
+	ON CONFLICT ON CONSTRAINT roomserver_user_room_keys_pk DO UPDATE SET pseudo_id_key = roomserver_user_room_keys.pseudo_id_key
```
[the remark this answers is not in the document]
S7evinK: `DO NOTHING` doesn't work, as this returns no result. We need to update, even if this just sets the key to the existing one.

roomserver/storage/tables/interface.go
```diff
@@ -186,10 +186,15 @@ type Purge interface {
...
+	// when creating keys locally.
+	InsertUserRoomPrivatePublicKey(ctx context.Context, txn *sql.Tx, userNID types.EventStateKeyNID, roomNID types.RoomNID, key ed25519.PrivateKey) (ed25519.PrivateKey, error)
+	// InsertUserRoomPublicKey inserts the given public key, this should be used for users NOT local to this server
 	InsertUserRoomPublicKey(ctx context.Context, txn *sql.Tx, userNID types.EventStateKeyNID, roomNID types.RoomNID, key ed25519.PublicKey) (ed25519.PublicKey, error)
+	// SelectUserRoomPrivateKey selects the private key for the given user and room combination
 	SelectUserRoomPrivateKey(ctx context.Context, txn *sql.Tx, userNID types.EventStateKeyNID, roomNID types.RoomNID) (ed25519.PrivateKey, error)
-	BulkSelectUserNIDs(ctx context.Context, txn *sql.Tx, senderKeys [][]byte) (map[string]types.EventStateKeyNID, error)
+	// BulkSelectUserNIDs selects all userIDs for the requested senderKeys. Returns a map from publicKey -> types.UserRoomKeyPair.
+	BulkSelectUserNIDs(ctx context.Context, txn *sql.Tx, senderKeys map[types.RoomNID][]ed25519.PublicKey) (map[string]types.UserRoomKeyPair, error)
```
kegsay, line 197: Need to doc what happens if some keys don't exist in the table.

roomserver/storage/postgres/user_room_keys_table.go
```diff
@@ -102,27 +102,31 @@ func (s *userRoomKeysStatements) SelectUserRoomPrivateKey(
...
-	rows, err := stmt.QueryContext(ctx, pq.Array(senderKeys))
+	roomNIDs := make([]types.RoomNID, 0, len(senderKeys))
+	var senders [][]byte
+	for roomNID := range senderKeys {
+		roomNIDs = append(roomNIDs, roomNID)
+		for _, key := range senderKeys[roomNID] {
+			senders = append(senders, key)
+		}
+	}
+	rows, err := stmt.QueryContext(ctx, pq.Array(senders))
```
kegsay, line 116: This isn't what I was expecting, as this code will give incorrect results in some cases. This code just queries on the set of sender keys, ignoring the room NID for each key. This means it's possible to query this function asking for the user NID in room A with sender key X, and get back the user NID for room B with sender key X.

We either need to `WHERE room_nid` to filter DB-side or we need to only add `UserRoomKeyPair` to `result` if the room NID was asked for in `senderKeys`.

roomserver/storage/shared/storage.go
```diff
@@ -1696,28 +1697,57 @@ func (d *Database) SelectUserRoomPrivateKey(ctx context.Context, userID spec.Use
...
 		}
-		nids := make([]types.EventStateKeyNID, 0, len(userNIDKeyMap))
-		for _, nid := range userNIDKeyMap {
-			nids = append(nids, nid)
+		nids := make([]types.EventStateKeyNID, 0, len(userRoomKeyPairMap))
+		for _, nid := range userRoomKeyPairMap {
+			nids = append(nids, nid.EventStateKeyNID)
 		}
+		// get the userIDs
 		nidMAP, seErr := d.EventStateKeys(ctx, nids)
```
kegsay, line 1736:
```suggestion
		nidMap, seErr := d.EventStateKeys(ctx, nids)
```

kegsay (changes requested):

PR #3100, opened 2023-06-03
PDU Sender split

devonh:
Initial cut of splitting PDU Sender into SenderID & UserID

### DIFF
--- a/clientapi/routing/directory.go
+++ b/clientapi/routing/directory.go
@@ -189,7 +189,7 @@ func SetLocalAlias(
 		}
 	}
 
-	deviceSenderID, err := rsAPI.QuerySenderIDForRoom(req.Context(), alias, *userID)
+	deviceSenderID, err := rsAPI.QuerySenderIDForUser(req.Context(), alias, *userID)
 	if err != nil {
 		return util.JSONResponse{
 			Code: http.StatusInternalServerError,
@@ -238,7 +238,7 @@ func RemoveLocalAlias(
 		}
 	}
 
-	deviceSenderID, err := rsAPI.QuerySenderIDForRoom(req.Context(), alias, *userID)
+	deviceSenderID, err := rsAPI.QuerySenderIDForUser(req.Context(), alias, *userID)
 	if err != nil {
 		return util.JSONResponse{
 			Code: http.StatusInternalServerError,
--- a/clientapi/routing/directory.go
+++ b/clientapi/routing/directory.go
@@ -189,7 +189,7 @@ func SetLocalAlias(
 		}
 	}
 
-	deviceSenderID, err := rsAPI.QuerySenderIDForUser(req.Context(), alias, *userID)
+	deviceSenderID, err := rsAPI.QuerySenderIDForUser(req.Context(), r.RoomID, *userID)
 	if err != nil {
 		return util.JSONResponse{
 			Code: http.StatusInternalServerError,
@@ -238,7 +238,17 @@ func RemoveLocalAlias(
 		}
 	}
 
-	deviceSenderID, err := rsAPI.QuerySenderIDForUser(req.Context(), alias, *userID)
+	roomIDReq := roomserverAPI.GetRoomIDForAliasRequest{Alias: alias}
+	roomIDRes := roomserverAPI.GetRoomIDForAliasResponse{}
+	err = rsAPI.GetRoomIDForAlias(req.Context(), &roomIDReq, &roomIDRes)
+	if err != nil {
+		return util.JSONResponse{
+			Code: http.StatusNotFound,
+			JSON: spec.NotFound("The alias does not exist."),
+		}
+	}
+
+	deviceSenderID, err := rsAPI.QuerySenderIDForUser(req.Context(), roomIDRes.RoomID, *userID)
 	if err != nil {
 		return util.JSONResponse{
 			Code: http.StatusInternalServerError,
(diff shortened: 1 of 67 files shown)


clientapi/routing/directory.go
```diff
@@ -215,9 +230,25 @@ func RemoveLocalAlias(
...
 ) util.JSONResponse {
+	userID, err := spec.NewUserID(device.UserID, true)
+	if err != nil {
+		return util.JSONResponse{
+			Code: http.StatusInternalServerError,
+			JSON: spec.InternalServerError{Err: "UserID for device is invalid"},
+		}
+	}
+
+	deviceSenderID, err := rsAPI.QuerySenderIDForUser(req.Context(), alias, *userID)
```
kegsay, line 241: Do we expect this to create the sender ID if one does not exist? I guess in this case no, but in some cases yes?

roomserver/internal/input/input_events.go
```diff
@@ -128,9 +128,9 @@ func (r *Inputer) processRoomEvent(
 	if roomInfo == nil && !isCreateEvent {
 		return fmt.Errorf("room %s does not exist for event %s", event.RoomID(), event.EventID())
 	}
-	_, senderDomain, err := gomatrixserverlib.SplitID('@', event.Sender())
+	sender, err := event.UserID()
 	if err != nil {
-		return fmt.Errorf("event has invalid sender %q", input.Event.Sender())
+		return fmt.Errorf("event has invalid sender %q", event.SenderID())
```
kegsay, line 133: Confusing if this ever happens as the error was caused by calling `.UserID` not `.SenderID` and we drop `err` to the floor.

roomserver/internal/input/input_events.go
```diff
@@ -128,9 +128,9 @@ func (r *Inputer) processRoomEvent(
 	if roomInfo == nil && !isCreateEvent {
 		return fmt.Errorf("room %s does not exist for event %s", event.RoomID(), event.EventID())
 	}
-	_, senderDomain, err := gomatrixserverlib.SplitID('@', event.Sender())
+	sender, err := event.UserID()
```
kegsay, line 131: This won't always work, but it will mostly work. We need to gracefully degrade when `err` is non nil.

roomserver/internal/input/input_events.go
```diff
@@ -193,9 +193,9 @@ func (r *Inputer) processRoomEvent(
 			serverRes.ServerNames = append(serverRes.ServerNames, input.Origin)
 			delete(servers, input.Origin)
 		}
-		if senderDomain != input.Origin && senderDomain != r.Cfg.Matrix.ServerName {
-			serverRes.ServerNames = append(serverRes.ServerNames, senderDomain)
-			delete(servers, senderDomain)
+		if sender.Domain() != input.Origin && sender.Domain() != r.Cfg.Matrix.ServerName {
+			serverRes.ServerNames = append(serverRes.ServerNames, sender.Domain())
+			delete(servers, sender.Domain())
```
kegsay, line 198: We should just optimistically do this if `sender` can be resolved (i.e we know an `mxid_mapping` for this sender key). We should NOT be failing to process the event if we have no `mxid_mapping`.

roomserver/internal/input/input_events.go
```diff
@@ -451,7 +451,7 @@ func (r *Inputer) processRoomEvent(
 	}
 
 	// Handle remote room upgrades, e.g. remove published room
-	if event.Type() == "m.room.tombstone" && event.StateKeyEquals("") && !r.Cfg.Matrix.IsLocalServerName(senderDomain) {
+	if event.Type() == "m.room.tombstone" && event.StateKeyEquals("") && !r.Cfg.Matrix.IsLocalServerName(sender.Domain()) {
```
kegsay, line 454: `sender` is obtained via `event.UserID()` which is the wrong way to check for "is this user local". We should be bouncing the `event.SenderID()` via `QueryUserIDForSender`.

kegsay: Come to think of it, it would be lovely if `event.UserID()` returned `*spec.UserID` so it makes it clear that hey sometimes this will be `nil`, which in and of itself is not an error.

kegsay (approved):
We need to be more rigorous to distinguish between `event.UserID()` and `QueryUserIDForSender(event.SenderID())`.

`QueryUserIDForSender` should do a DB hit to answer:
 - is this sender key mapped to a user ID at all? AND
 - is this user ID local? (aka do I have a private key for this sender key)

I am struggling to identify a legit use case for `event.UserID()` as I don't know how this field is actually getting set in GMSL (currently it just assumes sender ID is a `spec.UserID`).

My intuition here would be to:
 - remove `event.UserID()` entirely in GMSL.
 - Expect Dendrite to bounce via `QueryUserIDForSender` when it needs to map a `SenderID` to a `UserID`.

This keeps GMSL "pure" in that it dosn't need to care about "hey, I need to find the `m.room.member` event for this sender of this message event to serve queries for `event.UserID()`".

In the rare cases where GMSL needs the user ID for an event, we should interface up a function which Dendrite implements which ultimately calls `QueryUserIDForSender`.

userapi/consumers/roomserver_test.go
```diff
@@ -61,13 +61,13 @@ func Test_evaluatePushRules(t *testing.T) {
 		}{
 			{
 				name:         "m.receipt doesn't notify",
-				eventContent: `{"type":"m.receipt"}`,
+				eventContent: `{"type":"m.receipt","sender":"@test:remotehost"}`,
```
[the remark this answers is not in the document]
devonh: One of my iterations of changes required this. It is no longer required, which is comforting since the existing behaviour must be being maintained now.

PR #3101, opened 2023-06-03
feat: admin APIs for token authenticated registration

santhoshivan23:
### Pull Request Checklist

Signed-off-by: `Santhoshivan Amudhan santhoshivan23@gmail.com`


clientapi/routing/admin.go
```diff
@@ -24,6 +25,86 @@ import (
...
+	}{}
+
+	if err := json.NewDecoder(req.Body).Decode(&request); err != nil {
+		return util.JSONResponse{
+			Code: http.StatusBadRequest,
+			JSON: spec.Unknown("Failed to decode request body: " + err.Error()),
+		}
+	}
+	token := request.Token
+	if len(token) > 0 {
```
S7evinK, line 43: I'd probably remove this and only have something like:

```
if len(token) == 0 {
 // bad request
}
if len(token) > 64 {
// bad request
}
```
or:
```
if len(token) == 0 || len(token) > 64 {
// bad request
}
```
otherwise continue on the happy path and try to validate that the token is valid.

clientapi/routing/admin.go
```diff
@@ -24,6 +25,86 @@ import (
...
+		if len(token) > 64 {
+			return util.MatrixErrorResponse(
+				http.StatusBadRequest,
+				string(spec.ErrorInvalidParam),
+				"token must not be empty and must not be longer than 64")
+		}
+		is_token_valid, _ := regexp.MatchString("^[[:ascii:][:digit:]_]*$", token)
```
S7evinK, line 50: Please use `camelCase`, so `isTokenValid` instead. (Same for other variables, like `usesAllowed` instead of `uses_allowed`)

S7evinK (commented):
As there are no tests in Sytest, would be awesome if you could add tests for this in Dendrite. :)

userapi/storage/postgres/registration_tokens_table.go
```diff
@@ -0,0 +1,250 @@
...
+		stmt = s.listValidTokensStatement
+		rows, err = stmt.QueryContext(ctx, time.Now().UnixNano()/int64(time.Millisecond))
+	} else {
+		stmt = s.listInvalidTokenStatement
+		rows, err = stmt.QueryContext(ctx, time.Now().UnixNano()/int64(time.Millisecond))
+	}
+	if err != nil {
+		return tokens, err
+	}
+	for rows.Next() {
```
S7evinK, line 153: We should always close our rows, right before `rows.Next()`:
```go
defer internal.CloseAndLogIfError(ctx, rows, "ListRegistrationTokens: rows.close() failed")
```

userapi/storage/postgres/registration_tokens_table.go
```diff
@@ -0,0 +1,250 @@
...
+		return err
+	}
+	count, err := res.RowsAffected()
+	if err != nil {
+		return err
+	}
+	if count == 0 {
+		return fmt.Errorf("token: %s does not exists", tokenString)
+	}
+	return nil
```
S7evinK, line 223: Probably best to just ignore the result, no need to error if the statement no-ops anyway.

S7evinK (changes requested):
Needs a few changes, but this is promising :)
 
Comments for Postgres should also apply to SQLite.
As for the database functions, there is a bit of inconsistency here: Sometimes you're using `sqlutil.TxStmt()` and sometimes you're using the "raw" statement.

santhoshivan23:
> Needs a few changes, but this is promising :)
> 
> Comments for Postgres should also apply to SQLite. As for the database functions, there is a bit of inconsistency here: Sometimes you're using `sqlutil.TxStmt()` and sometimes you're using the "raw" statement.

Thanks @S7evinK - let me work on the comments

santhoshivan23:
@S7evinK  Thanks for your valuable comments. I've addressed all of them

santhoshivan23:
@S7evinK PR is ready for review!

clientapi/routing/admin.go
```diff
@@ -17,13 +21,274 @@ import (
...
+	}
+}
+
+func AdminUpdateRegistrationToken(req *http.Request, cfg *config.ClientAPI, userAPI userapi.ClientUserAPI) util.JSONResponse {
+	vars, err := httputil.URLDecodeMapValues(mux.Vars(req))
+	if err != nil {
+		return util.ErrorResponse(err)
+	}
+	tokenText := vars["token"]
+	request := make(map[string]interface{})
```
S7evinK, line 241: Given the request should be `map[string]*int64`, can we make this type safe please?
So if `uses_allowed` is present, it can either be `nil` or a `int64`.

clientapi/routing/admin.go
```diff
@@ -17,13 +21,274 @@ import (
...
+			Code: http.StatusBadRequest,
+			JSON: spec.BadJSON(fmt.Sprintf("Failed to decode request body: %s", err)),
+		}
+	}
+	newAttributes := make(map[string]interface{})
+	usesAllowed, ok := request["uses_allowed"]
+	if ok {
+		// Only add usesAllowed to newAtrributes if it is present and valid
+		// Non numeric values in payload will cause panic during type conversion. But this is the best way to mimic
+		// Synapse's behaviour of updating the field if and only if it is present in request body.
```
S7evinK, line 253: Making it type safe as mentioned above should avoid potential panics, and we still are able to distinguish between "not present" and "value is null"

S7evinK (commented):
Didn't have a look at the tests yet, doing that tomorrow.

santhoshivan23:
@S7evinK - addressed all the comments.

clientapi/routing/admin.go
```diff
@@ -16,14 +18,254 @@ import (
...
+		return util.JSONResponse{
+			Code: http.StatusBadRequest,
+			JSON: spec.BadJSON(fmt.Sprintf("Failed to decode request body: %s", err)),
+		}
+	}
+	newAttributes := make(map[string]interface{})
+	usesAllowed, ok := request["uses_allowed"]
+	if ok {
+		// Only add usesAllowed to newAtrributes if it is present and valid
+		if !(usesAllowed == nil || *usesAllowed >= 0) {
```
devonh, line 233: Can we flip the logic here too
```suggestion
		if usesAllowed != nil && *usesAllowed < 0 {
```

clientapi/routing/admin.go
```diff
@@ -16,14 +18,254 @@ import (
...
+				Code: http.StatusBadRequest,
+				JSON: spec.BadJSON("uses_allowed must be a non-negative integer or null"),
+			}
+		}
+		newAttributes["usesAllowed"] = usesAllowed
+	}
+	expiryTime, ok := request["expiry_time"]
+	if ok {
+		// Only add expiryTime to newAtrributes if it is present and valid
+		if !(expiryTime == nil || *expiryTime > time.Now().UnixNano()/int64(time.Millisecond)) {
```
devonh, line 244: And flip this as well
```suggestion
		if expiryTime != nil && spec.Timestamp(*expiryTime).Time().Before(time.Now()) {
```

devonh:
I added 2 comments about flipping logic around. Otherwise, once the linter errors are fixed this should be good to merge.

devonh (changes requested):

santhoshivan23:
thanks @devonh @S7evinK for the review. i've addressed the comments. hope we can merge this now

PR #3105, opened 2023-06-07
Use SenderID Type

devonh:


federationapi/internal/perform.go
```diff
@@ -147,27 +147,32 @@ func (r *FederationInternalAPI) performJoinUsingServer(
 	}
 
 	user, err := spec.NewUserID(userID, true)
-	if err != nil {
+	if err != nil || user == nil {
 		return err
 	}
```
S7evinK, line 152: I believe we can leave out the nil check, as `NewUserID` always returns a `UserID` if the error is nil? Also in other places.

syncapi/storage/shared/storage_consumer.go
```diff
@@ -224,9 +224,9 @@ func (d *Database) StreamEventsToEvents(device *userapi.Device, in []types.Strea
 	return out
 }
 
-func (d *Database) getSenderIDForUser(roomID string, userID spec.UserID) (string, error) { // nolint
-	// TODO: Repalce with actual logic for pseudoIDs
-	return userID.String(), nil
+func (d *Database) getSenderIDForUser(roomID string, userID spec.UserID) (spec.SenderID, error) { // nolint
+	// TODO: Replace with actual logic for pseudoIDs
+	return spec.SenderID(userID.String()), nil
```
kegsay, line 229: Surely we should be leaning on roomserver here, not syncapi's db?

federationapi/internal/perform.go
devonh: Yup, good catch. I realized the same thing last night, but didn't fix all the locations.

PR #3106, opened 2023-06-07
Cleanup remaining statekey usage for senderIDs

devonh:


clientapi/routing/account_data.go
```diff
@@ -145,8 +145,23 @@ func SaveReadMarker(
 	userAPI api.ClientUserAPI, rsAPI roomserverAPI.ClientRoomserverAPI,
 	syncProducer *producers.SyncAPIProducer, device *api.Device, roomID string,
 ) util.JSONResponse {
+	fullUserID, err := spec.NewUserID(device.UserID, true)
+	if err != nil {
+		return util.JSONResponse{
+			Code: http.StatusForbidden,
+			JSON: spec.Forbidden("userID doesn't have power level to change visibility"),
```
kegsay, line 152: Message here is confusing, this is about read markers.

clientapi/routing/aliases.go
```diff
@@ -55,9 +55,23 @@ func GetAliases(
...
+			return util.JSONResponse{
+				Code: http.StatusForbidden,
+				JSON: spec.Forbidden("userID doesn't have power level to change visibility"),
+			}
+		}
 		queryReq := api.QueryMembershipForUserRequest{
-			RoomID: roomID,
-			UserID: device.UserID,
+			RoomID:   roomID,
+			SenderID: senderID,
```
kegsay, line 74: So I can see this becoming a royal pain as we are constantly forced to do the dance of:
 - take device user ID -> make a spec.UserID
 - QuerySenderIDForUser
 - QueryMembershipForUser

Given both query functions use the rsAPI, it would be more ergonomic to just allow `QueryMembershipForUserRequest` to accept a `UserID` (like it did before) and resolve that in one place, rather than expecting the caller to do this.

We can then situationally add a `QueryMembershipForSender` function when we have a sender ID but no user ID.

clientapi/routing/directory.go
```diff
@@ -314,7 +314,22 @@ func SetVisibility(
 	req *http.Request, rsAPI roomserverAPI.ClientRoomserverAPI, dev *userapi.Device,
 	roomID string,
 ) util.JSONResponse {
-	resErr := checkMemberInRoom(req.Context(), rsAPI, dev.UserID, roomID)
+	fullUserID, err := spec.NewUserID(dev.UserID, true)
+	if err != nil {
+		return util.JSONResponse{
+			Code: http.StatusForbidden,
+			JSON: spec.Forbidden("userID doesn't have power level to change visibility"),
```
kegsay, line 321: This simply isn't true, we aren't checking PLs here..

clientapi/routing/leaveroom.go
```diff
@@ -29,10 +29,25 @@ func LeaveRoomByID(
...
+			Code: http.StatusBadRequest,
+			JSON: spec.Unknown("could not find senderID for this user"),
+		}
+	}
+
 	// Prepare to ask the roomserver to perform the room join.
 	leaveReq := roomserverAPI.PerformLeaveRequest{
 		RoomID: roomID,
-		UserID: device.UserID,
+		Leaver: roomserverAPI.SenderUserIDPair{SenderID: senderID, UserID: *userID},
```
kegsay, line 50: Again, it would be more ergonomic to just pass in the user ID and let the rsAPI do `QuerySenderIDForUser`.

kegsay (commented):
I'd like to see the API be a bit kinder to callers, as there is a lot of duplicated code of making `spec.UserID` then `QuerySenderIDForUser`. This would cut down the diff on this PR quite significantly.

PR #3112, opened 2023-06-13
Add `MXIDMapping` for pseudoID rooms

S7evinK:
Initial draft of setting the `MXIDMapping` on membership events when creating rooms.


syncapi/streams/stream_pdu.go
```diff
@@ -346,6 +348,46 @@ func (p *PDUStreamProvider) addRoomDeltaToResponse(
 	// Now that we've filtered the timeline, work out which state events are still
 	// left. Anything that appears in the filtered timeline will be removed from the
 	// "state" section and kept in "timeline".
+
+	for i, ev := range events {
```
S7evinK, line 352: This feels like a horrible hack atm, but this way we can at least invite other users to the room.
Changing powerlevels currently results in a weird behavior (alice made bob a moderator):

> alice changed the power level of alice from Default to Admin, bob from Default to Moderator, BzmJ3YUq9n1c5A8ZHTNSKmQNGNL8WT8q1sowxNd9XpE from Admin to Default.

Also changing the powerlevels afterwards is not possible anymore.

federationapi/routing/join.go
```diff
@@ -64,7 +64,13 @@ func MakeJoin(
 	}
 
 	createJoinTemplate := func(proto *gomatrixserverlib.ProtoEvent) (gomatrixserverlib.PDU, []gomatrixserverlib.PDU, error) {
-		identity, signErr := cfg.Matrix.SigningIdentityFor(request.Destination())
+		// TODO: remove this once the join dance understands pseudo IDs
+		var dummyUserID *spec.UserID
+		dummyUserID, err = spec.NewUserID(fmt.Sprintf("@dummy:%s", request.Destination()), true)
```
devonh, line 69: when does the db get updated with the actual userID after this point? wouldn't that be required for everything else to work?

devonh: and wouldn't this possibly result in user overlap if multiple users from the same server join the room?
seems like a nightmare

federationapi/routing/leave.go
```diff
@@ -59,14 +59,20 @@ func MakeLeave(
 	}
 
 	createLeaveTemplate := func(proto *gomatrixserverlib.ProtoEvent) (gomatrixserverlib.PDU, []gomatrixserverlib.PDU, error) {
-		identity, signErr := cfg.Matrix.SigningIdentityFor(request.Destination())
+		// TODO: remove this once the leave dance understands pseudo IDs
+		var dummyUserID *spec.UserID
+		dummyUserID, err = spec.NewUserID(fmt.Sprintf("@dummy:%s", request.Destination()), true)
```
devonh, line 64: shouldn't the userID mapping in the db have been updated by this point in time, in which case you won't be able to find the key for them in the db? not sure how this is working

devonh (commented):
For the most part it seems to be okay, other than the places you have already commented as being a little iffy.

roomserver/types/headered_event.go
```diff
@@ -18,13 +18,18 @@ import (
...
 	"github.com/matrix-org/gomatrixserverlib"
+	"github.com/matrix-org/gomatrixserverlib/spec"
 )
 
 // HeaderedEvent is an Event which serialises to the headered form, which includes
 // _room_version and _event_id fields.
 type HeaderedEvent struct {
 	gomatrixserverlib.PDU
 	Visibility gomatrixserverlib.HistoryVisibility
+	// TODO: Remove this. This is a temporary workaround to store the userID in the syncAPI.
```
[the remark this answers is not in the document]
S7evinK: Merge the syncAPI db with the roomserver DB? (which is/was on our agenda after removing polylith, iirc)

clientapi/routing/membership.go
```diff
@@ -459,6 +453,31 @@ func buildMembershipEvent(
 	if err != nil {
 		return nil, err
 	}
+
+	// If we're inviting a local user, we can generate the needed key here.
+	if targetSenderID == "" && cfg.Matrix.IsLocalServerName(targetID.Domain()) { // todo: remove
```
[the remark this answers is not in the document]
S7evinK: Good question :D Probably thought about moving this to the roomserver.

clientapi/routing/sendevent.go
```diff
@@ -121,6 +122,17 @@ func SendEvent(
 		delete(r, "join_authorised_via_users_server")
 	}
 
+	// for power level events we need to replace the userID with the pseudoID
```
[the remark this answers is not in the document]
S7evinK: Member events shouldn't have any userIDs in their event content. (Aside from 3PID invites, which are handled through a different endpoint)
Not sure which other event types would need replacing.

roomserver/internal/perform/perform_join.go
```diff
@@ -283,11 +286,39 @@ func (r *Joiner) performJoinRoomByID(
...
 
 	var buildRes rsAPI.QueryLatestEventsAndStateResponse
-	identity, err := r.Cfg.Matrix.SigningIdentityFor(userDomain)
+	identity, err := r.RSAPI.SigningIdentityFor(ctx, *roomID, *userID)
 	if err != nil {
 		return "", "", fmt.Errorf("error joining local room: %q", err)
 	}
 
+	// at this point we know we have an existing room
+	if inRoomRes.RoomVersion == gomatrixserverlib.RoomVersionPseudoIDs {
```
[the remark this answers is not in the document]
S7evinK: IMHO it's easier to read this way, otherwise we just call a function just for it to return immediately. Or when debugging, having to navigate to a function which returns immediately because of some condition.

federationapi/routing/join.go
S7evinK: Looks like we don't even need to sign this event (we still do, though), so going to switch back to getting the signing identity for the `request.Destination()`

PR #3113, opened 2023-06-13
Fix unsafe hotserving behaviour for multimedia uploads.

joshqou:
Return multimedia with a disposition type of attachment instead of inline. NVT#1548992

Signed-off-by: Josh Qou [jqou@icloud.com](mailto:jqou@icloud.com)


kegsay (approved):
LGTM, thanks!

PR #3114, opened 2023-06-14
rearrange order of sections about signing keys and configuring dendrite, fix a dead link

CicadaCinema:
I thought I would rearrange these pages since the configuration step requires that a signing key has been generated.


kegsay (approved):
Thanks!

CicadaCinema:
I have emailed a private sign-off to `dco@matrix.org`.

PR #3115, opened 2023-06-15
Fix commit info for #3113

joshqou:
Only noticed until I went back to check after merge that I forgot to properly fill in my commit info, this PR just recommits with the proper authorship info. Oops @kegsay

Signed-off-by: Josh Qou [jqou@icloud.com](mailto:jqou@icloud.com)


S7evinK:
Probably doesn't matter that much.

PR #3118, opened 2023-06-21
Fix syncAPI redactions

S7evinK:
Previously we were setting `redacted_because` to the PDU event, but as per the spec it should really be a client event.
This fixes it.


devonh:
Need to convert the stateKey into a userID before using `ToClientEvent`. At which point this should probably just use the `ToClientEventDefault` function which will move all the error handling out of here & handle getting both userIDs.

S7evinK:
As mentioned, the `stateKey` should be nil anyway, and `ToClientEventDefault` uses `FormatAll` instead of `FormatSync`.

PR #3119, opened 2023-06-21
Make it possible to join pseudo ID rooms

S7evinK:
This should allow us to join pseudo ID rooms over federation.
Also needs https://github.com/matrix-org/gomatrixserverlib/pull/397

### DIFF
--- a/roomserver/internal/perform/perform_join.go
+++ b/roomserver/internal/perform/perform_join.go
@@ -314,7 +314,7 @@ func (r *Joiner) performJoinRoomByID(
 		// sign the event with the pseudo ID key
 		identity = fclient.SigningIdentity{
 			ServerName: "self",
-			KeyID:      "ed25519",
+			KeyID:      "ed25519:1",
 			PrivateKey: pseudoIDKey,
 		}
 	}
--- a/roomserver/storage/shared/storage.go
+++ b/roomserver/storage/shared/storage.go
@@ -1032,7 +1032,8 @@ func (d *EventDatabase) MaybeRedactEvent(
 			return nil
 		}
 
-		validRoomID, err := spec.NewRoomID(redactedEvent.RoomID())
+		var validRoomID *spec.RoomID
+		validRoomID, err = spec.NewRoomID(redactedEvent.RoomID())
 		if err != nil {
 			return err
 		}
--- a/roomserver/storage/shared/storage.go
+++ b/roomserver/storage/shared/storage.go
@@ -10,6 +10,7 @@ import (
 	"sort"
 
 	"github.com/matrix-org/dendrite/internal/eventutil"
+	"github.com/matrix-org/dendrite/roomserver/api"
 	"github.com/matrix-org/gomatrixserverlib"
 	"github.com/matrix-org/gomatrixserverlib/spec"
 	"github.com/matrix-org/util"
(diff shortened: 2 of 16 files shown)


docs/installation/manual/3_signingkey.md
devonh, line 1: why is this change pulled into this PR?

docs/installation/manual/4_configuration.md
devonh, line 1: why is this change pulled into this PR?

roomserver/internal/perform/perform_join.go
```diff
@@ -314,7 +314,7 @@ func (r *Joiner) performJoinRoomByID(
 		// sign the event with the pseudo ID key
 		identity = fclient.SigningIdentity{
 			ServerName: "self",
-			KeyID:      "ed25519",
+			KeyID:      "ed25519:1",
```
devonh, line 317: what does the 1 mean?

roomserver/storage/sqlite3/user_room_keys_table.go
```diff
@@ -137,7 +138,7 @@ func (s *userRoomKeysStatements) BulkSelectUserNIDs(ctx context.Context, txn *sq
 	selectSQL := strings.Replace(selectUserNIDsSQL, "($2)", sqlutil.QueryVariadicOffset(len(senders), len(senderKeys)), 1)
 	selectSQL = strings.Replace(selectSQL, "($1)", sqlutil.QueryVariadic(len(senderKeys)), 1) // replace $1 with the roomNIDs
 
-	selectStmt, err := txn.Prepare(selectSQL)
+	selectStmt, err := s.db.Prepare(selectSQL)
```
devonh, line 141: why this change / did this not work before?

roomserver/storage/shared/storage.go
```diff
@@ -1757,58 +1762,54 @@ func (d *Database) SelectUserRoomPublicKey(ctx context.Context, userID spec.User
 // SelectUserIDsForPublicKeys returns a map from roomID -> map from senderKey -> userID
 func (d *Database) SelectUserIDsForPublicKeys(ctx context.Context, publicKeys map[spec.RoomID][]ed25519.PublicKey) (result map[spec.RoomID]map[string]string, err error) {
 	result = make(map[spec.RoomID]map[string]string, len(publicKeys))
-	err = d.Writer.Do(d.DB, nil, func(txn *sql.Tx) error {
```
devonh, line 1760: curious why move this out of the `Do` block

docs/installation/manual/3_signingkey.md
S7evinK: This branch has pulled main, while the other did not.

roomserver/internal/perform/perform_join.go
S7evinK: We have a `isValidID` in GMSL, which checks for a `:`, so added this. It's basically unused.

roomserver/storage/shared/storage.go
S7evinK: With SQLite this would deadlock, as there may be another writer waiting to create a key in the DB. And since we don't write anything in this method, it should be safe without a writer.

roomserver/storage/sqlite3/user_room_keys_table.go
S7evinK: It did work, while we had a writer, which we now don't have, given the change below. Maybe better to use `sqlutil.TxStmt`

devonh (approved):

Issue #2751, opened 2022-09-29
dendrite tries and fails to fetch its own keys


ashkitten:
### Background information

- **Dendrite version or git SHA**: e45ba35e979f2f8af1e75f5f16158973b790436b
- **Monolith or Polylith?**: monolith
- **SQLite3 or Postgres?**: postgres
- **Running in Docker?**: no
- **`go version`**: 1.18
- **Client used (if applicable)**: n/a

### Description

this showed up in the logs while trying to join a room:
```
Sep 29 19:24:11 xenia dendrite-monolith-server[1034867]: time="2022-09-29T19:24:11.913667350Z" level=error msg="Failed to fetch key for server" context=missing error="Post \"matrix://nyrina.link/_matrix/key/v2/query\": context deadline exceeded" fetcher=DirectKeyFetcher
Sep 29 19:24:11 xenia dendrite-monolith-server[1034867]: time="2022-09-29T19:24:11.913732990Z" level=warning msg="Failed to retrieve any keys" context=missing fetcher=DirectKeyFetcher
```

devonh:
Fixed in #3126

PR #3126, opened 2023-06-28
Add pseudoID compatibility to Invites

devonh:


devonh:
This should also fix #2751

clientapi/routing/membership.go
```diff
@@ -337,22 +337,55 @@ func sendInvite(
...
+		return util.JSONResponse{
+			Code: http.StatusBadRequest,
+			JSON: spec.InvalidParam("RoomID is invalid"),
+		}, err
+	}
+	inviter, err := spec.NewUserID(device.UserID, true)
+	if err != nil {
+		return util.JSONResponse{
+			Code: http.StatusInternalServerError,
+			JSON: spec.InternalServerError{},
```
S7evinK, line 351: Also `InvalidParam` as for the invitee below?

devonh: the device isn't a param, it's obtained internally to Dendrite by querying the database using the client info making the request. So I think internal error fits that case.

PR #3131, opened 2023-07-03
[pseudoIDs] Allow displayname changes, kick users on redaction

S7evinK:
This PR changes a few things to allow displayname/avatar changes in pseudo ID rooms.

- `GetRoomsByMembership` now queries the pseudo ID keys to get all rooms
- When redacting a join event containing a `mxid_mapping` in `content`, the user leaves the room or is kicked
- When redacting an event, we now also redact it in `current_room_state` (if needed)
- When storing a membership event (e.g. `join` -> `join`) we now remove the `mxid_mapping` from `unsigned.prev_content`)
- When receiving a membership event over federation, we check if it is a `join` -> `join` event and temporarily add a prev_content, so validating the event signature doesn't error because of a missing `mxid_mapping` 

A few things we need to think about:
Considering that users can leave rooms on their own and rejoin them, we now have two join events with a `mxid_mapping`. Currently redacting the _current_ join event results in a leave/kick, but we still have another unredacted `mxid_mapping`. It's up to the client to walk the `unsigned.replaces_state` field to redact other membership events containing a `mxid_mapping`

Kicking users, on redaction of a join event, may not always be possible, as the powerlevels could have been changed to allow redactions for PL50, but kicking is only allowed for PL100. How do we handle those cases? Temporarily fake/update the powerlevels to allow kicking?


internal/transactionrequest.go
```diff
@@ -208,6 +229,47 @@ func (t *TxnReq) ProcessTransaction(ctx context.Context) (*fclient.RespSend, *ut
...
+
+		// if the currentMembership is join and the new membership as well (e.g. displayname change),
+		// update unsigned so VerifyEventSignatures does not try to validate the mxid_mapping
+		if currentMembership == spec.Join && currentMembership == membership {
+			err = event.SetUnsignedField("prev_content", currentStateEv.Content())
+			if err != nil {
+				return false, err
+			}
+			return true, nil
+		}
```
devonh, line 268: Setting the unsigned field like this feels super hacky. Could we not just tell GMSL not to do the mxid mapping check? Probably not in a generic way that still makes sense with non-pseudoIDs huh?
We could pass `current_membership` to `VerifyEventSignatures` and do the `join == join` check in there, but then that would add another DB hit everywhere this is used. (only ~5-10 places it seems, but still not ideal)
Setting the unsigned field seems fragile if for whatever reason membership events start using this field in the future.

S7evinK: Indeed, this is super hacky. (Looking at where we verify event signatures, this needs to be in other places as well)
I'd like to avoid having to add yet another DB hit for every call to `VerifyEventSignatures`, given it is used in the hotpath when receiving and storing events.
Adding `currentMembership` as a parameter just for it to be used for pseudoIDs doesn't feel right either.

Need to think about what's the best way to avoid this hack.

S7evinK:
Not needed anymore, iirc.

Issue #3096, opened 2023-06-01
Move MSC2946 out of "MSC land"


S7evinK:
While we already have [MSC2946](https://github.com/matrix-org/dendrite/blob/main/setup/mscs/msc2946/msc2946.go), it is still needed to manually enable space summaries. 
We should move it to the fitting spots (e.g. clientapi/federationapi/roomserver), given it is in the spec since version 1.2.
It's also a source of confusion for new users, having to enable the MSC manually.

swedgwood:
Space summaries are no longer hidden behind an MSC flag :) (https://github.com/matrix-org/dendrite/pull/3134)

PR #3134, opened 2023-07-04
de-MSC-ifying space summaries (MSC2946)

swedgwood:
This PR moves and refactors the [code](https://github.com/matrix-org/dendrite/blob/main/setup/mscs/msc2946/msc2946.go) for [MSC2946](https://github.com/matrix-org/matrix-spec-proposals/pull/2946) ('Space Summaries') to integrate it into the rest of the codebase.

Solves #3096 

### Pull Request Checklist

* [ ] I have added Go unit tests or [Complement integration tests](https://github.com/matrix-org/complement) for this PR _or_ I have justified why this PR doesn't need tests

Signed-off-by: Sam Wedgwood <sam@wedgwood.dev>


roomserver/internal/query/query_room_hierarchy.go
```diff
@@ -0,0 +1,597 @@
...
+
+	done bool
+}
+
+const (
+	ConstCreateEventContentKey        = "type"
+	ConstCreateEventContentValueSpace = "m.space"
+	ConstSpaceChildEventType          = "m.space.child"
+	ConstSpaceParentEventType         = "m.space.parent"
+)
```
devonh, line 78: These should be moved to GMSL: https://github.com/matrix-org/gomatrixserverlib/blob/main/spec/eventtypes.go

clientapi/routing/room_hierarchy.go
```diff
@@ -0,0 +1,150 @@
...
+		Code: http.StatusOK,
+		JSON: MSC2946ClientResponse{
+			Rooms:     discoveredRooms,
+			NextBatch: nextBatch,
+		},
+	}
+
+}
+
+type MSC2946ClientResponse struct {
```
devonh, line 147: All these `MSC2946`'s should get swapped out for a more appropriate name.

federationapi/api/api.go
```diff
@@ -73,6 +72,8 @@ type RoomserverFederationAPI interface {
 	GetEventAuth(ctx context.Context, origin, s spec.ServerName, roomVersion gomatrixserverlib.RoomVersion, roomID, eventID string) (res fclient.RespEventAuth, err error)
 	GetEvent(ctx context.Context, origin, s spec.ServerName, eventID string) (res gomatrixserverlib.Transaction, err error)
 	LookupMissingEvents(ctx context.Context, origin, s spec.ServerName, roomID string, missing fclient.MissingEvents, roomVersion gomatrixserverlib.RoomVersion) (res fclient.RespMissingEvents, err error)
+
+	MSC2946Spaces(ctx context.Context, origin, dst spec.ServerName, roomID string, suggestedOnly bool) (res fclient.MSC2946SpacesResponse, err error)
```
devonh, line 76: here too, etc.

roomserver/internal/api.go
```diff
@@ -115,6 +108,15 @@ func (r *RoomserverInternalAPI) SetFederationAPI(fsAPI fsAPI.RoomserverFederatio
...
 	}
 
+	r.Queryer = &query.Queryer{
+		DB:                r.DB,
+		Cache:             r.Cache,
+		IsLocalServerName: r.Cfg.Global.IsLocalServerName,
+		ServerACLs:        r.ServerACLs,
+		Cfg:               r.Cfg,
+		FSAPI:             fsAPI,
+	}
```
devonh, line 118: I'm surprised this hadn't already been moved here with the rest of the `er`'s. It fits much nicer here :+1:

devonh (commented):
Overall things are looking really good and in the right place!

clientapi/routing/room_hierarchy.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/matrix-org/util"
+)
+
+type RoomHierarchyPaginationCache struct {
+	cache map[string]roomserverAPI.CachedRoomHierarchyWalker
+	mu    sync.Mutex
+}
+
+func NewRoomHierarchyPaginationCache() RoomHierarchyPaginationCache {
+	return RoomHierarchyPaginationCache{}
```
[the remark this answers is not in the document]
swedgwood: Done!

clientapi/routing/room_hierarchy.go
swedgwood: Swapped out the ones defined in Dendrite, will change the ones in GMSL separately

roomserver/internal/api.go
swedgwood: Thanks! I also had to move around some of the initialisation code to make it work (in a later commit, https://github.com/matrix-org/dendrite/pull/3134/commits/67f6e3e0ed63703c0f43c6ad54b0349fbc1b7236 )

clientapi/routing/room_hierarchy.go
```diff
@@ -0,0 +1,150 @@
+// Copyright 2021 The Matrix.org Foundation C.I.C.
```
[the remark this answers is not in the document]
swedgwood: Done + added a copyright header to the other new file

roomserver/internal/query/query_room_hierarchy.go
swedgwood: Opened PR here https://github.com/matrix-org/gomatrixserverlib/pull/402

swedgwood:
(rebase on main)

swedgwood:
I've opened this PR up to review now, but just a few things to note:
- I'm looking at adding tests for this PR to complement, so I haven't fully tested these changes yet
- Because I added room hierarchy querying to the `Queryer` struct in the roomserver API, and room hierarchy querying may require requests over federation, I moved around the initialisation order, such that `Queryer` is initialised when `SetFederationAPI` is called. However, as creating a new user API with `userapi.NewInternalAPI` depends on some methods implemented by `Queryer`, this meant I had to move around some lines of code to make this work. Though, in most of the tests, there is no fsAPI to set, so I added a lot of `rsAPI.SetFederationAPI(nil, nil)` to make this work. This feels like a bit of a hack, but not sure how else to do it. Do give me your thoughts when reviewing :)

roomserver/types/types.go
```diff
@@ -336,3 +338,36 @@ func (r *RoomInfo) CopyFrom(r2 *RoomInfo) {
...
+// Struct to represent a device or a server name.
+//
+// May be used to designate a caller for functions that can be called
+// by a client (device) or by a server (server name).
+//
+// Exactly 1 of Device() and ServerName() will return a non-nil result.
+type DeviceOrServerName struct {
+	device     *userapi.Device
+	serverName *spec.ServerName
+}
```
devonh, line 351: Do we need this type? NewDeviceNotServerName never seems to be used, we only use the servername variant. In which case the code that uses this should just take a spec.ServerName.
Or am I overlooking a use case somewhere? I'd prefer not to have a type like this.

cmd/dendrite/main.go
```diff
@@ -157,13 +157,14 @@ func main() {
...
-	userAPI := userapi.NewInternalAPI(processCtx, cfg, cm, &natsInstance, rsAPI, federationClient)
-	asAPI := appservice.NewInternalAPI(processCtx, cfg, &natsInstance, userAPI, rsAPI)
-
 	// The underlying roomserver implementation needs to be able to call the fedsender.
 	// This is different to rsAPI which can be the http client which doesn't need this
 	// dependency. Other components also need updating after their dependencies are up.
 	rsAPI.SetFederationAPI(fsAPI, keyRing)
+
+	userAPI := userapi.NewInternalAPI(processCtx, cfg, cm, &natsInstance, rsAPI, federationClient)
+	asAPI := appservice.NewInternalAPI(processCtx, cfg, &natsInstance, userAPI, rsAPI)
```
devonh, line 166: It would be good to document this ordering dependency somewhere. Maybe both in userapi.NewInternalAPI & appservice.NewInternalAPI?
It would help in the future to determine why this ordering of initialization matters.

The dependency sucks, but our service initialization is already terrible and needs a rework (which we have longer term plans for) so adding in this isn't exactly out of line with the existing usage :face_exhaling:

roomserver/types/types.go
swedgwood: I made this type because room hierarchy querying has slightly different authorisation logic depending on whether you call it from the client API or the server API. `NewDeviceNotServerName` is used in the client api `clientapi/routing/room_hierarchy.go`

cmd/dendrite/main.go
swedgwood: Good idea! And I definitely agree it needs a rework. Committed

docs/FAQ.md
```diff
@@ -64,16 +64,14 @@ Use [dendrite.matrix.org](https://dendrite.matrix.org) which we officially suppo
...
 ```
 mscs:
   mscs:
-    - msc2946
+    - msc2836
 ```
 
-Similarly, [msc2836](https://github.com/matrix-org/matrix-spec-proposals/pull/2836) would need to be added to mscs configuration in order to support Threading. Other MSCs are not currently supported.
-
 Please note that MSCs should be considered experimental and can result in significant usability issues when enabled. If you'd like more details on how MSCs are ratified or the current status of MSCs, please see the [Matrix specification documentation](https://spec.matrix.org/proposals/) on the subject.
```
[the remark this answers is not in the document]
swedgwood: Done!

devonh (approved):

PR #3145, opened 2023-07-07
Set max age for roomserver input stream to avoid excessive interior deletes

neilalexander:
If old messages build up in the input stream and do not get processed successfully, this can create a significant drift between the stream first sequence and the consumer ack floors, which results in a slow and expensive start-up when interest-based retention is in use.

If a message is sat in the stream for 24 hours, it's probably not going to get processed successfully, so let NATS drop them instead. Dendrite can reconcile by fetching missing events later if it needs to.


S7evinK (approved):
As already said, thanks!

PR #3149, opened 2023-07-11
Tweaks around `/messages`

S7evinK:
Try to mitigate some issues with `/messages`

### DIFF
--- a/syncapi/storage/shared/storage_sync.go
+++ b/syncapi/storage/shared/storage_sync.go
@@ -273,7 +273,13 @@ func (d *DatabaseTransaction) GetEventsInTopologicalRange(
 	// Check if we should be able to return events.
 	// If we received 0 events, this most likely means that the provided filter removed them.
 	if len(eIDs) > 0 && len(events) == 0 {
-		return nil, ErrNoEventsForFilter
+		// We try to fetch the events without a filter, so we can tell the client if there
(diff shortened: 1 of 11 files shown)


syncapi/storage/shared/storage_sync.go
```diff
@@ -264,6 +267,21 @@ func (d *DatabaseTransaction) GetEventsInTopologicalRange(
...
 	events, err = d.OutputEvents.SelectEvents(ctx, d.txn, eIDs, filter, true)
+	if err != nil {
+		return
+	}
+	// Check if we should be able to return events.
+	// If we received 0 events, this most likely means that the provided filter removed them.
+	if len(eIDs) > 0 && len(events) == 0 {
+		// We try to fetch the events without a filter, so we can tell the client if there
+		// are more events earlier than the requested and filtered.
+		events, err = d.OutputEvents.SelectEvents(ctx, d.txn, eIDs, nil, true)
```
kegsay, line 278: This doesn't make sense. If we are filtering out events, why are we returning some events?

kegsay (approved):
Dubious API shape with it returning unfiltered events if everything is filtered out, but it makes sense codewise.

Ideally `GetEventsInTopologicalRange` would return the start/end for you, so you don't need to jump through these hoops.

PR #3150, opened 2023-07-11
Optimise getting local members and membership counts

S7evinK:
The previous version was getting **ALL** membership events (as `ClientEvents`, so going through `NewEventFromTrustedJSONWithID`) for a given room.
Now we are querying only locally joined users as `ClientEvents`, which should **significantly** reduce allocations.

Take for example a large room with 2k membership events, but only 1 local user - avoiding 1999 `NewEventFromTrustedJSONWithID` calls just to calculate the `roomSize` which we can also query by other means.

This is also getting called for every `OutputRoomEvent` in the userAPI.

Benchmark with 1 local user and 100 remote users.
```
pkg: github.com/matrix-org/dendrite/userapi/consumers
cpu: 12th Gen Intel(R) Core(TM) i5-12500H
                    │   old.txt   │               new.txt               │
                    │   sec/op    │   sec/op     vs base                │
LocalRoomMembers-16   375.9µ ± 7%   327.6µ ± 6%  -12.85% (p=0.000 n=10)

                    │    old.txt    │               new.txt                │
                    │     B/op      │     B/op      vs base                │
LocalRoomMembers-16   79.426Ki ± 0%   8.507Ki ± 0%  -89.29% (p=0.000 n=10)

                    │   old.txt   │              new.txt               │
                    │  allocs/op  │ allocs/op   vs base                │
LocalRoomMembers-16   1015.0 ± 0%   277.0 ± 0%  -72.71% (p=0.000 n=10)
```

### DIFF
--- a/userapi/consumers/roomserver.go
+++ b/userapi/consumers/roomserver.go
@@ -408,12 +408,18 @@ func (s *OutputRoomEventConsumer) localRoomMembers(ctx context.Context, roomID s
 	req := &rsapi.QueryMembershipsForRoomRequest{
 		RoomID:     roomID,
 		JoinedOnly: true,
+		LocalOnly:  true,
 	}
 	var res rsapi.QueryMembershipsForRoomResponse
 
 	// XXX: This could potentially race if the state for the event is not known yet
 	// e.g. the event came over federation but we do not have the full state persisted.
 	if err := s.rsAPI.QueryMembershipsForRoom(ctx, req, &res); err != nil {
+		return nil, 0, err
+	}
+
+	totalCount, err := s.rsAPI.JoinedUserCount(ctx, roomID)
+	if err != nil {
 		return nil, 0, err
 	}
 
@@ -426,31 +432,18 @@ func (s *OutputRoomEventConsumer) localRoomMembers(ctx context.Context, roomID s
 		if *event.StateKey == "" {
 			continue
 		}
-		_, serverName, err := gomatrixserverlib.SplitID('@', *event.StateKey)
-		if err != nil {
-			log.WithError(err).Error("failed to get servername from statekey")
-			continue
-		}
-		// Only get memberships for our server
-		if serverName != s.serverName {
-			continue
-		}
+		// We're going to trust the Query from above to really just return
+		// local users
 		member, err := newLocalMembership(&event)
 		if err != nil {
 			log.WithError(err).Errorf("Parsing MemberContent")
-			continue
-		}
-		if member.Membership != spec.Join {
-			continue
-		}
-		if member.Domain != s.cfg.Matrix.ServerName {
 			continue
 		}
 
 		members = append(members, member)
 	}
 
-	return members, len(res.JoinEvents), nil
+	return members, totalCount, nil
 }
 
 // roomName returns the name in the event (if type==m.room.name), or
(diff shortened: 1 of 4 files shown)


userapi/consumers/roomserver.go
```diff
@@ -408,6 +408,7 @@ func (s *OutputRoomEventConsumer) localRoomMembers(ctx context.Context, roomID s
 	req := &rsapi.QueryMembershipsForRoomRequest{
 		RoomID:     roomID,
 		JoinedOnly: true,
+		LocalOnly:  true,
```
kegsay, line 411: This was just a bug that we omitted this?

userapi/consumers/roomserver.go
```diff
@@ -426,31 +432,18 @@ func (s *OutputRoomEventConsumer) localRoomMembers(ctx context.Context, roomID s
...
-		}
-		if member.Domain != s.cfg.Matrix.ServerName {
-			continue
-		}
 
 		members = append(members, member)
 	}
 
-	return members, len(res.JoinEvents), nil
+	return members, totalCount, nil
```
kegsay, line 446: How is this better than just doing `len(res.JoinEvents)`? We're still doing the same amount of work? Arguably doing more work as we now also call `s.rsAPI.JoinedUserCount(ctx, roomID)`?

userapi/consumers/roomserver.go
```diff
@@ -417,6 +418,11 @@ func (s *OutputRoomEventConsumer) localRoomMembers(ctx context.Context, roomID s
 		return nil, 0, err
 	}
 
+	totalCount, err := s.rsAPI.JoinedUserCount(ctx, roomID)
```
kegsay, line 421: Right, I _think_ I understand what you're doing here. This needs more comments. It is my understanding that previously we:
 - Called `QueryMembershipsForRoom` to get ALL joined members in the room.
 - Therefore could use `len(res.JoinEvents)` for the joined count.

But now we:
- Call `QueryMembershipsForRoom` to get LOCAL joined members only.
- Need to call this `JoinedUserCount` to get the total joined count incl. remote users?

If so, my questions are:
- Presumably this optimisation does nothing if there are no remote users, or if the bulk of the members are local to this server.
- Is there any difference in perf getting local vs remote users? Presumably no, but because we often want to do operations on local users, there's no point pulling out remote users too?

kegsay (approved):

userapi/consumers/roomserver.go
S7evinK: Yep, that's about it.
Previously we would parse the complete membership events for ALL joined users (which would `json.Unmarshal` for every event), now only for LOCAL users, which can be significantly lower than ALL users. (e.g. on my instance, there is only me in #dendrite, but 2k users over all - so instead of unmarshalling 2k times, we now do it once)
Once we unmarshalled the events, we also store it in our cache, which we may not even need there, so this possibly "pollutes" our cache with events we rarely care about.

If the majority of users is local, this doesn't do anything less than before, expect for one more DB hit to get the user count.

userapi/consumers/roomserver.go
S7evinK: `LocalOnly` was only added after we got push rules, so this may have been an oversight.

PR #3156, opened 2023-07-20
Fix event federation with pseudoID rooms

devonh:


roomserver/internal/input/input_events.go
```diff
@@ -448,6 +448,24 @@ func (r *Inputer) processRoomEvent(
 		return nil
 	}
 
+	// TODO: Revist this to ensure we don't replace a current state mxid_mapping with an older one.
+	if event.Type() == spec.MRoomMember {
```
S7evinK, line 452: Probably better to also check `event.Version() == gomatrixserverlib.RoomVersionPseudoIDs`, so we don't unmarshal every membership event we see.

S7evinK (approved):

PR #3159, opened 2023-07-27
[pseudoIDs] Fixes for room alias tests

swedgwood:
Some (deceptively) simple fixes for some bugs that caused room alias tests to fail (sytext `tests/30rooms/05aliases.pl`). Each commit has details about what it fixes.

Sytest results:

- Sytest before (79d4a0e): https://gist.github.com/swedgwood/972ac4ef93edd130d3db0930703d6c82
- Sytest after (4b09bed): https://gist.github.com/swedgwood/504b00ac4ee892acb757b7fac55fa28a

Room aliases go from `8/15` to `15/15`, but looks like these fixes also managed to fix about `4` other tests, which is a nice bonus :)

### DIFF
--- a/clientapi/routing/directory.go
+++ b/clientapi/routing/directory.go
@@ -181,10 +181,37 @@ func SetLocalAlias(
 		return *resErr
 	}
 
+	maybeRoomID, err := spec.NewRoomID(r.RoomID)
+	if err != nil {
+		return util.JSONResponse{
+			Code: http.StatusBadRequest,
+			JSON: spec.InvalidParam("invalid room ID"),
+		}
+	}
+	roomID := *maybeRoomID // should be safe due to error check
+
+	maybeUserID, err := spec.NewUserID(device.UserID, true)
+	if err != nil {
+		return util.JSONResponse{
+			Code: http.StatusInternalServerError,
+			JSON: spec.Unknown("internal server error"),
+		}
+	}
+	userID := *maybeUserID
+
+	senderID, err := rsAPI.QuerySenderIDForUser(req.Context(), roomID, userID)
+	if err != nil {
+		util.GetLogger(req.Context()).WithError(err).Error("QuerySenderIDForUser failed")
+		return util.JSONResponse{
+			Code: http.StatusInternalServerError,
+			JSON: spec.Unknown("internal server error"),
+		}
+	}
+
 	queryReq := roomserverAPI.SetRoomAliasRequest{
-		UserID: device.UserID,
-		RoomID: r.RoomID,
-		Alias:  alias,
+		SenderID: senderID,
+		RoomID:   r.RoomID,
+		Alias:    alias,
 	}
 	var queryRes roomserverAPI.SetRoomAliasResponse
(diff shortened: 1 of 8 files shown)


roomserver/internal/perform/perform_upgrade.go
```diff
@@ -188,7 +188,7 @@ func moveLocalAliases(ctx context.Context,
 			return fmt.Errorf("Failed to remove old room alias: %w", err)
 		}
 
-		setAliasReq := api.SetRoomAliasRequest{UserID: userID.String(), Alias: alias, RoomID: newRoomID}
+		setAliasReq := api.SetRoomAliasRequest{SenderID: senderID, Alias: alias, RoomID: newRoomID}
```
S7evinK, line 191: Not an issue with this, but would you mind changing the signature of `SetRoomAlias` to `SetRoomAlias(ctx, alias, roomID, senderID)`? 
Slowly moving away from polylith function signatures, also gets rid of `SetRoomAliasRequest`and `SetRoomAliasResponse`.

syncapi/streams/stream_pdu.go
```diff
@@ -519,7 +519,7 @@ func (p *PDUStreamProvider) updatePowerLevelEvent(ctx context.Context, ev *rstyp
 	}
 
 	var evNew gomatrixserverlib.PDU
-	evNew, err = gomatrixserverlib.MustGetRoomVersion(gomatrixserverlib.RoomVersionPseudoIDs).NewEventFromTrustedJSON(newEv, false)
+	evNew, err = gomatrixserverlib.MustGetRoomVersion(gomatrixserverlib.RoomVersionPseudoIDs).NewEventFromTrustedJSONWithEventID(ev.EventID(), newEv, false)
```
S7evinK, line 522: Oh, oops, good find!

S7evinK (approved):
LGTM, maybe we can get rid of some `Request`/`Response` structs with this PR as well? :)

clientapi/routing/directory.go
```diff
@@ -181,10 +181,37 @@ func SetLocalAlias(
 		return *resErr
 	}
 
+	maybeRoomID, err := spec.NewRoomID(r.RoomID)
```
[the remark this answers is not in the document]
swedgwood: Yeah fairs, my thought with it was to dereference as soon as possible, so if there ever was a nil-dereference, it gets caught early - and because go doesn't let you shadow variables I needed a name for the variable which might be a nil. But since we're dereferencing it in the same function anyway it's probably fine, I'll swap it over 👍

syncapi/streams/stream_pdu.go
swedgwood: This one took a long time to diagnose! It would send sytest into loops whilst it waits for a previously sent event ID in the sync response - but only for power level events on pseudo ID rooms. Eventually nailed it down to this though, it's always the one-line fixes that come after hours of debugging

roomserver/internal/perform/perform_upgrade.go
swedgwood: Done! And for removing aliases as well 👍

S7evinK (approved):

PR #3164, opened 2023-08-01
Use `*spec.SenderID` for `QuerySenderIDForUser`

swedgwood:
There are cases where a dendrite instance is unaware of a pseudo ID for a user, the user is not a member of that room. To represent this case, we currently use the 'zero' value, which is often not checked and so causes errors later down the line. To make this case more explict, and to be consistent with `QueryUserIDForSender`, this PR changes this to use a pointer (and `nil` to mean no sender ID).

Note this PR depends on https://github.com/matrix-org/gomatrixserverlib/pull/405 being merged first

### DIFF
--- a/clientapi/routing/redaction.go
+++ b/clientapi/routing/redaction.go
@@ -77,6 +77,7 @@ func SendRedaction(
 	// if user is member of room, and sender ID is nil, then this user doesn't have a pseudo ID for some reason,
 	// which is unexpected.
 	if senderID == nil {
+		util.GetLogger(req.Context()).WithField("userID", *deviceUserID).WithField("roomID", roomID).Error("missing sender ID for user, despite having membership")
 		return util.JSONResponse{
 			Code: http.StatusInternalServerError,
 			JSON: spec.Unknown("internal server error"),
--- a/clientapi/routing/redaction.go
+++ b/clientapi/routing/redaction.go
@@ -74,6 +74,15 @@ func SendRedaction(
 		return *resErr
 	}
 
+	// if user is member of room, and sender ID is nil, then this user doesn't have a pseudo ID for some reason,
+	// which is unexpected.
+	if senderID == nil {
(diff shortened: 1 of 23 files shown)


clientapi/routing/redaction.go
```diff
@@ -74,6 +74,15 @@ func SendRedaction(
 		return *resErr
 	}
 
+	// if user is member of room, and sender ID is nil, then this user doesn't have a pseudo ID for some reason,
+	// which is unexpected.
```
S7evinK, line 78: Probably best to log that as well then.

swedgwood:
After running pseudo ID tests, I've found that some tests that were previously passing are now failing - so I will look into these.

```
161	ok -> not ok	Users can't delete other's aliases
166	ok -> not ok	Only room members can list aliases of a room
227	ok -> not ok	Users cannot kick users from a room they are not in
235	ok -> not ok	m.room.history_visibility == "world_readable" allows/forbids appropriately for Guest users
236	ok -> not ok	m.room.history_visibility == "shared" allows/forbids appropriately for Guest users
239	ok -> not ok	m.room.history_visibility == "default" allows/forbids appropriately for Guest users
243	ok -> not ok	Guest non-joined users cannot send messages to guest_access rooms if not joined
249	ok -> not ok	m.room.history_visibility == "world_readable" allows/forbids appropriately for Real users
250	ok -> not ok	m.room.history_visibility == "shared" allows/forbids appropriately for Real users
253	ok -> not ok	m.room.history_visibility == "default" allows/forbids appropriately for Real users
257	ok -> not ok	Real non-joined users cannot send messages to guest_access rooms if not joined
268	ok -> not ok	/context/ on non world readable room does not work
699	ok -> not ok	AS can make room aliases
```

S7evinK (approved):
Aside from linter being unhappy, LGTM.

PR #3167, opened 2023-08-04
[pseudoID] More pseudo ID fixes

swedgwood:
4 fixes which brings total passing from 597 -> 629 passing

Totals before: `Totals: 597 passed, 47 expected fail, 138 failed`
Totals after: `Totals: 629 passed, 46 expected fail, 107 failed`


federationapi/consumers/presence.go
```diff
@@ -94,16 +95,23 @@ func (t *OutputPresenceConsumer) onMessage(ctx context.Context, msgs []*nats.Msg
...
 	}
 
-	var queryRes roomserverAPI.QueryRoomsForUserResponse
-	err = t.rsAPI.QueryRoomsForUser(t.ctx, &roomserverAPI.QueryRoomsForUserRequest{
-		UserID:         userID,
-		WantMembership: "join",
-	}, &queryRes)
+	parsedUserID, err := spec.NewUserID(userID, true)
+	if err != nil {
+		util.GetLogger(ctx).WithError(err).WithField("user_id", userID).Error("failed to extract domain from receipt sender")
```
devonh, line 100: Is the error message about domain extraction correct? should it just be a more generic "invalid user ID" or something?

roomserver/internal/query/query.go
```diff
@@ -830,13 +846,20 @@ func (r *Queryer) QueryCurrentState(ctx context.Context, req *api.QueryCurrentSt
...
+		return nil, err
 	}
-	res.RoomIDs = roomIDs
-	return nil
+	roomIDs := make([]spec.RoomID, len(roomIDStrs))
+	for i, roomIDStr := range roomIDStrs {
+		roomID, err := spec.NewRoomID(roomIDStr)
+		if err != nil {
+			return nil, nil
+		}
```
devonh, line 859: This logic feels wrong. I think we should be either skipping over this roomID & returning all valid roomIDs, or returning an error here.
Returning no rooms and no error doesn't seem right.
I think simply skipping over the invalid roomID would be best.

clientapi/routing/sendevent.go
```diff
@@ -316,11 +316,18 @@ func generateSendEvent(
 		}
 	}
 	senderID, err := rsAPI.QuerySenderIDForUser(ctx, *validRoomID, *fullUserID)
-	if err != nil || senderID == nil {
+	if err != nil {
 		return nil, &util.JSONResponse{
 			Code: http.StatusNotFound,
 			JSON: spec.NotFound("Unable to find senderID for user"),
 		}
```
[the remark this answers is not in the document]
swedgwood: Probably, yeah, I didn't change it to avoid touching too much existing code with this PR - but I can change it over

federationapi/consumers/presence.go
swedgwood: Ah yeah, you're right. I'm not actually sure where that error message came from. I must've copy/pasted it in by accident

roomserver/internal/query/query.go
swedgwood: Also true! I meant to return the error here - also would invalid room IDs be common place enough for us to want to skip them if they came up? Only because I could maybe see this causing obscure errors in the future if we silently ignore them (although I guess we could warn/error the log as well).

devonh: Unsure. I don't really have a strong preference either way. If you think returning the error is the way to go then let's just do that.

devonh (approved):

PR #3171, opened 2023-08-07
Add config key for default room version

swedgwood:
This PR adds a config key `room_server.default_config_key` to set the default room version for the room server.

Hopefully this should make testing different room versions (looking at you, pseudo IDs!) easier.

depends on https://github.com/matrix-org/gomatrixserverlib/pull/408


kegsay (approved):
LGTM!

PR #3182, opened 2023-08-18
[helm] Update Ingress hosts to account for IPv6 (server+client) and scheme (client)

Omar007:
This updates the matchers for deriving the host values from the dendrite config. The original version turned out to have 2 complications:
- It did not support IPv6 addresses as host value
- It failed for `well_known_client_host` which is a (base) URL instead of a hostname+port.

I've verified `well_known_server_name` with
```
dendrite.example.net:443
dendrite.example.net
192.168.1.1
192.168.1.1:1324
[dead::beef]:1234
[dead::beef]
[ffff:dead::beef]
```
and `well_known_client_name` with:
```
https://dendrite.example.net:443
https://dendrite.example.net
https://dendrite.example.net/
http://dendrite.example.net:8080/
http://192.168.1.1
http://192.168.1.1:8080/
http://[dead::beef]:1234
http://[dead::beef]/
http://[ffff:dead::beef]
```

Fixes #3175

### Pull Request Checklist

* [x] I have added Go unit tests or [Complement integration tests](https://github.com/matrix-org/complement) for this PR _or_ I have justified why this PR doesn't need tests
* [x] Pull request includes a [sign off below using a legally identifiable name](https://matrix-org.github.io/dendrite/development/contributing#sign-off) _or_ I have already signed off privately

Signed-off-by: `Omar Pakker <Omar007@users.noreply.github.com>`

PR #3188, opened 2023-08-24
[helm chart] Strategy wrongly duplicated at pod level

iMartyn:
A pod has no `strategy` field, only a `deployment` does and thus the deployment spec is broken by having `.spec.template.spec.strategy` as well as `.spec.strategy`.

### Pull Request Checklist

Signed-off-by: `Martyn Ranyard <m@rtyn.berlin>`

### DIFF
--- a/helm/dendrite/templates/deployment.yaml
+++ b/helm/dendrite/templates/deployment.yaml
@@ -26,13 +26,6 @@ spec:
       annotations:
         confighash: secret-{{ .Values.dendrite_config | toYaml | sha256sum | trunc 32 }}
     spec:
-      strategy:
-        type: {{ $.Values.strategy.type }}
-        {{- if eq $.Values.strategy.type "RollingUpdate" }}
-        rollingUpdate:
-          maxSurge: {{ $.Values.strategy.rollingUpdate.maxSurge }}
-          maxUnavailable: {{ $.Values.strategy.rollingUpdate.maxUnavailable }}
-        {{- end }}
       volumes:
       - name: {{ include "dendrite.fullname" . }}-conf-vol
         secret:


wrenix:
is already part of #3182

S7evinK:
Thanks for this, but as mentioned this is contained in #3182 which is now merged.

iMartyn:
So a little feedback here...
* @wrenix thumbs-downed my PR which works in argocd on my fork, when the original helm chart did not.
* @wrenix picks up on a duplication in `values.yaml` that is also nice (but not essential from an installation pov) to remove
* @wrenix doesn't even say thanks for the input
Run the project how you want, but #3182 sure is an open PR but not against a bug that is about the helm chart being *broken* (in modern setups it simply would not install).  You expect anyone proposing an enhancement to go in and look at every open PR?

Anyway, glad to see this fixed, but a little thanks (like @S7evinK did) for attempting to fix something would have gone a long way here.  As it is, I would think twice about submitting anything in future.

wrenix:
0. i am not an official person which has not all privileges (on review) and rules (be kind)
1. i make the thumbs down, because it solves not everything (see your point 2) - it was the only way i could take feedback / a review state ...
2. nice if it works with duplicated entries in `values.yaml` - but that should not be there ...
3. the PR #3182 solved two broken things in the helm-chart ( the ingress and this duplicated strategy entry)

I sorry, that i have not give you enough respect and thank you.
That i just push an earlier created PR, which contains multiple important fixed (and a cleanup of `strategy` everywhere) over your work.
No i am not looking in any PR, just in the list of related PRs (if an for me importent pr is not merged over a week or more).
So please be more tolerantly then i am was to not thank you for your work.
If you really want to harm somebody, then me and nobody of this project or the community by not creating PRs anymore.

PR #3196, opened 2023-09-07
Fix `user_id` query param breaking auth for non-appservices

### LINKED ISSUE #1738: Erroneous M_UNKNOWN_TOKEN when supplying a ?user_id for a non-appservice
The server should ignore the `user_id` query parameter if the token does not belong to a registered appservice. It appears as though Dendrite instead returns M_UNKNOWN_TOKEN.

tulir:
The `user_id` query param only has defined behavior when authenticating with an `as_token`. For any other tokens, the presence of the parameter should simply be ignored.

Fixes #1738


devonh (approved):
Thanks for fixing this!

PR #3199, opened 2023-09-13
Move pseudoID ClientEvent hotswapping to a common location

devonh:
Fixes a variety of issues where clients were receiving pseudoIDs in places that should be userIDs.
This change makes pseudoIDs work with sliding sync & element x.

### DIFF
--- a/syncapi/synctypes/clientevent.go
+++ b/syncapi/synctypes/clientevent.go
@@ -153,12 +153,10 @@ func ToClientEvent(se gomatrixserverlib.PDU, format ClientEventFormat, userIDFor
 		// TODO: Set Signatures & Hashes fields
 	}
 
-	if format != FormatSyncFederation {
-		if se.Version() == gomatrixserverlib.RoomVersionPseudoIDs {
-			err := updatePseudoIDs(&ce, se, userIDForSender, format)
-			if err != nil {
-				return nil, err
-			}
+	if format != FormatSyncFederation && se.Version() == gomatrixserverlib.RoomVersionPseudoIDs {
+		err := updatePseudoIDs(&ce, se, userIDForSender, format)
+		if err != nil {
+			return nil, err
 		}
 	}
 
@@ -282,7 +280,7 @@ func updateInviteEvent(userIDForSender spec.UserIDForSender, ev gomatrixserverli
 			return nil, err
 		}
 
-		newState, err := getUpdatedInviteRoomState(userIDForSender, inviteRoomState, ev, *validRoomID, eventFormat)
+		newState, err := GetUpdatedInviteRoomState(userIDForSender, inviteRoomState, ev, *validRoomID, eventFormat)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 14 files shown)


syncapi/synctypes/clientevent.go
```diff
@@ -78,63 +80,62 @@ func ToClientEvents(serverEvs []gomatrixserverlib.PDU, format ClientEventFormat,
 		if se == nil {
 			continue // TODO: shouldn't happen?
 		}
-		if format == FormatSyncFederation {
-			evs = append(evs, ToClientEvent(se, format, string(se.SenderID()), se.StateKey(), spec.RawJSON(se.Unsigned())))
+		ev, err := ToClientEvent(se, format, userIDForSender)
+		if err != nil {
+			logrus.Errorf("Failed converting event to ClientEvent: %s", err.Error())
```
S7evinK, line 85: Since we're continuing, maybe warn?
`logrus.WithError(err).Warn("Failed to convert event to ClientEvent")`?

syncapi/synctypes/clientevent.go
```diff
@@ -78,63 +80,62 @@ func ToClientEvents(serverEvs []gomatrixserverlib.PDU, format ClientEventFormat,
...
 			continue
 		}
+		evs = append(evs, *ev)
+	}
+	return evs
+}
 
-		sender := spec.UserID{}
-		validRoomID, err := spec.NewRoomID(se.RoomID())
+// ToClientEvent converts a single server event to a client event.
```
S7evinK, line 93:
```suggestion
// ToClientEventDefault converts a single server event to a client event.
```

syncapi/synctypes/clientevent.go
```diff
@@ -78,63 +80,62 @@ func ToClientEvents(serverEvs []gomatrixserverlib.PDU, format ClientEventFormat,
...
+	if len(stateKey) >= 1 && stateKey[0] == '@' {
+		parsedStateKey, err := spec.NewUserID(stateKey, true)
 		if err != nil {
-			continue
-		}
-		userID, err := userIDForSender(*validRoomID, se.SenderID())
-		if err == nil && userID != nil {
-			sender = *userID
+			// If invalid user ID, then there is no associated state event.
+			return nil, fmt.Errorf("Provided state key begins with @ but is not a valid user ID: %s", err.Error())
```
S7evinK, line 114: `%w` instead of `%s` please, so we can unwrap the error and use things like `errors.Is()`

syncapi/synctypes/clientevent.go
```diff
@@ -154,60 +155,276 @@ func ToClientEvent(se gomatrixserverlib.PDU, format ClientEventFormat, sender st
...
 
-// ToClientEvent converts a single server event to a client event.
-// It provides default logic for event.SenderID & event.StateKey -> userID conversions.
-func ToClientEventDefault(userIDQuery spec.UserIDForSender, event gomatrixserverlib.PDU) ClientEvent {
-	sender := spec.UserID{}
-	validRoomID, err := spec.NewRoomID(event.RoomID())
+func updatePseudoIDs(ce *ClientEvent, se gomatrixserverlib.PDU, userIDForSender spec.UserIDForSender, format ClientEventFormat) error {
+	ce.SenderKey = se.SenderID()
+
+	validRoomID, err := spec.NewRoomID(se.RoomID())
```
S7evinK, line 171: This has somewhat bothered me all the time.
We have, in theory, already a valid event, why do we need to parse the roomID it again as a `RoomID`.. `RoomID` in the `eventFields` struct should probably be a `spec.RoomID`. (at some point :D)

S7evinK (commented):
There are a few places where we could benefit from error wrapping using `%w` instead of `%s`.

syncapi/synctypes/clientevent.go
devonh: Done! :)

PR #3204, opened 2023-09-22
Support for room version v11

### LINKED ISSUE #3203: Room Version 11 Compatibility
**Description:** Dendrite users seem to currently be unable to view v11 rooms.

S7evinK:
Fixes #3203

### DIFF
--- a/syncapi/synctypes/clientevent.go
+++ b/syncapi/synctypes/clientevent.go
@@ -347,71 +347,81 @@ func updatePowerLevelEvent(userIDForSender spec.UserIDForSender, se gomatrixserv
 		return se, nil
 	}
 
-	pls, err := gomatrixserverlib.NewPowerLevelContentFromEvent(se)
-	if err != nil {
-		return nil, err
-	}
-	newPls := make(map[string]int64)
-	var userID *spec.UserID
-	for user, level := range pls.Users {
-		if eventFormat != FormatSyncFederation {
-			userID, err = userIDForSender(se.RoomID(), spec.SenderID(user))
-			if err != nil {
-				return nil, err
-			}
-			user = userID.String()
-		}
-		newPls[user] = level
-	}
-	var newPlBytes, newEv []byte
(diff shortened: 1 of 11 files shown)


syncapi/synctypes/clientevent.go
```diff
@@ -141,6 +141,18 @@ func ToClientEvent(se gomatrixserverlib.PDU, format ClientEventFormat, userIDFor
 		Redacts:        se.Redacts(),
 	}
 
+	// RoomVersion V11 moved the "redacts" field to "content", so add it there as well
+	// For compatibility with older clients, we keep the "redacts" top-level field.
+	// TODO: At which point do we remove "redacts" from the top-level?
+	if se.Type() == spec.MRoomRedaction && se.Version() == gomatrixserverlib.RoomVersionV11 && ce.Redacts != "" {
```
devonh, line 147: I don't think this should have the `se.Version() == gomatrixserverlib.RoomVersionV11` check.
I think the logic is inverted - ie. for any room version where there is a top level redacts field we should be copying that into the content for clients.
I think we can just remove the version check entirely.
```
For improved compatibility with newer clients, servers should add a redacts property to the content of m.room.redaction events in older room versions when serving such events over the Client-Server API.
```

syncapi/synctypes/clientevent.go
```diff
@@ -141,6 +141,18 @@ func ToClientEvent(se gomatrixserverlib.PDU, format ClientEventFormat, userIDFor
...
+		v11content, err := sjson.SetBytes(ce.Content, "redacts", ce.Redacts)
+		if err != nil {
+			logrus.WithError(err).Warn("failed to set content.redacts")
+		} else {
+			ce.Content = v11content
+		}
+	}
+
```
devonh, line 155: Do we cover this case for redactions:
```
For backwards-compatibility with older clients, servers should add a redacts property to the top level of m.room.redaction events in when serving such events over the Client-Server API.
```

devonh (approved):

Issue #3180, opened 2023-08-15
Do something about invalid well-known names


Elara6331:
**Description:**

The comments above `well_known_client_name` and `well_known_server_name` in the sample config have `localhost:443` as an example. This leads to people using things like `matrix.example.com:443` as their client/server names. That works for Element, Nheko, and some others, but many clients, including Fluffychat, NeoChat, Fractal, and Cinny require a scheme to be part of the name or they fail to connect. This appears to be based on the following sentence in the matrix spec:

> Parse it as a URL. If it is not a URL, then `FAIL_ERROR`.

which can be found at https://spec.matrix.org/v1.7/client-server-api/#well-known-uri

Either the examples should be changed to `https://localhost:443` and a warning or error logged if the name is invalid, or Dendrite should automatically add the scheme if it doesn't exist.

S7evinK:
https://github.com/matrix-org/dendrite/blob/1c4ec67bb6897859617fea263c658142a694e26d/dendrite-sample.yaml#L68-L73

Looks fine to me?
We could still check for a prefix here and log a warning if it's missing.

Elara6331:
Yeah, that looks fine. Looks like I had an old version of the sample config. That line was changed 3 months ago. I do think it'd be useful to log a warning because it took me a while to figure out what was wrong there.

S7evinK:
We're a bit more radical now than just logging a warning, we're failing to start up if `well_known_client_name` is not formatted in a proper way.

PR #3205, opened 2023-09-25
added a   warning log , for well_known_server_name,well_known_server_name when they dont have prefix 

Mothergoose31:
closing this https://github.com/matrix-org/dendrite/issues/3180

added a   warning log   when either well_known_server_name,  well_known_server_name:  dont have a prefix in them 

josephalvarengabeech@pm.me


clientapi/routing/routing.go
```diff
@@ -129,6 +129,10 @@ func Setup(
 		})).Methods(http.MethodGet, http.MethodOptions)
 	}
 
+	if !strings.HasPrefix(string(cfg.Matrix.ServerName), "http://") || !strings.HasPrefix(string(cfg.Matrix.ServerName), "https://") {
+		logrus.Warn("The well_known_server_name does not start with http:// or https:// does not start with 'http://' or 'https://'. Some clients may fail to connect.")
```
S7evinK, line 133: Looks like accidentally duplication? (as well as `well_known_client_name` instead of `well_known_server_name`)
Also better to put this inside the `if cfg.Matrix.WellKnownClientName != "" {` block above. (same for the federation change below)

S7evinK:
Right, updated to only check client well-known, as that's the only well-known that actually needs a http/https prefix.

S7evinK:
Thanks!

PR #3206, opened 2023-09-26
Complement fixes

devonh:


syncapi/synctypes/clientevent.go
```diff
@@ -363,14 +363,17 @@ func updatePowerLevelEvent(userIDForSender spec.UserIDForSender, se gomatrixserv
...
-	newPlBytes, err = json.Marshal(newPls)
-	if err != nil {
-		return nil, err
-	}
-	newEv, err = sjson.SetRawBytes(se.JSON(), "content.users", newPlBytes)
-	if err != nil {
-		return nil, err
+	var newPlBytes []byte
+	newEv := se.JSON()
+	if len(pls.Users) > 0 {
```
kegsay, line 368: I think you mean nil check here? Else it's impossible to set `users: {}`? Not that I'm sure why you'd want to..

roomserver/internal/perform/perform_leave.go
```diff
@@ -93,11 +93,21 @@ func (r *Leaver) performLeaveRoomByID(
...
+			return nil, fmt.Errorf("failed looking up userID for sender %q: %w", senderUser, serr)
 		}
-		if !r.Cfg.Matrix.IsLocalServerName(sender.Domain()) {
-			return r.performFederatedRejectInvite(ctx, req, res, *sender, eventID, *leaver)
+
+		var domain spec.ServerName
+		if sender == nil {
+			// TODO: Currently a federated invite has no way of knowing the mxid_mapping of the inviter.
+			// Should we add the inviter's m.room.member event (with mxid_mapping) to invite_room_state to allow
+			// the invited user to leave via the inviter's server?
```
kegsay, line 104: Seems reasonable.

kegsay (approved):

syncapi/synctypes/clientevent.go
devonh: For the particular test where there was no users field, I wanted to not present an empty users field to the client. With our logic we were adding in the field even if it wasn't present in the first place.
I forgot that someone may actually want to have an empty users field.

I've changed the logic now to allow an empty users field while also not adding it in if it isn't there in the first place.

PR #3220, opened 2023-09-29
fix(helm): repo index

wrenix:
Something goes wrong with the chart-releaser:
it should not upload an *.tgz into gh-pages (and if it should be stored under /docs).

@S7evinK do you like to take a look?


S7evinK (approved):
Thanks! Hoping to get https://github.com/helm/chart-releaser-action/pull/163 merged soon.

PR #3227, opened 2023-10-02
Tweaks around the device list updater

S7evinK:
I hope the comments explain the changes.

`notifyWorkers` notifies a worker which then calls `processServer`, which in turn gets all users and calls `processServerUser`. There is no need to call `processServer` for the same domain on startup.


devonh:
The changes look reasonable. Is there any way to add tests for this?

Issue #2508, opened 2022-06-01
Users can no longer write after changing their display names


mpldr:
### Background information

- **Dendrite version or git SHA**: 0.8.5
- **Monolith or Polylith?**: Monolith
- **SQLite3 or Postgres?**: Postgres
- **Running in Docker?**: yes
- **`go version`**: /shrug
- **Client used (if applicable)**: Element/Schildichat

### Description

 - **What** is the problem: Users can't write in their encrypted rooms anymore after changing their displaynames
 - **Who** is affected: Everyone who has tried so far
 - **How** is this bug manifesting: Messages can't be sent ("Some of your messages have not been sent. You can select all or individual messages to retry or delete")
 - **When** did this first appear: after setting the server up some days ago

The logs are too long to just "find" related events… anything I should look out for?

### Steps to reproduce

- Create user
- Invite user to encrypted room
- User logs in for the first time
- User joins Room
- User sends "Hello" which is successfully received
- User changes displayname
- User cannot send messages to the room anymore

<!--
Describe how what happens differs from what you expected.

If you can identify any relevant log snippets from server logs, please include
those (please be careful to remove any personal or private data). Please surround them with
``` (three backticks, on a line on their own), so that they are formatted legibly.

Alternatively, please send logs to @kegan:matrix.org or @neilalexander:matrix.org
with a link to the respective Github issue, thanks!
-->

**Comment by neilalexander** [comment]
Please try the latest `main` commit, since this should be fixed by #2507.

**Comment by mpldr** [comment]
Updated to 0.8.7 and still no message sent. Only log message generated:
```
matrix-dendrite-1  | time="2022-06-02T10:16:02.991790972Z" level=info msg=PerformPusherCreation display_name="FluffyChat ios" localpart=user pushkey="cuohufdMj0UMrcSAjVptnC:APA91bHCV70RATc4b2RZLRj5F6xvBQ9PMph-duZNGmjzdQqCn9g5pWLNMxhB0fKNjK1F_y3PcaEcQjI7u0GoxEIX7_lEE59ToHj4pgkxNGsc6ZzYVDQnbme0ZtTXKrBXTb3JxJeKq0nc" req.id=Ok1Y7UjL9rzQ req.method=POST req.path=/_matrix/client/r0/pushers/set user_id="@user:server.com"
```

**Comment by mpldr** [comment]
Leaving said private chat is also not possible:

```
[...]

Issue #1760, opened 2021-02-09
stopped getting new events from room, can't rejoin it


zephryn:
### Background information
- **Dendrite version or git SHA**: still on 0.3.8
- **Monolith or Polylith?**: monolith
- **SQLite3 or Postgres?**: sqlite3
- **Running in Docker?**: nope
- **`go version`**: 1.15.8 linux/amd64

### Description
a few days ago i noticed that a room i was in wasn't receiving new events.

when trying to leave the room, i get the following error:
```
Error leaving room
User "(matrix account)" is not a member of room "(id of room i can't join)"
```
attempting to rejoin the room returns a different error:
```
InputRoomEvents auth failed: eventauth: eventauth: "(matrix account)" is not allowed to change their membership from "leave" to "join"
```
this message is similar to the one shown in issue #1023.

trying to rejoin also produces these logs server-side:
```
level=error msg="helpers.CheckAuthEvents failed for event, rejecting event" func="processRoomEvent\n\t" file=" [github.com/matrix-org/dendrite/roomserver/internal/input/input_events.go:106]" auth_event_ids="[(event id)]" error="eventauth: \"(matrix account)\" is not allowed to change their membership from \"leave\" to \"join\"" event_id="(event id)"
level=info msg="Error authing soft-failed event" func="processRoomEvent\n\t" file=" [github.com/matrix-org/dendrite/roomserver/internal/input/input_events.go:120]" error="eventauth: \"(matrix account)\" is not allowed to change their membership from \"leave\" to \"join\"" event_id="(event id)" room="(id of room i can't join)" type=m.room.member
http: TLS handshake error from (client-side ip goes here): EOF
```

zephryn:
short update on this issue: these room errors came and went for a while in two rooms i was in and have not reoccurred for about a month now. not sure if the issue was resolved with a new release or if it just happened to go away after some time.

kegsay:
Possibly related: https://github.com/matrix-org/dendrite/issues/1662

PR #3231, opened 2023-10-06
Fix state resets

S7evinK:
Needs https://github.com/matrix-org/gomatrixserverlib/pull/419

May fix: https://github.com/matrix-org/dendrite/issues/2508, https://github.com/matrix-org/dendrite/issues/1760


clientapi/routing/profile.go
```diff
@@ -300,7 +300,7 @@ func updateProfile(
 		}, e
 	}
 
-	if err := api.SendEvents(ctx, rsAPI, api.KindNew, events, device.UserDomain(), domain, domain, nil, true); err != nil {
+	if err := api.SendEvents(ctx, rsAPI, api.KindNew, events, device.UserDomain(), domain, domain, nil, false); err != nil {
```
S7evinK, line 303: This slows down changing profiles, but makes sure that all events have been sent to the roomserver.
This is needed, as e.g. bridges may change the avatar and displayname in short succession.

devonh (approved):
Changes look great & test is clear!

PR #3243, opened 2023-10-24
Check event is not rejected 

S7evinK:
Companion PR to https://github.com/matrix-org/gomatrixserverlib/pull/421


kegsay (approved):
I have concerns here that because we do not recalculate rejected-ness, we may cascade fail later events incorrectly.

A reminder of what sets rejected = true:

> - Passes authorization rules based on the event’s auth events, otherwise it is rejected.
> - Passes authorization rules based on the state before the event, otherwise it is rejected.

Rejected-ness is not a pure yes/no (unlike say hash checks) because the server may fail to fetch the auth events / state before the event due to connectivity issues. At that point, the event is neither accepted nor rejected, it's indetermined.

It's unclear what the best course of action is when there are network connectivity issues as this can become an attack vector. It's clearly not correct to _accept_ it (else an adversary in control of the network can just firewall off the server and send rubbish to it, and have the server merrily accept it). The opposite is a denial of service attack though: valid events can be made invalid by preventing the server talking to other servers. This isn't the only place where this can happen in federation today (/gme dishing up prev_events is another) so it's less bad I guess, but we definitely do need to keep retrying to fetch the events if we cannot.

PR #3191, opened 2023-08-26
fix(helm): empty storage class in pvcs

wrenix:
fix #3103 

---

not yet tested

PR #3245, opened 2023-10-25
chore(helm): use empty/nil storageClass for helm-docs

wrenix:
i believe that `nil` would be false in the if :
```yaml
storageClass:
```
is still handled correct.

---
In past ( #3191 ), will have the problem with an empty string `""`:
```yaml
storageClass: ""
```

---
do you take another look @S7evinK ?


S7evinK (approved):
LGTM, thanks!

PR #3254, opened 2023-11-01
Use `IsBlacklistedOrBackingOff` to determine if we should try to fetch devices

S7evinK:
Use `IsBlacklistedOrBackingOff` from the federation API to check if we should fetch devices.

To reduce back pressure, we now only queue retrying servers if there's space in the channel.


federationapi/routing/profile_test.go
```diff
@@ -67,11 +66,8 @@ func TestHandleQueryProfile(t *testing.T) {
 		keyRing := serverKeyAPI.KeyRing()
 		fedapi := fedAPI.NewInternalAPI(processCtx, cfg, cm, &natsInstance, &fedClient, nil, nil, keyRing, true)
 		userapi := fakeUserAPI{}
-		r, ok := fedapi.(*fedInternal.FederationInternalAPI)
-		if !ok {
```
kegsay, line 71: Why remove?

userapi/internal/device_list_update.go
```diff
@@ -362,13 +368,20 @@ func (u *DeviceListUpdater) notifyWorkers(userID string) {
 	if err != nil {
 		return
 	}
+	_, err = u.isBlacklistedOrBackingOffFn(remoteServer)
+	var federationClientError *fedsenderapi.FederationClientError
+	if errors.As(err, &federationClientError) {
+		if federationClientError.Blacklisted {
+			return
```
kegsay, line 375: Needs unit tests.

userapi/internal/device_list_update.go
```diff
@@ -362,13 +368,20 @@ func (u *DeviceListUpdater) notifyWorkers(userID string) {
...
+		}
+	}
+
 	hash := fnv.New32a()
 	_, _ = hash.Write([]byte(remoteServer))
 	index := int(int64(hash.Sum32()) % int64(len(u.workerChans)))
 
 	ch := u.assignChannel(userID)
 	deviceListUpdaterBackpressure.With(prometheus.Labels{"worker_id": strconv.Itoa(index)}).Inc()
-	defer deviceListUpdaterBackpressure.With(prometheus.Labels{"worker_id": strconv.Itoa(index)}).Dec()
```
kegsay, line 371: Why remove? Using a `defer` is much cleaner than what we have now.

userapi/internal/device_list_update.go
```diff
@@ -405,24 +418,40 @@ func (u *DeviceListUpdater) worker(ch chan spec.ServerName, workerID int) {
...
+			// The channel is at capacity, don't try to send more work
+			if len(ch) == cap(ch) {
+				continue
+			}
 			serversToRetry = serversToRetry[:0] // reuse memory
-			time.Sleep(time.Second)
+
+			// -2, so we have space for incoming device list updates over federation
+			maxServers := (cap(ch) - len(ch)) - 2
+			if maxServers <= 0 {
```
kegsay, line 432: This check is also effectively doing `if len(ch) == cap(ch)` because if they are `==` then `N-N-2 <= 0`, maybe just have this check and drop the earlier one?

userapi/internal/device_list_update.go
```diff
@@ -405,24 +418,40 @@ func (u *DeviceListUpdater) worker(ch chan spec.ServerName, workerID int) {
...
 		var serversToRetry []spec.ServerName
 		for {
+			deviceListUpdaterServersRetrying.With(prometheus.Labels{"worker_id": strconv.Itoa(workerID)}).Set(float64(len(retries)))
+			time.Sleep(time.Second * 2)
+
+			// The channel is at capacity, don't try to send more work
+			if len(ch) == cap(ch) {
+				continue
+			}
 			serversToRetry = serversToRetry[:0] // reuse memory
```
kegsay, line 428: This is super unclear that it's nuking the `serversToRetry` slice, and why. Can we add more comments here please?

userapi/internal/device_list_update.go
```diff
@@ -431,6 +460,7 @@ func (u *DeviceListUpdater) worker(ch chan spec.ServerName, workerID int) {
 		_, exists := retries[serverName]
 		retriesMu.Unlock()
 		if exists {
+			deviceListUpdaterBackpressure.With(prometheus.Labels{"worker_id": strconv.Itoa(workerID)}).Dec()
```
kegsay, line 463: I am unconvinced that this is going to track counts correctly.  It is incremented for every call to `notifyWorkers(userID)` but only decremented under very certain scenarios (when not full, etc).

kegsay (changes requested):
Mostly good, few things.

userapi/internal/device_list_update.go
S7evinK: Just because we sent the work to the worker doesn't mean the back pressure is gone, the worker might still be processing the server. So decrementing doesn't make sense here.
We're incrementing right before trying to send the work and the worker decrements it once it is done processing the server. (either when skipping or actually done)

userapi/internal/device_list_update.go
S7evinK: See above.

kegsay (approved):

PR #3256, opened 2023-11-02
Use port number instead of name for k8s service port

Xinayder:
I've found an issue when deploying Dendrite's Helm chart on my local cluster. The template for generating an Ingress resource tries to find the service port using a name (`http`), but the template that generates the Service resource, instead, identifies the resource with a port number.

According to the [Kubernetes ServiceSpec](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/service-v1/), `ports.targetPort` can be either a number or a string; if it's the latter, it will be looked up as a named port in the pod's container ports.

### Pull Request Checklist


Xinayder:
I did some more testing and it turns out that even if you set the `targetPort` property on the Service definition, the ingress controller won't find the pod.

However, if you set the `service.port` property on the ingress resource to the port number, it works. If it's set to `http`, it doesn't.

InezMc:
Confirm that private signoff recieved.

Xinayder:
Still haven't received an email back from Matrix regarding my private signoff request.

Xinayder:
Sign-off confirmed.

S7evinK (approved):
Makes sense. Thanks!

PR #3296, opened 2024-01-06
Fix notary keys requests for all keys

S7evinK:
This should be more spec compliant:
> If no key IDs are given to be queried, the notary server should query for all keys.

### DIFF
--- a/federationapi/federationapi_test.go
+++ b/federationapi/federationapi_test.go
@@ -5,11 +5,14 @@ import (
 	"crypto/ed25519"
 	"encoding/json"
 	"fmt"
+	"net/http"
+	"net/http/httptest"
 	"strings"
 	"sync"
 	"testing"
 	"time"
 
+	"github.com/matrix-org/dendrite/federationapi/routing"
 	"github.com/matrix-org/dendrite/internal/caching"
 	"github.com/matrix-org/dendrite/internal/httputil"
 	"github.com/matrix-org/dendrite/internal/sqlutil"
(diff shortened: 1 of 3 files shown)


federationapi/federationapi_test.go
```diff
@@ -362,3 +368,118 @@ func TestRoomsV3URLEscapeDoNot404(t *testing.T) {
...
+		httpBody      string
+		pubKeyRequest *gomatrixserverlib.PublicKeyNotaryLookupRequest
+		validateFunc  func(t *testing.T, response util.JSONResponse)
+	}{
+		{
+			name: "empty httpBody",
+			validateFunc: func(t *testing.T, resp util.JSONResponse) {
+				want := util.JSONResponse{
+					Code: http.StatusBadRequest,
+					JSON: spec.BadJSON("The request body could not be decoded into valid JSON. unexpected end of JSON input"),
```
kegsay, line 386: It would be nicer to assert the error code and not the entire error message, as that is pretty brittle.

federationapi/internal/query.go
```diff
@@ -43,6 +43,15 @@ func (a *FederationInternalAPI) fetchServerKeysFromCache(
 	ctx context.Context, req *api.QueryServerKeysRequest,
 ) ([]gomatrixserverlib.ServerKeys, error) {
 	var results []gomatrixserverlib.ServerKeys
+
+	// We got a request for _all_ server keys, return them.
+	if len(req.KeyIDToCriteria) == 0 {
```
kegsay, line 48: Technically `{"server_keys":{"servera":null}}` will then be treated as getting all keys.

federationapi/routing/keys.go
```diff
@@ -217,10 +221,7 @@ func NotaryKeys(
 		}
 	}
 
-	var response struct {
-		ServerKeys []json.RawMessage `json:"server_keys"`
-	}
-	response.ServerKeys = []json.RawMessage{}
```
kegsay, line 223: Bad refactor. You forgot to do this, so no keys will be `server_keys: null` not `server_keys: []`.

kegsay (changes requested):
Please add additional testing for the `null` vs `[]` edge case.

PR #3305, opened 2024-01-17
Update sentry reporting

S7evinK:
This hopefully reduces the garbage we currently produce.
(Using [GlitchTip](https://glitchtip.com/) on my personal instance, this seems to look better)

### DIFF
--- a/federationapi/routing/routing.go
+++ b/federationapi/routing/routing.go
@@ -647,6 +647,8 @@ func MakeFedAPI(
 		// add the user to Sentry, if enabled
 		hub := sentry.GetHubFromContext(req.Context())
 		if hub != nil {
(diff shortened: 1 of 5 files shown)


devonh (commented):
Wouldn't we also want to do this here as well: https://github.com/matrix-org/dendrite/blob/bebf701dce01d47264d694b118c81dcb84a37b04/federationapi/routing/routing.go#L648

devonh (commented):
Should we be doing a similar thing everywhere we use CaptureException or CaptureMessage?
The docs would suggest we should https://docs.sentry.io/platforms/go/usage/concurrency/

S7evinK:
Created https://github.com/matrix-org/dendrite/issues/3306 as a follow up task

Issue #3273, opened 2023-11-24
Complement flake `TestPushSync/Adding_a_push_rule_wakes_up_an_incremental`


S7evinK:
Most recent: https://github.com/matrix-org/dendrite/actions/runs/6984545139/job/19007575361?pr=3078#step:7:123

PR #3307, opened 2024-01-18
Don't send device list updates upon registration

S7evinK:
Fixes https://github.com/matrix-org/dendrite/issues/3273

As we otherwise send down device list updates which are merely useful for the user and causes tests to be flakey:

```
❌ TestPushSync/Adding_a_push_rule_wakes_up_an_incremental_/sync (10ms)
      push_test.go:57: no pushrules found in sync response: {"next_batch":"s0_0_0_0_0_1_1_0_1","device_lists":{"changed":["@user-1:hs1"]}}
```

What this does: If a `PerformDeviceCreation` request is coming from registering an account, it does **not** send device list updates, as they are merely useful (no joined rooms, no one to inform) . In all other cases, the behavior is unchanged and device list updates are sent as usual.

### DIFF
--- a/.github/codecov.yaml
+++ b/.github/codecov.yaml
@@ -7,7 +7,7 @@ coverage:
     project:
       default:
         target: auto
-        threshold: 0%
+        threshold: 0.1%
         base: auto
         flags:
           - unittests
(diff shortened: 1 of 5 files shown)


S7evinK:
Also sneakily changes the project coverage threshold to 0.1%, as it is quite annoying to have CI failing just because the a nightly run had slightly higher coverage.

Issue #3226, opened 2023-10-02
Declare support for Matrix 1.8


S7evinK:
### Client-Server API

**Backwards Compatible Changes**

- [ ] Require callers to be joined to the room to report its events, as per [MSC2249](https://github.com/matrix-org/matrix-spec-proposals/pull/2249).
  - No `/report` endpoint at all (see https://github.com/matrix-org/dendrite/issues/3216)

**Spec Clarifications**

- [ ] Allow `null` in `room_types` in `POST /publicRooms` endpoints schemas. ([#1564](https://github.com/matrix-org/matrix-spec/issues/1564))
  - No filtering atm (see https://github.com/matrix-org/dendrite/issues/3222)

### Server-Server API

**Deprecations**
- [x] Deprecate `matrix` SRV lookup steps during server discovery, as per [MSC4040](https://github.com/matrix-org/matrix-spec-proposals/pull/4040).
  - https://github.com/matrix-org/gomatrixserverlib/pull/411

**Backwards Compatible Changes**
- [x] Add `matrix-fed` SRV lookup steps to server discovery, as per [MSC4040](https://github.com/matrix-org/matrix-spec-proposals/pull/4040).
  - https://github.com/matrix-org/gomatrixserverlib/pull/411

**Spec Clarifications**
- [x] Document why `/state_ids` can respond with a 404. ([#1521](https://github.com/matrix-org/matrix-spec/issues/1521))
  - We already 404 if we can't find the event

### Room Versions

**Backwards Compatible Changes**

- [x] Add room version 11 as per [MSC3820](https://github.com/matrix-org/matrix-spec-proposals/pull/3820).
    - Move `redacts` from top level to `content` on `m.room.redaction` events in room version 11, as per [MSC2174](https://github.com/matrix-org/matrix-spec-proposals/pull/2174).
    - Remove `creator` from `m.room.creator` events in room version 11, as per [MSC2175](https://github.com/matrix-org/matrix-spec-proposals/pull/2175).
    - Remove remaining usage of `origin` from events in room version 11, as per [MSC3989](https://github.com/matrix-org/matrix-spec-proposals/pull/3989).
    - Update the redaction rules in room version 11, as per [MSC2176](https://github.com/matrix-org/matrix-spec-proposals/pull/2176) and [MSC3821](https://github.com/matrix-org/matrix-spec-proposals/pull/3821).

### Appendices

**Backwards Compatible Changes**

[...]

PR #3313, opened 2024-01-25
Allow + in MIDs as per MSC4009

RosstheRoss:
This PR adds `+` to the username regex, per MSC4009.

### Pull Request Checklist

Signed-off-by: `Matt Strapp <matt@mattstrapp.net>`


RosstheRoss:
The original version of this PR had a bug where the regex was too loose and would allow some not allowed characters in a range, I added another test so no one else makes the same mistake I did 😅

S7evinK (approved):
LGTM, thanks for this!
One task down on #3226 :)

PR #3334, opened 2024-02-20
Speed up start up time by batch querying ACL events

S7evinK:
This should significantly speed up start up times on servers with many rooms.


roomserver/storage/tables/interface.go
```diff
@@ -235,6 +235,8 @@ func ExtractContentValue(ev *types.HeaderedEvent) string {
 		key = "topic"
 	case "m.room.guest_access":
 		key = "guest_access"
+	case "m.room.server_acl":
+		return string(content)
```
kegsay, line 239: This feels wrong, but I understand why you're doing it (all keys in content are required). Needs a comment.

kegsay (approved):

PR #3337, opened 2024-02-26
Make usage of relays optional, avoid DB roundtrips

S7evinK:
This should avoid 2 additional DB roundtrips if we don't want to use relays.

So instead of possibly doing roughly 20k trips to the DB, we are now "only" doing ~6600.

### DIFF
--- a/federationapi/federationapi.go
+++ b/federationapi/federationapi.go
@@ -113,7 +113,7 @@ func NewInternalAPI(
 		_ = federationDB.RemoveAllServersFromBlacklist()
 	}
 
-	stats := statistics.NewStatistics(federationDB, cfg.FederationMaxRetries+1, cfg.P2PFederationRetriesUntilAssumedOffline+1, false)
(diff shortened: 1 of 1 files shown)


federationapi/federationapi.go
```diff
@@ -113,10 +113,7 @@ func NewInternalAPI(
 		_ = federationDB.RemoveAllServersFromBlacklist()
 	}
 
-	stats := statistics.NewStatistics(
-		federationDB,
-		cfg.FederationMaxRetries+1,
-		cfg.P2PFederationRetriesUntilAssumedOffline+1)
+	stats := statistics.NewStatistics(federationDB, cfg.FederationMaxRetries+1, cfg.P2PFederationRetriesUntilAssumedOffline+1, false)
```
devonh, line 116: This should use the new config option.

```suggestion
	stats := statistics.NewStatistics(federationDB, cfg.FederationMaxRetries+1, cfg.P2PFederationRetriesUntilAssumedOffline+1, cfg.EnableRelays)
```

devonh (approved):

federationapi/federationapi.go
S7evinK: D'oh, that was the whole point of this PR. :D

Issue #3358, opened 2024-04-09
Edited messages appear twice in fulltext search


S7evinK:
When a message is edited, they appear twice in the results of `/search`.
While Element Web seems to aggregate the results and only returns one results, Element Android returns both. This may be a bit confusing.

IMO, Dendrite should only return the latest message.

The responsible code for this:
https://github.com/matrix-org/dendrite/blob/14a6c100977be6484eec419fdc8f496eb27b0677/syncapi/consumers/roomserver.go#L593-L627

We'd need to check if `m.relates.to` in `content` contains a `rel_type` of `m.replace` and then delete the `event_id` from the fulltext index before indexing the new event.

adnull:
Hi! I'd like to take this.

S7evinK:
Sure, go for it! :)

S7evinK:
Fixed with #3363.

PR #3363, opened 2024-04-18
Fix: Edited messages appear twice in fulltext search

adnull:
As stated in https://github.com/matrix-org/dendrite/issues/3358 the search response contains both original and edited message.
This PR fixes it by removing of the original message from the fulltext index after indexing the edit message event.
I also made some cosmetic changes/fixes i found in the code

Signed-off-by: `Alexander Dubovikov <d.lexand@gmail.com>`

### DIFF
--- a/syncapi/consumers/roomserver.go
+++ b/syncapi/consumers/roomserver.go
@@ -622,6 +622,23 @@ func (s *OutputRoomEventConsumer) writeFTS(ev *rstypes.HeaderedEvent, pduPositio
 		if err := s.fts.Index(e); err != nil {
 			return err
 		}
+		// If the event is an edited message we remove the original event from the index
+		// to avoid duplicates in the search results.
+		relatesTo := gjson.GetBytes(ev.Content(), "m\\.relates_to")
+		if relatesTo.Exists() {
+			relatedData := relatesTo.Value().(map[string]interface{})
+			if _, ok := relatedData["rel_type"]; ok && relatedData["rel_type"] == "m.replace" {
+				// We remove the original event from the index
+				if srcEventID, ok := relatedData["event_id"]; ok {
+					if err := s.fts.Delete(srcEventID.(string)); err != nil {
+						log.WithFields(log.Fields{
+							"event_id": ev.EventID(),
+							"src_id":   srcEventID,
+						}).WithError(err).Error("Failed to delete edited message from fulltext index")
+					}
+				}
+			}
+		}
 	}
 	return nil
 }
--- a/syncapi/consumers/roomserver.go
+++ b/syncapi/consumers/roomserver.go
@@ -601,9 +601,11 @@ func (s *OutputRoomEventConsumer) writeFTS(ev *rstypes.HeaderedEvent, pduPositio
 	}
 	e.SetContentType(ev.Type())
 
+	var relatesTo gjson.Result
 	switch ev.Type() {
 	case "m.room.message":
 		e.Content = gjson.GetBytes(ev.Content(), "body").String()
+		relatesTo = gjson.GetBytes(ev.Content(), "m\\.relates_to")
 	case spec.MRoomName:
 		e.Content = gjson.GetBytes(ev.Content(), "name").String()
 	case spec.MRoomTopic:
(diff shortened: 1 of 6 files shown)


syncapi/consumers/roomserver.go
```diff
@@ -622,6 +622,23 @@ func (s *OutputRoomEventConsumer) writeFTS(ev *rstypes.HeaderedEvent, pduPositio
...
+				if srcEventID, ok := relatedData["event_id"]; ok {
+					if err := s.fts.Delete(srcEventID.(string)); err != nil {
+						log.WithFields(log.Fields{
+							"event_id": ev.EventID(),
+							"src_id":   srcEventID,
+						}).WithError(err).Error("Failed to delete edited message from the fulltext index")
+					}
+				}
+			}
+		}
```
S7evinK, line 641: What do you think about:
```go
		relType := gjson.GetBytes(ev.Content(), "m\\.relates_to.rel_type").Str
		if relType == "m.replace" {
			eventID := gjson.GetBytes(ev.Content(), "m\\.relates_to.event_id").Str
			if err := s.fts.Delete(eventID); err != nil {
				log.WithFields(log.Fields{
					"event_id": ev.EventID(),
					"src_id":   eventID,
				}).WithError(err).Error("Failed to delete edited message from the fulltext index")
			}
		}
```

imo this is easier to read than type assertions (and probably less error prone)

syncapi/syncapi_test.go
```diff
@@ -1324,6 +1327,95 @@ func TestUpdateRelations(t *testing.T) {
...
+
+	if err := api.SendEvents(processCtx.Context(), rsAPI, api.KindNew, room.Events(), "test", "test", "test", nil, false); err != nil {
+		t.Fatalf("failed to send events: %v", err)
+	}
+
+	ev1 := room.CreateAndInsert(t, user, "m.room.message", map[string]interface{}{"body": "first"})
+	ev2 := room.CreateAndInsert(t, user, "m.room.message", map[string]interface{}{
+		"body": " * first",
+		"m.new_content": map[string]interface{}{
+			"body":    "first",
```
S7evinK, line 1366:
```suggestion
		"body": " * first second",
		"m.new_content": map[string]interface{}{
			"body":    "first second",
```

maybe? So we can also check that we actually only get "first second" as the result?

S7evinK (approved):
LGTM otherwise!

syncapi/consumers/roomserver.go
adnull: Indeed, i don't like assertions too, but I thought to avoid double parsing so didn't find another way to do this.

syncapi/syncapi_test.go
adnull: We check the ID of the replacing event as the test condition, so the event content isn't important in this case, but if you think it is better to control both I will add an appropriate test condition.

adnull: btw I remembered, when I was testing that on android element client, double resulting were happening when the source and the edited message were the same.

S7evinK: Yea, it probably doesn't make a huge difference.

In my test case I was searching for "Fulltext" on an edited message, where I simply added a ("on Dendrite") and both the original and new content showed up. I assume if the text is replaced entirely, it wouldn't result in duplicates.

syncapi/consumers/roomserver.go
S7evinK: We can also move this part of the code up to the `case "m.room.message"`, so we don't parse the event for `m.room.name` etc (even though that shouldn't make a huge difference, except when someone changes the topic/name continuously)

adnull: Sure, I'll fix it shortly

adnull: I've looked into it with fresh eyes and replaced type asserions with Map() method of gjson.Result. 
I've also moved parsing into "m.room.message" section leaving deletion on it's place as we can delete the source message only after the replacing message was indexed successfully.

S7evinK:
Thanks for this!

PR #3372, opened 2024-05-10
Fix media DB possibly leaking connections

S7evinK:
Grafana Pyroscope unveiled that we are hitting https://github.com/golang/go/blob/ad10fbd3c4ec7413775028213f4d5089b18926f7/src/database/sql/sql.go#L2739-L2742 for media DB queries.

Making the methods pointer receivers fixes this.

(Also some minor `error` cosmetic updates)


devonh (approved):
Seems weird that it ever wasn't pointer receivers :shrug: 
Looks good!

PR #3376, opened 2024-05-16
Don't bother hitting the DB if we got everything from cache

S7evinK:
Yet another finding where we hit the DB with a useless query.


devonh (approved):
:+1:

PR #3380, opened 2024-05-26
refactoring user interactive auth

hkohlsaat:
- making members of `auth.UserInteractive` private as they aren't used outside of the package and in order to protect from concurrent writes
- making the mutex protect the sessions only as only they are written to concurrently

This prepares future contributions I plan, in particular adding functionality for multi-stage flows.

### Pull Request Checklist

* [x] I have justified why this PR doesn't need tests: No functionality added or removed, only refactoring tested behaviour.
* [x] I have already signed off privately.

### DIFF
--- a/clientapi/auth/user_interactive.go
+++ b/clientapi/auth/user_interactive.go
@@ -106,9 +106,9 @@ type UserInteractive struct {
 	flows []userInteractiveFlow
 	// Map of login type to implementation
 	types map[string]Type
-	// Map of session ID to completed login types, will need to be extended in future
 
-	mutex    sync.RWMutex
+	mutex sync.RWMutex
+	// Map of session ID to completed login types, will need to be extended in future
 	sessions map[string][]string
 }
 
@@ -160,13 +160,11 @@ func (u *UserInteractive) challenge(sessionID string) *util.JSONResponse {
 	completed := u.sessions[sessionID]
 	u.mutex.RUnlock()
 
-	flows := u.flows
-
 	return &util.JSONResponse{
 		Code: 401,
 		JSON: Challenge{
 			Completed: completed,
-			Flows:     flows,
+			Flows:     u.flows,
 			Session:   sessionID,
 			Params:    make(map[string]interface{}),
 		},
--- a/clientapi/auth/user_interactive.go
+++ b/clientapi/auth/user_interactive.go
@@ -103,12 +103,13 @@ type userInteractiveFlow struct {
 // the user already has a valid access token, but we want to double-check
 // that it isn't stolen by re-authenticating them.
 type UserInteractive struct {
-	sync.RWMutex
-	Flows []userInteractiveFlow
+	flows []userInteractiveFlow
 	// Map of login type to implementation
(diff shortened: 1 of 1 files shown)


clientapi/auth/user_interactive.go
```diff
@@ -103,12 +103,13 @@ type userInteractiveFlow struct {
...
-	Flows []userInteractiveFlow
+	flows []userInteractiveFlow
 	// Map of login type to implementation
-	Types map[string]Type
+	types map[string]Type
 	// Map of session ID to completed login types, will need to be extended in future
-	Sessions map[string][]string
+
+	mutex    sync.RWMutex
+	sessions map[string][]string
```
S7evinK, line 112: We've lost the comment for `sessions` here.

clientapi/auth/user_interactive.go
```diff
@@ -157,10 +156,11 @@ type Challenge struct {
...
 func (u *UserInteractive) challenge(sessionID string) *util.JSONResponse {
-	u.RLock()
-	completed := u.Sessions[sessionID]
-	flows := u.Flows
-	u.RUnlock()
+	u.mutex.RLock()
+	completed := u.sessions[sessionID]
+	u.mutex.RUnlock()
+
+	flows := u.flows
```
S7evinK, line 163: Minor nit: No need to create `flows` when we can use `u.flows` below directly.

S7evinK (approved):
Other than the minor nits, looks fine to me.

Can we test the changes? Or do you plan to add tests when adding multi-stage flows?

clientapi/auth/user_interactive.go
hkohlsaat: - [x] using `u.flows` directly

clientapi/auth/user_interactive.go
hkohlsaat: - [x] moving comment

hkohlsaat:
> Can we test the changes?
As I wrote in the description, I only refactored internals of a tested module.
Almost all behavior of this file is tested. Especially the lines I changed are tested already as far as I checked. All I did is slightly refactoring the internals with tests passing before and after - without any intent to change behaviour. I don't see how one would add more tests without overdoing it, given the overseeable amount of changes. Is this okay with you? If not, could you give me an example of what you would like to have tested, additionally?
(Of course, conversely, I would add tests for multi-stage flows when I work on them, because that would be a change in behaviour. But I guess that's beside the point of these changes and your comment?)

S7evinK:
Thanks for contributing to Dendrite!

The repo has been cloned into [element-hq/dendrite](https://github.com/element-hq/dendrite) and made the switch to AGPL as per https://matrix.org/blog/2023/11/06/future-of-synapse-dendrite/.
For more information, feel free to read [this](https://element.io/blog/element-to-adopt-agplv3/) and [this](https://element.io/blog/sustainable-licensing-at-element-with-agpl/)

If you still wish to contribute under the new license, feel free to create a new PR over at [element-hq/dendrite](https://github.com/element-hq/dendrite).

PR #3381, opened 2024-05-31
Speed up purging rooms

S7evinK:


S7evinK:
Since I've successfully used this multiple times, merging.

PR #3382, opened 2024-05-31
Use mautrix for tag events instead of deprecated matrix/go

dgrr:
The iOS crashes because dendrite is using an old framework that is deprecated. For reference [look here](https://github.com/matrix-org/matrix-ios-sdk/issues/1857).
This PR just fixes this.

Signed-off-by: `Dario <me@dgrp.es>`

### DIFF
--- a/clientapi/routing/room_tagging.go
+++ b/clientapi/routing/room_tagging.go
@@ -21,9 +21,9 @@ import (
 	"github.com/matrix-org/dendrite/clientapi/httputil"
 	"github.com/matrix-org/dendrite/clientapi/producers"
 	"github.com/matrix-org/dendrite/userapi/api"
-	"github.com/matrix-org/gomatrix"
 	"github.com/matrix-org/gomatrixserverlib/spec"
 	"github.com/matrix-org/util"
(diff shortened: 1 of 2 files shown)


S7evinK:
The Spec states `order` is a number in the range of 0-1, while the iOS SDK is using string. 
https://spec.matrix.org/latest/client-server-api/#mtag

So imo this should be fixed in the iOS SDK.

dgrr:
> The Spec states `order` is a number in the range of 0-1, while the iOS SDK is using string. https://spec.matrix.org/latest/client-server-api/#mtag
> 
> So imo this should be fixed in the iOS SDK.

Indeed, but the new library (mautrix) is also using a string or json.Number, so it might be serialized as string and used as a number. Probably because a valid value could be `0.0000004200000000000069`, and the precision of decoding into a float32 might not be enough. Therefore the usage of strings.

I think the change is legit, but both libraries use string (or json.Number) to improve the precision of encoding/decoding the message.

S7evinK:
Well, sorry for that, but Element iOS is not spec conform.

PR #3383, opened 2024-06-10
claim to support v1.3 and v1.4

jjj333-p:
We already claim to support v1.1 and v1.2 which dendrite supports about as much. Maunium bridges now require v1.4  and will refuse to run otherwise.

i dont think it needs additional tests given the existing test for the endpoint seems to just be if it responds which it still does

### Pull Request Checklist

Signed-off-by: `Joseph Winkie <jjj333.p.1325@gmail.com>`

### DIFF
--- a/clientapi/routing/routing.go
+++ b/clientapi/routing/routing.go
@@ -147,6 +147,8 @@ func Setup(
 					"v1.0",
 					"v1.1",
 					"v1.2",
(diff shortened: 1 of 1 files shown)


S7evinK:
As much as I'd like to approve this..
I'm not going to claim support for versions we don't really have support for. Sorry :(

PR #3384, opened 2024-06-11
Blacklist sytests that require MSC3967

richvdh:
https://github.com/matrix-org/sytest/pull/1383 updates some sytests in line with MSC3967. Dendrite does not support MSC3967, so these tests fail.


devonh (approved):
Seems like a sane approach to me for now.

PR #3397, opened 2024-07-27
Implement MSC3916

S7evinK:
Needs https://github.com/matrix-org/gomatrixserverlib/pull/437


S7evinK:
Complement PR: https://github.com/matrix-org/complement/pull/731

mediaapi/routing/download.go
```diff
@@ -369,8 +384,49 @@ func (r *downloadRequest) respondFromLocalFile(
 		" object-src 'self';"
 	w.Header().Set("Content-Security-Policy", contentSecurityPolicy)
```
tulir, line 385: Wouldn't it be better to set these headers in the `!forFederation` block rather than deleting them in the federation block?

mediaapi/routing/download.go
```diff
@@ -767,19 +823,68 @@ func (r *downloadRequest) fetchRemoteFile(
...
+
+		// Get the actual media content
+		p, multipartErr = mr.NextPart()
+		if multipartErr != nil {
+			return "", false, multipartErr
+		}
+
+		contentLength, reader, parseErr = r.GetContentLengthAndReader(p.Header.Get("Content-Length"), p, maxFileSizeBytes)
+		// For multipart requests, we need to get the Content-Type of the second part, which is the actual media
+		r.MediaMetadata.ContentType = types.ContentType(p.Header.Get("Content-Type"))
```
tulir, line 881: I think you're missing redirect handling here? https://spec.matrix.org/v1.11/server-server-api/#get_matrixfederationv1mediadownloadmediaid

> The second part is either:
> 1. [...]
> 2. or a `Location` header to redirect the caller to where the media can be retrieved. The URL at `Location` SHOULD have appropriate `Content-Type` and `Content-Disposition` headers which describe the media.
>
>    When `Location` is present, servers SHOULD NOT cache the URL. The remote server may have applied time limits on its validity. If the caller requires an up-to-date URL, it SHOULD re-request the media download.

mediaapi/routing/download.go
```diff
@@ -767,19 +823,68 @@ func (r *downloadRequest) fetchRemoteFile(
...
+
+	var contentLength int64
+	var reader io.Reader
+	var parseErr error
+	if isMultiPart {
+		r.Logger.Debug("Downloaded file using authenticated endpoint")
+		var params map[string]string
+		_, params, err = mime.ParseMediaType(resp.Header.Get("Content-Type"))
+		if err != nil {
+			panic(err)
```
neilalexander, line 855: Panic may not be ideal here.

mediaapi/routing/download.go
```diff
@@ -767,19 +823,68 @@ func (r *downloadRequest) fetchRemoteFile(
...
+		// Get the first, JSON, part
+		p, multipartErr := mr.NextPart()
+		if multipartErr != nil {
+			return "", false, multipartErr
+		}
+
+		if p.Header.Get("Content-Type") != "application/json" {
+			return "", false, fmt.Errorf("first part of the response must be application/json")
+		}
+		// TODO: Once something is defined, parse the JSON content
```
neilalexander, line 871: IMO parse it anyway, just to be sure that Dendrite always fails properly for garbage content.

mediaapi/routing/download.go
S7evinK: Good point.

mediaapi/routing/download.go
S7evinK: Hu, how did that get there?

mediaapi/routing/download.go
S7evinK: Sounds like a good idea.

mediaapi/routing/download.go
S7evinK: Thanks for the fix! :)

clientapi/routing/routing.go
```diff
@@ -94,6 +94,7 @@ func Setup(
 	unstableFeatures := map[string]bool{
 		"org.matrix.e2e_cross_signing": true,
 		"org.matrix.msc2285.stable":    true,
+		"org.matrix.msc3916":           true,
```
S7evinK, line 97: Wondering if this should be `.stable`?

tulir: Probably yes since the unstable endpoints aren't supported. I think clients will only use new endpoints if v1.11 is advertised though

S7evinK: From a short test with Element Web: Even with v1.11 advertised, the new endpoints aren't used. :/

tulir: Element Web specifically is buggy: element-hq/element-web#27759

S7evinK: Works fine with ED nightly when advertising v1.11.

federationapi/routing/routing.go
```diff
@@ -678,6 +679,53 @@ func MakeFedAPI(
 	return httputil.MakeExternalAPI(metricsName, h)
 }
 
+// MakeFedAPI makes an http.Handler that checks matrix federation authentication.
+func MakeFedAPIHTML(
```
kegsay, line 683: It ain't HTML though?

mediaapi/routing/download.go
```diff
@@ -61,6 +64,9 @@ type downloadRequest struct {
 	ThumbnailSize      types.ThumbnailSize
 	Logger             *log.Entry
 	DownloadFilename   string
+	forFederation      bool // whether we need to return a multipart/mixed response
```
kegsay, line 67: Flag is unclear. Is this implying that the download is over fed? Or the response will be sent via fed?

mediaapi/routing/download.go
```diff
@@ -767,19 +826,94 @@ func (r *downloadRequest) fetchRemoteFile(
...
-			return "", false, fmt.Errorf("File with media ID %q does not exist on %s", r.MediaMetadata.MediaID, r.MediaMetadata.Origin)
+		isMultiPart = false
+		// try again on the unauthed endpoint
+		// create request for remote file
+		resp, err = client.CreateMediaDownloadRequest(ctx, r.MediaMetadata.Origin, string(r.MediaMetadata.MediaID))
+		if err != nil || (resp != nil && resp.StatusCode != http.StatusOK) {
+			if resp != nil && resp.StatusCode == http.StatusNotFound {
+				return "", false, fmt.Errorf("File with media ID %q does not exist on %s", r.MediaMetadata.MediaID, r.MediaMetadata.Origin)
+			}
+			return "", false, fmt.Errorf("file with media ID %q could not be downloaded from %s", r.MediaMetadata.MediaID, r.MediaMetadata.Origin)
```
kegsay, line 841: `err` is dropped to the floor. Please don't do this.

kegsay (changes requested):
A few things:
 - restructure the download functions to make the fallback logic clearer,
 - double check error handling / messages as numerous ones aren't logging the right thing,
 - add tests, at the _very least_ for the multipart stuff as it seems pretty complex and fiddly. There's examples in the MSC itself, so I suggest you use them for the tests.

mediaapi/routing/download.go
```diff
@@ -767,19 +826,94 @@ func (r *downloadRequest) fetchRemoteFile(
...
+		// Get the actual media content
+		p, multipartErr = mr.NextPart()
+		if multipartErr != nil {
+			return "", false, multipartErr
+		}
+
+		redirect := p.Header.Get("Location")
+		if redirect != "" {
+			if !strings.HasPrefix(redirect, "https://") {
+				return "", false, fmt.Errorf("redirect URL must be HTTPS")
```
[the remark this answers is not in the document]
tulir: Federation has to be all https, but I guess redirects could allow http too (should maybe have that in the spec 🤔). The main thing I wanted to prevent is someone using non-http things like the special `matrix://` protocol

mediaapi/routing/download.go
S7evinK: Whoops, fixed the shadowing of `err` but forgot to update `fmt.Errorf`.

S7evinK: Also, this is everywhere in the MediaAPI, we're dropping `err` in quite a few places. Here it is not great, but in other places it is fine, as we otherwise may disclose system internals (i.e. returning an error about permission issues for a specific path)

mediaapi/routing/download.go
S7evinK: That it's the response for an incoming federation request.

federationapi/routing/routing.go
S7evinK: While true, we also have `MakeHTMLAPI`, with basically the same function signature. That's why I named it that way, could be `MakeFedHTMLAPI`, though.

mediaapi/routing/download.go
S7evinK: Oh, turns out that this is the "original" error message, which I just copy & pasted.

mediaapi/routing/download.go
```diff
@@ -767,19 +826,94 @@ func (r *downloadRequest) fetchRemoteFile(
 ) (types.Path, bool, error) {
 	r.Logger.Debug("Fetching remote file")
 
-	// create request for remote file
-	resp, err := client.CreateMediaDownloadRequest(ctx, r.MediaMetadata.Origin, string(r.MediaMetadata.MediaID))
+	// Attempt to download via authenticated media endpoint
```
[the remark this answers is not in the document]
S7evinK: That would be an exercise when refactoring/rewriting the entire MediaAPI, imho.

mediaapi/routing/download.go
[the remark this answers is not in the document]
S7evinK: `Location` header is now removed, due to not having any mechanism to allow/block access to certain resources.

mediaapi/routing/download.go
```diff
@@ -767,19 +823,69 @@ func (r *downloadRequest) fetchRemoteFile(
...
+	isMultiPart := true
+	resp, err := r.fedClient.DownloadMedia(ctx, r.origin, r.MediaMetadata.Origin, string(r.MediaMetadata.MediaID))
 	if err != nil || (resp != nil && resp.StatusCode != http.StatusOK) {
-		if resp != nil && resp.StatusCode == http.StatusNotFound {
-			return "", false, fmt.Errorf("File with media ID %q does not exist on %s", r.MediaMetadata.MediaID, r.MediaMetadata.Origin)
+		isMultiPart = false
+		// try again on the unauthed endpoint
```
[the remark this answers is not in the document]
S7evinK: True, but what if e.g. matrix.org (or CF in that matter), returns a 520? (or whatever the response was for "CF is upset").
Going to revisit this in another PR, though.

mediaapi/routing/download.go
```diff
@@ -367,10 +382,50 @@ func (r *downloadRequest) respondFromLocalFile(
...
-	if _, err := io.Copy(w, responseFile); err != nil {
-		return nil, fmt.Errorf("io.Copy: %w", err)
+	if !r.multipartResponse {
+		w.Header().Set("Content-Security-Policy", contentSecurityPolicy)
+		if _, err = io.Copy(w, responseFile); err != nil {
+			return nil, fmt.Errorf("io.Copy: %w", err)
+		}
+	} else {
+		// Update the header to be multipart/mixed; boundary=$randomBoundary
+		boundary := uuid.NewString()
```
kegsay, line 393: I think this can be improved a bit. Please factor out the else block into a new function:
```go
func sendMultipartResponse(w http.ResponseWriter, boundary, dataContentType string, data io.Reader) {

}
```

This helps cut down the size of `respondFromLocalFile` but more importantly the new function can be unit tested with the `httptest` package. Please add a test with the exact example from the MSC (the one with `This media is plain text. Maybe somebody used it as a paste bin.`)

mediaapi/routing/download.go
```diff
@@ -767,19 +826,78 @@ func (r *downloadRequest) fetchRemoteFile(
...
-	contentLength, reader, parseErr := r.GetContentLengthAndReader(resp.Header.Get("Content-Length"), &resp.Body, maxFileSizeBytes)
+	// If this wasn't a multipart response, set the Content-Type now. Will be overwritten
+	// by the multipart Content-Type below.
+	r.MediaMetadata.ContentType = types.ContentType(resp.Header.Get("Content-Type"))
+
+	var contentLength int64
+	var reader io.Reader
+	var parseErr error
+	if isAuthed {
+		r.Logger.Debug("Downloaded file using authenticated endpoint")
```
kegsay, line 854: Similar to my other comment, this new code can easily be put in a separate function which can be unit tested.

kegsay (changes requested):
Still needs tests. Unit tests for new code is not a big ask here and should land before this PR does. Ideally we'd test control flow here as it's complex, but that is a bigger ask and would need more refactoring to do correctly.

kegsay (approved):

PR #3399, opened 2024-07-28
Fix Postgres CI runs

S7evinK:
For some reason Postgres wouldn't allow connections over TCP. Since the tests are run in isolation, this allows all "peer" authed connections without a password.

For CI to be fully happy, we need #3397


S7evinK:
Fixed with https://github.com/matrix-org/dendrite/pull/3360, as it was touching those files anyway.

PR #3407, opened 2024-08-02
Bump yggdrasil

S7evinK:


S7evinK:
Not seeing any issues, so approved from my side. (This patch has been sent to me by @neilalexander)

PR #3410, opened 2024-08-16
All might legend/issue3333

AllMightLegend:
### Pull Request Checklist

Signed-off-by: `Srinjoy Sen Chowdhury  srinjoysen123@gmail.com`

The issue involved ensuring that the ts parameter in Dendrite is correctly handled when the request is from an appservice, and that it is ignored or handled differently when not. The resolution was achieved through the following steps:

Refactoring the Logic:
The logic for processing the ts parameter was refactored into a separate function named HandleEventTimestamp. This function determines whether the request is from an appservice by calling isAppService(req). If it is, the function parses the ts parameter using httputil.ParseTSParam(req). If the request is not from an appservice, the function defaults to using the current time or another appropriate handling.

Updating sendevent.go:
The inline logic in sendevent.go that handled the ts parameter was replaced with a call to the new HandleEventTimestamp function. This ensures that the logic is centralized and can be easily tested and maintained.

Creating Tests:
A new test file, ts_param_test.go, was created in the testing directory. This file includes tests that cover various scenarios:

When the ts parameter is valid and the request is from an appservice.
When the ts parameter is invalid or missing.
When the request is not from an appservice.


S7evinK:
Sorry for closing this, feel free to join [#dendrite-dev:matrix.org](https://matrix.to/#/%23dendrite-dev%3Amatrix.org?via=matrix.org&via=envs.net&via=element.io) for further discussion.

PR #3418, opened 2024-09-10
Update NATS to 2.10.20, use `SyncAlways`

neilalexander:
The internal NATS instance is definitely convenient but it does have one problem: its lifecycle is tied to the Dendrite process. That means if Dendrite panics or OOMs, it takes out NATS with it. I suspect this is sometimes contributing to what people see with stuck streams, as some operations or state might not be written to disk fully before it gets interrupted.

Using `SyncAlways` means that NATS will effectively use `O_SYNC` and block writes on flushes, which should improve resiliency against this kind of failure considerably. It might affect performance a little but shouldn't be significant.

Also updates NATS to 2.10.20 as there have been all sorts of fixes since 2.10.7, including better `SyncAlways` handling.

Signed-off-by: Neil Alexander <git@neilalexander.dev>


S7evinK (approved):
Thanks!

PR #3427, opened 2024-09-18
Ensure Helm packages generated end up in docs/ subpath that's hosted

Arkaniad:
I've noticed that Chart Releaser is behaving improperly, and not successfully putting the charts where gh-pages will hots them and generating invalid index.

This change should ensure built charts end up in the `docs/` subpath, which should ensure that `gh-pages` branch isn't getting an invalid `index.yaml` and new charts should be automatically available.

### Pull Request Checklist

Signed-off-by: `Rhea Danzey <rdanzey@element.io>`

### DIFF
--- a/helm/ct.yaml
+++ b/helm/ct.yaml
@@ -5,3 +5,5 @@ chart-repos:
 chart-dirs:
   - helm
 validate-maintainers: false
+# this should ensure the tarballs are in the appropriate location for GH pages, rather than repo root
+package-path: docs/


S7evinK (approved):
Ohh, thank you for taking a look at that, I meant to do that after the release on Friday.

S7evinK:
Looks like that didn't do the right thing. :(
Wondering if moving back to `latest` fixes that?

https://github.com/matrix-org/dendrite/blob/152626c87e419a9b345a91f390f999212f420ff2/.github/workflows/helm.yml#L34-L41

Arkaniad:
It seems worth a try! I'm puzzled why it's not generating links to the release tarball as it seems to have been..

Arkaniad:
Trying [here](https://github.com/matrix-org/dendrite/pull/3429)

I'll merge it unless there's any objections, and if it doesn't fix I'll only spend a little more time pushing it and failing that just fix up the index manually and leave it there for now

Issue #3424, opened 2024-09-16
Authenticated media requires authentication even on OPTIONS method


arenekosreal:
### Background information

- **Dendrite version or git SHA**: 0.13.8
- **SQLite3 or Postgres?**: postgres
- **Running in Docker?**: no
- **`go version`**: 1.23.1
- **Client used (if applicable)**: Cinny web 4.2.1 on ungoogled-chromium 128.0.6613.113

### Description

- **What** is the problem: Cinny fails to load all media through MSC3916 endpoints.
- **Who** is affected: Me
- **How** is this bug manifesting: In chromium's developer tools, all media requests through MSC3916 endpoints failed to perform correctly.
- **When** did this first appear: after Cinny upgraded to 4.2.1 and dendrite upgraded to 0.13.8

### Steps to reproduce

- Ensure Cinny web version is 4.2.1 and dendrite version is 0.13.8
- Simply login and check media like avatars or images in chat.
- You will find those media failed to load.
- Check developer tools in browser for the requests to the media
- You will find GET request failed with `net::ERR_FAILED`, fetch request failed with `CORS error`, preflight request failed with `401 Unauthorized`

<!--
Describe how what happens differs from what you expected.

If you can identify any relevant log snippets from server logs, please include
those (please be careful to remove any personal or private data). Please surround them with
``` (three backticks, on a line on their own), so that they are formatted legibly.

Alternatively, please send logs to @kegan:matrix.org, @s7evink:matrix.org or @devonh:one.ems.host
with a link to the respective Github issue, thanks!
-->
### Screenshots of developer tools
![image](https://github.com/user-attachments/assets/0e678ae4-0cd4-433f-a73a-0dae56772a59)

### Notes
According to [this](https://stackoverflow.com/questions/68143581/how-to-force-authentication-in-preflight-request/68148781#68148781) stackoverflow answer, authentication in preflight requests is not possible, so this may be the cause.

PR #3431, opened 2024-09-20
Make OPTIONS method on MSC3916 endpoints available without auth

arenekosreal:
OPTIONS method is usually sent by browser in preflight requests, most of the time we cannot control preflight request to add auth header.

Synapse will return a 204 response directly without authentication for those OPTIONS method.

According to firefox's documentation, both 200 and 204 are acceptable so I think there is no need to change handler in dendrite.

This closes https://github.com/matrix-org/dendrite/issues/3424

No need to add a test because this is just a fix and I have tested on my Cinny Web client personally.

### Pull Request Checklist

Signed-off-by: `arenekosreal <17194552+arenekosreal@users.noreply.github.com>`

### DIFF
--- a/internal/httputil/httpapi.go
+++ b/internal/httputil/httpapi.go
@@ -210,6 +210,12 @@ func MakeExternalAPI(metricsName string, f func(*http.Request) util.JSONResponse
 // This is used to serve HTML alongside JSON error messages
 func MakeHTTPAPI(metricsName string, userAPI userapi.QueryAcccessTokenAPI, enableMetrics bool, f func(http.ResponseWriter, *http.Request), checks ...AuthAPIOption) http.Handler {
 	withSpan := func(w http.ResponseWriter, req *http.Request) {
+		if req.Method == http.MethodOptions {
+			util.SetCORSHeaders(w)
+			w.WriteHeader(http.StatusOK) // Maybe http.StatusNoContent?
+			return
+		}
+
 		trace, ctx := internal.StartTask(req.Context(), metricsName)
 		defer trace.EndTask()
 		req = req.WithContext(ctx)


S7evinK (changes requested):
While this works, I'd prefer having the `OPTIONS` check in the `MakeHTTPAPI` function, i.e.

```go
// MakeHTTPAPI adds Span metrics to the HTML Handler function
// This is used to serve HTML alongside JSON error messages
func MakeHTTPAPI(metricsName string, userAPI userapi.QueryAcccessTokenAPI, enableMetrics bool, f func(http.ResponseWriter, *http.Request), checks ...AuthAPIOption) http.Handler {
	withSpan := func(w http.ResponseWriter, req *http.Request) {
		if req.Method == http.MethodOptions {
			util.SetCORSHeaders(w)
			w.WriteHeader(http.StatusOK)
			return
		}
...
```

arenekosreal:
I have changed the code to check `OPTIONS` in `MakeHTTPAPI`, I must say the diff is much cleaner than the old one.

internal/httputil/httpapi.go
```diff
@@ -210,6 +210,12 @@ func MakeExternalAPI(metricsName string, f func(*http.Request) util.JSONResponse
 // This is used to serve HTML alongside JSON error messages
 func MakeHTTPAPI(metricsName string, userAPI userapi.QueryAcccessTokenAPI, enableMetrics bool, f func(http.ResponseWriter, *http.Request), checks ...AuthAPIOption) http.Handler {
 	withSpan := func(w http.ResponseWriter, req *http.Request) {
+		if req.Method == http.MethodOptions {
+			util.SetCORSHeaders(w)
+			w.WriteHeader(http.StatusOK) // Maybe http.StatusNoContent?
```
S7evinK, line 215: We're using `http.StatusOK` in [MakeJSONAPI](https://github.com/matrix-org/util/blob/399730281e66400bbfa38220eeb874609fca6c55/json.go#L136-L140), so I'd keep that the same.

S7evinK (approved):
Thanks!

PR #3432, opened 2024-09-20
Add release trigger for gh-pages rebuild

Arkaniad:
I've noticed that newly released charts don't appear immediately until I manually run the `gh-pages` workflow - Lets see if we can dispatch that on a release event

### Pull Request Checklist

Signed-off-by: `Rhea Danzey <rdanzey@element.io>`

### DIFF
--- a/.github/workflows/helm.yml
+++ b/.github/workflows/helm.yml
@@ -39,3 +39,13 @@ jobs:
           config: helm/cr.yaml
           charts_dir: helm/
           mark_as_latest: false
+
+  invoke_pages_build:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/helm.yml
```diff
@@ -39,3 +39,13 @@ jobs:
           config: helm/cr.yaml
           charts_dir: helm/
           mark_as_latest: false
+
+  invoke_pages_build:
+    runs-on: ubuntu-latest
```
S7evinK, line 44: I think we need a `needs: [ release ]`, as otherwise this might run concurrently with the release.

S7evinK (approved):
Again, something I wanted to do for a long time :D So thanks!

